PR #2, opened 2018-12-04
PR For the purpose of code reviewing the parser -- This should not be merged

VMatthijs:


lib/lexer.mll
```diff
@@ -1,301 +0,0 @@
...
-                                  <- new_lexbuf.lex_start_p ;
-                                if dup_exists (Str.split (Str.regexp
-                                  "\" included from \"") path) then
-                                  raise (Errors.SyntaxError
-                                    (Includes (
-                                      "Found cyclical include structure.\n",
-                                      (lexeme_end_p (Stack.top include_stack)
-                                         ))))          ;
-                                Stack.push new_lexbuf include_stack ;
-                                token new_lexbuf }
```
seantalts: Could you unindent this a little, maybe by moving the { to the next line?

lib/lexer.mll
```diff
@@ -1,301 +0,0 @@
...
-let real_constant1 = integer_constant? '.' ['0'-'9']* exp_literal? 
-let real_constant2 = '.' ['0'-'9']+ exp_literal?
-let real_constant3 = integer_constant exp_literal 
-let real_constant = real_constant1 | real_constant2 | real_constant3
-
-rule token = parse
-(* White space, line numers and comments *)
-    '\n'                      { lexer_logger "newline" ;
-                                incr_linenum lexbuf ; token lexbuf }
-  | [' ' '\t' '\012' '\r']    { lexer_logger "space" ;
```
seantalts: might be worth setting up a separate whitespace definition

lib/Parse.ml
```diff
@@ -1,68 +0,0 @@
...
-
-open Errors
-
-let parse parse_fun lexbuf =
-  (* see the Menhir manual for the description of
-     error messages support *)
-  let open MenhirLib.General in
-  let module Interp = Parser.MenhirInterpreter in
-  let _ = Stack.push lexbuf Lexer.include_stack in
-  let input _ =
```
seantalts: I think you can get rid of the _ and the wrapping () and the empty () and use partial application

lib/parser.mly
```diff
@@ -1,599 +0,0 @@
...
-    ae=option(pair(ASSIGN, expression)) SEMICOLON
-    { grammar_logger "var_decl" ;
-      let sizes = match d with None -> [] | Some l -> l in
-      match ae with
-      | Some a ->
-          UntypedStmt
-            ( VDeclAss
-                {sizedtype= reducearray (sbt, sizes); identifier= id; value= snd a}
-            , initialize_stmt_meta $startpos $endpos )
-      | _ ->
```
seantalts: I think might be better to write `None ->` just to be explicit

lib/Ast.ml
```diff
@@ -1,223 +0,0 @@
...
-      { loop_variable: identifier
-      ; lower_bound: 'e
-      ; upper_bound: 'e
-      ; loop_body: 's }
-  | ForEach of identifier * 'e * 's
-  (* TODO: maybe make separate nodes for matrix and array foreach loops? *)
-  | Block of 's list
-  | VDecl of 'e sizedtype * identifier
-  | VDeclAss of {sizedtype: 'e sizedtype; identifier: identifier; value: 'e}
-  | TVDecl of 'e sizedtype * 'e transformation * identifier
```
seantalts: T for Top could probably be changed to be a little more descriptive - maybe constrainable or even global? not sure

seantalts: I'm also still wondering if we would be served just as well with a single Decl that encapsulated all the possibilities orthogonally:
```
VDecl of {
sizedtype: 'e sizedtype;
id: identifier;
value: 'e option;
transformation: 'e transformation option;
write_out: boolean;
global: boolean;
}
```
Or something like that? My reasoning is around the idea that we might eventually want to orthogonally toggle whether we write a variable out or not, and that the code for generating these things would probably be easier in this form (though we'll be working from the MIR). But maybe the semantic analysis is better in the current form?

lib/Semantic_check.ml
```diff
@@ -1,2163 +0,0 @@
...
-
-(* TODO: check that variables are assigned to before they are used! *)
-
-open Symbol_table
-open Ast
-open Stan_math_signatures
-open Operators
-open Errors
-open Type_conversion
-open Pretty_printing
```
seantalts: We had some discussion and think it might make sense to try opening Core_kernel in most places to essentially change to that standard library in a consistent fashion.

lib/Semantic_check.ml
```diff
@@ -1,2163 +0,0 @@
...
-let context_flags =
-  { current_block= Functions
-  ; in_fun_def= false
-  ; in_returning_fun_def= false
-  ; in_rng_fun_def= false
-  ; in_lp_fun_def= false
-  ; in_loop= false }
-
-(* Some helper functions *)
-let dup_exists l =
```
seantalts: Check out Core_kernel.List.find_a_dup

lib/lexer.mll
```diff
@@ -1,301 +0,0 @@
...
-  | '['                       { lexer_logger "[" ;
-                                Parser.LBRACK }
-  | ']'                       { lexer_logger "]" ;
-                                Parser.RBRACK }
-  | '<'                       { lexer_logger "<" ;
-                                Parser.LABRACK }
-  | '>'                       { lexer_logger ">" ;
-                                Parser.RABRACK }
-  | ','                       { lexer_logger "," ;
-                                Parser.COMMA }
```
seantalts: could do all of these on the same line

lib/parser.mly
```diff
@@ -1,599 +0,0 @@
...
-%start <Ast.untyped_program> program
-
-%%
-
-
-(* Grammar *)
-
-(* program *)
-program:
-  | obf=option(function_block)
```
seantalts: consider ofb?

lib/parser.mly
```diff
@@ -1,599 +0,0 @@
...
-  | bt=basic_type ud=option(unsized_dims)
-    {  grammar_logger "unsized_type" ;
-       let reparray n x =
-         let rec repeat n f x =
-           if n <= Int64.zero then x else repeat (Int64.pred n) f (f x) in
-         repeat n (fun y -> Array y) x in
-       let size =
-         match ud with Some d -> Int64.succ (Int64.of_int d) | _ -> Int64.zero
-       in
-       reparray size bt    }
```
seantalts: Can we simplify this? Seems like a fold might work.

lib/parser.mly
```diff
@@ -1,599 +0,0 @@
...
-    {  grammar_logger "printable string" ; [PString s] }
-  | p1=printables COMMA p2=printables
-    { grammar_logger "printables" ; p1 @ p2 }
-
-(* L-values *)
-lhs:
-  | id=identifier
-    {  grammar_logger "lhs_identifier" ; (id, []) }
-  | l=lhs LBRACK id=indexes RBRACK
-    {  grammar_logger "lhs_index" ; (fst l, (snd l)@id) }
```
seantalts: Can you unpack in the binding? e.g.
`| (fstl, sndl) = lhs LBRACK id=indexes RBRACK`

lib/parser.mly
```diff
@@ -1,599 +0,0 @@
...
-
-let initialize_expr_meta startpos endpos =
-  {expr_untyped_meta_loc= Location (startpos, endpos)}
-
-let initialize_stmt_meta startpos endpos =
-  {stmt_untyped_meta_loc= Location (startpos, endpos)}
-
-(* Takes a sized_basic_type and a list of sizes and repeatedly applies then
-   SArray constructor, taking sizes off the list *)
-let reducearray (sbt, l) =
```
seantalts: There are some List module functions (maybe more in Core_kernel) to do various kinds of folds and reduces

lib/Ast.ml
```diff
@@ -1,223 +0,0 @@
...
-
-and unsizedtype =
-  | Int
-  | Real
-  | Vector
-  | RowVector
-  | Matrix
-  | Array of unsizedtype
-  | Fun of (originblock * unsizedtype) list * returntype
-  | PrimitiveFunction
```
seantalts: MathLibraryFunction but also could you add a comment on why this is in here (that it is used to annotate expressions)

lib/Ast.ml
```diff
@@ -1,223 +0,0 @@
...
-      ; arguments: (originblock * unsizedtype * identifier) list
-      ; body: 's }
-
-and statement_untyped_metadata =
-  {stmt_untyped_meta_loc: location sexp_opaque [@compare.ignore]}
-
-and statement_returntype =
-  | NoReturnType
-  | Incomplete of returntype
-  | Complete of returntype
```
seantalts: Need comments or renaming to indicate what these mean

lib/Ast.ml
```diff
@@ -1,223 +0,0 @@
...
-  | InfixOp of 'e * infixop * 'e
-  | PrefixOp of prefixop * 'e
-  | PostfixOp of 'e * postfixop
-  | Variable of identifier
-  | IntNumeral of string
-  | RealNumeral of string
-  | FunApp of identifier * 'e list
-  | CondFunApp of identifier * 'e list
-  | GetLP
-  (* deprecated *)
```
seantalts: put one line up and/or indicate which is deprecated

lib/Ast.ml
```diff
@@ -1,223 +0,0 @@
...
-
-and 'e index =
-  | All
-  | Single of 'e
-  | Upfrom of 'e
-  | Downfrom of 'e
-  | Between of 'e * 'e
-  | Multiple of 'e
-
-and 'e expression =
```
seantalts: We talked briefly about how if, in the body here, we just wrote `'e expression` everywhere we wrote `'t` or something then `'t` becomes the type of the metadata. this might be easier to understand and only a little bit more annoying to write.

lib/Ast.ml
```diff
@@ -1,223 +0,0 @@
...
-(* == Statements == *)
-and assignmentoperator =
-  | Assign
-  | PlusAssign
-  | MinusAssign
-  | TimesAssign
-  | DivideAssign
-  | EltTimesAssign
-  | EltDivideAssign
-  | ArrowAssign
```
seantalts: move deprecated

lib/Ast.ml
```diff
@@ -1,223 +0,0 @@
...
-  ; expr_typed_meta_loc: location sexp_opaque [@compare.ignore] }
-
-and untyped_expression =
-  | UntypedExpr of (untyped_expression expression * expression_untyped_metadata)
-
-and typed_expression =
-  | TypedExpr of (typed_expression expression * expression_typed_metadata)
-
-(* == Statements == *)
-and assignmentoperator =
```
seantalts: Assignment operators are basically the same as the infix operators (minus the arrow) - so we could probably factor this such that an assignment takes an infix operator (where conditional operators are separated out).

lib/Semantic_check.ml
```diff
@@ -1,2163 +0,0 @@
-(** Semantic validation of AST*)
-
-(* Idea: check many of things related to identifiers that are hard to check
-during parsing and are in fact irrelevant for building up the parse tree *)
-
-(* To check:
-Variable in scope if used
```
seantalts: Move these comments to the places where these actions occur

lib/Semantic_check.ml
```diff
@@ -1,2163 +0,0 @@
...
-User defined functions cannot be overloaded
-Function ending in _lp only where target is available
-TODO: Test that user defined functions with probability suffixes have right type.
-(Mutual) recursive functions have a definition
-Make sure return types of statements involving continue and break are correct.
-Make sure data only arguments to functions are checked properly.
-Sizes should be of level at most data.
-*)
-
-(* TODO: check that variables are assigned to before they are used! *)
```
seantalts: go through the files and update any TODOs or other notes

lib/Semantic_check.ml
```diff
@@ -1,2163 +0,0 @@
...
-   actually get evaluated in that phase. *)
-let update_originblock name ob =
-  match Symbol_table.look vm name with
-  | Some (old_ob, ut) ->
-      let new_ob = lub_originblock [ob; old_ob] in
-      Symbol_table.unsafe_replace vm name (new_ob, ut)
-  | _ -> fatal_error ()
-
-(* The actual semantic checks for all AST nodes! *)
-let rec semantic_check_program p =
```
seantalts: You can destructure the record above or refer to `p.functionblock`

lib/Semantic_check.ml
```diff
@@ -1,2163 +0,0 @@
...
-    in
-    let _ = context_flags.current_block <- Model in
-    let _ = Symbol_table.begin_scope vm in
-    let ubm =
-      Core_kernel.Option.map ~f:(List.map semantic_check_statement) bm
-    in
-    let _ = Symbol_table.end_scope vm in
-    let _ = context_flags.current_block <- GQuant in
-    let ubgq =
-      Core_kernel.Option.map ~f:(List.map semantic_check_statement) bgq
```
seantalts: `open Core_kernel` at top

lib/Semantic_check.ml
```diff
@@ -1,2163 +0,0 @@
...
-    in
-    { functionblock= ubf
-    ; datablock= ubd
-    ; transformeddatablock= ubtd
-    ; parametersblock= ubp
-    ; transformedparametersblock= ubtp
-    ; modelblock= ubm
-    ; generatedquantitiesblock= ubgq }
-
-(* This could also be dealt with during lexing. That would probably be more efficient. *)
```
seantalts: should we do this instead?

lib/Semantic_check.ml
```diff
@@ -1,2163 +0,0 @@
...
-  | Void -> Void
-  | ReturnType ut -> ReturnType (semantic_check_unsizedtype ut)
-
-(* Probably nothing to do here *)
-and semantic_check_unsizedtype = function
-  | Array ut -> Array (semantic_check_unsizedtype ut)
-  | Fun (l, rt) ->
-      Fun
-        ( List.map
-            (function
```
seantalts: could this be `(fun (ob, ut) -> ...`

lib/Semantic_check.ml
```diff
@@ -1,2163 +0,0 @@
...
-                ^ uid.name ^ " was supplied." )
-          | None ->
-              semantic_error ~loc
-                ( "A returning function was expected but an undeclared \
-                   identifier " ^ uid.name ^ " was supplied." ) )
-      (* TODO: Insert informative error message in case identifier is found but not with appropriate type. *)
-      )
-  | GetLP ->
-      let _ =
-        if
```
seantalts: this indentation is crazy

lib/Semantic_check.ml
```diff
@@ -1,2163 +0,0 @@
...
-      in
-      LocationScale (ue1, ue2)
-  | Ordered -> Ordered
-  | PositiveOrdered -> PositiveOrdered
-  | Simplex -> Simplex
-  | UnitVector -> UnitVector
-  | CholeskyCorr -> CholeskyCorr
-  | CholeskyCov -> CholeskyCov
-  | Correlation -> Correlation
-  | Covariance -> Covariance
```
seantalts: Isn't this all summarized as `x -> x`?

lib/Semantic_check.ml
```diff
@@ -1,2163 +0,0 @@
...
-      if rt1 <> rt2 then
-        semantic_error ~loc
-          ( "Branches of conditional need to have the same return type. \
-             Instead, found return types "
-          ^ pretty_print_returntype rt1
-          ^ " and "
-          ^ pretty_print_returntype rt2
-          ^ "." )
-      else Incomplete rt1
-  | AnyReturnType, NoReturnType
```
seantalts: Is there any better way to express this?

VMatthijs:
Ooh, this is great! Thanks for the comments, @seantalts !

lib/lexer.mll
VMatthijs: OK. Fixed this.

lib/lexer.mll
VMatthijs: Done.

lib/parser.mly
VMatthijs: Agreed. Done.

lib/lexer.mll
VMatthijs: Done.

lib/parser.mly
VMatthijs: Done.

lib/parser.mly
VMatthijs: Done.

lib/Semantic_check.ml
VMatthijs: Yes! Done.

VMatthijs: Ah. Actually, we can't, as we're going from an untyped to a typed AST (which happen to share many labels). The type checker can't figure it out if you write x -> x.

lib/Semantic_check.ml
```diff
@@ -1,2163 +0,0 @@
...
-      match get_stan_math_function_return_type_opt uid.name argumenttypes with
-      | Some Void ->
-          TypedStmt
-            ( NRFunApp (uid, ues)
-            , {stmt_typed_meta_type= NoReturnType; stmt_typed_meta_loc= loc} )
-      | Some (ReturnType _) ->
-          semantic_error ~loc
-            ( "A non-returning function was expected but a returning function "
-            ^ uid.name ^ " was supplied." )
-      | _ -> (
```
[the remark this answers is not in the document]
VMatthijs: Agreed. Done.

lib/parser.mly
VMatthijs: Sadly, no. Menhir does not let you do that.

lib/Ast.ml
VMatthijs: Done.

lib/Ast.ml
```diff
@@ -1,223 +0,0 @@
...
-(* for auto generating s-exp *)
-
-(* == Source code locations for semantic errors == *)
-type location =
-  | Location of Lexing.position * Lexing.position  (** delimited location *)
-  | Nowhere  (** no location *)
-
-(* == Unsized types == *)
-type originblock =
-  | Primitives
```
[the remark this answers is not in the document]
VMatthijs: Done.

lib/Semantic_check.ml
VMatthijs: Done.

lib/Semantic_check.ml
```diff
@@ -1,2163 +0,0 @@
...
-      let uid = semantic_check_identifier id in
-      let ues = List.map semantic_check_expression es in
-      let ut = semantic_check_truncation t in
-      let argumenttypes =
-        List.map
-          (fun z ->
-            (snd (typed_expression_unroll z)).expr_typed_meta_origin_type )
-          (ue :: ues)
-      in
-      let _ =
```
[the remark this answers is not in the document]
VMatthijs: Not sure if I can actually, unless I'd put in a lot of auxiliary definitions everywhere. It's just the trade-off of using 80 character wide lines and auto-formatting tools, I think.

lib/Semantic_check.ml
VMatthijs: It's what ocamlformat does. How would you expect to see it indented? Are there other formatting tools that you know of?

lib/Semantic_check.ml
VMatthijs: Sure. It means the same thing. I've changed it now.

lib/Semantic_check.ml
VMatthijs: I don't think so actually. I take back that comment. I don't see how to throw good errors as easily that way. If you feel like rewriting it the other way, go ahead though!

lib/Semantic_check.ml
VMatthijs: Done.

lib/parser.mly
VMatthijs: Done.

lib/parser.mly
seantalts: Boo, haha. Oh well, it does much else for us.

lib/Semantic_check.ml
seantalts: ocamlformat does have settings as well, we might try some of them. Is there no way to influence what it spits out? I've seen it sometimes respect my choices in certain places.

lib/Semantic_check.ml
```diff
@@ -1,2163 +0,0 @@
...
-            , {stmt_typed_meta_type= NoReturnType; stmt_typed_meta_loc= loc} )
-      | Some (ReturnType _) ->
-          semantic_error ~loc
-            ( "A non-returning function was expected but a returning function "
-            ^ uid.name ^ " was supplied." )
-      | _ -> (
-          let _ =
-            if is_stan_math_function_name uid.name then
-              semantic_error ~loc
-                ( "Ill-typed arguments supplied to function " ^ uid.name
```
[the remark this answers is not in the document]
seantalts: the multi-line strings help with long strings, but they won't do interpolation. Assuming they are similar enough to multi-line strings in other languages, they should be used everywhere that you need to write long strings, so this
```
("Some long" ^
    " string.")
```
becomes
```
{|Some long
 string |}
```
. The `^` concat operator is linear in the size of its arguments as it copies both to a new string. `sprintf` is probably preferred if there are a lot of variables to interpolate.

lib/Semantic_check.ml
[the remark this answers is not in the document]
seantalts: Aren't semicolons equivalent to `let _ = ` sometimes? And perhaps putting things on previous lines?

lib/Semantic_check.ml
seantalts: These transformations or constraints or whatever we call them are not shared? Could we share them?

lib/Ast.ml
VMatthijs: Done. 👍

lib/Ast.ml
VMatthijs: You may be right. I could be convinced. It would require some refactoring though.

Not sure how the write_out flag would work in code_gen. Could you figure that out? Should I already add it?

It seems to me like we only ever want to write out global though, right? So then, in a sense, we'd like the write_out flag only to occur for globals?

Also, do we need the globals flag? Isn't something with a transformation simply the same as a global var?

lib/Semantic_check.ml
VMatthijs: I've been trying to switch to this, but it's been giving me more trouble than expected. I'm getting a lot of clashes. I'm starting to wonder whether it's worth it. Will try for a bit longer to give it a chance.

lib/Ast.ml
VMatthijs: Could you make the proposal more precise? The version I can imagine would really be quite verbose, but I could well be missing something.

lib/Semantic_check.ml
VMatthijs: Could you explain what you would like to see changed here? Not sure I understand.

lib/Semantic_check.ml
VMatthijs: They are shared, but for some reason the type checker complains if you write x -> x.

lib/Ast.ml
VMatthijs: Done.

lib/Ast.ml
```diff
@@ -1,223 +0,0 @@
...
-  | TVDecl of 'e sizedtype * 'e transformation * identifier
-  | TVDeclAss of
-      { tsizedtype: 'e sizedtype
-      ; transformation: 'e transformation
-      ; tidentifier: identifier
-      ; tvalue: 'e }
-  | FunDef of
-      { returntype: returntype
-      ; funname: identifier
-      ; arguments: (originblock * unsizedtype * identifier) list
```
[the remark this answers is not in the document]
VMatthijs: Ok done.

lib/Semantic_check.ml
VMatthijs: Done.

lib/Semantic_check.ml
```diff
@@ -1,2163 +0,0 @@
...
-          Core_kernel.String.is_suffix uid.name ~suffix:"_cdf"
-          || Core_kernel.String.is_suffix uid.name ~suffix:"_ccdf"
-        then
-          semantic_error ~loc
-            ( "CDF and CCDF functions may not be used with sampling notation. \
-               Use increment_log_prob(" ^ uid.name ^ "_log(...)) instead." )
-      in
-      let _ =
-        if
-          get_stan_math_function_return_type_opt (uid.name ^ "_lpdf")
```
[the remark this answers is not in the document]
VMatthijs: Done.

lib/Ast.ml
VMatthijs: This has been fixed now.

Issue #7, opened 2018-12-12
Improve include paths backtrace


VMatthijs:
Add line and character numbers to the files we are including from, rather than just listing the chain of file names.

VMatthijs:
Note that the current syntax in stanc2 is

```
 error in 'parse-err.stan' at line 2, column 9
 included from 'stan/src/test/test-models/included/test.stan' at line 2
```

(but recursive includes are not allowed).

Issue #31, opened 2019-01-12
Create more code reuse in error reporting


VMatthijs:
Syntax and semantic errors should be able to share the same code; similarly for lexing and include errors. There should be some sharing possible with warnings as well.

VMatthijs:
This should be fixed by PR https://github.com/stan-dev/stanc3/pull/36

bob-carpenter:
Thanks for doing all the housekeeping in keeping the issues up to date.

seantalts:
Those emojis were supposed to signal my agreement and support! I'm not really sure it worked.

Issue #20, opened 2018-12-18
AST locations should support #includes


seantalts:
Right now AST locations are:

```ocaml
type location =
  | Location of Lexing.position * Lexing.position  (** delimited location *)
  | Nowhere  (** no location *)
...
type position = {
  pos_fname : string;
  pos_lnum : int;
  pos_bol : int;
  pos_cnum : int;
}
```

This doesn't seem to support the include use case where a statement comes from a file but may have been included by other files (unless I'm thinking about this wrong).

VMatthijs:
It currently has all the include information as well, but appended to the filename. I did it that way because ocamllex and menhir are very restrictive in their communication channel. (Basically, a single value of type Lexing.position.) It's a bit hacky, but would do the trick. I still plan to insert the line number and column into the include backtrace (currently, it's just a backtrace of file names). I could clean things up a bit by just writing encoding and decoding functions for storing the whole include backtrace in the filename string.

Feel free to rewrite it to something better if you have ideas, though.

If this is something you urgently need, please let me know. Otherwise, I will continue working on preparing my POPL talks for the next few days.

seantalts:
As we spoke about, we need to alter the existing stuff to produce backtraces. Even if we go with a single location string from the AST for now (to be fixed later if performance is an issue), I'm not sure why there are two Lexing.positions (almost certainly not the data structure we want to end up using).

seantalts:
This won't block anything, really - I will just assume we are using a single string and fake it for now.

seantalts:
(But if we will end up with a different data type, knowing now would save me time).

VMatthijs:
If you're happy with a single string, let's stick with that.

I opened an issue for this a while back: https://github.com/stan-dev/stanc3/issues/7

seantalts:
That works. I want to also track that we should update the type of locations to more closely match what we need when generating error messages.

VMatthijs:
After discussion, we decided that just the AST should be updated to not be hacky. For now the hack will remain in the parser.

VMatthijs:
This should be fixed by PR https://github.com/stan-dev/stanc3/pull/36

PR #32, opened 2019-01-12
Fixing include trace

VMatthijs:
This should hopefully fix issue

https://github.com/stan-dev/stanc3/issues/7

### DIFF
--- a/lib/Errors.ml
+++ b/lib/Errors.ml
@@ -1,6 +1,15 @@
 (** Setup of our compiler errors *)
 open Ast
 
+(** Insert the line and column number string in a filename string before the first
+    include, after the first filename *)
+let append_position_to_filename fname pos_string =
+  let split_fname = Str.split (Str.regexp ", included from\nfile ") fname in
+  match split_fname with
+  | [] -> ""
+  | fname1 :: fnames ->
+      String.concat ", included from\nfile " ((fname1 ^ pos_string) :: fnames)
+
 (** Our type of syntax error information *)
 type parse_error =
   | Lexing of string * Lexing.position
@@ -27,7 +36,7 @@ let position {Lexing.pos_fname; pos_lnum; pos_cnum; pos_bol} =
 let error_context file line column =
   try
     let bare_file =
-      List.hd (Str.split (Str.regexp "\" included from\n\"") file)
+      List.hd (Str.split (Str.regexp ", included from\nfile ") file)
     in
     let input = open_in bare_file in
     for _ = 1 to line - 3 do
@@ -73,9 +82,9 @@ let report_syntax_error = function
           sprintf "columns %d-%d" start_column curr_column
         else sprintf "column %d" start_column
       in
-      Printf.eprintf
-        "\nSyntax error at file \"%s\", %s, %s, parsing error:\n%!" file lines
-        columns ;
+      Printf.eprintf "\nSyntax error at file %s, parsing error:\n%!"
+        (append_position_to_filename file
+           (Printf.sprintf ", %s, %s" lines columns)) ;
       ( match error_context file curr_line curr_column with
       | None -> ()
       | Some line -> Printf.eprintf "%s\n" line ) ;
@@ -84,18 +93,18 @@ let report_syntax_error = function
       | Some error_message -> prerr_endline error_message )
   | Lexing (_, err_pos) ->
       let file, line, column = position err_pos in
-      Printf.eprintf
-        "\nSyntax error at file \"%s\", line %d, column %d, lexing error:\n"
-        file line (column - 1) ;
+      Printf.eprintf "\nSyntax error at file %s, lexing error:\n"
+        (append_position_to_filename file
+           (Printf.sprintf ", line %d, column %d" line (column - 1))) ;
       ( match error_context file line (column - 1) with
       | None -> ()
       | Some line -> Printf.eprintf "%s\n" line ) ;
       Printf.eprintf "Invalid character found. %s\n\n" ""
   | Includes (msg, err_pos) ->
       let file, line, column = position err_pos in
-      Printf.eprintf
-        "\nSyntax error at file \"%s\", line %d, column %d, includes error:\n"
-        file line column ;
+      Printf.eprintf "\nSyntax error at file %s, includes error:\n"
+        (append_position_to_filename file
+           (Printf.sprintf ", line %d, column %d" line column)) ;
       ( match error_context file line column with
       | None -> ()
       | Some line -> Printf.eprintf "%s\n" line ) ;
@@ -111,8 +120,10 @@ let print_location loc ppf =
       let begin_line = begin_pos.Lexing.pos_lnum in
       let filename = begin_pos.Lexing.pos_fname in
       if String.length filename != 0 then
-        Format.fprintf ppf "file \"%s\", line %d, columns %d-%d" filename
-          begin_line begin_char end_char
+        Format.fprintf ppf "file %s"
+          (append_position_to_filename filename
+             (Printf.sprintf ", line %d, columns %d-%d" begin_line begin_char
+                end_char))
       else
         Format.fprintf ppf "line %d, columns %d-%d" (begin_line - 1) begin_char
           end_char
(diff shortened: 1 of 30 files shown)


VMatthijs:
@seantalts , could you maybe have a look at this at some point? (no rush!) This could also be used to produce a good location string for your MIR. I can do that after I have a quick chat with you to know what you'd like it to look like.

test/integration/included/stanc.expected
```diff
@@ -145,20 +175,45 @@ Syntax error at file "sem-err.stan", line 1, columns 2-6, parsing error:
...
+  y ~ normal(0, 1);
+}
+
+
+Warning: deprecated language construct used at file ../included/dep-warning.stan, line 2, column 2, included from
+file ../included/incl_stanc_helper_deprecated_warning.stan, line 2, column 27, included from
+file stanc_helper_with_bad_include_deprecated_warning.stan, line 2, column 50:
+   -------------------------------------------------
+     1:    real z;
+     2:    # I am a deprecated comment.
```
seantalts: Woo!

lib/Errors.ml
```diff
@@ -145,11 +156,9 @@ let fatal_error ?(msg = "") _ =
 (* Warn that a language feature is deprecated *)
 let warn_deprecated (pos, msg) =
   let file, line, column = position pos in
-  Printf.eprintf
-    "\n\
-     Warning: deprecated language construct used at file \"%s\", line %d, \
-     column %d:\n"
-    file line (column - 1) ;
+  Printf.eprintf "\nWarning: deprecated language construct used at file %s:\n"
+    (append_position_to_filename file
```
seantalts: Seeing a lot of these warning calls, maybe abstract to a function that emits a warning, file name + line nums, and text at the location?

VMatthijs: Do you mean to achieve more code sharing between the different kinds of errors and the warnings? They're all a tiny bit different, but I agree there probably could be more code sharing there.

VMatthijs:
1. If you have a look at the error messages on example models (e.g. in the included folder), then you'll see that you get a full include backtrace on each error or warning. It is still done by storing the include backtrace in the filename, currently.
2. I agree that that would be nicer. I don't know how to pass that state through Menhir, though. What could be done more easily is to convert into one of these data structures after parsing when we construct the AST and then to convert back to a string when we throw an error. That would mean though that the lexing and parsing errors take differently structured location data than the semantic errors. My sense is that I'd rather work on other things at the moment, mostly because I just don't know how to do this at the moment, so it would cost me a lot of time and I'm not sure if the payoff justifies that investment. If you know how to do it in not very much time, please go ahead! I totally agree that it would be the better design.

lib/Errors.ml
seantalts: It's related. This is just one specific area where we need to DRY up the code.

seantalts:
I'm not sure I understand your answer to how this adds the information to the AST. Is it stored in the `Location` variant where you've re-purposed the Lexing.position type such that one of its string records now stores additional information?

"pass that state through Menhir" - I think I see what you mean - since the includes are just a rule that replaces the lexbuf with another lexbuf pointing to the new file (and on EOF it just pops one off the global stack). Why not use that global stack (or one like it) to add the position information in the parsing stage?

lib/Errors.ml
VMatthijs: What does it my to DRY up the code? Is that an acronym?

VMatthijs:
> I'm not sure I understand your answer to how this adds the information to the AST. Is it stored in the `Location` variant where you've re-purposed the Lexing.position type such that one of its string records now stores additional information?

Yes, the filename together with the include backtrace are stored in the Lexing.position.pos_fname field as a single string. 

> "pass that state through Menhir" - I think I see what you mean - since the includes are just a rule that replaces the lexbuf with another lexbuf pointing to the new file (and on EOF it just pops one off the global stack). Why not use that global stack (or one like it) to add the position information in the parsing stage?
If you know how to do that, please go for it! The only way I know how to get position information out of Menhir for a syntax error or for building the AST is by using `$startpos` and `$endpos`, which gives you the start and end position of a rule (they keep getting updated by Menhir). Of course, you can also manage your own include trace state separately by modifying it in every production, by hand (in the braces after each rule), but that'd have to happen in every single rule and could be ugly and I still wouldn't be 100% sure how to do it: you'd want to get the appropriate include trace for every token and somehow make the include trace for the first and last token a rule matches available.

lib/Errors.ml
seantalts: yep - "Don't Repeat Yourself." http://wiki.c2.com/?DontRepeatYourself. People usually moderate that with something like "write everything (twice | thrice | ? ) before abstracting". So it's a balance.

seantalts:
I'm pretty sure that even if we end up not being able to find a way to capture structured location information that we do not want to expose as the AST's permanent location API a hacked Lexing.position :P So it's probably best if we go ahead and do as you suggested and parse out the structured data from the string before inserting it into the AST as something structured - I'm imagining a list/stack of something similar to (but easier to use than) a Lexing.position.

Re: global stack - let me try to understand more. The problem here is that in the parser.mly, you only by default get access to the Lexer's Lexing.position automatically. Here is my question: From the parser, can you not access `lexer.include_stack`?

bob-carpenter:
I like this idea, too.   No need to expose and carry around parser/lexer
internals if a simple stack will suffice (it should given the way includes
work).  It does add some conversion code, though.

VMatthijs:
> I'm pretty sure that even if we end up not being able to find a way to capture structured location information that we do not want to expose as the AST's permanent location API a hacked Lexing.position :P So it's probably best if we go ahead and do as you suggested and parse out the structured data from the string before inserting it into the AST as something structured - I'm imagining a list/stack of something similar to (but easier to use than) a Lexing.position.
> 
> Re: global stack - let me try to understand more. The problem here is that in the parser.mly, you only by default get access to the Lexer's Lexing.position automatically. Here is my question: From the parser, can you not access `lexer.include_stack`?

Sure, if you rewrite the code a bit (i.e. move the include structure to its own file or to parser to avoid a circular dependency between the parser and lexer), you can access `include_stack`, but you'll have to manage it manually if you want to access the include stack of both the start and end position of what a rule matches on. It is easy to just get the include stack at the end of the rule, but note that the start and end might be included from different files. Currently, I'm capturing both the start and end include stack. If we only cared about the include stack at the end of the rule I would've done it this way from the start.

VMatthijs:
@seantalts , do you agree if I merge this and close issue https://github.com/stan-dev/stanc3/issues/7.

Issues https://github.com/stan-dev/stanc3/issues/20 and https://github.com/stan-dev/stanc3/issues/31 will then remain open.

lib/Errors.ml
VMatthijs: Cool! I agree that the error messages could be condensed further. Hope to address it in a separate issue: https://github.com/stan-dev/stanc3/issues/31 .

seantalts:
> Sure, if you rewrite the code a bit (i.e. move the include structure to its own file or to parser to avoid a circular dependency between the parser and lexer), 

This seems fine. Is the lexer dependency on the parser just on the tokens? I've see those separated out into their own files pretty often and wasn't sure why - maybe this answers that question.

> you'll have to manage it manually in some clever way while parsing if you want to access the include stack of both the start and end position of what a rule matches on. It is easy to just get the include stack at the end of the rule, but note that the start and end might be included from different files. Currently, I'm capturing both the start and end include stack. If we only cared about the include stack at the end of the rule I would've done it this way from the start.

Good point. Is this important just to print error messages indicating the entire cross-file span correctly? Do you have any example tests illustrating this I could look at?

> Getting the beginning and end include stacks into the AST as a stack would be easy. However, that would still leave us with the issue that this hack remains in the parser and lexer and their error messages.

For me this is the minimum for fixing #20 before launch. The way to think about this is that the AST and MIR data structures are not just the API for huge chunks of the code base, but also for other users who might like to develop tools on top of Stan or contribute additional tools (linters, optimizers, etc). For example, the IBM Research team behind YAPS would like to start using our ocaml parser already likely via the S-expression output and stop maintaining their own ANTLR parser. Generally, even if you hack something together in an ugly way and you decide you want to launch with it, you want to wrap that hack up in a nice interface/API that hides those details.

seantalts:
Re: Merging this - I'm not sure this code is code we will use after fixing #20, but I will merge as it's an improvement and sets the correct standard for include messages.

PR #33, opened 2019-01-23
New special statements in the MIR; pull out top var decls in MIR blocks

seantalts:
It seems like a good idea to represent some of the lower level code that we currently directly generate in the MIR so that we can reap the benefit of our (future) generalized optimization passes. Right now it seems to make sense to add the following new statement nodes to the MIR:
* Checks (e.g.`check_positive_ordered` et al calls; right now just generated from data declarations right after reading in the variables, but we've already talked a bit about pulling those checks up to a higher level in the Stan MIR from their current place in the Math functions so we avoid checking things too often)
* ZeroInit (maybe should just be Zero or SZero or ZeroOut or something? not just using normal assignment because the right-hand side depends on the type and there may be more efficient methods to call to set everything to zero depending on the type and backend). This call can be totally elided if we know that we don't use the variable until it's (fully) assigned to.
* MarkLocation - here I'm thinking we might use this to take the location out of the metadata and allow the optimization pass to operate on this as well, since we're talking about making it much less efficient by using basically full error location strings for this instead of integers. 

The last two came from discussions with @mitzimorris today about existing inefficiencies in the generated code as well as some previous work she'd done to remove some inefficiencies like these. 

Pulling out the topvardecls then seemed like a good idea because we can pull them out and then generate MIR statements that are paired with these declarations. It was also necessary because we do very different things for the same AST VarDecl in different blocks, so either the function that translates AST statements to MIR statements (or the one that emits code) needs to know what block its currently operating in, or we need to deal with top var declarations separately. A side effect is that we can simplify the two structures since we don't need a single structure to represent both (I think this is going back on something Matthijs and I talked about with the AST early on where he had different types for local and global variables and I asked him to unify them, woops).


seantalts:
Going to merge this for now - hooking up the code generator tests now even though code generation isn't finished just so we can track progress! :) 

Please still feel free to review.

PR #35, opened 2019-01-28
context_flags immutable with functional updates

bob-carpenter:
* Remove global variable `context_flags` from `Semantic_check.ml`

* Plumb immutable `context_flags` through as argument

* Use [functional updates](http://dev.realworldocaml.org/records.html#functional-updates) in place of mutability

No behavior changes, but this touches a lot of lines of code.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,7 +7,7 @@ pipeline {
     agent none
     stages {
         stage("Build & Test") {
-            agent {
+            agent { 
                 dockerfile {
                     filename 'docker/dev-ubuntu/Dockerfile'
                     args "-u root --privileged --entrypoint=''" // TODO: set up a proper user in Dockerfile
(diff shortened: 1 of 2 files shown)


VMatthijs:
Looks great! I'm so happy you're able to work on this codebase, @bob-carpenter ! Will merge.

seantalts:
This is kind of silly, but @bob-carpenter what do you think about using `cf` instead of `context_flags` given that it's used so often throughout? The benefit would be reduction in the vertical space used in a few of these places (I think [these](https://github.com/stan-dev/stanc3/pull/35/files#diff-3b64248472843a9cb69a14706dd34024R268) and [these](https://github.com/stan-dev/stanc3/pull/35/files#diff-3b64248472843a9cb69a14706dd34024R1396)).

I don't feel strongly either way, just curious since we have done lots of these variable name shorternings in the rest of the code.

PR #37, opened 2019-01-29
Moving tests to bottom of file

VMatthijs:
@seantalts , could you have a look at the unit tests here? Is there anything else you think should be tested?

I've now moved the tests to the bottom of the file, as a compromise between your style and Bob's. Keeping them in the same file has the advantage that we can test internals that we hide behind the interface.

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -3,10 +3,13 @@ open Core_kernel
 
 (** Source code locations *)
 type location =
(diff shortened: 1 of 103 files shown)


VMatthijs:
To be clear: this can definitely wait until you're better.

VMatthijs:
Ah, I did work in the wrong branch. Will try again in a new PR.

Issue #16, opened 2018-12-14
Don't elide offset and multiplier of 0 and 1 when pretty-printing


seantalts:

seantalts:
I think maybe the right phrasing of this issue is that we might want the AST to be able to represent a partially specified offset/multiplier. So it's easy to not generate code for a default-value offset or multiplier.

VMatthijs:
This should now be fixed by PR https://github.com/stan-dev/stanc3/pull/38

PR #38, opened 2019-02-01
Splitting offset multiplier nodes

VMatthijs:
This PR addresses issue https://github.com/stan-dev/stanc3/issues/16 .

It has the following side effects:
- slightly more precise error messages for bounds and offset-multiplier;
- made the `originblock` type a local data structure to the semantic check, rather than exposing it in the AST;
- slight refactoring and more code reuse in semantic check.


lib/Ast.ml
```diff
@@ -11,19 +11,8 @@ type location =
 (** Delimited locations *)
 type location_span = {begin_loc: location; end_loc: location}
 
-(** Origin blocks, to keep track of where variables are declared *)
```
bob-carpenter: There's lingering doc on line 94 that refers to origin blocks, but there are no more origin blocks.  This is, by the way, what I meant about it being hard to keep the doc up to date in these cases and why I prefer not to have doc that repeats what the code's doing.

lib/Pretty_printing.ml
```diff
@@ -171,6 +161,8 @@ and pretty_print_transformation = function
   | LowerUpper (e1, e2) ->
       "<lower=" ^ pretty_print_expression e1 ^ ", upper="
       ^ pretty_print_expression e2 ^ ">"
+  | Offset e -> "<offset=" ^ pretty_print_expression e ^ ">"
```
bob-carpenter: [question]
Is this going to get refactored to use the `Format` library?  I'd think our pretty printing is more important for readability than our C++ code generation (for things like line breaking).

lib/Pretty_printing.ml
```diff
@@ -213,7 +205,8 @@ and pretty_print_transformed_type st trans =
   in
   match trans with
   | Identity -> pretty_print_sizedtype st
-  | Lower _ | Upper _ | LowerUpper _ | OffsetMultiplier _ ->
+  | Lower _ | Upper _ | LowerUpper _ | Offset _ | Multiplier _
+   |OffsetMultiplier _ ->
```
bob-carpenter: Space-vbar should be transposed so it lines up and is `| OffsetMultiplier`

lib/Semantic_check.ml
```diff
@@ -19,6 +19,17 @@ open Pretty_printing
    on (1) a global symbol table vm, and (2) structure of type context_flags_record
    to communicate information down the AST. *)
 
+(** Origin blocks, to keep track of where variables are declared *)
+type originblock =
```
bob-carpenter: I see `originblock` moved to here, but the thing with doc in the other file commented above didn't seem to have an origin block in it.

lib/Ast.ml
VMatthijs: Right. That makes sense. I'll remove it.

lib/Pretty_printing.ml
VMatthijs: Some day, if anyone is brave enough / feels like it. I agree it should be refactored.

lib/Pretty_printing.ml
VMatthijs: Sadly, this is what OCamlformat makes of it.

lib/Semantic_check.ml
VMatthijs: Sorry. That comment was outdated.

VMatthijs:
Thanks for reviewing, @bob-carpenter !

bob-carpenter:
Should we file a bug report with OCamlformat?  This can't be their intended output, can it?  

Either way, no big deal for this PR if the autoformatter takes it out of our control.

seantalts:
I think that's intentional from the formatter - the weirdly indented line is actually a continuation of the line above it, which doesn't have an action specified yet.

VMatthijs:
Could someone maybe reapprove so this can be merged?

VMatthijs:
Thanks!

PR #39, opened 2019-02-01
DO NOT MERGE - Parser unit test review

VMatthijs:
This PR serves to have an interface to discuss the unit tests for the parser.

Currently, I have only written unit tests for cases where the BNF grammar is ambiguous and I wanted to make sure the correct parse tree was constructed using the precedence encoded in the Menhir file, using associativity/precedence rules as well as inlining. The rest seemed to be covered by the test model suite.

@seantalts , could you have a look to see if they suffice to complement the end-to-end test models? Are there any other examples of phrases you'd like to see a unit test for?


seantalts:
Niceee, I love the github PR hack. Thanks :)

lib/Parse.ml
```diff
@@ -79,455 +79,3 @@ let parse_file parse_fun path =
...
-                    ((expr_untyped (IntNumeral 2)) (expr_untyped_loc <opaque>))))))
-                (expr_untyped_loc <opaque>))))))
-           (stmt_untyped_loc <opaque>)))))
-       (generatedquantitiesblock ())) |}]
-
-let%expect_test "parse operator precedence" =
-  let ast =
-    parse_string Parser.Incremental.program
-      "model {  \
-       print({a,b?c:d||e&&f==g!=h<=i<j>=k>l+m-n*o/p%q.*s./t\\r^u[v]'}); }"
```
seantalts: Whoa. Did you manually verify the output below? This may or may not be better if broken up into smaller unit tests, especially since there is a pairwise element to precedence. But not too sure about that.

lib/Parse.ml
```diff
@@ -79,455 +79,3 @@ let parse_file parse_fun path =
...
   in
   parse parse_fun lexbuf
-
-(* TESTS *)
-let%expect_test "parse conditional" =
-  let ast =
-    parse_string Parser.Incremental.program
-      "model { if (1 < 2) { print(\"hi\");}}"
-  in
-  print_s [%sexp (ast : Ast.untyped_program)] ;
```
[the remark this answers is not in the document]
VMatthijs: agreed. will change.

lib/Parse.ml
```diff
@@ -79,455 +79,3 @@ let parse_file parse_fun path =
...
-                 (expr_untyped_loc <opaque>))))
-              (expr_untyped_loc <opaque>))))
-           (is_global false)))
-         (stmt_untyped_loc <opaque>)))))
-     (generatedquantitiesblock ())) |}]
-
-let%expect_test "parse indices, two different colons" =
-  let ast =
-    parse_string Parser.Incremental.program
-      "model { matrix[5, 5] x; print(x[2 - 3 ? 3 : 4 : 2]); }"
```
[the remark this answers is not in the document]
VMatthijs: Will change.

lib/Parse.ml
VMatthijs: Yes. I manually verified it.

Honestly, I wasn't entirely sure how best to write it. I guess you could break it up into groups of three variables (i.e. two operations) if you trust that that determines the parse tree of larger expressions. It wasn't 100% clear to me that would suffice or what precisely would suffice as some of these operations associate left and others right (mostly left). How could we convince ourselves that we've found the right correctness criterion here?

seantalts: Nice. I was mostly just commenting on the sort of ergonomics of the test more than its completeness, though that's also a good concern. My answers to that are very unsatisfying, usually something like "if you run tests on every program you care about you'll probably be ok." I'm hoping @bob-carpenter or @mitzimorris can weigh in here with something more ergonomic and/or complete.

Actually, for completeness, I think I have an idea. We should use quickcheck (or the idea behind it) to generate possible subtrees of this form, pretty print them, and then parse them again and check equality. That doesn't get you any illegal parses, but that's not what this test is looking for anyway right?

bob-carpenter:
> Nice. I was mostly just commenting on the sort of ergonomics of the test more than its completeness, though that's also a good concern. My answers to that are very unsatisfying, usually something like "if you run tests on every program you care about you'll probably be ok." I'm hoping @bob-carpenter or @mitzimorris can weigh in here with something more ergonomic and/or complete.

We never managed good non-trivial unit testing for either
AST generation or code generation from the AST.

In some cases, we check that certain expressions or statements
generate certain code, but there's not much of that kind of
testing.

Also, very little end-to-end testing that you get the right
density, etc., in models.  I think this kind of end-to-end
testing might allay a lot of of our fears due to lack of unit
testing coverage.


> Actually, for completeness, I think I have an idea. We should use quickcheck (or the idea behind it) to generate possible subtrees of this form, pretty print them, and then parse them again and check equality. That doesn't get you any illegal parses, but that's not what this test is looking for anyway right?

That's already being done with the top-level pretty printer, I think.

seantalts:
> That's already being done with the top-level pretty printer, I think.
I don't think there's any generation of test cases happening. I'm actually not sure what tests are running on pretty printing other than that a bunch of Stan models pretty print to something that doesn't change and looks good (via expect tests).

VMatthijs:
We've got
`
val pretty_print_program : Ast.untyped_program -> string
`
and
`
val parse_program : string -> Ast.untyped_program 
`.

What I've done is verified *once* on all of our test models `str` that `(parse_program (pretty_print_program (parse_program str)) = (parse_program str)`. Since, the changes to the pretty printer are being tracked by the expect tests.

What Sean seems to be advocating here is that we randomly generate `Ast.untyped_program`s `ast` (satisfying the precedence constraints presumably) and then verify that `parse_program (pretty_print_program ast) = ast`. That is a bit different. It would require a function to randomly generate valid `Ast.untyped_program`s.

VMatthijs:
@seantalts , shall I close this PR for now?

seantalts:
yep

PR #44, opened 2019-02-04
Removing monadic notation

VMatthijs:
I changed the monadic notation in the semantic check to plain `Option.map`, as I think that's easier to Google and more self-explanatory than the infix operator `>>|`.


seantalts:
Looks good! I think this is probably for the best.

VMatthijs:
Thanks!

bob-carpenter:
I'm happy to review, but not between 7 am and 9 am :-)

PR #45, opened 2019-02-04
Simplifying semantic check

VMatthijs:
Slight refactoring to simplify semantic check.

I also added some extra test models.

### DIFF
--- a/lib/Semantic_check.ml
+++ b/lib/Semantic_check.ml
@@ -107,40 +107,27 @@ let probability_distribution_name_variants id =
       [name; drop_suffix name 4 ^ "_lpmf"; drop_suffix name 4 ^ "_lpdf"]
     else [name] )
 
+let lub_rt loc rt1 rt2 =
+  match (rt1, rt2) with
+  | ReturnType UReal, ReturnType UInt | ReturnType UInt, ReturnType UReal ->
+      ReturnType UReal
+  | rt1, rt2 when rt1 = rt2 -> rt2
+  | _ ->
+      semantic_error ~loc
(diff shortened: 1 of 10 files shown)


lib/Semantic_check.ml
```diff
@@ -107,40 +107,27 @@ let probability_distribution_name_variants id =
...
     else [name] )
 
+let lub_rt loc rt1 rt2 =
+  match (rt1, rt2) with
+  | ReturnType UReal, ReturnType UInt | ReturnType UInt, ReturnType UReal ->
+      ReturnType UReal
+  | rt1, rt2 when rt1 = rt2 -> rt2
+  | _ ->
+      semantic_error ~loc
+        ( "Branches of function definition need to have the same return type. \
```
seantalts: This function is used for conditionals as well as function definitions. A fix might be as simple as including a string parameter to `lub_rt` to put in the error message.

VMatthijs: Ultimately, I think every use case comes down to having a function definition with multiple branches in it because of the use of conditionals and having different return types in the different branches.

I guess there's the distinction between
```
if (1)
  return 1;
real x = 24;
return {1};
```
and
```
if (1)
  return 1;
else {
  real x = 24;
  return {1};
}
```

seantalts: Oh good point. Never mind!

VMatthijs:
Thanks, @seantalts ! I feel like it could probably still be condensed further, but I couldn't quite see how to do it right now. Gradually, this code base will shrink a bit, hopefully.

PR #48, opened 2019-02-04
Setting up stanc.ml to output generated code

seantalts:
I haven't turned on the expect tests in this PR, but we're about ready to. I was kind of torn because right now they're mostly garbage, but it would be nice to keep track of how the garbage is evolving, and what some of our design decisions will look like (like the "locations are strings" is kinda funny lookin). 

I would like to merge this in before finishing because it has a refactor of `stanc.ml` (and I probably should have just done that separately but it came about while I was in there to add the generated code output and I didn't think I'd be modifying as much as I did).


dune
```diff
@@ -7,4 +7,8 @@
 
 (env
  (static
-  (flags (-ccopt -static))))
+  (flags (-ccopt -static)))
+  (dev
+  (flags (:standard -g)))
+ (release
+  (flags (:standard -O3))))
```
seantalts: Not super high confidence on these

seantalts:
Could one of you two (@VMatthijs @bob-carpenter) take a look at this? It's mostly changes to stanc.ml to hook up code generation.

VMatthijs:
Will have a look at this after breakfast. Sorry this took a while! I thought Bob might want to do it but I think he's been held up in meetings.

dune
VMatthijs: The -O3 flag seems to only work for the compilers with Flambda optimizers. See https://github.com/ocaml/dune/issues/1427 . In particular, it crashes on my machine as I do not have Flambda installed. You could change it to `(ocamlopt_flags (:standard -O3))` or we could instruct people to install Flambda. Not even sure if the compiler needs much optimization as it seems to run fast enough as is. Then again, if it doens't make the compilation take much longer, it might be worth it. Have you looked at compile and runtimes at all, when using -O3?

lib/Ast_to_Mir.ml
```diff
@@ -196,18 +196,6 @@ let rec trans_checks cvarname ctype t =
...
-let tvdecl_checks {tvident; tvtrans; tvtype; tvloc} =
-  let with_sloc stmt = {sloc= tvloc; stmt} in
-  let check_stmts =
-    List.map ~f:with_sloc (trans_checks tvident tvtype tvtrans)
-  in
-  with_sloc (SList check_stmts)
-
 let trans_tvdecl {Ast.stmt_typed; stmt_typed_loc; _} =
   match stmt_typed with
   | Ast.VarDecl
```
VMatthijs: Presumably, we will throw an error later if we find a `None` for the `Mir.tvdecl option`?

lib/Ast_to_Mir.ml
```diff
@@ -251,6 +251,7 @@ let pull_tvdecls = function
    * model -> no constraints allowed (not even corr_matrix et al); not visible
    * gq -> declared and checked in write_array, shows up as param in some methods
 
+   Local vs. Global vardecls
    So there are "local" (i.e. not top-level; not read in or written out anywhere) variable
    declarations that do not allow transformations. These are the only kind allowed in
    the model block. There are also then top-level ones, which are the only thing you can
```
VMatthijs: Basically, I think the rule is that variable declarations are always global unless these occur inside a block statement or in the model-block. You can check on a `Ast.VarDecl` whether it is global by inspecting the `is_global` flag.

lib/Ast_to_Mir.ml
```diff
@@ -294,7 +295,7 @@ let trans_prog filename
     let merge_maps maps =
       List.map ~f:Map.Poly.to_alist maps
       |> List.concat |> Map.Poly.of_alist_exn
-    and tvtables, stmts = List.map ~f:pull_tvdecls blocks |> List.unzip in
```
VMatthijs: I quite like this style of left-to-right composition. I also tried Googling the `|>` operator and it's easy to find. I hope it won't confuse anyone else though.

lib/Debugging.ml
```diff
@@ -25,10 +25,3 @@ let typed_ast_to_string x =
 
 let typed_ast_logger t =
   if !typed_ast_printing then print_endline (typed_ast_to_string t)
-
-(* Controls whether an auto-formatted program will be pretty printed *)
-let pretty_print_program = ref false
-
-let auto_formatter t =
-  if !pretty_print_program then
-    print_endline (Pretty_printing.pretty_print_program t)
```
VMatthijs: +1

lib/Mir.ml
```diff
@@ -93,7 +93,7 @@ type 's prog =
   ; prog_path: string }
 [@@deriving sexp]
 
-type stmt_loc = {sloc: string; stmt: stmt_loc statement}
+type stmt_loc = {sloc: string sexp_opaque; stmt: stmt_loc statement}
```
VMatthijs: Realistically, we'll also want to derive a `compare` operation on MIRs for the purpose of optimizations. In that case, you also want to add `[@compare.ignore]` to the location fields, I would guess.

lib/Stan_math_backend.ml
```diff
@@ -163,19 +157,16 @@ let emit_error_wrapper ppf (emit_err, err_arg, emit_contents, contents_arg) =
...
-let emit_runtime_exception ppf =
-  fprintf ppf "std::runtime_error(std::string(%s) + e.what())"
+let emit_located_msg ppf msg =
+  fprintf ppf
+    {|stan::lang::rethrow_located(
+    std::runtime_error(std::string(%s) + e.what(), current_statement__);
+// Next line prevents compiler griping about no return
+throw std::runtime_error("*** IF YOU SEE THIS, PLEASE REPORT A BUG ***");
+|}
+  @@ Option.value ~default:"e" msg
```
VMatthijs: What's going on with the formatting here?

stanc.ml
```diff
@@ -9,30 +9,35 @@ let version = "stanc version 3.0 alpha"
...
+let pretty_print_program = ref false
+let dump_mir = ref false
+let output_file = ref ""
 
 (** Some example command-line options here *)
 let options =
   Arg.align
     [ ( "--debug-lex"
-      , Arg.Unit (fun () -> Debugging.lexer_logging := true)
+      , Arg.Set Debugging.lexer_logging
```
VMatthijs: Cool! I didn't know these were already defined in the `Arg` module.

stanc.ml
```diff
@@ -64,50 +64,49 @@ let options =
...
       exit 1
   in
   let _ = Debugging.typed_ast_logger typed_ast in
-  let mir = Ast_to_Mir.trans_prog filename typed_ast in
-  (*let _ = Stan_math_backend.emit_prog Format.str_formatter mir in
-    let cpp = Format.flush_str_formatter () in
-    print_string cpp
-  *)
-  ignore mir
+  if not !pretty_print_program then (
```
VMatthijs: Hmm. I think @bob-carpenter expressed a preference for all command line arguments to be independent/orthogonal.
That is, he suggested we introduce an extra command line argument for suppressing the code gen for in case we only want to pretty print and not generate code. That way, we can also pretty print and generate code at the same time.
I don't have any strong feelings about this, myself, however.

test/integration/examples-good/dune
```diff
@@ -18,4 +18,4 @@
 ;;
 ;; (alias
 ;;  (name runtest)
-;;  (action (diff cpp.expected cpp.output)))
\ No newline at end of file
+;;  (action (diff cpp.expected cpp.output)))
```
VMatthijs: Am I right in thinking that this doesn't do anything yet? As far as I could tell the cpp output is not yet being tracked, correct?

dune
seantalts: No. I just copied this from some tutorial on ocamldebug, haha. I can just remove all the flags from the release build.

lib/Ast_to_Mir.ml
seantalts: It's just used to filter below: 
```
 List.(filter_opt (map ~f:trans_tvdecl lst))
```

lib/Ast_to_Mir.ml
seantalts: Yep, that's right. I can add a note about the fact that if they're nested in a Block they are also local.

lib/Ast_to_Mir.ml
VMatthijs: That's fine.

lib/Stan_math_backend.ml
seantalts: Not sure. Do you want me to move line 167 up to after the end of the multi-line string on line 166? That might pass the formatter, I forget. Just in case - the @@ is essentially `$` in Haskell - groups everything to its right as if we'd put parens around it.

dune
VMatthijs: Sounds good! If we feel like we need some optimization, we can add them back in and throw in a dependency on Flambda.

stanc.ml
seantalts: Ah, I'll have to talk about it with him at some point. I think it makes more sense to think of the compiler as having different actions associated with it. It's unlikely someone wants to generate C++ and pretty print, especially if both are going to stdout or something.

test/integration/examples-good/dune
seantalts: Yes that's right - `;;` is line comment.

lib/Mir.ml
seantalts: Woops, didn't see this in time. Will add to the next PR.

PR #51, opened 2019-02-05
Switch to Fmt library (ready now)

### LINKED ISSUE #46: Investigate switching from Format to `fmt`
Based on recommendations from the ocaml community [here](https://discuss.ocaml.org/t/debugging-memory-issues/3223/9) and [here](https://discuss.ocaml.org/t/format-module-from-the-standard-library/2254/6?u=sean). It's supposed to have a much nicer, more ergonomic API and I have personally experienced a bunch of pain with the Format API so far, so it seems directionally appropriate so far.

https://erratique.ch/software/fmt/doc/Fmt.html

seantalts:
Fixes #46

### DIFF
--- a/docker/static/Dockerfile
+++ b/docker/static/Dockerfile
@@ -17,8 +17,8 @@ RUN opam switch create 4.07 ocaml-base-compiler.4.07.1 && \
   opam switch 4.07
 
 # Stanc3 also needs
-COPY ./scripts/install_dev_deps.sh ./
-RUN bash -x install_dev_deps.sh
+COPY ./scripts/install_build_deps.sh ./
+RUN bash -x install_build_deps.sh
 
 
 ENTRYPOINT [ "opam", "config", "exec", "--" ]
--- a/docker/static/Dockerfile
+++ b/docker/static/Dockerfile
@@ -17,6 +17,8 @@ RUN opam switch create 4.07 ocaml-base-compiler.4.07.1 && \
   opam switch 4.07
 
 # Stanc3 also needs
-RUN opam install -y menhir core_kernel ppx_deriving
+COPY ./scripts/install_dev_deps.sh ./
+RUN bash -x install_dev_deps.sh
+
 
 ENTRYPOINT [ "opam", "config", "exec", "--" ]
--- a/lib/Stan_math_backend.ml
+++ b/lib/Stan_math_backend.ml
@@ -120,9 +120,7 @@ let pp_arg ppf ((adtype, name, st), idx) =
   pf ppf "const %a& %s" (pp_unsizedtype adstr) st name
 
 let pp_template_decls ppf ts =
-  pf ppf "@[<hov>template <%a>@]@ "
-    (list ~sep:comma (fun ppf t -> pf ppf "typename %s" t))
-    ts
+  pf ppf "@[<hov>template <%a>@]@ " (list ~sep:comma (fmt "typename %s")) ts
 
 let with_idx lst = List.(zip_exn lst (range 0 (length lst)))
 
--- a/lib/Stan_math_backend.ml
+++ b/lib/Stan_math_backend.ml
@@ -6,57 +6,55 @@ open Fmt
 
 let zero = Lit (Int, "0")
 
-let rec emit_unsizedtype ad ppf = function
+let rec pp_unsizedtype ad ppf = function
   | Ast.UInt | UReal -> string ppf ad
-  | UArray t -> pf ppf "std::vector<%a>" (emit_unsizedtype ad) t
+  | UArray t -> pf ppf "std::vector<%a>" (pp_unsizedtype ad) t
   | UMatrix -> pf ppf "Eigen::Matrix<%s, -1, -1>" ad
   | URowVector -> pf ppf "Eigen::Matrix<%s, 1, -1>" ad
   | UVector -> pf ppf "Eigen::Matrix<%s, -1, 1>" ad
   | x -> raise_s [%message (x : unsizedtype) "not implemented yet"]
 
 let%expect_test "emit function type raises" =
-  ( match emit_unsizedtype "sars" stdout Ast.UMathLibraryFunction with
+  ( match pp_unsizedtype "sars" stdout Ast.UMathLibraryFunction with
   | () -> ()
   | exception e -> print_s [%sexp (e : exn)] ) ;
   [%expect {| ((x UMathLibraryFunction) "not implemented yet") |}]
 
-let emit_call ppf (name, emit_args, args) =
-  pf ppf "%s(@[<hov>%a@])" name emit_args args
+let pp_call ppf (name, pp_args, args) =
+  pf ppf "%s(@[<hov>%a@])" name pp_args args
 
-let rec emit_expr ppf s =
+let rec pp_expr ppf s =
   match s with
   | Var s -> string ppf s
   | Lit (Str, s) -> pf ppf "\"%s\"" s
   | Lit (_, s) -> string ppf s
-  | FnApp (fname, args) ->
-      emit_call ppf (fname, list ~sep:comma emit_expr, args)
+  | FnApp (fname, args) -> pp_call ppf (fname, list ~sep:comma pp_expr, args)
   | BinOp (e1, op, e2) ->
-      pf ppf "%a %s %a" emit_expr e1 (Operators.operator_name op) emit_expr e2
+      pf ppf "%a %s %a" pp_expr e1 (Operators.operator_name op) pp_expr e2
   | TernaryIf (cond, ifb, elseb) ->
-      pf ppf "(%a) ? (%a) : (%a)" emit_expr cond emit_expr ifb emit_expr elseb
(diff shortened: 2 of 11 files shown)


VMatthijs:
Happy to review this once the other one is merged.

docker/static/Dockerfile
```diff
@@ -17,6 +17,8 @@ RUN opam switch create 4.07 ocaml-base-compiler.4.07.1 && \
   opam switch 4.07
 
 # Stanc3 also needs
-RUN opam install -y menhir core_kernel ppx_deriving
+COPY ./scripts/install_build_deps.sh ./
+RUN bash -x install_build_deps.sh
```
seantalts: This caught me - I had added the new dependency on fmt to the build script but not to this Dockerfile, so I figured it was best to keep the authoritative list in one place.

scripts/install_dev_deps.sh
```diff
@@ -1,5 +1,3 @@
 #!/bin/bash
 # Merlin, utop, ocp-indent, ocamlformat, and patdiff are all for developer assistance
-# Menhir is our parsing library and annoyingly its module name does not match
-# its library name, so we install it manually here.
-opam install -y core_kernel menhir ppx_deriving merlin utop ocp-indent patdiff ocamlformat
+opam install -y merlin utop ocp-indent patdiff ocamlformat
```
seantalts: I separated the deps for developers from the deps needed to build because patdiff has some dependency that wasn't building properly in the static environment.

docker/static/Dockerfile
VMatthijs: Good idea!

lib/Stan_math_backend.ml
```diff
@@ -2,77 +2,59 @@
...
-  | Some x -> fprintf ppf "%a" emitter x
-  | None -> emit_str ppf default
-
-let rec emit_unsizedtype ad ppf = function
-  | Ast.UInt | UReal -> emit_str ppf ad
-  | UArray t -> fprintf ppf "std::vector<%a>" (emit_unsizedtype ad) t
-  | UMatrix -> fprintf ppf "Eigen::Matrix<%s, -1, -1>" ad
-  | URowVector -> fprintf ppf "Eigen::Matrix<%s, 1, -1>" ad
-  | UVector -> fprintf ppf "Eigen::Matrix<%s, -1, 1>" ad
+let rec pp_unsizedtype ad ppf = function
```
seantalts: The convention in Fmt is to create "pretty printers" for your type named `pp_<typename>`: https://erratique.ch/software/fmt/doc/Fmt.html#nameconv

VMatthijs: Not sure how I feel about the naming `pp_`. To me, that sounds like pretty printing. How about `gen_` or `cpp_`?

seantalts:
I think @bob-carpenter might want to take a look at this too since he's going to be working on the generator soon and had been getting familiar with Format, though we went over the general idea together.

lib/Stan_math_backend.ml
VMatthijs: OK! Would it make sense to change the filename to `Code_generation` then? Or something similar? Not a big deal either way.

seantalts: I'm following the convention for Fmt: https://erratique.ch/software/fmt/doc/Fmt.html#nameconv. And we are pretty printing C++.

seantalts:
> I wonder if slightly less abbreviated variable names might help or a bit more documentation.

Which variable names are confusing?

lib/Stan_math_backend.ml
seantalts: I'm not sure I understand... It's called Stan_math_backend rather than the more generic `Code_generation` because we will probably have multiple code generation backends. Right now we just have the C++ Stan Math backend, but we want to keep it easy to support Tensorflow Probability as a backend as well.

VMatthijs: Isn't `backend` a bit more general? In my mind, the backend of a compiler does more than just code generation. It may also do various low-level optimizations.

How about `Stan_math_code_gen`?

VMatthijs: OK!

seantalts:
I wouldn't waste too much time trying to understand the code of the code generation yet - it's a mess right now and I haven't cleaned it up yet. The MIR is important to get right, and good names help, so it's good to bring those up.

> - in MIR.ml, `cfname`, `vident`, `st`, `tvdecl`, `tvident`, `tvtype`, `tvtrans`, `tvloc`, `tvtable`, `prog`, `gqb`: I'd have a preference for slightly longer names in an important data structure like the IR.

This is a good point. I generally try to strike a balance because as the name increases in length, you can fit less code on the screen at once, which makes it harder to understand again. I'm also generally following the abbreviations you created in the AST where applicable - that's where TV (for TopVar) came from (though those are now gone). I think record field names like 'identifier' in the AST are generally both difficult to understand (because identifiers are all over the place, not just for VarDecls) AND overly long ('ident' is, as far as I'm aware, an unambiguous abbreviation). That said there are definitely engineers who think you should never use abbreviations (though we have quite a few in the AST already), I think rightfully so (like 'S' for 'sized', U for 'unsized', 'pe' for 'plus equals', etc). I'm a bit torn. I think `cfname` is particularly pure and can be changed to `check_name` without much ado, and `vident` can become `varident`. I think `tvdecl` can become `top_var_decl[aration?]` as well, because it's a type name that isn't referred to often. But I don't think `tvtrans` should become `top_var_transformation`.Do you have suggestions for these? @bob-carpenter would you mind weighing in here? I'm never sure about these things.

>  With autocomplete, there isn't much of a cost to that, right?

The cost is that you can fit less code on the screen at once (which limits readability) and I think it also takes up more space in working memory, but that's a theory of mine. A counter-point here is that with type definitions in the editor, it usually becomes clear what something stands for (though we should change some that aren't clear that are just strings, for example). But we know that all of these `tv*` names are part of a named record type that should help explain the abbreviation inline for those new to the code.

> - in Stan_math_backend.ml, a bunch of names, mostly function arguments: `ad`, `pf` (a Fmt name, I know), `st`, `ts`.

`ts` was just left in by mistake - was trying to debug things for the memory leak and added the explicit parameter instead of using the point-free style. Oops. But `st` was intentional as a consistent abbreviation for `sizedtype`. What else should we call that? I'm uncomfortable calling it `sizedtype` as that's the name of the type (though of course OCaml would let us, I think it's confusing). 

The general consensus seems to be the Fmt API is much better than the Format one, and I think we want to program with it unadulterated as we assume those looking at the code know how the dependencies of the code work already.

> Having nice names in the MIR seems more important to me than in the code gen.

+1 

I will go ahead and make some changes and maybe the melange will be readable enough.

bob-carpenter:
I'll just pull out answers rather than going inline.

1.
Did we want to even go with something like Fmt rather than just writing strings out directly using the lower level printf facilities?  It seems the main purpose of Format/Fmt is breaking and whitespace prettification, but we can probably do that all ourself as I don't really care how long the lines are.  I'd like to choose based on which way will be easier to understand and maintain.

2.  
I strongly prefer shorter names as long as the abbreviation conventions are consistent.  I don't like seeing _ast_ in one place and _abstract_syn_tree_ in another and _abs_syntax_tr_ in a third.

In fact, I'd just go down to id from ident. 

As Sean points out, it's not the cost to create, it's the cost to read.  I like to be able to see the structure of code from looking at it, and long names get in the way.  I think that's the main thing that long names prevent---quick scanning for structure.

  a + b * x

may not tell you the contents of the variables, but the form is clear, exactly the opposite of what you get with 

  mean_log_income + log_income_change_per_cm * height_in_cm

The operators get buried in the names, even in simple cases like this.


3.  
I also dislike naming variables the same thing as the type, as in:

  foobar foobar(x);

in C++, where foobar is both the type name and the name of the instance.
We've used a lot of

  foobar fb(x);

in the Stan code to try to distinguish intances from class names without
having to go to foobar_t or something heavier like that.

lib/Stan_math_backend.ml
seantalts: Sounds good!

seantalts:
> Did we want to even go with something like Fmt rather than just writing strings out directly using the lower level printf facilities? It seems the main purpose of Format/Fmt is breaking and whitespace prettification, but we can probably do that all ourself as I don't really care how long the lines are. I'd like to choose based on which way will be easier to understand and maintain.

Even if we don't use the indentation tools, it would be nice to 1) print to files a la `fprintf` and 2) handle nested string situations (provided by `%a` in Format or Fmt). I'm feeling a little invested in `%a` (via Fmt now) already but that's probably just sunk cost fallacy. We should have a discussion about this in person probably.

Given the rest of the comments I will try to create longer type names and then continue using consistent abbreviations for values. I think treating record field names more like values might make sense given field punning, and assuming the record type shows what the abbreviations stand for.

seantalts:
Okay, merging for now with a decent attempt at naming according to this conversation. We can talk tomorrow about Fmt vs. linked lists of strings vs ?

PR #54, opened 2019-02-09
[Code gen] Clean up, small MIR change, rethink pp_statement, mother model integration test

seantalts:
I'm going to try to make smaller, more frequent PRs whenever tests are passing and anything in MIR or AST changes (or anything else slightly major, like the pp_statement change here). Nothing is "now working" with this PR but tests are passing and I think it contains progress.


lib/Stan_math_code_gen.ml
```diff
@@ -191,35 +194,31 @@ let pp_statement pp_statement_with_meta ppf s =
...
-        , fdbody )
-
-let rec pp_statement_loc ppf {sloc; stmt} =
-  (*
-  Sexp.to_string_hum [%sexp (stmt : stmt_loc statement)] |> print_endline ;
-*)
-  pf ppf "current_statement_loc__ = \"%s\";@;" sloc ;
-  pp_statement pp_statement_loc ppf stmt
+      match fdbody.stmt with
+      | Skip -> pf ppf ";@ "
```
seantalts: This right here is the reason I had to convert pp_statement from taking in an `'a statement` to knowing exactly what `'a` would be. Needed to dissect `fdbody` to see if it was just a Skip and not emit code, and we still don't care about the metadata in this function but apparently the typechecker needs us to know what it will be so it can inspect. Not sure if @VMatthijs can think of any ways around this, it probably eventually reduces to needing row polymorphism. I tried to hack it by just pretty printing the body and checking if it's empty, but unfortunately the locations will always be printed out even if the statement is a Skip (which, actually thinking about it now, maybe that shouldn't be true). 

Relatedly, I'm wondering if it makes sense to just put the locations in the MIR as a variant of statement. The trouble I perceive would be that it could make inter-statement dependency tracking more difficult than just checking where Vars are used and where they come from or are modified. I'd love it if anyone could chime in here and hash this out with me now, or maybe we should just try it for now.

lib/Ast.ml
```diff
@@ -131,6 +131,7 @@ type 'e sizedtype =
   | SArray of 'e sizedtype * 'e
 [@@deriving sexp, compare, map, hash]
 
+(** remove_size [st] converts st from a sizedtype to an unsizedtype. *)
```
bob-carpenter: I'd prefer to document the return.  And I don't think we need to repeat the name of the function.
```
(** Return the unsized type corresponding to the argument type. *)
```
It's a bit confusing here because there's no real argument `st` anywhere to document.

lib/Stan_math_code_gen.ml
```diff
@@ -73,14 +73,11 @@ let pp_block ppf (pp_body, body) = pf ppf "{@;<1 4>@[<v>%a@]@,}" pp_body body
 let pp_for_loop ppf (loopvar, lower, upper, pp_body, body) =
   pf ppf "@[<hov>for (@[<hov>size_t %s = %a;@ %s < %a;@ %s++@])" loopvar
     pp_expr lower loopvar pp_expr upper loopvar ;
-  pf ppf "@;<0 4>@[<v>%a@]@]" pp_body body
+  pf ppf "@ @;<0 4>@[<v>%a@]@]" pp_body body
```
bob-carpenter: I'm not goin to pretend I understand this difference yet!

lib/Stan_math_code_gen.ml
```diff
@@ -157,24 +164,20 @@ let pp_statement pp_statement_with_meta ppf s =
   | MarkLocation _ -> () (* XXX *)
   | Check _ -> () (* XXX *)
   | IfElse (cond, ifbranch, elsebranch) ->
-      let pp_else ppf x =
-        pf ppf " else {\n %a;\n}" pp_statement_with_meta x
-      in
-      pf ppf "if (%a) %a %a\n" pp_expr cond pp_block
-        (pp_statement_with_meta, ifbranch)
+      let pp_else ppf x = pf ppf " else {\n %a;\n}" pp_statement x in
```
bob-carpenter: [question]
Aren't we supposed to get rid of all the hard `\n` instances in favor of their breaks?

lib/Stan_math_code_gen.ml
bob-carpenter: I may need a face-to-face explanation of the problem.

lib/Stan_math_code_gen.ml
```diff
@@ -478,18 +453,63 @@ let trans_checks s =
...
 
 (* XXX *)
-let pp_model ppf p = ignore (ppf, p)
+let pp_model ppf p =
+  pf ppf
+    {|
+  class %s_model : public prob_grad {
+@ @[<v 1>
+@ private:
+@ @[<v 1>%a@]
```
bob-carpenter: I'm going to need more time to figure all these out.  Was there doc somewhere for this fmt thing other than just the command-level doc?

lib/Ast.ml
seantalts: For some reason, the Jane Street style guide repeats the name of the function: https://opensource.janestreet.com/standards/ (under Comments). It doesn't seem to show up when they generate the ocamldoc webpages at all, which is weird. It shows up like this in my editor: 
image

Do you think we should deviate from their style guide? I was kind of going with it for now just because it's nice not to have to choose things.

seantalts: I think that repetition is related to your other comment in that you might want to e.g. show that a point-free function definition actually takes a specific arg, or something?

lib/Stan_math_code_gen.ml
seantalts: Yeah, I think so. This was just a formatting change - the code here is left over from before I knew that about the Format library. (I was going to say "before I knew the library" but I still am learning a lot). I'll go ahead and change it.

lib/Stan_math_code_gen.ml
seantalts: Well, it seems like `Fmt` mostly just changed the available functions but not the format specifiers/strings/whatever you call them. So `%a` is still that magic compositional format specifier from the [`Format`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Format.html) library. And `@` is a pretty-print specifier that works similarly to `%` for format strings - the first character that follows a `@` is always part of the pretty-printing command, most of which are documented on the ocaml Format module page linked above. There is a 1-to-1 correspondence between `@x` and the functions available in the Format library. On this line we have a 'space' break hint, equivalent to `print_space ()` followed by an `open_vbox 1`, followed by %a and the closing of the vbox. The complete correspondence is here:
image


We could also switch to stop using all of the `@` pretty-printing specifiers, instead using explicit named functions instead (with a bit of an increase in verbosity). I had been thinking about that and don't feel strongly either way (maybe an experiment is warranted). 

Or just switch to lists of strings 😓

lib/Stan_math_code_gen.ml
seantalts: Sounds good. Are you free tomorrow afternoon?

lib/Stan_math_code_gen.ml
seantalts: Just added a `print_cut ()` effectively.

seantalts:
I'll just merge for now... Let's meet in person to talk about pp_statement, adding locations to the MIr, whether we should just switch to lists of strings instead of Format/Fmt or try using explicit named functions for `print_cut ()` etc,  and how to doc functions.

bob-carpenter:
I have no problem sticking with the simpler forms---I just have to learn them.  I read about them (or something like them) in the Format lib doc.

PR #56, opened 2019-02-12
Fix integration tests to run on Windows too

VMatthijs:
This PR makes sure that the integration tests can be run on Windows as well. It leads to the following behaviour:
- `dune runtest` will try to run the integration tests with the binary built in the `default.windows` context, if it exists;
- if not, then it simply uses the binary in the `default` context.

That is, on OSX / Linux, your workflow will stay the same and on WSL it will be such that
- `make clean`;  `dune runtest` will build an Ubuntu binary and run the tests on it;
- `make clean`; `dune build -x windows @runtest` will build a Windows binary and run the tests on it.

### DIFF
--- a/test/integration/run_bin_on_args.ml
+++ b/test/integration/run_bin_on_args.ml
@@ -1,8 +1,22 @@
 open Core_kernel
 
+let maybe_convert_cmd_to_windows cmd =
+  let pattern = "/install/default/bin/" in
+  let to_windows str =
+    String.substr_replace_first ~pattern ~with_:"/install/default.windows/bin/"
+      str
+  in
+  let path =
+    String.prefix cmd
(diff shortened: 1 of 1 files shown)


VMatthijs:
@seantalts , could you maybe have a look at this at some point? We can also discuss it in person later today. Jenkins seems to have problems again, though...

seantalts:
Wow. Docker is such a shitshow. Also this looks good but I'm curious as to why dune doesn't switch to the windows binary to run the tests when you use `-x windows`? Maybe we can figure out some command line flag for that to work?

seantalts:
Restarting the daemon on Ben's linux box seemed to fix it. Maybe the package upgraded at some point.

PR #60, opened 2019-02-16
Another chunk of code gen

seantalts:
This PR is not too much progress but slightly more significantly has obvious stubs for many of the remaining methods (not that the ones that have implementations are perfect yet). I'm hoping this makes it a little easier for folks to see what needs to be done. I'm hoping y'all (especially @bob-carpenter) can also comment on which stubs I'm missing.

Other minor changes:
* Switched to targeting 8 schools for now, as this is a smaller target and once we hit that we can get RStan and PyStan engaged on figuring out how to distribute
* Made the git commit hook exit if formatting failed (slightly easier to use as a result)
* Changed MIR a little - now we use the Errors function for printing locations
It's sort of almost exciting: there is a cpp file! With lots of stubs and mistakes! https://github.com/stan-dev/stanc3/blob/read-data-gen/test/integration/examples-good/code-gen/cpp.expected


seantalts:
@bob-carpenter or @VMatthijs do you have time to take a look at this? I was also thinking of just merging for now and leaving the PR around for the sake of future commentary - already building another branch on top of this one and it has some changes to MIR etc.

lib/Stan_math_code_gen.ml
```diff
@@ -1,4 +1,20 @@
...
+    the Format library[0]. The boxing system is best described in this explainer
+    pdf [2], particularly Section 3 ("Format basics"). It's worth noting that someone
+    was able to make a decent-looking pretty-printer for a subset of Javascript[3] that
+    might serve as a good reference. Good luck!
+
+    [0] Format module doc: https://caml.inria.fr/pub/docs/manual-ocaml/libref/Format.html
+    [1] Fmt module doc: https://erratique.ch/software/fmt/doc/Fmt.html
+    [2] Format Unraveled: https://hal.archives-ouvertes.fr/hal-01503081/file/format-unraveled.pdf
+    [3] Javascript pretty-printer https://github.com/Virum/compiler/blob/28e807b842bab5dcf11460c8193dd5b16674951f/JavaScript.ml#L112
+*)
```
VMatthijs: Thank you for adding this! This really helps to get started with Format+Fmt. :-)

VMatthijs:
@seantalts , I've had enough of a look at this that I'm happy with you merging this if it helps you get on with the code gen. I can hopefully do an actual code review tomorrow night, if that's in time. (Am on holidays, so not doing much work this week.) If not, then maybe @bob-carpenter can do it earlier.

bob-carpenter:
I may have more time tomorrow.  I anticipate being stuck at the airport.

VMatthijs:
Some random comments/questions:
a) I see you throw exceptions in a couple of places, for code branches that should never happen. So far, I had been using the `fatal_error` function from `Errors.ml`. Should we have a consistent policy for this sort of thing? E.g. make `fatal_error` such that it produces a suitable error message for users to report a bug?

lib/Ast_to_Mir.ml
```diff
@@ -44,12 +44,7 @@ let targetpe e =
   let t = Var "target" in
   Assignment (t, BinOp (t, Plus, e))
```
VMatthijs: Is this not slow?

I see that you've chosen to treat `target` as a normal variable, rather than doing what the current generator does.
Also, you are translating compound operator assignments to plain assignments followed by an operator application.
Are you planning to rely on an optimizing compiler or do you think that these things would amount to premature optimization?

lib/Ast_to_Mir.ml
```diff
@@ -64,11 +59,21 @@ let truncate_dist ast_obs t =
...
+    String.drop_suffix (String.drop_prefix s 1) 1
+  else s
+
+let trans_printable (p : Ast.typed_expression Ast.printable) =
+  match p with
+  | Ast.PString s -> Lit (Str, unquote s)
+  | Ast.PExpr e -> trans_expr e
 
 let rec trans_stmt {Ast.stmt_typed; stmt_typed_loc; _} =
   let or_skip = Option.value ~default:Skip in
```
VMatthijs: In the operator assignment below, you are treating a binary operator as a function application. Before, it seemed like binary operators were not treated as functions. What is the reason for that inconsistency?

Also, why is the lhs of an assignment treated as a plain expression?

lib/Ast_to_Mir.ml
```diff
@@ -64,11 +59,21 @@ let truncate_dist ast_obs t =
...
   match t with
-  | Ast.NoTruncate -> None
-  | Ast.TruncateUpFrom lb -> Some (trunc Less lb None)
-  | Ast.TruncateDownFrom ub -> Some (trunc Greater ub None)
+  | Ast.NoTruncate -> []
+  | Ast.TruncateUpFrom lb -> [trunc Less lb None]
+  | Ast.TruncateDownFrom ub -> [trunc Greater ub None]
   | Ast.TruncateBetween (lb, ub) ->
-      Some (trunc Less lb (Some (trunc Greater ub None)))
+      [trunc Less lb (Some (trunc Greater ub None))]
```
VMatthijs: This makes sense to me. Have you checked that it is in fact what the current Stan compiler does?

lib/Ast_to_Mir.ml
```diff
@@ -64,11 +59,21 @@ let truncate_dist ast_obs t =
...
+
+let unquote s =
+  if s.[0] = '"' && s.[String.length s - 1] = '"' then
+    String.drop_suffix (String.drop_prefix s 1) 1
+  else s
+
+let trans_printable (p : Ast.typed_expression Ast.printable) =
+  match p with
+  | Ast.PString s -> Lit (Str, unquote s)
```
VMatthijs: Oh whoops. I should've probably done the unquoting while building the AST.

lib/Ast_to_Mir.ml
```diff
@@ -100,11 +105,8 @@ let rec trans_stmt {Ast.stmt_typed; stmt_typed_loc; _} =
...
-        Block
-          [ Option.value
-              ~default:(bind_loc stmt_typed_loc Skip)
-              (truncate_dist arg truncation)
-          ; bind_loc stmt_typed_loc add_dist ]
+        SList
+          (truncate_dist arg truncation @ [bind_loc stmt_typed_loc add_dist])
     | Ast.Print ps -> NRFunApp ("print", List.map ~f:trans_printable ps)
     | Ast.Reject ps -> NRFunApp ("reject", List.map ~f:trans_printable ps)
     | Ast.IfThenElse (cond, ifb, elseb) ->
```
VMatthijs: The translation of ForEach loops does not seem right to me.

lib/Ast_to_Mir.ml
```diff
@@ -281,16 +283,15 @@ let trans_prog filename
...
     let flattened = List.(concat (map ~f:lbind stmts)) in
-    with_no_loc (match flattened with [] -> Skip | _ :: _ -> SList flattened)
+    with_no_loc (SList flattened)
   in
   let pull_tvdecls_multi blocks =
-    let merge_maps maps =
-      List.map ~f:Map.Poly.to_alist maps
-      |> List.concat |> Map.Poly.of_alist_exn
-    and tvtables, stmts = blocks |> List.map ~f:pull_tvdecls |> List.unzip in
+    let tvtables, stmts = blocks |> List.map ~f:pull_tvdecls |> List.unzip in
```
VMatthijs: I'm pretty confused about what is going on here in `trans_prog`.

Why are you removing block structure from the statements? How do we make sure code gets generated for local variable scope then?

lib/Mir.ml
```diff
@@ -50,11 +54,12 @@ type unsizedtype = Ast.unsizedtype [@@deriving sexp, hash]
 type constraint_check =
   {ccfunname: string; ccvid: string; cctype: sizedtype; ccargs: expr list}
 
+and formal_params = (adtype * string * unsizedtype) list
```
VMatthijs: Could you give `formal_params` a more informative name? E.g. `fun_arg_decl`.

test/integration/examples-good/code-gen/dune
```diff
@@ -2,11 +2,11 @@
 
 (rule
  (targets cpp.output)
- (deps (package stanc) (:stanfiles (glob_files *.stan)))
+ (deps (package stanc) (:stanfiles eight_schools_ncp.stan))
```
VMatthijs: Good idea to start there.

lib/Ast_to_Mir.ml
```diff
@@ -44,12 +44,7 @@ let targetpe e =
   let t = Var "target" in
```
VMatthijs: I see you throw exceptions in a couple of places, for code branches that should never happen. So far, I had been using the `fatal_error` function from `Errors.ml`. Should we have a consistent policy for this sort of thing? E.g. make `fatal_error` such that it produces a suitable error message for users to report a bug?

lib/Ast_to_Mir.ml
VMatthijs: Generally, why is there both a List of statements and Block of statements in the MIR?

lib/Mir.ml
```diff
@@ -17,6 +17,10 @@ let gensym () =
   _counter := !_counter + 1 ;
   sprintf "sym%d" !_counter
 
+let gensym_enter () =
+  let old_counter = !_counter in
+  (gensym (), fun () -> _counter := old_counter)
+
```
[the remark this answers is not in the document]
seantalts: I don't think the old code generator generated symbols ?

And yeah, generating fresh symbols should happen centrally somewhere and it's more of a middle/backend thing rather than a front-end thing. The only time I reset the gensym is when I know it's safe to do so, right now that's just when generating for loops for data input checks or similar stuff that doesn't have user code inside it.

lib/Mir.ml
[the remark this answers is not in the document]
seantalts: postfix operators aren't a thing in any of our potential target languages; someone once convinced me that binary operators in C++ are more efficient than function calls (but I'm inclined to just switch back to treating them like fn calls for now), prefix operators idk

lib/Ast_to_Mir.ml
seantalts: 100% relying on an optimizing compiler, for sure for sure. That's an important idea behind the design of the MIR - break down the Stan concepts into simpler ones and optimize holistically. I realized today during meditation that we should also be breaking down the complex indexing operations we have into straightforward MIR.

>  Also, you are translating compound operator assignments to plain assignments followed by an operator application.

yes

> Is this not slow?

What specifically? the compiler is still fairly quick to compile things, now that that Format bug is gone.

> I see that you've chosen to treat `target` as a normal variable, rather than doing what the current generator does.

Yeah. Ryan was saying it would be handy if it was structurally different, because from the perspective of tracking probabilistic dependencies it's the only Var that introduces those. I'll have to think about that a little more but it makes sense prima facie to turn it back into a MIR node for that reason, unfortunately.

lib/Ast_to_Mir.ml
seantalts: So obviously a var or indexed thingy are expressions, so that works out, but of course you can put nonsense stuff in the MIR as the lhs of assignment if it's just "expressions." I think @rybern  suggested making an LHS type in the MIR somewhere that would fix this. I think I would also like to just remove the concept of Indexed as an Expr from the MIR completely, to be replaced with an explicit For loop. Then we'd just have to use string identifiers everywhere we wanted to just refer to a var, like here on the lhs of assignment.

> In the operator assignment below, you are treating a binary operator as a function application. Before, it seemed like binary operators were not treated as functions. What is the reason for that inconsistency?

It would help me if you put the PR comments below the code in question as github shows only the code directly above the comment. And yeah, this is left over from when I was just treating binops as fnapps. We should probably have a larger discussion about that topic - I'm kind of ambivalent either way and would sort of rather it just be simpler if we can get away with it.

lib/Ast_to_Mir.ml
seantalts: I forget, maybe

lib/Ast_to_Mir.ml
seantalts: What seems wrong about it?

lib/Ast_to_Mir.ml
seantalts: Blocks are a mechanism for local variable scoping. Lists are just lists of statements without such a mechanism. I'm not removing all blocks, just extra ones at the top level. It may have a bug and be removing one level too many, or something.

lib/Mir.ml
seantalts: that's literally what a formal parameter is, haha. https://chortle.ccsu.edu/java5/Notes/chap34A/ch34A_3.html

seantalts: actually I guess those are just for identifiers, technically. I'll change it to fun_arg_decl, which has no pre-existing meaning and might be easier for people to deal with.

lib/Ast_to_Mir.ml
seantalts: Hmm, I like the Core_kernel tools, but it might be useful to have a standardized message with a link to the bugtracker. `fatal_error` is not a good name for those - there are a lot of fatal errors that are totally normal and users should NOT submit bugs, like specifying an invalid filename. Maybe we can write something like `raise_impossible` that takes in a `[%message ]` thing (looks like that's just a `Sexp.t`)?

lib/Mir.ml
VMatthijs: Ah, right! I'd just never heard that term.

lib/Ast_to_Mir.ml
VMatthijs: Sure, `raise_impossible` seems like a fine name to me. I see what you mean about not all fatal errors being the programs fault. We should just change the name of `fatal_error` I think, as I've been using it exclusively for this sort of situation. Adding a link to a bug tracker seems like a good idea.

lib/Ast_to_Mir.ml
VMatthijs: Where are there extra blocks at the top level? Have looked at the code again and still don't understand why this `coalesce` is necessary. Could you explain a bit more the thinking behind it?

lib/Ast_to_Mir.ml
VMatthijs: You're translating the foreach loop into a for loop. Say that you're looping over an array of type `int[,]` As it currently stands, the loop variable will change type from `int[]` to `int`. Maybe have a look at what the current code generator does? It also needs to generate different code depending on whether you're looping over an array or over an Eigen type.

lib/Ast_to_Mir.ml
VMatthijs: > It would help me if you put the PR comments below the code in question as github shows only the code directly above the comment.
I know, sorry. I wanted to comment on something that wasn't new in this PR and github wouldn't let me comment on the actual line that I wanted to comment on. It will only let me comment on lines that were changed in this PR.

VMatthijs: > I think I would also like to just remove the concept of Indexed as an Expr from the MIR completely, to be replaced with an explicit For loop.

Not sure I understand how that would work. Can you please explain? We are in a language with arrays, right? Don't you need indexing? Or do you have a way in mind to treat arrays as syntactic sugar?

lib/Mir.ml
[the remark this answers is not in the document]
VMatthijs: OK. I think the old code generator just used some naming scheme right? As long as it is used carefully, I'm happy with it.

lib/Mir.ml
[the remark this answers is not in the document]
VMatthijs: Could you maybe discuss the binary operator question with @bob-carpenter ? I think he'd be the most likely person to know the best course of action here, right? I don't know what's best myself. When in doubt, I'd be inclined to do what the old code generator does, myself.

lib/Ast_to_Mir.ml
VMatthijs: > 100% relying on an optimizing compiler, for sure for sure. That's an important idea behind the design of the MIR - break down the Stan concepts into simpler ones and optimize holistically.

Sure. However, for instance an optimization pass to exploit variables that are only ever added to (like `target`) is not one that we are likely to write very soon. It's not clear to me that it would have very many more use cases and would be at the top of the priority list. Of course, I'm not sure if it would make a significant difference at all, in speed. Myself, I think we'll want to implement
- constant folding
- loop unrolling
- dead code elimination
- function inlining
- loop-invariant code motion
- CSE
- tail-call optimization (?).
However, the new compiler should be able to replace the old one before we have any of these optimization passes, as they might take a while to implement.

lib/Stan_math_code_gen.ml
```diff
@@ -20,15 +36,17 @@ let%expect_test "emit function type raises" =
   | exception e -> print_s [%sexp (e : exn)] ) ;
   [%expect {| ((x UMathLibraryFunction) "not implemented yet") |}]
 
-let pp_call ppf (name, pp_args, args) =
-  pf ppf "%s(@[<hov>%a@])" name pp_args args
+let pp_call ppf (name, pp_arg, args) =
+  pf ppf "%s(@[<hov>%a@])" name (list ~sep:comma pp_arg) args
```
VMatthijs: Is this enough to deal with functions which take some arguments which are strings are others that are expressions, which is allowed in Stan, like
`
print("hello", {1, 2, 3});
`.
Arguably, we should just add a string type to the language and treat strings are expressions.

lib/Stan_math_code_gen.ml
```diff
@@ -20,15 +36,17 @@ let%expect_test "emit function type raises" =
   | exception e -> print_s [%sexp (e : exn)] ) ;
```
VMatthijs: Maybe use a more informative name than `ad` in the above for the string that will represent the base type?

lib/Stan_math_code_gen.ml
```diff
@@ -20,15 +36,17 @@ let%expect_test "emit function type raises" =
...
   match s with
   | Var s -> string ppf s
-  | Lit (Str, s) -> pf ppf "\"%s\"" s
+  | Lit (Str, s) -> pf ppf "%S" s
   | Lit (_, s) -> string ppf s
-  | FunApp (fname, args) -> pp_call ppf (fname, list ~sep:comma pp_expr, args)
+  | FunApp (fname, args) -> pp_call ppf (fname, pp_expr, args)
   | BinOp (e1, op, e2) ->
       pf ppf "%a %s %a" pp_expr e1 (Operators.operator_name op) pp_expr e2
   | TernaryIf (cond, ifb, elseb) ->
```
VMatthijs: This all seems fine with me. The only one I cannot comment on is the Indexed expressions. Have you checked it does the same as the old code gen?

lib/Ast_to_Mir.ml
VMatthijs: Also, if you'd want to get rid of language constructs, would it be worth encoding the while loop as a for-loop or vice versa? Not really sure whether that's a good idea. We'll definitely want to identify statically sized for-loops for the loop-unrolling, but not sure if a uniform representation of all loops would get in the way for that.

lib/Stan_math_code_gen.ml
```diff
@@ -70,10 +88,12 @@ let rec stantype_prim_str = function
 let pp_prim_stantype ppf st = pp_unsizedtype (stantype_prim_str st) ppf st
 let pp_block ppf (pp_body, body) = pf ppf "{@;<1 2>@[<v>%a@]@,}" pp_body body
 
+(** [pp_for_loop ppf (loopvar, lower, upper, pp_body, body)] tries to
+    pretty print a for-loop from lower to upper given some loopvar.*)
```
VMatthijs: why tries?

lib/Stan_math_code_gen.ml
```diff
@@ -116,61 +136,115 @@ let pp_arg ppf ((adtype, name, st), idx) =
...
+      let fname, extra_args = trans_math_fn fname in
+      pf ppf "%s(@[<hov>%a@]);" fname (list ~sep:comma pp_expr)
+        (extra_args @ args)
   | Break -> string ppf "break;"
   | Continue -> string ppf "continue;"
   | Return e -> pf ppf "return %a;" (option pp_expr) e
   | Skip -> ()
-  | MarkLocation _ -> () (* XXX *)
-  | Check _ -> () (* XXX *)
+  | Check {ccfunname; ccvid; ccargs; cctype} ->
```
VMatthijs: Does this also do the right thing if `cctype` is an Eigen-type?

lib/Mir.ml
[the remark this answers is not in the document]
VMatthijs: Also, shouldn't the generated names end in `__` or something to make sure they never clash with user variable names?

lib/Ast_to_Mir.ml
seantalts: We also need to change the type signature to take S-Expressions, or add a fn with a `_s` suffix that does so we can use it with `[%message "bad statement" (s: stmt_loc)]` or whatever.

lib/Ast_to_Mir.ml
```diff
@@ -160,9 +162,6 @@ let rec trans_stmt {Ast.stmt_typed; stmt_typed_loc; _} =
   in
```
VMatthijs: Apologies about the comment in the wrong place, but it won't let me put it where I want it: the translation of VarDecls seems not quite right to me.

We do not want the adtype to be autodiffable if we are in the transformed data or generated quantities block, right?

Also, in case we are a global variable, do we want to generate an MIR statement for that at all? Or do we just want to put it in the global var map?

lib/Ast_to_Mir.ml
VMatthijs: Could you explain the use case of s-expressions in these errors? Is the idea that these error messages should face the developers rather than users so it makes sense to serialize internal data structures in them? Is this use of `[%message ...]` documented anywhere? I've been Googling but cannot find it. I really wish Core_kernel was documented a bit better.

lib/Stan_math_code_gen.ml
```diff
@@ -116,61 +136,115 @@ let pp_arg ppf ((adtype, name, st), idx) =
...
+  pf ppf "%a@ " (option ~none:(const string "void") (pp_unsizedtype scalar)) rt
 
-let pp_decl ppf (vident, st) =
-  pf ppf "%a %s;" pp_prim_stantype (Ast.remove_size st) vident
+let pp_location = fmt "current_statement__ = %S;@;"
 
-let with_no_loc stmt = {stmt; sloc= ""}
+(** [pp_located_error ppf (body_block, err_msg)] surrounds [body_block]
+    with a C++ try-catch that will rethrow the error with the proper source location
+    from the [body_block] (required to be a [stmt_loc Block] variant).*)
```
VMatthijs: What is a `stmt_loc Block` variant? What happens if you feed it something else?

lib/Ast_to_Mir.ml
seantalts: That's exactly it! I can try to find where I found this raise_s pattern soon... But yes, these are all unexpected situations the programmer failed to accommodate, so it's likely we want to serialize the offending data (as we mostly encode "situations" with data types).

lib/Stan_math_code_gen.ml
```diff
@@ -183,19 +257,16 @@ let rec pp_statement ppf {stmt; sloc} =
       let argtypetemplates =
```
VMatthijs: You use `ignore decl_adtype` above. Should that be our preferred style rather than just matching on `Decl {decl_id; decl_type; _}` ?

lib/Stan_math_code_gen.ml
```diff
@@ -214,20 +285,20 @@ let rec pp_statement ppf {stmt; sloc} =
...
   [%expect
     {|
-      {
-        current_statement_loc__ = "lo";
-        break;
-      } |}]
+    {
+      current_statement__ = "lo";
+      stan_print(pstream__);
+    } |}]
```
VMatthijs: Why should the outer location not be printed? Is it set up such that statements always need to be wrapped in blocks for the location to be printed? Note that top level statements are not wrapped in a block. How would we get error locations for those?

lib/Stan_math_code_gen.ml
```diff
@@ -116,61 +136,115 @@ let pp_arg ppf ((adtype, name, st), idx) =
...
+      throw std::runtime_error("*** IF YOU SEE THIS, PLEASE REPORT A BUG ***"); |}
   @@ Option.value ~default:"e" msg
 
-let pp_located_error ppf (pp_contents, contents_arg, msg) =
-  pp_error_wrapper ppf (pp_located_msg, msg, pp_contents, contents_arg)
+let maybe_templated_arg_types (args : formal_params) =
+  let is_autodiff (adtype, _, _) =
+    match adtype with Ast.AutoDiffable -> true | _ -> false
+  in
+  List.filter ~f:is_autodiff args |> List.mapi ~f:(fun i _ -> sprintf "T%d__" i)
```
VMatthijs: Getting these template parameters to match (and not clash with each other!) everywhere seems like one of the hard tasks in code gen for me. Is it obvious to you that your current strategy will work? Are you using the same naming scheme as in the existing code gen? (Just asking because I cannot assess the correctness of this myself right now so wanted to at least flag it as a potentially subtle point.)

VMatthijs:
@seantalts , I've now reviewed this as well as I can without diving deeply into the existing code generator. It mostly seems to make sense to me. I am a bit nervous though as I find it so hard to convince myself that this is all correct, mostly because correctness is a global property of the code gen (e.g. name binding matching across long distances) and I do not have a good enough global picture in my head of what it should look like.

What is the current plan for showing that the code gen in correct? Is it still just running on test models and seeing that it gives the same results? What if it doesn't? How will we find the source of the error then? Is the code gen as currently written similar enough to the old code gen such that a diff would be informative?

Do you have suggestions on how you would like to see this reviewed or on how I can help with the effort?

lib/Ast_to_Mir.ml
VMatthijs: >What specifically? the compiler is still fairly quick to compile things, now that that Format bug is gone.

I meant slow at runtime, rather than slow to compile.

lib/Ast_to_Mir.ml
VMatthijs: The model block should be treated as being a local scope. Otherwise, all program blocks are global scope. Note however that a program block does not have a Block AST node at the top (even though it's written with curly braces).

I like that you merged the transformed parameters and model blocks in the MIR by the way. I hope we can do the same in the Stan language soon. In my mind, target should be available in transformed parameters and the model block should just be a local scope within transformed parameters. (At which point we should probably change the name of the compound block.)

bob-carpenter:
> I like that you merged the transformed parameters and model blocks in the MIR by the way. I hope we can do the same in the Stan language soon. In my mind, target should be available in transformed parameters and the model block should just be a local scope within transformed parameters. (At which point we should probably change the name of the compound block.)

I'm OK with the change, but discussions about changing the language in the past have been contentious.  

The design I've lobbied for is to make the Jacobian available in the transformed parameters block as

jacobian += <expression>;

and then treat it like the Jacobian contributions of the transforms (e.g., turning it off for optimization).  This would let us define all the transforms we use directly in the Stan language.  As is, we can't replicate our built-in constraints because we can't define a Jacobian that's optional (turned off during optimization).

lib/Ast_to_Mir.ml
VMatthijs: Example we talked about

```
//ORIGINAL stan
real xs[4];
for (x in xs)
  print(x);
  
  
  
  
//generated code (stanc2)
real xs[4];
for (i__ in 1:4) {
  real x = xs[i__];
  print(x);
}



//generated code (stanc3)
real xs[4];
for (x in 1:4) {
  print(x);
}
```

lib/Mir.ml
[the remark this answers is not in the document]
seantalts: Ended up talking with Matthijs a lot today and decided to just go with trying to keep things close to existing code gen until we release. BinOp could have gone either way, but the math library doesn't support calling `add(int, int)` so rather than figuring out a way around it, it seems easier to just go back to BinOp. I want to also make tickets or otherwise keep track of opportunities for future simplification...

lib/Ast_to_Mir.ml
seantalts: We talked about this in person, but just to record - the Single index we still need something like, but we could represent all of the multi indexing as For loops. But for now, since the target languages have more efficient ways of dealing with multi indexing than just for loops and we don't want to take the chance that it's not as performant without an optimization pass, we'll leave them in.

lib/Stan_math_code_gen.ml
seantalts: If I understand you right, that's already happening in the MIR and Ast_to_Mir - printables become Lit(Str, unquote p) and expressions are also translated correctly, and we print string literals correctly.

lib/Stan_math_code_gen.ml
seantalts: it does for at least some cases. Haven't done much comprehensive testing in the module yet.

lib/Stan_math_code_gen.ml
seantalts: I don't have much faith in myself! :P

lib/Stan_math_code_gen.ml
seantalts: Yep.

lib/Ast_to_Mir.ml
seantalts: Globals should already be removed from the MIR statements with this code. Not sure what to do about autodiffable - I enjoy that almost all of the passes are not context sensitive like this yet. Maybe we need to do a pass over transformed data and gen quant to fix the vardecls, then... Not sure the best way to organize that.

lib/Stan_math_code_gen.ml
seantalts: Hm, this has been a thorny thing in a few places - trying to know when we will be getting a Block (and thus printing braces) vs. not getting one. I put in a comment to try to refactor it generally (also mixes with your coalesce comments).

lib/Stan_math_code_gen.ml
seantalts: Not sure.

lib/Stan_math_code_gen.ml
seantalts: Locations for Blocks are not printed as all of their member statements have individual locations.

lib/Stan_math_code_gen.ml
seantalts: I don't have a global strategy on that. The trickiest seeming ones are to make sure each model method uses the same template parameters, and the easiest thing there is to just always use `T__` for the autodiff type. I think only within FunDefs do we need to worry, and here we probably need to note the interaction between VarDecl's being subtly incorrect (i.e. always being autodiffable) and FunDef's translation creating a variable number of template parameters and then not passing them into pp_statement anywhere.

seantalts:
Re: testing: I think for now it's most useful to compare the generated code for eight schools against the one from stan 2. Eventually we will then do that with the mother model from @mitzimorris, and then we will write end-to-end tests running optimization with known init values for enough iterations to converge for all of the stat comp benchmarks models.

lib/Stan_math_code_gen.ml
VMatthijs: Right! Thanks! That makes sense.

PR #64, opened 2019-02-26
Switch the MIR to the subset of binary operators we can actually generate

seantalts:
Also deleted some html files we probably don't need, renamed Ast.Exp to Pow (`Exp` is often short for "expression" or the mathematical function e^x).


lib/Ast.ml
```diff
@@ -46,7 +46,7 @@ and operator =
   | LDivide
   | EltTimes
   | EltDivide
-  | Exp
+  | Pow
```
VMatthijs: +1

lib/Ast_to_Mir.ml
```diff
@@ -3,11 +3,36 @@
...
+  | Ast.Or -> Or
+  | Ast.And -> And
+  | Ast.Equals -> Equals
+  | Ast.NEquals -> NEquals
+  | Ast.Less -> Less
+  | Ast.Leq -> Leq
+  | Ast.Greater -> Greater
+  | Ast.Geq -> Geq
+  | _ ->
+      raise_s [%message "Should never try to translate " (op : Ast.operator)]
```
VMatthijs: Is that because they should be represented with function calls instead?

lib/Ast_to_Mir.ml
```diff
@@ -3,11 +3,36 @@
...
   match expr_typed with
   | Ast.TernaryIf (cond, ifb, elseb) ->
       TernaryIf (trans_expr cond, trans_expr ifb, trans_expr elseb)
-  | Ast.BinOp (lhs, op, rhs) -> BinOp (trans_expr lhs, op, trans_expr rhs)
+  | Ast.BinOp (lhs, op, rhs) -> (
+      let lhs, rhs = (trans_expr lhs, trans_expr rhs) in
+      match op with
+      | Ast.LDivide | Ast.EltTimes | Ast.EltDivide | Ast.Pow | Ast.Not
+       |Ast.Transpose ->
+          let fnname = Sexp.to_string_hum [%sexp (op : Ast.operator)] in
```
VMatthijs: Wouldn't you get capitalization in the function name this way? Is that the name we use in the Math library?

lib/Ast_to_Mir.ml
```diff
@@ -92,7 +117,7 @@ let rec trans_stmt {Ast.stmt_typed; stmt_typed_loc; _} =
         let rhs =
           match assign_op with
           | Ast.Assign | Ast.ArrowAssign -> rhs
-          | Ast.OperatorAssign op -> BinOp (assignee, op, rhs)
+          | Ast.OperatorAssign op -> BinOp (assignee, trans_op op, rhs)
```
VMatthijs: As discussed yesterday, this is OK for now. But let's keep it in mind as a potential source of optimization.

test/integration/example-models/education/copyright.html
```diff
@@ -1,13 +0,0 @@
-<hr>
-
-<p> Copyright (c) 2016, the author(s). All rights reserved. </p>
```
VMatthijs: Would we not get in trouble for removing this copyright?

seantalts: No idea, haha. I'll put it back just in case.

lib/Ast_to_Mir.ml
seantalts: Yep. I can add that to the message.

lib/Ast_to_Mir.ml
seantalts: Yeah, there will need to be a separate, Stan Math-specific translation step. We get to choose what they are in the MIR, so it seemed to make sense to follow the AST etc. We can lowercase them though, that might make more sense (some will maybe still need translation).

seantalts: on second thought I think I kind of like that they are upper cased as you can tell that they're a bit different from your typical math functions. Leaving for now...

PR #67, opened 2019-03-07
Ast to mir autodifftype fix

VMatthijs:
This fixes one of @seantalts' todos that we discussed

### DIFF
--- a/lib/Ast_to_Mir.ml
+++ b/lib/Ast_to_Mir.ml
@@ -104,6 +104,7 @@ let trans_printable (p : Ast.typed_expression Ast.printable) =
   | Ast.PExpr e -> trans_expr e
 
 let rec trans_stmt adt {Ast.stmt_typed; stmt_typed_loc; _} =
+  let recurse = trans_stmt adt in
   let or_skip = Option.value ~default:Skip in
   let s =
     match stmt_typed with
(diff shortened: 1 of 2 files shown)


lib/Ast_to_Mir.ml
```diff
@@ -103,7 +103,7 @@ let trans_printable (p : Ast.typed_expression Ast.printable) =
   | Ast.PString s -> Lit (Str, unquote s)
   | Ast.PExpr e -> trans_expr e
 
-let rec trans_stmt {Ast.stmt_typed; stmt_typed_loc; _} =
+let rec trans_stmt adt {Ast.stmt_typed; stmt_typed_loc; _} =
   let or_skip = Option.value ~default:Skip in
```
seantalts: might be slightly prettier to define here `let recurse = trans_stmt adt in`

seantalts: or, and maybe this is crazy, `let trans_stmt = trans_stmt adt in` 🐙  but idk if that's more of less confusing...

VMatthijs: went with the first

VMatthijs: @seantalts , are you happy with the fix?

VMatthijs:
In the current code generator, autodiff types for local vars in function bodies are determined by the autodiff types of the function arguments. We could mimic that. Would that make sense? Perhaps have a look at an example of the code generated for a Stan model with a udf to see what I mean.

I have made sure that if functions are inlined, they also get the right autodiff types in their body.

Some day, we could treat autodiff types of local variables in a smarter way. The AST currently has the autodiff type for the rhs expressions that get assigned to variables. From that, we should be able to decide the optimal autodiff type.

bob-carpenter:
The distinction isn't autodiff/not autodiff, it's double-only (transformed data, generated quantities), vs. template parameter (parameters, transformed parameters).  The template parameters can be instantiated with anything for which the underlying functions are implemented.

I think this distinction is even more relevant when considering multiple back ends.

I also think we can get away with using a typedef for the scalar type in all the generated code.  There should always be an appropriate typedef at the top of each scoped member function.

seantalts:
What would you call that in a non-C++ dependent way? How is it more relevant when considering multiple backends?

seantalts:
I just realized having it on the expression level doesn't really help us 😂

bob-carpenter:
It's not C++.  In Stan, variables declared in data, transformed data,
and generated quantities can only be double.  Variables declared in
parameters, transformed parameters, and the model are polymorphic.  They
could be long double, they could be some symbolic autodiff type,
they could be plain old double, or they could be one of our autodiff
types.  

In the current C++ generator, we use a typedef `local_scalar_t__` to
indicate the type of local variables.  In code for the data-only blocks (the constructor for data, transformed data,  `write_array` for generated quantities), it's set to `double`.  Otherwise, it winds up being a template parameter in the generated `log_prob` method.

This can overpromote.  For example:

```
model {
  real sigma = 1;
  y ~ normal(mu, sigma);
}
```

In this case, `sigma` will be assigned type `local_scalar_t__`, which will be instantiated to an autodiff type when calculating the log density.  It only needs to be declared as `double`.  

With more data flow analysis, we should be able to figure out if a variable like `sigma` ever gets set to a templated type (to be instantiated with an autodiff type), and if not, assign it `double`.

seantalts:
(I think 'polymorphic' was the key word I was looking for!) If `sigma` never gets set to a polymorphic type, does that mean we can *always* lift that computation out of that section into a section with only `double`s (e.g. lift out of log_prob into the model constructor for example)? If so maybe we really don't need autodiff level on expressions, we just move code to the correct place?

bob-carpenter:
Afraid not.  You need to do what Maria Gorinova did with SlicStan and worry about dependencies.  For example, it's totally OK to make `x` double here:

```
parameters {
  real<lower = 0, upper = 1> theta;
}
model {
  real x = 0;
  if (theta > 0.5)
    x = 3.2;
}
```
Now the value of `x` is always going to be `double`, but it depends on `theta`, so it can't be evaluated as transformed data.

seantalts:
Ah yeah, that makes sense. Sorry, I'm trying to keep track of a dozen things at once :D

PR #68, opened 2019-03-08
Monotone framework

VMatthijs:
This implements the monotone framework, as described in 2.3-2.4 of
Nielson, Nielson, and Hankin or 9.3 of Aho et al..
This gives a modular way of implementing many static analyses.

We give nine different instances of the framework:
- constant propagation/folding [used for constant propagation optimization]
- expression propagation/folding (aka forward substitution)  [used for forward substitution optimization]
- copy propagation [used for copy propagation optimization]
- reaching definitions [used for e.g. dependence analysis]
- live variables [used for dead code elimination optimization]
- anticipated expressions [used for lazy code motion optimization]
- available expressions [used for lazy code motion optimization]
- postponable expressions [used for lazy code motion optimization]
- used expressions [used for lazy code motion optimization].

The constant and expression propagation phases make use of a partial evaluator to apply algebraic simplifications (e.g. the compound functions in the Stan Math library and arithmetic and logical identities).

To give an instance of the framework means to give a triple of
1. a flowgraph
2. a lattice of properties
3. a transfer function.

We have implemented transfer functions (3.) for all eight analyses above as well as the corresponding property lattices (2.). Further, we have implement a function for computing the inverse flowgraph (1.), given a flowgraph. This lets us do reverse analyses like live variables, anticipated expressions, and used expressions.

The missing part in the puzzle is code to compute a flowgraph (1.) from the Mir. @rybern is writing that code. Once we have that missing part in place, we can test many instances of these analyses easily.

As an application, showing the use of this framework, we have implemented the following optimizations relying on it:
1. a constant propagation/folding optimization (which evaluates arithmetic and logical operators)
2. an expressions propagation/folding optimization (which partially evaluates expressions, using algebraic identities e.g. in operators and math library functions); this means we (almost) do not need to write compound math library function anymore in our Stan code;
3. a dead code elimination optimization.
Missing still is the complex lazy code motion optimization, which will use the remaining analyses that have been implemented.

This PR is currently clearly lacking in tests. However, it is very hard to write sensible tests without this bit of @rybern's code.

### DIFF
--- a/lib/Mir.ml
+++ b/lib/Mir.ml
@@ -120,7 +120,7 @@ type stmt_loc =
 
 type stmt_loc_num =
   { slocn: string sexp_opaque [@compare.ignore]
-  ; stmtn: stmt_loc statement
+  ; stmtn: stmt_loc_num statement
   ; num: int }
 [@@deriving sexp, hash]
 
--- a/lib/Mir.ml
+++ b/lib/Mir.ml
@@ -45,11 +45,11 @@ and expr =
   | BinOp of expr * operator * expr
   | TernaryIf of expr * expr * expr
   | Indexed of expr * index list
-[@@deriving sexp, hash]
+[@@deriving sexp, hash, map]
 
-type adtype = Ast.autodifftype [@@deriving sexp, hash]
-type sizedtype = expr Ast.sizedtype [@@deriving sexp, hash]
-type unsizedtype = Ast.unsizedtype [@@deriving sexp, hash]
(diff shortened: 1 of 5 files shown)


seantalts:
https://github.com/stan-dev/stanc3/wiki/Software-Engineering-best-practices#usable-history :P 

If you have a habit of saving partial progress with commit, you can always `git commit --amend` or rebase commits to squash them together.

PR #70, opened 2019-03-12
Static analysis and optimization

VMatthijs:
This contains both @rybern 's work on the dependence analysis and flowgraph construction, as well as @VMatthijs 's work on the monotone framework and optimization.


seantalts:
I brought this up at the meeting but I don't think everyone was there. While you guys are designing this it would be good to keep in mind that code generation facilities might also want to generate MIR as long as the semantics are similar and then use the optimizer. This is weird for two reasons I can think of:
1) there's MIR with Vars that can't be defined in the MIR like `jacobian__` and `function__`
2) there will be function and method calls that are undefined like `in__.scalar_lb_constrain`

It would be great if there were APIs allowed you to essentially provide your own declarations or otherwise ignore things that aren't defined, something like that.

lib/Monotone_framework.ml
```diff
@@ -0,0 +1,942 @@
...
+  let (module Transfer) = propagation_transfer flowgraph_to_mir in
+  let (module Mf) =
+    monotone_framework (module Flowgraph) (module Lattice) (module Transfer)
+  in
+  Mf.mfp ()
+
+let reaching_definitions_mfp (mir : Mir.typed_prog)
+    (module Flowgraph : Monotone_framework_sigs.FLOWGRAPH
+      with type labels = int)
+    (flowgraph_to_mir : (int, Mir.stmt_loc_num) Map.Poly.t) =
```
rybern: I have a could of questions/concerns about this signature and some similar ones:
- `flowgraph_to_mir` is has lots of redundancy, since each label stores the entire corresponding subtree, right? There's another type, `(label, ('e, label) statement * 'm) Map.Poly.t` that's already constructed when Flowgraph is constructed, and doesn't have redundancy, and can also be converted back into the normal mir via `build_recursive_statement`
- It seems like `Flowgraph` and `flowgraph_to_mir` can both be directly computed from `mir`, and are therefore redundant information, as far as I can tell. That makes me wonder if this could instead just take the `mir` and compute `Flowgraph` and `flowgraph_to_mir`, or if it could take `Flowgraph`, `flowgraph_to_mir`, and `initial`

VMatthijs: On the first point, `flowgraph_to_mir : (int, Mir.stmt_loc_num) Map.Poly.t`, not `flowgraph_to_mir : (int, Mir.stmt_loc) Map.Poly.t`, so it shouldn't be storing subtrees, right?

On your second point, definitely! That is a refactor we should do.

VMatthijs:
@rybern , question: for the while and for-loops, what does the flow graph currently look like? Do you have the end of the loop pointing back to the first statement in the loop or to the loop (while/for) node itself? I think the latter is what we'd really want, right, as the loop guards will get evaluated on every iteration of the loop?

rybern:
Having the loops point to the loop statement itself is:
- obviously correct
- far easier to implement
- not what I'm currently doing, because I didn't think of it
... I'll fix that :)

VMatthijs:
Thanks! And thanks for the suggestions above. I'm sure there are many ways of refactoring the code like that, which would make things much easier to understand and maintain. I've just been changing things so often while I was exploring how to get stuff to work, which has resulted in some mess. :-/

lib/Monotone_framework.ml
rybern: Oh, `Mir.stmt_loc_num` isn't recursive on statements, got it. I lose track of the names of the various combinations of expression and statement types, but they seem necessary under the current scheme. I think it would be really nice to have a type where the type variables actually corresponded to the metadata, but that change might not be in the cards.

VMatthijs: I wouldn't exclude the possibility of that change. It's mostly that everything would need to be refactored then, which is some work.

VMatthijs:
@rybern , @seantalts , @bob-carpenter : question about the spec for optimizations. How strictly should the optimizations respect the program semantics? Many optimizations (e.g. dead code elimination, lazy code motion) might decrease the number of times an expression is evaluated. There is no problem with that if the expression is pure. However, if it has side effects, those optimizations are not semantics preserving.

Expressions currently can have three kinds of side effects in Stan:
- add to target (_lp functions)
- print to std::out
- throw an error.
I've currently written the optimizations in such a way that only the contributions to target are a part of the semantics that needs to be preserved. Optimizations can currently change the printing and error behaviour of a program. Is that OK or should the printing and error behaviour also strictly be respected? (In my mind, you should just not run the optimizer if you care about that sort of debugging behaviour. But maybe people use it for more than just debugging?)

It wouldn't be very hard (but some work) to also start respecting printing behaviour, as only user defined functions can do that (and print). Errors would be harder as all our library functions can in principle throw (when given arguments with the wrong runtime properties). I guess you could have a rule for error throwing functions that they number of evaluations on the same arguments cannot be decreased beyond once, as throwing an error is an idempotent effect. However, that seems a bit complicated to put in everywhere. In particular, it would mean no dead code elimination of math library functions, but we could still use lazy code motion on them.

What do you think is the right thing to do here?

VMatthijs:
> Having the loops point to the loop statement itself is:
> 
> * obviously correct
> * far easier to implement
> * not what I'm currently doing, because I didn't think of it
> 
> ... I'll fix that :)

Thanks for fixing it! Works like a charm.

Question: could you have another look at what you're doing for the flowgraph construction of `continue`? I suspect it's not quite right. Also, `break` is very conservative, but that's less important. I've added tests in dataflow_utils for these and have added comments about the answers I would expect.

rybern:
Makes sense to me, I'll change the loop behavior and have a look at continue/break now.

rybern:
My current understanding:
- Loop statements are preceded by:
  - The statements that come before the loop
  - The natural exit points of the loop body
  - Continue statements in the loop body
- Loop exits are:
  - The loop node itself, since the last action of a typical loop execution is to check if there are any iterations remaining
  - The statements that come before the loop, in case the loop didn't execute
  - Break statements in the loop body, since broken loops don't execute the loop statement

Is this in line with your thinking?

rybern:
I think I'm meeting your "conservative" expect test results now.

I think your non-conservative expectations basically eliminate dead code after breaks and continues, right? I could add that to the dataflow analysis with some complexity cost, but are you sure that that's where we should do it?

lib/Monotone_framework.ml
```diff
@@ -0,0 +1,958 @@
...
+      Set.Poly.singleton x
+  | Mir.Assignment (_, _) -> Errors.fatal_error ()
+  | Mir.TargetPE _ -> Set.Poly.singleton "target"
+  | Mir.NRFunApp (s, _) when String.suffix s 3 = "_lp" ->
+      Set.Poly.singleton "target"
+  | Mir.NRFunApp (_, _)
+   |Mir.Check _ | Mir.Break | Mir.Continue | Mir.Return _ | Mir.Skip
+   |Mir.IfElse (_, _, _)
+   |Mir.While (_, _)
+   |Mir.For _ | Mir.Block _ | Mir.SList _ ->
```
rybern: For should declare and assign the loop variable, right?

VMatthijs:
> My current understanding:
> 
> * Loop statements are preceded by:
>   
>   * The statements that come before the loop
>   * The natural exit points of the loop body
>   * Continue statements in the loop body
> * Loop exits are:
>   
>   * The loop node itself, since the last action of a typical loop execution is to check if there are any iterations remaining
>   * The statements that come before the loop, in case the loop didn't execute
>   * Break statements in the loop body, since broken loops don't execute the loop statement
> 
> Is this in line with your thinking?

Mostly. I'm not sure I see why the statement before the loop could be a loop exit though. Could you explain that a bit more? Otherwise, it's in line with what I had in mind.

VMatthijs:
> I think I'm meeting your "conservative" expect test results now.
> 
> I think your non-conservative expectations basically eliminate dead code after breaks and continues, right? I could add that to the dataflow analysis with some complexity cost, but are you sure that that's where we should do it?

Fantastic! Thank you! I think I agree that getting optimal treatment of breaks and continues might not be worth the investment. Of course, we can revisit that decision later if it proves to be important for applications. As long as the analysis is safe (conservative), I am happy. :-)

lib/Monotone_framework.ml
VMatthijs: Oh whoops! Thanks for spotting that. Will fix it now.

rybern:
Oh, I see, it would still check the predicate. I haven't really been thinking about the predicate as a statement. Glad you're double checking me! I'll change that next

lib/Monotone_framework.ml
rybern: I added it so that my tests would work, but I think it should also be in the kill set like an assignment, which I didn't do

VMatthijs: Ah, I see you already fixed it. Sweet!

rybern:
Okay, made that change. I left the failing expect test in Optimize so that you can confirm it's now doing what you expect.

By the way, is elimination of unrunnable for/while/if branches already included in your dead code elimination?

VMatthijs:
> By the way, is elimination of unrunnable for/while/if branches already included in your dead code elimination?

Yes! Dead code elimination already does all of that type of stuff. Is that a good design do you think or should it be separated out? I guess the one thing it doesn't do yet is eliminate the declaration of dead variables. I'll fix it to do that now.

rybern:
Nope, that sounds good, just checking if that made it in.

Under the other scheme, I was doing a little optimization when I could prove that a loop ran >0 times, but there isn't an obvious way to include that information when the loop can't be entirely skipped. It's all pretty simple now, which is nice.

VMatthijs:
@rybern , we just had an interesting discussion about the semantics that we care about (for the purposes of static analysis and optimization). The conclusions were:
- printing is not an important part of the semantics (can be discarded or rearranged by optimizations)
- rejecting (including what math library functions can do when fed a wrong input) isn't either (so it's free can to discard for optimizations).

The part about rejecting might be important for your factor graph construction as well, as it probably means it can just be ignored. @bob-carpenter explained that reject only affects what model we are calculating because we need to have it reject the metropolis proposal in order to satisfy detailed balance. However, he was saying it should never be used on purpose to specify models; instead it should just be there for error checking (the expectation should be that it never gets called). Indeed, if we have a model where reject gets evaluated often, we get terrible diagnostics anyway, so the user shouldn't trust the outputs of the model. Would that simplify your life?

rybern:
Okay, good to know about the intentions around reject. It's an interesting question though - reject _shouldn't_ be used that way, but if it is accidentally or against advice, it does affect the semantics. Doesn't ignoring it (no pun intended) still mean we'd be creating edge cases where our transformations and analysis could give incorrect results?

I don't (yet) think that removing ignore would make the factor graph stuff much simpler, I think most of the added complexity would be not moving it around in optimizations.

VMatthijs:
Yes, I'm a little bit concerned about that as well, @rybern . Maybe we should discuss it once more with @bob-carpenter . The problem is that I don't fully grasp the behaviour it would lead to if you have a model where you reject a significant number of iterations. I suspect it would just mean that every rejected iteration is listed as a divergence. I guess we tell our users not to trust their results if they get a significant proportion of divergences.
The real problem here is that most Stan Math library functions can reject as well. E.g. normal(0, sigma) rejects if sigma < 0. If we start respecting them in optimizations, that would mean that we basically cannot optimize Math library function calls, which is where most models spends most of their time. 
Not sure what the right course of action is here.

VMatthijs:
@rybern , I think this is in a decent state again since the monster merge. I've split out the analysis_and_optimization into its own library to be in line with the design that was chosen for splitting up the compiler. As we had unit tests for the optimizations that used parts of the frontend, those needed to be put in their own unit_tests library. (As we did not want a dependency between the analysis library and the frontend.)
I really hope I've not messed up your code too badly. The MIR was changed pretty drastically, and I did my best to patch all your code. It could be that some of the unit tests have nonsensical node numbering in their comments now though due to the changes to the MIR.
At least I think it's in a state again such that we can work on this simultaneously. I think the next step will be to DRY up some of our code. Feel free to have a stab at it if you like doing some coding.

VMatthijs:
@seantalts , I've isolated the contributions of this PR into its own library `analysis_and_optimization`. (This PR is dependent on https://github.com/stan-dev/stanc3/pull/132 though.)
Would it be an idea to review the contributions of this PR that fall outside `analysis_and_optimization` and then to merge it? My idea would be that `analysis_and_optimization` do not make it into the stanc binary yet and are still slightly work in progress. I think it's not any more work in progress than some of the other stuff on master though. And at least it would be isolated. @rybern and I can then take responsibility for tidying up `analysis_and_optimization` further in a second PR.

VMatthijs:
@seantalts , could you maybe have a look at this so we can merge it and @enetsee can start contributing from a new PR? Like I said, everything in the `analysis_and_optimization` and `unit_tests` folders can be ignored.
You'll note that some of the files in `test/integration/example-models` are changing. They got corrupted in the past (I think because of line ending conversion). I am resetting them to what they were before.

VMatthijs:
@bob-carpenter , I've added an optimization pass here that you might like, but I have a couple of remaining questions about it to make sure I did the right thing.
The pass calculates the optimal autodiff level (DataOnly vs Autodiffable) for each local variable in the program, using a combination of forward and reverse dataflow analyses. This means that we could generate more optimal code for local variables in some cases where there is no need to treat them as vars. 

My two questions:
1. If we have a control flow dependency (as opposed to a data flow dependency) like
```
if (some_param > 1)
  my_local_var += 1.0;
```
does that then also mean that we need to treat `my_local_var` as a `var` rather than a `double`? (I know this particular case is not differentiable, but we do allow people to write programs like that and they can be made differentiable with some care.)
I think the answer is yes, right?
2. Should transformed parameters always be treated as `var`s? Even if they never depend on parameters? For example, what about the following program:
```
parameters {
  real x;
}
transformed parameters {
  real y = 1.0;
}
```

bob-carpenter:
> On May 30, 2019, at 10:12 AM, Matthijs Vákár <notifications@github.com> wrote:
> 
> @bob-carpenter , I've added an optimization pass here that you might like, but I have a couple of remaining questions about it to make sure I did the right thing.
> The pass calculates the optimal autodiff level (DataOnly vs Autodiffable) for each local variable in the program, using a combination of forward and reverse dataflow analyses. This means that we could generate more optimal code for local variables in some cases where there is no need to treat them as vars.
> 
> My two questions:
> 
> 	• If we have a control flow dependency (as opposed to a data flow dependency) like
> if (some_param > 1)
>   my_local_var += 1.0;
> 
> does that then also mean that we need to treat my_local_var as a var rather than a double? (I know this particular case is not differentiable, but we do allow people to write programs like that and they can be made differentiable with some care.)
> I think the answer is yes, right?

Why would the control flow be different for my_local_var?  If it's not defined as a function of autodiff variables, there's nothing to pass autodiff information down to in the reverse pass.  So I would think no.

> 2. Should transformed parameters always be treated as vars? Even if they never depend on parameters? For example, what about the following program:
> 
> parameters {
>   real x;
> }
> transformed parameters {
>   real y = 1.0;
> }

I don't see any reason this couldn't be moved back up to the transformed data block.  But right now, users would expect to see that y = 1 printed out in every iteration.  But it's never going to change, so it can be fixed as data.

seantalts:
Yep! As we talked about I'll just focus on the changes to frontend, middle, and code gen and you and @enetsee and @rybern can keep hacking at the analysis stuff in master.

src/middle/Middle.ml
```diff
@@ -131,6 +149,7 @@ let gensym_enter () =
   let old_counter = !_counter in
   (gensym (), fun () -> _counter := old_counter)
 
+let gensym_reset_danger_use_cautiously () = _counter := 0
```
seantalts: Without digging into the use cases, I'm wondering if this is necessary because there is also the gensym_enter function that returns a reset function that restores the old state of the counter.

src/middle/Mir.ml
```diff
@@ -164,26 +157,39 @@ type ('e, 's) prog =
...
+
+type stmt_no_meta = (expr_no_meta, unit) stmt_with
+
 type stmt_loc =
   (mtype_loc_ad, (location_span sexp_opaque[@compare.ignore])) stmt_with
 [@@deriving sexp]
 
-type expr_no_meta = unit with_expr [@@deriving sexp]
-type stmt_no_meta = (expr_no_meta, unit) stmt_with [@@deriving sexp]
+type stmt_loc_num =
```
seantalts: Should the num variants be left inside the optimizer? Not sure if y'all envision use cases by third parties.

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -376,8 +381,10 @@ let pp_ctor ppf (p : typed_prog) =
   in
   let rec pp_statement_zeroing ppf s =
     match s.stmt with
-    | Decl {decl_id; decl_type; _} ->
-        pp_set_size ppf (decl_id, decl_type, DataOnly)
+    | Decl {decl_id; decl_type; _} -> (
+      match decl_type with
+      | Sized st -> pp_set_size ppf (decl_id, st, DataOnly)
+      | Unsized _ -> () )
```
seantalts: Should this throw an exception? Or ... idk, somehow () seems unlikely to be correct, but not sure

src/middle/Middle.ml
VMatthijs: This is only meant for unit tests for optimizations. The idea is to reset it at the start of a unit test, so the results of the unit test are not dependent on the order that the unit tests are run in. Maybe it should have something like: only for use in testing in its name or doc?

I guess one could also use gensym_enter before running the test and then call reset again after the test is done. You're kind of relying on someone not forgetting to reset though and it seems harder to remember to write two things, rather than one, but don't feel so strongly about this.

src/middle/Mir.ml
VMatthijs: I'll move them to the optimizer.

src/middle/Middle.ml
seantalts: Hmm, gotcha. This makes sense, though ordinarily you'd use a test fixture for this. We might even consider making a monad for it (tagging @enetsee :P)

seantalts:
Hey! I'm gonna go ahead and merge this to master along with the other 2 PRs and then do that reorganization we talked about, breaking out Statement_gen.ml and some kind of Utils module (probably in Middle). I'm thinking that will be easier for us all to have all the code in one place and I'll trust each of us to keep on top of whatever little TODOs we have left for each PR :)

VMatthijs:
Thanks! Will submit a PR for the minor fixes you asked for.

Issue #69, opened 2019-03-11
Add type (and location?) information to the Mir.


VMatthijs:
Currently, the type information for expressions that is collected during type checking is thrown away when passing to the Mir. For some optimizations, like certain algebraic rewrites, this type information is needed, so it is probably a good information to put type annotations on expressions in the Mir.

We want to e.g. do an algebraic simplification
```
 | "log", [BinOp (FunApp ("exp", [x]),Minus,FunApp ("exp", [y]))] ->
    FunApp ("log_diff_exp", [x;y])
```
but can only do that if `x` and `y` are of type `real`, as `log_diff_exp` is otherwise not implemented.

Another example where we even would like (optional) container sizes on expressions is when we want to statically evaluate some calls to `size`, `rows` and `cols`.

seantalts:
unsized types on all expressions? I am wrestling with something that could use that too - particularly indices. But I was thinking of trying to convert to just object-level variants as needed, e.g. a `MatrixSingle` index so we know to use `()` instead of `[]` in codegen. I might still want to do that even if we add type information to the MIR, not sure.

Would you mind giving an example in the top-level comment in this PR that illustrates how it would be useful?

VMatthijs:
Sure. We want to e.g. do an algebraic simplification
```
 | "log", [BinOp (FunApp ("exp", [x]),Minus,FunApp ("exp", [y]))] ->
    FunApp ("log_diff_exp", [x;y])
```
but can only do that if `x` and `y` are of type `real`, as `log_diff_exp` is otherwise not implemented.

Another example where we even would like (optional) container sizes on expressions is when we want to statically evaluate some calls to `size`, `rows` and `cols`.

bob-carpenter:
>  We want to e.g. do an algebraic simplification
> 
>  | "log", [BinOp (FunApp ("exp", [x]),Minus,FunApp ("exp", [y]))] ->
>     FunApp ("log_diff_exp", [x;y])
> 
> but can only do that if x and y are of type real, as log_diff_exp is otherwise not implemented.

Real or assignable to real.  Is there a name for the set of things that can be assigned?  If we had a lattice of types, the inverse relation's a filter.

seantalts:
[...]

PR #73, opened 2019-03-16
Add unsizedtypes and location_spans to the MIR; get rid of MIR operators

seantalts:


seantalts:
Matthijs do you wanna take a look? I guessed at a bunch of stuff here and would love a second pair of eyes who knows the semantic checker etc better.

lib/Ast_to_Mir.ml
```diff
@@ -3,51 +3,46 @@
...
-      let msg = " should have been transformed to a FunApp by this point." in
-      raise_s [%message (op : Ast.operator) msg]
+(* XXX fix exn *)
+let unwrap_return_exn = function
+  | Some (Ast.ReturnType ut) -> ut
+  | x ->
+      raise_s [%message "Unexpected return type " (x : Ast.returntype option)]
+
+let rec op_to_funapp op args =
+  { texpr= FunApp (Operators.operator_name op, trans_exprs args)
```
rybern: One downside of this will be when we want to split up terms in a target expression to build the factor graph, we'll want to match on Plus, but instead we can string match on the name.

Also, just checking, can the user define a "plus" function that might be being shadowed?

seantalts: Yeah, I think that's okay. We can always add a constant string that has the operator names if that helps, or something like that.

> Also, just checking, can the user define a "plus" function that might be being shadowed?

Good catch, that is a problem (well, "Plus"). I guess they need to end with __?

lib/Mir.ml
```diff
@@ -101,35 +84,48 @@ and 's statement =
...
   *)
   ; log_prob: 's list
-  ; gen_quant_vars: top_var_table
+  ; gen_quant_vars: 'e top_var_table
   ; generate_quantities: 's list
   ; prog_name: string
   ; prog_path: string }
 [@@deriving sexp]
 
+type expr_typed_located =
```
rybern: It might be useful to define a generic recursive record type for expression trees with metadata and statement trees with metadata, so that functions can be polymorphic across types of metadata. Something like:

```
type 'm expr_with = { expr: (expr_with 'm) expr; with: 'm }
type 'm stmt_with = { stmt: (stmt_with 'm) statement; with: 'm }
```

So then e.g. expr_typed_located could be written as
`expr_type_location expr_with`
where expr_type_location is a type that contains the type, loc, and adlevel. @VMatthijs I think you might have run into this too?

lib/Ast_to_Mir.ml
rybern: > I guess they need to end with __?
I think so? Not sure how that's done.

lib/Mir.ml
seantalts: Ahhhh yeah we were trying to figure out what to do about this!!!! We even had a long conversation about it. Does this solve that recursive problem we keep having where e.g. pp_statement used to not give a shit about the metadata but when I had to do something different for UDFs with a `Skip` body, that broke down: https://github.com/stan-dev/stanc3/blob/master/lib/Stan_math_code_gen.ml#L246

rybern: I think it would, if I understand the issue

lib/Ast_to_Mir.ml
VMatthijs: User defined functions cannot shadow any function names in our math library. Semantic checking would reject that. So as long as the function names are chosen such that users would never choose them, eg ending in __, we'd be fine

lib/Mir.ml
VMatthijs: +1. This should give you some limited sense of row polymorphism at the cost of one extra level of in direction in the meta data.

seantalts: Yeah, I don't love that extra level. And if we keep using records and just constellation of metadata, we can just keep adding and removing stuff and using the `; _}` pattern to avoid caring about the rest of the record. But on the other hand testing these functions would be much easier if they were polymorphic over metadata. But also even for `pp_expr` we need to know that we have the type of the expression for printing the indices, so I'm not sure any of them can be polymorphic over any metadata.

VMatthijs: Right. It does sound like this wouldn't solve the problem for expressions. That's a pity.

lib/Mir.ml
```diff
@@ -13,76 +13,59 @@ open Core_kernel
 
 type litType = Int | Real | Str [@@deriving sexp, hash]
```
VMatthijs: I am not sure it is a good idea to fully remove the operator type. In particular, `Or` and `And` do not behave like functions semantically, as they are lazy in their second argument. A lot of optimizations need to special case on them for that reason to make sure the optimizations are semantics preserving.

bob-carpenter:
And these are the only two that are lazy in Stan (in addition to the ternary/conditional).

lib/Mir.ml
seantalts: We talked about this in the meeting though - we can roughly as easily pattern match on `| FunApp("Or", lhs :: rhs :: [])` as `BinOp(lhs, Or, rhs)`. I assume the optimizer is keeping track of which functions are commutative and generally not assuming that order doesn't matter right?

lib/Ast_to_Mir.ml
seantalts: ok, pushed a commit to have them end in __

lib/Mir.ml
seantalts: I think it's probably worth spending some time playing around with it, so I will do that now and report back...

lib/Mir.ml
```diff
@@ -13,76 +13,59 @@ open Core_kernel
...
-  | Return of expr option
+  | Return of 'e option
   | Skip
-  | IfElse of expr * 's * 's option
-  | While of expr * 's
+  | IfElse of 'e * 's * 's option
+  | While of 'e * 's
   (* XXX Collapse with For? *)
-  | For of {loopvar: expr; lower: expr; upper: expr; body: 's}
+  | For of {loopvar: 'e; lower: 'e; upper: 'e; body: 's}
```
VMatthijs: I still don't understand why the loopvar is an expression here, rather than a plain string.

lib/Mir.ml
```diff
@@ -13,76 +13,59 @@ open Core_kernel
...
   (* A Block for now corresponds tightly with a C++ block:
      variables declared within it have local scope and are garbage collected
      when the block ends.*)
   | Block of 's list
   (* An SList does not share any of Block's semantics - it is just multiple
      (ordered!) statements*)
   | SList of 's list
-  | Decl of {decl_adtype: adtype; decl_id: string; decl_type: unsizedtype}
+  | Decl of {decl_adtype: autodifftype; decl_id: string; decl_type: unsizedtype}
```
VMatthijs: Still not 100% sure whether we'll need sizes on these types. My sense now is that the sizes here should be optional. For function inlining (which can be important in combination with other optimizations and dependence analysis), it is important to be able to have decls without sizes. However, for optimization purposes, it could be good to also have optional sizes. Do we want to add those sizes later?

@bob-carpenter , @seantalts , what do you think?

lib/Semantic_check.mli
```diff
@@ -12,3 +12,9 @@ val check_that_all_functions_have_definition : bool ref
 val model_name : string ref
 (** A reference to hold the model name. Relevant for checking variable
     clashes and used in code generation. *)
+
+val inferred_unsizedtype_of_indexed :
```
VMatthijs: Could you add a docstring here?

lib/Stan_math_code_gen.ml
```diff
@@ -20,22 +20,13 @@ open Core_kernel
...
-  | Equals -> "=="
-  | NEquals -> "!="
-  | Less -> "<"
-  | Leq -> "<="
-  | Greater -> ">"
-  | Geq -> ">="
-
-let zero = Lit (Int, "0")
+let internal_expr =
+  { texpr= Var "UHOH"
```
VMatthijs: What's the use of this definition?

lib/Mir.ml
```diff
@@ -101,35 +84,48 @@ and 's statement =
    We'll need to scan the list of declarations for top-level ones and essentially remove them
    from the block. The AST has an `is_global` flag that also tracks this.
 *)
-type top_var_decl =
-  {tvident: string; tvtype: sizedtype; tvtrans: transformation; tvloc: string}
+type 'e top_var_decl =
+  { tvident: string
+  ; tvtype: 'e Ast.sizedtype
+  ; tvtrans: 'e Ast.transformation
+  ; tvloc: Ast.location_span sexp_opaque }
```
VMatthijs: Could you also make sure that all these location fields are set to compare ignore? (See Ast.ml for an example.) It's useful to be able to use comparison functions on the MIR in such a way that locations do not matter.

lib/Mir.ml
seantalts: I'm forgetting the previous conversation, sorry! I can't think of a reason a string wouldn't work right now. Anyone else?

lib/Mir.ml
seantalts: These are just for local variables, which aren't declared with sizes from what I remember, right? So any sizes we'd be adding would be ones the optimizer figured out. Wouldn't that go in the metadata somewhere then? Either in the 'e expr metadata or whatever graphs and other structures y'all are building up in the dataflow analysis and optimizer passes?

VMatthijs:
Thanks for doing this, @seantalts ! I think this'll make the optimization code much nicer too. Because the expressions are now phrased as the fixpoint of a parameterized type, I will be able to use ppx_deriving.map to get a map.expr which I get then combine with a fixpoint to get a recursive map/fold operation over expressions. That'll allow me to hugely condense the optimization code, making it easier to read and maintain. Hopefully that'll be the sort of high level abstraction for the optimization passes you've been asking me for. Will do the refactor of optimization once this one is merged and will run the design by you.

Could you have a look at my questions/comments and decide how/whether you'd like to address them? Approving so you can then merge whenever you see fit.

lib/Stan_math_code_gen.ml
seantalts: There is a lot of boilerplate to declaring a new expression now, so it's nice to have something to generate from as you can see where this is used.

lib/Mir.ml
seantalts: I did all the rest of them, I just left this one out.

seantalts:
>  Because the expressions are now phrased as the fixpoint of a parameterized type, I will be able to use ppx_deriving.map to get a map.expr which I get then combine with a fixpoint to get a recursive map/fold operation over expressions.

Unfortunately I think you might be confused about what ppx_deriving.map actually provides... it just provides 'a expr -> ('a -> 'b) -> 'b expr`. I don't think that can help much? I tried do use that when we first added ppx_deriving and came away pretty disappointed. I was actually thinking we might want to get rid of that dependency... But maybe it's providing more than that function? I couldn't find others when I was messing with it.

lib/Semantic_check.mli
seantalts: Yep. I wasn't sure this was the right function so was waiting to run it by you and get it all okayed.

VMatthijs:
It's just that function that I want. I could also write it myself, but that'd be more maintenance.
You can then form a kind of recursive map from that that recursively maps a function over the fixpoint of the parameterized type constructor. Can point you to an example.
If you'd rather not have that dependence, we could also write all of our map constructs ourselves, but I think it'd be easier for maintenance and correctness to just use the ones ppx_deriving gives us. Is there a reason you'd rather not have that dependency?

lib/Ast_to_Mir.ml
VMatthijs: Thanks!

lib/Mir.ml
VMatthijs: Sure, you can also pattern match on those. You just need to remember to do so, because they will usually need special treatment.
It's also a bit slower as it will require string comparison, but that presumably doesn't matter.

lib/Mir.ml
VMatthijs: No prob! Not 100% sure I'd asked you about it before. Maybe I just wondered about it in my head.

seantalts:
I don't care too much about the dependency, sorry for the red herring. I am really curious how you're planning to use it in the optimizer. Are we adding the optimizer's metadata to the tree now?

lib/Mir.ml
VMatthijs: Local variables are declared with a size as far as I'm aware, just not with a transformation. I guess we need to enforce the size at runtime if it's declared. That would suggest optional sizes might be the way to go. For purposes of code gen/optimization, there's even `auto` in C++ if we want to be really lazy and we don't want to enforce any type constraints.

lib/Operators.ml
```diff
@@ -4,37 +4,39 @@ open Core_kernel
...
-let _ = Hashtbl.add_multi operator_names ~key:"Equals" ~data:"logical_eq"
-let _ = Hashtbl.add_multi operator_names ~key:"NEquals" ~data:"logical_neq"
-let _ = Hashtbl.add_multi operator_names ~key:"Less" ~data:"logical_lt"
-let _ = Hashtbl.add_multi operator_names ~key:"Leq" ~data:"logical_lte"
-let _ = Hashtbl.add_multi operator_names ~key:"Greater" ~data:"logical_gt"
-let _ = Hashtbl.add_multi operator_names ~key:"Geq" ~data:"logical_gte"
-let _ = Hashtbl.add_multi operator_names ~key:"PNot" ~data:"logical_negation"
-let _ = Hashtbl.add_multi operator_names ~key:"Transpose" ~data:"transpose"
-let _ = Hashtbl.add_multi operator_names ~key:"TernaryIf" ~data:"if_else"
+let _ = Hashtbl.add_multi operator_names ~key:"PMinus__" ~data:"minus"
```
VMatthijs: Hmm, I was expecting the rhs to have the underscores here and to be used as the name for code gen. (It would also need to be changed in Stan_math_signatures then.) I need to think about this a bit more.

lib/Semantic_check.mli
VMatthijs: This is definitely the function I was thinking of. It seems right to me.

lib/Operators.ml
seantalts: if it was on the RHS or just in code gen, we wouldn't be able to tell between our operator and a user calling their own "Plus" function in the MIR.

VMatthijs:
> Are we adding the optimizer's metadata to the tree now?
What metadata? So far, the optimizer doesn't collect any metadata. I guess the static analysis passes collect metadata, but they do so in a Map and that data is used for an optimization immediately after it is collected and then discarded because the optimization invalidates it.
(Example of the metadata that is collected during static analysis for optimization and stored in Maps: reaching definitions, available expressions, constants, live variables.)

seantalts:
Yep, that metadata (I think "optimizer" typically also refers to the associated data flow analyses etc, but I could be wrong). When do we get sizes on local variables?

seantalts:
Wait, sorry, wrong thread! Can you just show me how you're planning to use the map? I'm just curious- I haven't found any use for it besides updating metadata on a tree.

seantalts:
Oh, I see it now above. Yay github not refreshing the page.

seantalts:
I see your example now - this makes it look like `ppx_deriving.map` is not providing `'a statement -> ('a -> 'b) -> 'b statement` but rather `'a statement -> ('a statement -> 'b statement) -> 'b statement`? Or something like that? what does the "rec" in all of the function names refer to?

seantalts:
My editor is still telling me it's only 
image

lib/Mir.ml
seantalts: Oh right it's just the function parameters that don't have sizes! We can make a new type type that has optional sizes and change the Decl to have one of those.

seantalts:
I think I addressed most of these, going to merge now, but we can keep the conversation going. Still trying to understand how the map_statement function becomes useful as I would like to use a version like that myself :P

VMatthijs:
@seantalts , @rybern , @bob-carpenter , could you weigh in on whether you think it is a good idea to have locations on expressions in the MIR? This PR added them because I think Sean interpreted Bob as saying that he needed/wanted them, but now Bob seems to say he doesn't need them after all. (To be clear: we definitely want types and adlevels. They have already been super useful for partial evaluation, if nothing else.)

My reason for doubting whether we want locations is that we often want to compare expressions for equality and forms Sets and Maps of them in the optimization passes (e.g. CSE or lazy code motion). The relevant notion of equality then is one that ignores locations. That means that if we have locations on expressions, we'd have to jump through hoops to get the location ignored or to just create a new type of expression without location for optimization purposes.

Any thoughts/feelings? Thanks, guys!

VMatthijs:
I reasked this question at issue https://github.com/stan-dev/stanc3/issues/69. Perhaps that's a better place to have this discussion.

PR #82, opened 2019-03-22
MIR pretty printer

enetsee:
Not sure if this will be terrifically useful but I put together a `fmt` based pretty printer for the data types in the MIR module. I wasn't sure how to represent a couple of things (e.g. `MutliIndex`) so I freestyled.

I got things like indenting to work pretty well so, if nothing else, that might be a useful point of reference.

### DIFF
--- a/lib/Mir.ml
+++ b/lib/Mir.ml
@@ -11,7 +11,11 @@ open Core_kernel
        - mark FnApps as containing print or reject
 *)
 
-type litType = Int | Real | Str [@@deriving sexp, hash]
+type litType =
+  | Int
+  | Real
+  | Str
+[@@deriving sexp, hash]
 
 and 'e index =
   | All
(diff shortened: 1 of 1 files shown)


bob-carpenter:
Thanks, @enetsee.  It's useful to see this spelled out and I'm sure it's going to be useful in testing, as everything is string based.

But I'm a little reluctant to try to review this given my inexperience with OCaml.

lib/Mir.ml
```diff
@@ -73,59 +210,355 @@ and ('e, 's) statement =
...
+  | IfElse(pred,s_true,Some s_false) ->
+      Fmt.pf ppf
+        {|@[<v2>@[%a(%a)@] {@;%a@]@;@[<v2>@[} %a@] {@;%a@]@;}|}
+        pp_builtin_syntax "if"
+        pp_e pred
+        pp_s s_true
+        pp_builtin_syntax "else"
+        pp_s s_false
+
+  | IfElse(pred,s_true,_) ->
```
seantalts: Can we make the _ a None just to be explicit?

seantalts:
@enetsee If you don't mind I'll wait to review this until it's been formatted - I find it a little faster that way and I don't have to think about comments on style.

enetsee:
Sure - sorry about that, I have ocamlformat running after save now! I'll wait until your changes to MIR are merged and send a new PR.

PR #85, opened 2019-03-24
Restore order to the MIR 

seantalts:
Matthijs brought up at the meeting on Monday that we need to keep the declarations in order, and that it would generally be more elegant for everything in the optimizer to deal with declarations inline rather than separately even for global variables. This PR restores order and puts the declarations back inline at the cost of a bunch of additional "internal functions" with special names that will pass through the optimizer untouched but be special-cased in code gen, e.g. this is now the MIR for a parameter in the model block (`matrix<lower=0>[10, 20] mat[5];`):
```
(((Decl (decl_adtype AutoDiffable) (decl_id mat)
       (decl_type (SArray (SMatrix (Lit Int 10) (Lit Int 20)) (Lit Int 5))))
      (For (loopvar sym1__) (lower (Lit Int 0))
       (upper (FunApp Length__ ((Var mat))))
       (body
        (Block
         ((Assignment (Indexed (Var mat) ((Single (Var sym1__))))
           (FunApp Constrain__
            ((FunApp ReadParam__ ((Lit Str mat) (Lit Int 10) (Lit Int 20)))
             (Lit Str lb) (Lit Str matrix) (Lit Int 0))))))))))
```

(that's a function application for a function named "Constrain__" and one named "ReadParam__").

There are regressions in code gen but putting this up to get eyes on it early since it's kind of a large change.

Other changes: 
* Made S-Expression output for the MIR much simpler, elliding locations and types all the time. Makes for much less verbose testing output.
* Adds sizes back to Decls on statements. We will need to add another statement variant for a decl with unsizedtype.
* play test for "internal functions" and pattern matching on them, see

```
+  | NRFunApp (fname, {texpr= Lit (Str, check_name); _} :: args)
+    when fname = fnCheck ->
+      let args = {internal_expr with texpr= Var "function__"} :: args in
+      pp_statement ppf {stmt= NRFunApp (check_name, args); sloc}
   | NRFunApp (fname, args) ->
       let fname, extra_args = trans_math_fn fname in
       pf ppf "%s(@[<hov>%a@]);" fname (list ~sep:comma pp_expr)
         (extra_args @ args)
-  | Check (fname, args) ->
-      let args = {internal_expr with texpr= Var "function__"} :: args in
-      pp_statement ppf {stmt= NRFunApp (fname, args); sloc}
```


lib/Mir.ml
```diff
@@ -65,49 +63,39 @@ and ('e, 's) statement =
...
   ; prepare_params:
       's list
-      (* XXX too intimately tied up with stan reader.hpp and writer.hpp in codegen
-     TODO: codegen parameter constraining and unconstraining in prepare_params
-  *)
-  ; log_prob: 's list
-  ; gen_quant_vars: 'e top_var_table
-  ; generate_quantities: 's list
+      (* param & transformed param decls and statements *)
+  ; log_prob: 's list (* model block, assuming above blocks run in scope*)
```
VMatthijs: Not sure what this comment means

lib/Mir.ml
```diff
@@ -65,49 +63,39 @@ and ('e, 's) statement =
...
+  ; generate_quantities:
+      's list
+      (* also assumes prepare_data and prepare_params run*)
```
VMatthijs: Could you make these comments a bit more clear? Not sure I understand them.

lib/Mir.ml
```diff
@@ -65,49 +63,39 @@ and ('e, 's) statement =
   (* An SList does not share any of Block's semantics - it is just multiple
      (ordered!) statements*)
   | SList of 's list
-  | Decl of {decl_adtype: autodifftype; decl_id: string; decl_type: unsizedtype}
+  | Decl of
+      { decl_adtype: autodifftype
+      ; decl_id: string
+      ; decl_type: 'e sizedtype }
```
VMatthijs: Could you also add an unsized decl here?

lib/Mir.ml
```diff
@@ -65,49 +63,39 @@ and ('e, 's) statement =
...
-  ; log_prob: 's list
-  ; gen_quant_vars: 'e top_var_table
-  ; generate_quantities: 's list
+      (* param & transformed param decls and statements *)
+  ; log_prob: 's list (* model block, assuming above blocks run in scope*)
+  ; generate_quantities:
+      's list
+      (* also assumes prepare_data and prepare_params run*)
+  ; transform_inits: 's list
+  ; output_vars: 'e io_var list
```
VMatthijs: Cool! Is the new idea that global variables will only be written out to the sample if they are in this list?

lib/Ast_to_Mir.ml
```diff
@@ -35,8 +33,8 @@ and trans_expr
         | FunApp ({name; _}, args) | Ast.CondDistApp ({name; _}, args) ->
             FunApp (name, trans_exprs args)
         | GetLP | GetTarget -> Var "target"
-        | ArrayExpr eles -> FunApp ("make_array", trans_exprs eles)
-        | RowVectorExpr eles -> FunApp ("make_rowvec", trans_exprs eles)
+        | ArrayExpr eles -> FunApp (fn_make_array, trans_exprs eles)
```
VMatthijs: +1

lib/Ast_to_Mir.ml
```diff
@@ -98,21 +95,245 @@ let unquote s =
...
+  in
+  let texpr_type =
+    Semantic_check.inferred_unsizedtype_of_indexed e.texpr_loc e.texpr_type
+      [(i, UInt)]
+  in
+  {e with texpr; texpr_type}
+
+(** [mkfor] returns a MIR For statement that iterates over the given expression
+    [iteratee]. *)
+let mkfor ut bodyfn iteratee sloc =
```
VMatthijs: why are you passing in the type separately? why not just use iteratee.texpr_type?

lib/Mir.ml
seantalts: Can I let you do that since you need it for the optimizer?

lib/Mir.ml
seantalts: log_prob code assumes that data and params are already in scope and ready, maybe that's a better way to say it. This happens by running prepare_params and prepare_data first.

lib/Mir.ml
seantalts: Yep.

lib/Ast_to_Mir.ml
seantalts: this code predates expressions having types. I'll see if I can update it.

lib/Ast_to_Mir.ml
```diff
@@ -98,21 +95,245 @@ let unquote s =
...
+  | UArray t -> mkfor ut (fun e -> for_scalar t bodyfn e sloc) var sloc
+  | UFun _ | UMathLibraryFunction ->
+      raise_s [%message "Can't for over " (ut : unsizedtype)]
+
+(** [for_eigen unsizedtype...] generates a For statement that loops
+    over the eigen types in the underlying [unsizedtype]; i.e. just iterating
+    overarrays and running bodyfn on any eign types found within.*)
+let rec for_eigen ut bodyfn var sloc =
+  match ut with
+  | Ast.UInt | UReal | UVector | URowVector | UMatrix -> bodyfn var
```
VMatthijs: Not sure I understand what bodyfn is going to be in practice, here.

lib/Ast_to_Mir.ml
```diff
@@ -98,21 +95,245 @@ let unquote s =
...
+
+let internal_read_fn dread args =
+  match dread with
+  | ReadData -> FunApp (fn_read_data, args)
+  | ReadParam -> FunApp (fn_read_param, args)
+
+type decl_context =
+  { dread: readaction option
+  ; dconstrain: constrainaction option
+  ; dadlevel: autodifftype }
```
VMatthijs: dread and dadlevel, classic names!

lib/Mir.ml
VMatthijs: Right! Maybe put that in the code comment?

lib/Mir.ml
VMatthijs: Sure. There would need to be code gen for it at some point though. I guess we can figure that out later.

seantalts:
Also what do you think about the high level design here? I think two things I forgot to mention are that I still needed to keep track of lists of the inputs and outputs to 1) generate the class's private (data) fields and 2) know what to write out. I'm not 100% sure I need the io_var block thing to have Data as an option, probably the input variables are always data so I could remove that. We might also change the way I'm storing size to destructure the sizedtype a little once someone gets around to code gen for write_array.

Also the transform_inits is new - it just reads constrained params in, unconstrains them, and returns the unconstrained versions.

lib/Mir.ml
VMatthijs: That's great. That immediately solves the problem that variables introduced by lazy code motion for instance should not be written out.

VMatthijs:
I really like the high level design, as far as I understand it thus far. It's still a lot of code to wrap my head around, so not sure to which level of detail I should review.

seantalts:
Okay that's good! I think any level you have appetite for is fine, just with an eye towards getting Bob and Michael unblocked at least on the MIR stuff.

VMatthijs:
Updated my comment above.

The places that are more difficult for me to follow are for instance the `mkfor`, `for_eigen` and `gen_check` functions in Mir_to_ast. I think that there partly just needs to be some complexity there that I haven't quite wrapped my head around.

But I think I understand your high level design and I like it.

seantalts:
Hmm, yeah. Those do seem too complicated.

PR #88, opened 2019-03-27
MIR Pretty Printer

enetsee:
Adding a pretty printer for MIR. There are a couple of choices I made which need review:

~~1.`pp_io_var` adds the name of the `io_block` in double angle brackets after the declaration e.g. `int x <<data>>;`~~
1. `pp_io_var` now pretty prints the `io_block` before the declaration in line with @bob-carpenter 's proposal [here](https://discourse.mc-stan.org/t/generalizing-the-stan-language-for-stan-3/1599?source_topic_id=1781) and SlicStan
2. `pp_index` represents a `MultiIndex` with a tilde (`~`) followed by the index expression

### DIFF
--- a/lib/Mir.ml
+++ b/lib/Mir.ml
@@ -282,21 +282,21 @@ let pp_transform_inits pp_s ppf {transform_inits; _} =
 
 let pp_prog pp_e pp_s ppf prog =
   Format.open_vbox 0 ;
-  (pp_functions_block pp_s) ppf prog ;
+  pp_functions_block pp_s ppf prog ;
   Fmt.cut ppf () ;
-  (pp_input_vars pp_e) ppf prog ;
+  pp_input_vars pp_e ppf prog ;
   Fmt.cut ppf () ;
-  (pp_prepare_data pp_s) ppf prog ;
+  pp_prepare_data pp_s ppf prog ;
   Fmt.cut ppf () ;
-  (pp_prepare_params pp_s) ppf prog ;
+  pp_prepare_params pp_s ppf prog ;
   Fmt.cut ppf () ;
-  (pp_log_prob pp_s) ppf prog ;
+  pp_log_prob pp_s ppf prog ;
   Fmt.cut ppf () ;
-  (pp_generate_quantities pp_s) ppf prog ;
+  pp_generate_quantities pp_s ppf prog ;
   Fmt.cut ppf () ;
-  (pp_transform_inits pp_s) ppf prog ;
+  pp_transform_inits pp_s ppf prog ;
   Fmt.cut ppf () ;
-  (pp_output_vars pp_e) ppf prog ;
+  pp_output_vars pp_e ppf prog ;
   Format.close_box ()
 
 type expr_typed_located =
--- a/lib/Mir.ml
+++ b/lib/Mir.ml
@@ -236,10 +236,10 @@ let pp_io_block ppf = function
 type 'e io_var = string * ('e sizedtype * io_block) [@@deriving sexp]
 
 let pp_io_var pp_e ppf (name, (sized_ty, io_block)) =
-  Fmt.pf ppf "@[<h>%a %s %a;@]" (pp_sizedtype pp_e) (sized_ty, Ast.Identity)
+  Fmt.pf ppf "@[<h>%a %a %s;@]" 
+    pp_io_block io_block
+    (pp_sizedtype pp_e) (sized_ty, Ast.Identity)
     name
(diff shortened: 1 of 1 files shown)


lib/Mir.ml
```diff
@@ -74,15 +128,127 @@ and ('e, 's) statement =
...
+      Fmt.pf ppf {|%a%a %s;|} pp_autodifftype decl_adtype (pp_sizedtype pp_e)
+        (decl_type, Ast.Identity) decl_id
+  | FunDef {fdrt; fdname; fdargs; fdbody} -> (
+    match fdrt with
+    | Some rt ->
+        Fmt.pf ppf {|@[<v2>%a %s%a {@ %a@]@ }|} pp_unsizedtype rt fdname
+          Fmt.(list pp_fun_arg_decl ~sep:comma |> parens)
+          fdargs pp_s fdbody
+    | _ ->
+        Fmt.pf ppf {|@[<v2>%s%a {@ %a@]@ }|} fdname
```
seantalts: Don't you want to print "void" or something here? or

lib/Mir.ml
```diff
@@ -96,18 +262,66 @@ type ('e, 's) prog =
...
+  ; prepare_params: 's list (* param & tparam decls and statements *)
+  ; log_prob: 's list (*assumes data & params are in scope and ready*)
+  ; generate_quantities: 's list (* assumes data & params ready & in scope*)
+  ; transform_inits: 's list
+  ; output_vars: 'e io_var list
+  ; prog_name: string
+  ; prog_path: string }
+   *)
+let pp_prog pp_e pp_s ppf prog =
+  Fmt.pf ppf "@[<v>@;%a@;%a@;%a@;%a@;%a@;%a@;%a@;%a@]"
```
seantalts: I always wonder if there's some more readable way to format this. For example I think I might find this pattern slightly more digestible if also more repetitive:
https://github.com/stan-dev/stanc3/blob/master/lib/Stan_math_code_gen.ml#L472

lib/Mir.ml
```diff
@@ -35,10 +35,64 @@ and 'e expr =
...
+and pp_index pp_e ppf = function
+  | All -> Fmt.char ppf ':'
+  | Single index -> pp_e ppf index
+  | Upfrom index -> Fmt.pf ppf {|%a:|} pp_e index
+  | Downfrom index -> Fmt.pf ppf {|:%a|} pp_e index
+  | Between (lower, upper) -> Fmt.pf ppf {|%a:%a|} pp_e lower pp_e upper
+  | MultiIndex index ->
+      (* TODO: I'm not sure what a multi-index is so this formatting probably
+          makes no sense... *)
+      Fmt.pf ppf {|~%a|} pp_e index
```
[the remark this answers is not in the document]
enetsee: Ok - I guess there is nothing special to do here so I will go ahead and remove the tilde.

lib/Mir.ml
enetsee: Sure; that will also simplify the pretty printing logic since I can just default the return type to `void` when it is `None`.

lib/Mir.ml
enetsee: Agreed; I actually hit this being hard to read when implementing. I'll have a look at alternatives and follow the style in the linked file if I don't have anything better.

lib/Mir.ml
```diff
@@ -96,18 +257,66 @@ type ('e, 's) prog =
...
+
+let pp_generate_quantities pp_s ppf {generate_quantities; _} =
+  pp_block "generate_quantities" pp_s ppf generate_quantities
+
+let pp_transform_inits pp_s ppf {transform_inits; _} =
+  pp_block "transform_inits" pp_s ppf transform_inits
+
+let pp_prog pp_e pp_s ppf prog =
+  Format.open_vbox 0 ;
+  (pp_functions_block pp_s) ppf prog ;
```
seantalts: why this partial application like this?

enetsee: Do you mean why have functions for each element of the `prog` type? If so, this is is to support pretty printing only parts of the MIR.

If you mean partial application in the FP sense, I can't see where I'm doing it in the lines highlighted and it wouldn't have intentionally do this!

seantalts: FP sense. without partial application
```
   (pp_functions_block pp_s) ppf prog ;
```
becomes
```
   pp_functions_block pp_s ppf prog ;
```
unless I'm missing something?

enetsee: Gotcha - I editted from the previous version using the format string and didn't drop the parentheses. I'll remove

PR #89, opened 2019-03-29
Move types common to AST and MIR into a separate module

enetsee:
**Note**: this PR is based on #88 

This PR moved the types common to the AST and MIR into a separate module; the types are:

- `location`
- `location_span`
- `autodifftype`
- `unsizedtype`
- `returntype`
- `sizedtype`
- `transformation`

I have also moved functions acting on those types to this module. The AST and MIR have been updated to refer to these types directly rather than aliasing them.

This refactoring is the first step in moving towards the project structure proposed in #83 .


seantalts:
Sweet. What's the plan for this Ast_Mir_common file when we make the MIR its own library? It would make sense to me if those things were just part of the MIR starting now...?

enetsee:
I was looking at breaking this and other types currently in MIR out into multiple modules rather than having a monolithic `MIR.ml`. I don't know whether it's cleaner to do it now or after I've added the MIR project and don't mind either way.

seantalts:
Ah. I have PTSD from a bunch of places in Stan and Stan Math where we've really over aggressively broken stuff out into different files, so I'd rather just keep the MIR as a single file.

enetsee:
Understood; I still think it makes sense to at least call these out as it highlights the dual use. Again, I don't mind either way.

enetsee:
@seantalts I've been thinking about this a little more and have put together a proposal on 
how types *could* be separated into modules in a sane way. I put a sketch indicating what I'm thinking here: https://sketch.sh/s/hxtpcLAmIlNuw2WM1ZjTcj/

The result would be distinct modules for:

- Location
- Transformation
- Type
- Expr
- Statement
- Prog

For me, this organization is easier to follow without causing too much file-hopping.

I also read a little more on Jane Street's approach [here](https://blog.janestreet.com/core-principles-uniformity-of-interface/)  and the motivation for putting types in their own modules where appropriate.

Looking at the Stan-Math library, it looks like many of the files actually correspond to union cases if they were implemented in OCaml so (hopefully) we would never get to that level of granularity.

enetsee:
Note that all these modules could still live in a single file; this is the pattern used [here](https://github.com/ocaml/merlin/blob/master/src/lsp/protocol.ml).

seantalts:
Nice. Will review today, sorry!

seantalts:
Hey, I actually like what you've done in the sketch a lot. I think having them in separate files is a good idea since we're going to have the MIR be in its own library, right? Do you want to switch this PR over to do that?

seantalts:
Question: what's the time frame on the above do you think? Should I merge #91 before or after? I feel bad making you deal with merge conflicts again so I'm happy to wait on that.

enetsee:
I would say go ahead and merge #91. I think  agreeing how to do this might take a bit more discussion (and getting to that point first will ultimately entail less work). I also spotted that `Transformation` should be AST only so the PR needs changing even as-is...

seantalts:
Word. I like the idea of an MIR lib with separate files for the major types and having the AST open the relevant ones. Is that basically where you want this to go?

enetsee:
Ok - I have ended up with this: https://sketch.sh/s/hxtpcLAmIlNuw2WM1ZjTcj/

There would be two files, `mir.ml` and `ast.ml`, each of which would contain multiple modules for related types. The modules in `ast.ml` reuse types defined in `mir.ml` as before.

I have followed the [`Real World OCaml` advice](https://v1.realworldocaml.org/v1/en/html/files-modules-and-programs.html):

> Core itself is a library that works hard to create uniform interfaces. Here are some of the guidelines that are used in Core.
> 
>  - A module for (almost) every type. You should mint a module for almost every type in your program, and the primary type of a given module should be called t.
> -  ...

It looks a little bit crazy at first but the uniform interfaces do actually help (I think!).

In addition, I have named the modules containing pattern functors (i.e. our top-level two-level types) with a trailing 'F' e.g. `ExprF`, `StatementF`, `ProgF`. I think it helps to make explicit what these types are

If you are happy with the approach, I'll rebase and update the PR.

seantalts:
> It looks a little bit crazy at first but the uniform interfaces do actually help (I think!).

Do you have any examples where this interface would help?

I think I want to keep the pretty printing separate from the definitions of the types - they really clutter it up and aren't useful when trying to understand the interface presented by the AST and MIR. 

I don't see anything in Core where they do anything complicated with multiple recursive types that only have meaning when presented together. The closest thing is maybe Blang which is just like 6 boolean variants. Do you know of any examples of wrapping up pieces of an AST into individual modules out in the wild? I think these types might be special because they are essentially just a shared data format that you then can write multiple pretty-printers and parsers for.

enetsee:
> I don't see anything in Core where they do anything complicated with multiple recursive types that only have meaning when presented together.

Yeah, you’re right. I’ll revert back to types in two modules.

enetsee:
@seantalts I've made the changes so we have types in either `Mir.ml` or `Ast.ml`. `transformation` now lives in the AST module.

seantalts:
Thanks for going through all possible designs with me 😂 Will take a look at this this afternoon! And I'll get it merged in before the other PR that changes the MIR again. Also if you have any time and wouldn't mind taking a look at this other PR, there is a very confusing change in there that I'm not sure how to best express in OCaml / functional style - basically switching on function name where we want to generate many of the names and not use string literals: https://github.com/stan-dev/stanc3/pull/91/files/aec2a82f5972f84ad7637b7ab84bfd017d1e9b67#diff-8ba34bdbcf669b58cacd824684203c87R214

seantalts:
Looks good, thanks!

Issue #78, opened 2019-03-19
Refactor FunDef


rybern:
As I understand it, function definitions will always (and only) appear as the top-level statements in the functions_block list. If this is the case, how about moving FunDef from a general statement to its own type?

```
type 's fun_def = FunDef of
  { fdrt: unsizedtype option
  ; fdname: string
  ; fdargs: fun_arg_decl
  ; fdbody: 's }

type ('e, 's) prog =
  { functions_block: 's fun_def list
  ; ...
```

This has three main advantages:
- It clearly expresses and enforces the invariant that function definitions only appear in the function_block,
- It shrinks the number of variants in statement, so that we don't have to handle the function definition case whenever we match on a statement, which is especially a problem when we don't expect a function definition in e.g. the model block,
- It makes it easier to pull out function definitions from a problem without writing partial functions

I'd be happy to try to work on a pull request if I get the go-ahead.

bob-carpenter:
We want to be thinking forward when they can be defined anywhere using static lexical scoping and/or lambdas with closures.  All the ones that go up front as they do now, just have nothing they can access in scope.  But if we defined such a function in the transformed data block, it should have access to data variables.

At least that's my two cents.

VMatthijs:
The question is whether closures should ever make their way into the MIR though. They'd definitely be in the AST, but we might want to get rid of them through lambda lifting before building the MIR. I'd say we want to keep the MIR as simple as possible. So ideally no closures, higher order functions, structs. Anything that can easily be treated as syntactic sugar should be translated away before the MIR.

seantalts:
So if we got rid of closures, we'd generate a functions block FunDef with more arguments, essentially, and pass in whatever data or transformed data (or stuff that's actually more of a closure like local variables) that it needs, right? Is that acceptable 100% of the time when you might otherwise write a function inline? If so then it seems fine coupled with the appropriate Ast_to_Mir code to do that conversion.

rybern:
Agh, sorry @increasechief! I forgot you mentioned you wanted to try this, I just submitted a pull request. It's possibly incomplete, I'll leave future work to you.

increasechief:
Only in the absence of anyone else caring enough or needing to jump on it. I take no possession here. :)

seantalts:
[...]

PR #92, opened 2019-04-02
Moved FunDef into its own type

rybern:
I recognize this is a low priority change, feel free to ignore this until convenient :)

I took a stab at changing the MIR to disallow inline function definitions. There's a chance I broke some stuff, but all the tests do still pass. I raised errors when a non-FunDef was found in the function block, and when a FunDef was found outside of the function block, since I don't think we support closures yet.

The one question I ran into - when function definitions were statements, they had locations (and other metadata). Is it worth it to figure out how to include the location of the function definition itself?

Note that I ran the formatter in the second commit (I should do that, right?), which makes it harder to find the substantive changes I made in the diff - it might be clearer to look at the diff for the first commit, rather than the whole pull request.

### DIFF
--- a/lib/Ast_to_Mir.ml
+++ b/lib/Ast_to_Mir.ml
@@ -335,6 +335,7 @@ let trans_decl {dread; dconstrain; dadlevel} sloc sizedtype transform
   in
   SList ((decl :: read_stmts) @ checks)
 
+
 let rec trans_stmt declc {Ast.stmt_typed; stmt_typed_loc= sloc; _} =
   let trans_stmt = trans_stmt {declc with dread= None; dconstrain= None} in
   let texpr_loc = sloc in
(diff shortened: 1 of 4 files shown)


seantalts:
> The one question I ran into - when function definitions were statements, they had locations (and other metadata). Is it worth it to figure out how to include the location of the function definition itself?

I think we talked about this in the meeting and we wanted to go ahead and just put a location on a function definition in the MIR.

rybern:
Issue: https://github.com/stan-dev/stanc3/issues/78

rybern:
@seantalts Yes! I'll spend some time on this this afternoon

I think the last commit here is ocamlformat, does it not look right? I'll try the hook as well

seantalts:
Oh, my bad. I didn't realize it would put multiple `|>` on the same line, but it totally does as long as all of them in the expression fit on the line. Never mind :)

rybern:
Resubmitted as https://github.com/stan-dev/stanc3/pull/108

PR #95, opened 2019-04-05
Modified Dockerfile to no longer need the --priviledged argument. Mod…

serban-nicusor-toptal:
…ified Jenkinsfile to reflect Dockerfile changes. Added small bat script to install opam dependencies on windows. Added readme.md in /scripts to aid in creating windows/linux build environments.


docker/dev-ubuntu/Dockerfile
```diff
@@ -1,15 +1,25 @@
-FROM ubuntu:bionic
+#Use Official Docker Images for OCAML/OPAM
+#https://github.com/ocaml/infrastructure/wiki/Containers
+FROM ocaml/opam2:debian-stable
```
VMatthijs: This just means that we're not testing our install_ocaml script anymore, but that's fine with me. @seantalts , are you also OK with that?

scripts/README.MD
```diff
@@ -0,0 +1,101 @@
...
+   opam init --disable-sandboxing -y
+   opam switch create 4.07.0
+   opam switch 4.07.0
+   eval $(opam env)
+```
+4. Add the opam repository for the cross-compile windows 
+```
+opam repository add windows git://github.com/ocaml-cross/opam-cross-windows
+```
+5. If desired, request the compiler to be built with flambda optimizers:
```
VMatthijs: It's not just if desired. We actually need to use conf-flambda-windows. For some reason, I cannot get core_kernel-windows to compile without its optimizations. I suspect that its tail-call optimization or function inlining helps avoid a stack overflow I otherwise get.

By the way, I suspect that 5. 6. and 7. can also be summarized as
```
   opam install -y core_kernel-windows menhir-windows ppx_deriving-windows fmt-windows
```
as conf-flambda-windows and ocaml-windows are already listed as dependencies for core_kernel-windows. Doesn't matter hugely either way though.

scripts/README.MD
```diff
@@ -0,0 +1,101 @@
...
+   sudo add-apt-repository ppa:avsm/ppa
+   sudo apt-get update
+```
+2. Install `opam` `mingw-w64` `gcc` `m4` dependencies required
+```
+   sudo apt-get install opam mingw-w64 gcc m4 -y
+```
+3. Init opam, create and switch to 4.07.0, update environment shell
+```
+   opam init --disable-sandboxing -y
```
VMatthijs: I'm not sure it's a good idea to encourage people to disable sandboxing by default (though it's probably fine for our build servers). I just had to resort to that because WSL and Alpine didn't seem to support bwrap.

VMatthijs:
Nice! I added some more comments. Not sure if it's useful.

serban-nicusor-toptal:
Noted! Will do the requested changes now.

scripts/README.MD
serban-nicusor-toptal: I will leave it as separate so it can be easier to understand what's under the hood. I hope that's okay with you.

scripts/README.MD
serban-nicusor-toptal: Specified this in the readme now.

docker/dev-ubuntu/Dockerfile
serban-nicusor-toptal: I've done this so we use the latest images build by the official team. This means a stable image, always up to date and reviewed by the devs of ocaml. Besides this, for us it means faster build/test times since we're not spending time building it from source, just using an already pre-build image.

serban-nicusor-toptal:
Updated: https://github.com/stan-dev/stanc3/tree/jenkins-stable/scripts

scripts/README.MD
VMatthijs: Absolutely!

VMatthijs:
Thanks, @serban-nicusor-toptal ! Looks great

docker/dev-ubuntu/Dockerfile
VMatthijs: Makes sense! Naive question [I know little to nothing about Jenkins/Docker]: in the new setup, how often would all the library dependencies have to be installed by the build/test machine? Are the installations of those being cached or would they be installed every time we run tests?

serban-nicusor-toptal: They're in cache :D

VMatthijs: Great!

seantalts: Oh yeah woops, we need to make sure the scripts are under test in at least one of these test runs. Can you revert it back to using the `install_ocaml` script?

PR #96, opened 2019-04-05
Add `_opam` to git ignore.

enetsee:
Using and opam switch creates directory named `_opam` which should be ignored by git.


seantalts:
There appear to be no changes in this PR?

PR #99, opened 2019-04-06
Create separate libraries for MIR, optimization, frontend and backend

enetsee:
This follows on from #89 and is another partial step to #83 

This PR separates the library `stanclib` into four:

- `stan_mir`: the MIR module created in #89 
- `stan_optimize`: currently empty but ready for program analysis work
- `stan_math_backend`: basically moves `Stan_math_code_gen.ml` it's own library
- `stan_frontend`: the AST and parser

The new structure is a stepping stone to a fully modular structure outlined in #83. The idea is to support some of the stated aims of the project, in particular:

- Multiple phases: having separate libraries makes clear these phases and allows us to more easily abstract over them
- Research platform: exposing the phases as separate libraries would allow research implementations of e.g. probabilistic optimizations to be dropped into the compiler easily
- Modular: this structure allows us to expose relevant parts of the library to other tools e.g. the language server work only need know about frontend and MIR.


seantalts:
Hey, sorry, finally merged that code gen PR. I started working on the merge conflicts (nothing serious!) but realized Expression_gen relies on Stan_math_signatures and is part of the stan math backend. This is a conceptual issue we need to think about. tagging @VMatthijs. Do we want to say that we are typechecking against the Stan Language function specs, and we just so happen to use mostly the same names in the Stan Math library? Or do we want to say that we will typecheck against whatever functions the backend provides? Leaning towards the former, in which case there's just a code organization issue (where do Stan function type signatures live if we need them in both the front and backend?)

enetsee:
> Do we want to say that we are typechecking against the Stan Language function specs, and we just so happen to use mostly the same names in the Stan Math library? Or do we want to say that we will typecheck against whatever functions the backend provides? 

Compiling against a language spec would be especially cool if that spec were broken up into various capabilities which backend's _may_ implement and we would statically know what capabilities a program required. This would be a nice way of introducing experimental language features which required new capabilities in the backend.

VMatthijs:
For now, the Stan documentation strongly suggests that the whole math library is an essential part of any Stan implementation. That suggests to me that it should be bundled with the frontend, rather than backend. But I can imagine that if we're ever targeting say tensorflow, that it would just not support the same library functions, so we'd have to change the signatures supported. At the same time, TensorFlow also has a pretty different type system (higher dimensional tensors) and doesn't support control flow primitives, so to target that backend we'd have to change the language anyway. So basically, not sure. When in doubt, I'd have a slight preference for not doing big refactors that would hold up the launch. But perhaps there's a good case to move the Math Signatures to the backend.

seantalts:
I guess it would be a bigger refactor to have the frontend and backend talk for the purposes of typechecking somehow, or to move the typechecking to the backend or something. So it seems easier for now / less of a change to assume that the signatures file, even though it was derived heavily from Stan Math, is actually the spec for the Stan language. That said, I don't think we can leave Stan_math_signatures in the frontend because the backend also relies on them, so I think we probably need to move that file to the MIR (and maybe rename it to Stan_signatures if we're going with that-is-the-spec).

> Compiling against a language spec would be especially cool if that spec were broken up into various capabilities which backend's may implement and we would statically know what capabilities a program required. This would be a nice way of introducing experimental language features which weren't purely syntactic sugar and required new capabilities in the backend.

I have some half-baked thoughts about this but don't really know what it would look like. How would you organize the code for this? Would the front end be a parameterized module that took in the necessary typechecking functions, and stanc.ml would wire the appropriate part of the backend to the front end?

enetsee:
> I have some half-baked thoughts about this but don't really know what it would look like. How would you organize the code for this? Would the front end be a parameterized module that took in the necessary typechecking functions, and stanc.ml would wire the appropriate part of the backend to the front end?

This is the sort of thing I was thinking but my thoughts are pretty vague at the moment too. Short term we could pull the `Stan_math_signatures` into another lib (`Stan_spec`?) and expose that way.

> But I can imagine that if we're ever targeting say tensorflow, that it would just not support the same library functions, so we'd have to change the signatures supported. 

Exactly, if/when multiple back-ends are supported, the spec would have to live outside of any particular implementation and we'd need to figure out what made sense given the peculiarities of each back-end.

seantalts:
Word.

Okay, so for now take it that those signatures are part of Stan. Expression_gen is the only place in the backend where the signatures are used, I think, and it's just to tell if a function is a Stan UDF or not. We could also just include that information in the MIR, since we have it in a less hacky way much earlier in the front-end. What do you all think about that?

seantalts:
just to update status - I think this is waiting for #103.

enetsee:
Sorry - I should be able to make progress on #103 tomorrow all being well!

seantalts:
No rush or anything, just was trying to help myself keep track of all the balls in the air :)

enetsee:
Superceded by #109

Issue #79, opened 2019-03-19
Assignment needs type information on its lhs in AST, Ast_to_Mir


VMatthijs:
Try running the following expect test. It crashes. That is because currently an operator assignment is translated (in Ast_to_mir.ml) into a plain assignment which inherits the type of the RHS, but it should be inheriting the type of the LHS.

```
let%expect_test "Operator-assign example" =
  let ast =
    Parse.parse_string Parser.Incremental.program
      {|
        model {
          real r;
          vector[2] x;
          x ./= r;
        }
      |}
  in
  let mir = trans_prog "" (Semantic_check.semantic_check_program ast) in
  print_s [%sexp (mir : Mir.typed_prog)] ;
  [%expect
    {| |} ]
```

VMatthijs:
There is a difficulty here that in the MIR the lhs of an assignment needs to have a type, while it doesn't in the AST.
To fix this issue, we could either add these types to the AST or remove them from the MIR. It seems not quite right to me that currently in the MIR an arbitrary expression can occur to the left of an assignment, which is related.
@seantalts , how shall we resolve this?

bob-carpenter:
The only lvalue in the current version of Stan is a variable with indexing.  Unless we have reference returns, this is likely to remain the case.

seantalts:
Hm, I don't think this is solved just by removing types from the MIR assignee - if we're using an operator assign, that desugars eg `a += b` to `a = a + b`, so we still need a type for `a`. Ideas?

seantalts:
I'm going to rename this issue to put the type into the AST and transfer it in Ast_to_Mir.

VMatthijs:
@seantalts , @enetsee , are you guys OK with adding type and AD-level to identifiers to fix this? There are also other places where I need to have that information on left-hand-sides. I feel like this would be the cleanest way of adding that information to the AST/MIR? We'd just have typed and untyped identifiers, similarly to how it works for expressions.

@enetsee , do you have a sense of when you'll merge your semantic check PR?

seantalts:
I don't adding type and AD-level to identifiers solves this either - the identifier may be indexed arbitrarily. I think we need the type of the lhs in Assignment specifically (which might be a good opportunity to extract out the `lhs` type we've talked about).

VMatthijs:
Hmm. I see what you mean. I guess one option is to go to the less type safe situation again where the LHS is just an arbitrary expression. Not sure that's ideal though.
[...]

PR #102, opened 2019-04-09
Add necessary information to Assignment in MIR

seantalts:
But still need to generate the correct type in the AST. Fixes the MIR half down for #79, renamed the issue to deal with the frontend half of the fix.

### DIFF
--- a/lib/Mir.ml
+++ b/lib/Mir.ml
@@ -94,7 +94,7 @@ and 'e expr =
 type fun_arg_decl = (autodifftype * string * unsizedtype) list
 
 and ('e, 's) statement =
-  | Assignment of 'e * 'e
+  | Assignment of string * 'e index list * 'e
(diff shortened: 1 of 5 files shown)


lib/Mir.ml
```diff
@@ -94,7 +94,7 @@ and 'e expr =
 type fun_arg_decl = (autodifftype * string * unsizedtype) list
 
 and ('e, 's) statement =
-  | Assignment of 'e * 'e
+  | Assignment of string * 'e index list * 'e
```
rybern: It might be useful to name this type, since it's also used in reaching definitions and dependency analysis

seantalts:
@rybern would you mind taking another look and approving if you think it's okay? granted the whole statement generation is a work-in-progress but just if you don't mind double checking the MIR. thanks!

PR #103, opened 2019-04-10
[WIP] Function application / Stan Math functions

enetsee:
I have made the minimal set of changes to:

- Remove the dependency on `Stan_math_signatures` in `Expression_gen`
- Indicate when a function is a 'Stan Math` function during semantic checking of the AST
- Change expect tests in order to 'pass' 

My initial read through of the code makes me think that there are actually quite a few rules to determine when something is 'user defined' but I haven't got a clear picture of them and this initial commit is therefore (probably) horribly wrong.

@seantalts  @VMatthijs This feels quite hacky; was this anywhere near the approach you had in mind? If not, I'll change direction. If so, can we set out the various classes of functions we have within Stan and how we should be treating them during code gen?


Jenkinsfile
```diff
@@ -15,13 +15,22 @@ pipeline {
         stage("Build & Test") {
             agent {            
                 dockerfile {
+<<<<<<< HEAD
```
seantalts: Looks like a bad merge here for some reason. I'd just take the files from master

lib/Mir.ml
```diff
@@ -65,6 +65,7 @@ let rec remove_size = function
   | SArray (t, _) -> UArray (remove_size t)
 
 type litType = Int | Real | Str [@@deriving sexp, hash]
+type funKind = StanMath | UserDefined [@@deriving compare, sexp, hash]
```
seantalts: Unfortunately there are more options here than just StanMath or UserDefined... Maybe for now I'd put UserDefined or Builtin, assuming it's not too crazy to use the word builtin to refer to math library functions, internal mir functions, Stan functions, etc?

lib/Ast.ml
```diff
@@ -50,7 +50,7 @@ type 'e expression =
   | Variable of identifier
   | IntNumeral of string
   | RealNumeral of string
-  | FunApp of identifier * 'e list
+  | FunApp of Mir.funKind * identifier * 'e list
```
seantalts: I think we also need to do NRFunApp :(

lib/Ast_to_Mir.ml
```diff
@@ -305,7 +317,10 @@ let mkdataread id var sizedtype sloc =
   let read_base var =
     { var with
       texpr=
-        FunApp (string_of_internal_fn FnReadData, [mkstring var.texpr_loc id])
+        FunApp
+          ( Mir.StanMath
```
seantalts: I wonder if we should just go ahead and have a separate funkind for these internal functions? That could be handy for switching on later, actually.

seantalts:
what feels hacky about it to you? Is it just that we're shoehorning a bunch of stuff into StanMath kind? I feel that. We could probably go ahead and add the 4 kinds I can think of if that helps (most of the code in this PR will stay the same, but the places where we generate internal function calls will change):
1. StanLibrary (i.e. Math library and Stan functions, conflated because that distinction doesn't really make conceptual sense in most cases at this level of the abstraction)
2. UserDefined
3. CompilerInternal (name possibly needs work) - referring to all those special ones we have in the MIR... some of which do actually get translated to math functions...

enetsee:
> what feels hacky about it to you? Is it just that we're shoehorning a bunch of stuff into StanMath kind?

Yeah, it's mainly that (and more precisely my lack of knowledge about what the functions are!) and the parts of `Expression_gen` where we're relying on prefixes and suffixes of strings. 

Categorization of functions seems like a good direction, I'll update and fix the merge conflict.

bob-carpenter:
The difference with user-defined functions is that they need an additional output stream argument at the end of their signature because user-defined functions can have print statements in them.

enetsee:
@seantalts I've updated `fun_kind` so that we have three kinds of functions:

- `StanLib` - Stan library functions
- `CompilerInternal` - the internal functions defined by `internal_fn` in `Mir.ml`
- `UserDefined` - all other functions

The function kind is determined by the identifier using (this function)[https://github.com/enetsee/stanc3/blob/function-application/lib/Semantic_check.ml#L509].

General checks are applied to all function kinds and then specific checks are applied once the kind is determined.

I have also refactored the part of `Semantic_check.ml` dealing with function application quite a bit. This was initially just to make it easier to read but I ended up using the `Or_error` module to deal with the semantic error exceptions. There is a bit more about it (here)[https://blog.janestreet.com/how-to-fail-introducing-or-error-dot-t/].

Does this way or writing it make sense to you? (As a sidenote, it _might_ be worth considering extending the `Or_error` validation approach to all of the semantic check code. The main benefit as I understand it, is that we could report multiple semantic errors in one sweep.)

I haven't carried these changes across to `Expression_gen.ml` yet and I haven't started work in `NRFunApp` but can make a start when I have a steer on the above.

seantalts:
Gotcha, will take a look! In the meantime if you have a moment would you mind looking over the double parameterization stuff to see if there are any suggestions you'd make? https://github.com/stan-dev/stanc3/pull/104 The main replacement types are here: https://github.com/stan-dev/stanc3/pull/104/files#diff-a40d5abea93039602b4488ca1343d674R157

docker/static/Dockerfile
```diff
@@ -3,16 +3,9 @@ FROM alpine:3.8
...
-RUN apk update && apk upgrade && \
-  apk add build-base bzip2 git tar curl ca-certificates openssl \
-  patch rsync bash libx11-dev nano bubblewrap coreutils xz ncurses-dev m4 pcre-dev
 
-# Build ocaml and then opam from scratch
-RUN git config --global user.email "docker@example.com" && \
-  git config --global user.name "Docker" && \
-  git clone -b 2.0 git://github.com/ocaml/opam /tmp/opam && \
-  sh -c "cd /tmp/opam && make cold && mkdir -p /usr/local/bin && cp /tmp/opam/opam /usr/local/bin/opam && cp /tmp/opam/opam-installer /usr/local/bin/opam-installer && chmod a+x /usr/local/bin/opam /usr/local/bin/opam-installer && rm -rf /tmp/opam" && \
-  strip /usr/local/bin/opam*
```
seantalts: Possibly another bad merge on this file, assuming you didn't mean to change it on this PR. You can check out the one on master with `git checkout origin/master docker/static/Dockerfile`.

lib/Semantic_check.ml
```diff
@@ -6,10 +6,10 @@ during parsing and are in fact irrelevant for building up the parse tree *)
 open Core_kernel
 open Symbol_table
 open Ast
-open Stan_math_signatures
```
seantalts: Why not open this one?

enetsee:
I'll try and take a look double parameterization stuff on Sunday. Also, as a heads up, I've almost finished with the reparameterization of the AST too; the expect tests are taking a while to fix but I'll also try to make a PR on Sunday.

lib/Semantic_check.ml
```diff
@@ -497,7 +304,221 @@ and inferred_unsizedtype_of_indexed loc ut typed_indexl =
...
+           signatures: %s\n\
+           Instead supplied arguments of incompatible type: %s."
+          id.name
+          (Stan_math_signatures.pretty_print_all_math_lib_fn_sigs id.name)
+          (List.map es ~f:type_of_expr_typed |> pretty_print_unsizedtypes)
+      in
+      Or_error.error_string err
+
+let fn_kind_from_identifier id =
+  if Stan_math_signatures.is_stan_math_function_name id.name then Mir.StanLib
```
seantalts: All of the names in that module are super long and redundant especially paired with the module name... Can we go back to opening it?

lib/Semantic_check.ml
```diff
@@ -573,153 +594,45 @@ and semantic_check_expression cf {expr_untyped_loc= loc; expr_untyped} =
...
+      ; expr_typed_type= UReal
+      ; expr_typed_loc= loc }
+  | FunApp (_, id, es) ->
+      let uid = semantic_check_identifier id
+      and ues = List.map ~f:(semantic_check_expression cf) es in
+      semantic_check_fn_map_rect ~loc (uid, ues)
+      |> Or_error.bind ~f:(semantic_check_fn_conditioning ~loc)
+      |> Or_error.bind ~f:(semantic_check_fn_target_plus_equals cf ~loc)
+      |> Or_error.bind ~f:(semantic_check_fn_rng cf ~loc)
+      |> Or_error.bind ~f:(semantic_check_fn ~loc)
```
seantalts: seems like we could use `val Or_error.combine_errors_unit : unit Or_error.t list -> unit Or_error.t` maybe?

seantalts:
I really like the blog post on error and or_error, especially around tagging errors. I am wondering or hoping we can find something a little less verbose - they mention that they include two functions at the top level of Core to ease the lexical burden; maybe we can do something similar.

seantalts:
Or wait, Ok is probably from opening Or_error.Monad_infix, which we don't want to do at the top level. But you get the idea. Anyway, I think this is a good idea and I like that we can present all of the type errors at once with this pattern, though in practice I feel like languages that do that actually don't do anyone any favors (looking at you, C++). 

> The function kind is determined by the identifier using this function.

At this point in the AST, there are no compiler internal functions I believe. I was under the impression that Semantic_check is responsible for basically resolving the function name to a specific type signature at this stage and that we could use that same code to determine if it's a UDF or library function. That might address the code smell too.

lib/Semantic_check.ml
enetsee: There is definitely no need to `bind` the first 4 checks since they don't depend on one another. Since we aren't operating on the input in these functions, the return type of these _should_ be `unit Or_error`.

The last function `semantic_check_fn` is actually doing something and returning a `typed_expression` with the correct `fun_kind` so we can't use `Or_error.combine_errors_unit`.

What we want is to have the effectful functions composed to drop the `unit Or_error.t` but keep the result of the last function. 

There is an infix function `*>` from `Applicative` which does what we want - sequences actions and ignores the first argument. It's a bit messy to [write in terms of `apply`](https://github.com/janestreet/base/blob/master/src/applicative.ml#L50) so I think it's worth paying the price here.

The code would look like:

```
      Or_error.(
        semantic_check_fn_map_rect ~loc (uid, ues)
        *> semantic_check_fn_conditioning ~loc (uid, ues)
        *> semantic_check_fn_target_plus_equals cf ~loc (uid, ues)
        *> semantic_check_fn_rng cf ~loc (uid, ues)
        *> semantic_check_fn ~loc (uid, ues)
        |> ok_exn 
      )
```

enetsee:
> At this point in the AST, there are no compiler internal functions I believe. I was under the impression that Semantic_check is responsible for basically resolving the function name to a specific type signature at this stage and that we could use that same code to determine if it's a UDF or library function. That might address the code smell too.

Yes, this is very true! To be clear with our types, I think we should have as AST `fun_kind` which only has UDF and Stab Library and another in MIR which includes Compiler Internal. (We could use polymorphic variants but they can introduce quite confusing types and error messages.)

enetsee:
> Anyway, I think this is a good idea and I like that we can present all of the type errors at once with this pattern, though in practice I feel like languages that do that actually don't do anyone any favors (looking at you, C++).

True; I guess having the information is one thing but reporting it usefully is a different matter..

lib/Ast.ml
enetsee: I don't know why, but I didn't think users could define side-effecting functions. I can't seem to work out how this is being handled at the moment in `Stan_math_code_gen.ml` - do you have any pointers?

seantalts:
> Yes, this is very true! To be clear with our types, I think we should have as AST fun_kind which only has UDF and Stab Library and another in MIR which includes Compiler Internal. (We could use polymorphic variants but they can introduce quite confusing types and error messages.)

That sounds good, but let's just have two types and not get into polymorphic variants yet (only duplicating two variants here).

lib/Semantic_errors.ml
```diff
@@ -0,0 +1,226 @@
+open Core_kernel
+open Pretty_printing
+open Mir
+
+(** Low-level semantic errors used for validation *)
+type t =
+  | IdentifierIsKeyword of location_span * string
+  | IdentifierIsModelName of location_span * string
```
VMatthijs: This is a super welcome refactor! I should've done it that way the first time around, obviously. Thank you!

lib/Semantic_errors.ml
```diff
@@ -0,0 +1,226 @@
...
+  | FnTargetPlusEquals of location_span
+  | FnRng of location_span
+
+let string_of_operator = mk_string_of Ast.sexp_of_operator
+let ternary_if = "TernaryIf__"
+
+(** A hash table to hold some name conversions between the AST nodes and the
+    Stan Math name of the operator *)
+let string_of_operators =
+  Map.Poly.of_alist_multi
```
VMatthijs: Is this the best place for this function? Do we not also need something similar to it elsewhere? How does it compare to the functionality in the Operators.ml file?

lib/Semantic_errors.ml
```diff
@@ -0,0 +1,226 @@
...
+  | IdentifierInUse (loc, name) ->
+      Errors.semantic_error ~loc
+        (Format.sprintf "Identifier '%s' is already in use." name)
+  | IdentifierIsModelName (loc, name) ->
+      Errors.semantic_error ~loc
+        (Format.sprintf "Identifier '%s' clashes with model name." name)
+  | IdentifierIsKeyword (loc, name) ->
+      Errors.semantic_error ~loc
+        (Format.sprintf "Identifier '%s' clashes with reserved keyword." name)
+  | IdentifierNotInScope (loc, name) ->
```
VMatthijs: Nice!

enetsee:
Ok - I think I actually managed the merge :tada:! Summary of changes:

## AST 

- I introduced a `fun_kind` type in `Ast.ml` which allows for either `StanLib` or `UserDefined` functions
- I added a `fun_kind` to the `FunApp` and `NRFunApp` data constructors of `expression` and `statement` respectively 
- I updated `parser.mly` to add `UserDefined` to both `expressions` and `statements`

## Semantic Check

I reordered function definitions so we aren't using `let rec` everywhere. This was mainly to help me understand the structure of the file but my understanding is that limiting their use to actual mutually recursive functions is preferred.

I also refactored the semantic check on `FunApp` quite significantly. A lot of the refactoring is based around the use of a data type to accumulate the results of individual semantic checks. Initially I was using `Or_error` but it didn't quite do what we want. Specifically, you are stuck using `Error.t` for your error type which meant using our own `SemanticError` exception was a bit tricky. Instead, I created a new datatype `Semantic_errors.t` which captures the granular semantic errors and a simple module for validation which accumulates errors in `Validation.ml`. At the moment, since I am only using it for `FunApp`, I eventually raise an exception on the first semantic error encountered but, ask discussed above, we could easily move all semantic checks into this and report all errors in one pass.

Semantic check is also the place where we determine if a function is `StanLib` or `UserDefined`. This boils down to checking whether a function identifier is a know stan math function using `is_stan_math_function_name`. This determination is made in the functions `semantic_check_fn` and `semantic_check_nrfn` for `FunApp` expressions and `NRFunApp` statements respectively.

## AST to MIR 

- I introduced a `fun_kind` type in `Mir.ml` which allows for either `StanLib`, `UserDefined` or `CompilerInternal` functions
- The only place `CompilerInternal` `fun_kind`s should be introduced is in `Ast_to_Mir.ml`

## Codegen

The point of this PR is to eliminate the dependency in code gen on the Stan math signatures. *I feel less comfortable that I am doing the right thing here.*

For `FunApp` expressions, we now pattern match on the `fun_kind` in `pp_expr` (within `Expression_gen.ml`) and e.g. generate the appropriate additional arguments for user defined functions

For `NRFunApp` statements we are using the `fun_kind` to change name / add extra arguments when using the `FnPrint` and `FnLength` internal functions. I don't know whether this is the intended behavior. If so, the matching I'm doing in `pp_statement` (within `Stan_math_code_gen.ml` is redundant.

lib/Semantic_errors.ml
enetsee: It definitely isn't the right place for it; I think `Operators.ml` is gone on master but I agree that this kind of thing probably lives with the type definition.

lib/Semantic_errors.ml
enetsee: I think we should consider rewriting the whole of the semantic check using these low level errors in combination with the custom validation module I've added (see my comment below on why I'm using a custom approach).

lib/Semantic_errors.ml
seantalts: This functionality still exists in the AST: https://github.com/stan-dev/stanc3/blob/c559488e688c316115650f12d26a2b5e27ed4072/lib/Ast.ml#L292

seantalts: Or no, this is for math function names, that one lives in Semantic_check.ml : https://github.com/stan-dev/stanc3/pull/103/files#diff-3b64248472843a9cb69a14706dd34024R30

enetsee: Yeah, I left it in there too since most of semantic check still uses the exception based approach. If we fully committed to the validation approach it wouldn't be needed since _all_ exception messages would be rendered in 'Semantic_errors.ml' . I can expose from there for the time being which is probably a bit cleaner

enetsee: Ok - I moved the operator related stuff to `Semantic_errors.ml`. Again, the idea would be to have all exception message rendering functions in here and have `Semantic_check.ml` simply construct the low level semantic error data type.

lib/Ast_to_Mir.ml
```diff
@@ -61,7 +73,7 @@ and trans_exprs = List.map ~f:trans_expr
 let trans_sizedtype = map_sizedtype trans_expr
 
 let neg_inf =
-  { expr= FunApp (string_of_internal_fn FnNegInf, [])
+  { expr= FunApp (CompilerInternal, string_of_internal_fn FnNegInf, [])
```
seantalts: Hmm, this negative_infinity() is actually math library function... but on the other hand FnNegInf is defined in the MIR as one of these compiler internals. A small part of me wants to just mark this as StanLib, but not sure if that'd have weird repercussions.

lib/Ast_to_Mir.ml
```diff
@@ -129,7 +141,8 @@ let mkfor bodyfn iteratee smeta =
   let loopvar, reset = Util.gensym_enter () in
   let lower = {expr= Lit (Int, "0"); emeta= internal_meta} in
   let upper =
-    { expr= FunApp (string_of_internal_fn FnLength, [iteratee])
+    { expr=
+        FunApp (CompilerInternal, string_of_internal_fn FnLength, [iteratee])
```
seantalts: Same for this - `length` is a math function at the end of the day.

lib/Expression_gen.ml
```diff
@@ -19,12 +19,11 @@ let stan_namespace_qualify f =
 
 (* return true if the types of the two expression are the same *)
 let types_match e1 e2 = e1.emeta.mtype = e2.emeta.mtype
+let is_stan_math f = ends_with "__" f || starts_with "stan::math::" f
 
 (* "__" is an illegal suffix for user functions, used for built-in operators not in signatures *)
 let is_user_defined f =
```
seantalts: Isn't this just checking whether it's Mir.UserDefined or not?

lib/Semantic_check.ml
```diff
@@ -497,23 +250,182 @@ and inferred_unsizedtype_of_indexed loc ut typed_indexl =
...
+(* -- Operators ------------------------------------------------------------- *)
+let semantic_check_operator i = i
+
+(* -- Expressions ----------------------------------------------------------- *)
+
+(* Function application validation checks *)
+let semantic_check_fn_map_rect ~loc id es =
+  match (id.name, es) with
+  | "map_rect", {expr= Variable arg1_name; _} :: _ ->
+      if
```
seantalts: Can this be written with a `when` clause more succinctly? 
```ocaml
let semantic_check_fn_map_rect ~loc id es =
  match (id.name, es) with
  | "map_rect", {expr= Variable arg1_name; _} :: _ 
      when
         String.(
          is_suffix arg1_name.name ~suffix:"_lp"
          || is_suffix arg1_name.name ~suffix:"_rng")
     -> Semantic_errors.FnMapRect (loc, arg1_name.name) |> Validate.error
  | _ -> Validate.ok ()
```

lib/Validation.ml
```diff
@@ -0,0 +1,55 @@
...
+    | Error e1, Error e2 -> Error (append e1 e2)
+
+  let apply_const a b = apply b ~f:(map a ~f:(fun _ x -> x))
+  let liftA2 f x y = apply y ~f:(apply x ~f:(pure f))
+  let consA next rest = liftA2 List.cons next rest
+  let sequence ts = List.fold_right ~init:(pure []) ~f:consA ts
+
+  module Validation_infix = struct
+    let ( <*> ) x f = apply x ~f
+    let ( *> ) x y = apply_const x y
```
seantalts: Let's stick to the 4 monadic punctuations that they use everywhere in Core_kernel? `|>`, `@@`, `>>=` and `>>|` and use names for everything else if you don't mind.

lib/Stan_math_code_gen.ml
```diff
@@ -439,7 +453,11 @@ let%expect_test "udf" =
   ; fdname= "sars"
   ; fdargs= [(DataOnly, "x", UMatrix); (AutoDiffable, "y", URowVector)]
   ; fdbody=
-      Return (Some (w @@ FunApp ("add", [w @@ Var "x"; w @@ Lit (Int, "1")])))
+      Return
+        (Some
+           ( w
+           @@ FunApp (UserDefined, "add", [w @@ Var "x"; w @@ Lit (Int, "1")])
```
seantalts: this one should be StanLib :p

lib/Stan_math_code_gen.ml
```diff
@@ -19,7 +19,7 @@
 open Core_kernel
 open Mir
 open Fmt
-open Expression_gen
+module ExprGen = Expression_gen
```
seantalts: Why shouldn't we just open it? It's basically only in a separately file because that's how it was added - it's very much integral with this module. 

Separately (in another PR) I think we might want to move the pretty printing of the MIR out of that module and just keep the types in there, which will help with name clashes as a byproduct.

lib/Semantic_errors.ml
seantalts: I love having a list of all of the error types in one place! Amazing. 

Agreed about your refactor assuming Matthijs also agrees.

lib/Validation.mli
```diff
@@ -0,0 +1,2 @@
+(** @inline *)
```
seantalts: What does this do?

lib/Expression_gen.ml
```diff
@@ -307,5 +321,7 @@ let%expect_test "pp_expr10" =
 
 let%expect_test "pp_expr11" =
   printf "%s"
-    (pp_unlocated (FunApp ("poisson_rng", [dummy_locate (Lit (Int, "123"))]))) ;
-  [%expect {| poisson_rng(123, base_rng__) |}]
+    (pp_unlocated
+       (FunApp
+          (Mir.UserDefined, "poisson_rng", [dummy_locate (Lit (Int, "123"))]))) ;
```
seantalts: Should be StanLib

lib/Semantic_check.ml
enetsee: Yes it can! I'll update

lib/Validation.ml
enetsee: Sure - I do agree the infix operators don't buy us a huge amount and will look cryptic unless you happen to know what an applicative functor is...

lib/Stan_math_code_gen.ml
enetsee: This is mainly habit - I usually try and avoid opening modules altogether but I take your point on them having the same purpose. I'll un-qualify it and add and look at merging the files in #99.

lib/Validation.mli
enetsee: It shouldn't be there; I'll remove.

enetsee:
@seantalts  @VMatthijs  thank you both for taking a look. 

> I think it would have been a little more optimal to have that as a separate PR though

Agreed; I started with the validation approach so I could follow the code more easily but it probably makes sense to introduce it completely or not at all.

> It wouldn't be easy to pull that out separately, would it?

This is basically reverting the semantic check code for `FunApp` and `NRFunApp` back to the exception based approach. I might be able to do this today.

> Hmm, this negative_infinity() is actually math library function... but on the other hand FnNegInf is defined in the MIR as one of these compiler internals.

At the moment it looks like some of the `internal_fn` functions are actually Stan math so really this type just represents functions introduced during translation to MIR. Would it be worth defining separate types so they also have the meaning we want?

lib/Expression_gen.ml
enetsee: Yeah, this is now only called via `gen_extra_fun_args` which itself is only called for `UserDefined`. I'll remove.

enetsee:
I've removed the validation stuff so the PR just adds function kinds now. I'm still not completely sure what is going on in `Expression_gen` with stan lib function kinds so I'd appreciate another pair of eyes.

lib/Expression_gen.ml
```diff
@@ -169,17 +167,17 @@ and gen_misc_special_math_app f =
...
         (fun ppf es ->
           if List.length es = 2 then pp_binary ppf "std::min(%a, %a)" es
-          else pp_ordinary_fn ppf f es )
+          else pf ppf "%s(@[<hov>%a@])" f (list ~sep:comma pp_expr) es )
   | "ceil" ->
       Some
         (fun ppf es ->
           if is_scalar (first es) then pp_unary ppf "std::ceil(%a)" es
-          else pp_ordinary_fn ppf f es )
+          else pf ppf "%s(@[<hov>%a@])" f (list ~sep:comma pp_expr) es )
```
seantalts: why prefer the duplication?

seantalts: To answer my own question, seems like the original abstraction wasn't working for some of the use cases and we weren't gaining a ton with it. This seems fine.

enetsee: Yeah, I pulled out the function for exactly that reason. I think the above could still be tidied up a bit though

seantalts: Yeah. Doesn't have to all be perfect all in one go especially at this stage in the project. Don't worry about it this one. Let me take a look at the expression gen stuff for you...

lib/Expression_gen.ml
```diff
@@ -238,7 +248,11 @@ and pp_expr ppf e =
   | Var s -> pf ppf "%s" s
   | Lit (Str, s) -> pf ppf "%S" s
   | Lit (_, s) -> pf ppf "%s" s
-  | FunApp (f, es) -> gen_fun_app ppf e.emeta.mtype f es
+  | FunApp (Mir.StanLib, f, es) -> gen_fun_app ppf e.emeta.mtype f es
+  | FunApp (Mir.CompilerInternal, f, es) ->
+      pp_compiler_internal_fn ppf (stan_namespace_qualify f) es
+  | FunApp (Mir.UserDefined, f, es) ->
```
seantalts: chef kiss

lib/Expression_gen.ml
```diff
@@ -238,7 +248,11 @@ and pp_expr ppf e =
...
+      pp_ordinary_fn ppf (stan_namespace_qualify f) es
```
seantalts: Don't think we need the stan_namespace_qualify on user defined ones

seantalts:
Ran tests locally since Jenkins is down. Merging.

seantalts:
Thanks! Next up modularizing? and then error variants?

PR #104, opened 2019-04-10
Double down on the parameterization of the MIR

seantalts:
Curious what you all think about the results here. We seem to gain approximately nothing in code gen, it seems - we really can't construct functions that are unaware of at least the expression metadata for easier testing or anything. But curious if anyone can spot mistakes or anti-patterns here and has suggestions.

### DIFF
--- a/lib/Mir.ml
+++ b/lib/Mir.ml
@@ -145,24 +145,24 @@ type ('e, 's) prog =
   ; prog_path: string }
 [@@deriving sexp]
 
-type 'm expr_with = {expr: 'm expr_with expr; emeta: 'm}
+type 'm with_expr = {expr: 'm with_expr expr; emeta: 'm}
 
-and typed_and_located =
-  { expr_type: unsizedtype
-  ; expr_loc: location_span sexp_opaque [@compare.ignore]
-  ; expr_adlevel: autodifftype }
-
-and expr_typed_located = typed_and_located expr_with [@@deriving sexp]
-
-type ('e, 'm) stmt_with = {stmt: ('e, ('e, 'm) stmt_with) statement; smeta: 'm}
-
-and stmt_loc =
-  (expr_typed_located, (location_span sexp_opaque[@compare.ignore])) stmt_with
+and mtype_loc_ad =
+  { mtype: unsizedtype
+  ; mloc: location_span sexp_opaque [@compare.ignore]
+  ; madlevel: autodifftype }
 [@@deriving sexp]
 
-type expr_no_meta = unit expr_with
+type ('e, 'm) stmt_with =
+  {stmt: ('e with_expr, ('e, 'm) stmt_with) statement; smeta: 'm}
+
+and stmt_loc =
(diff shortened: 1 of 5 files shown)


seantalts:
(also, this depends on #102, so might be worth waiting for that to get merged first).

seantalts:
@rybern would you mind taking a look at this? I think this is your idea and I want to make sure it's as you envisioned, plus I'm sure there is room for improvement in the implementation.

rybern:
Yes, definitely, unfortunately will be tomorrow at the earliest

lib/Mir.ml
```diff
@@ -145,18 +145,23 @@ type ('e, 's) prog =
...
-  ; stmt: (expr_typed_located, stmt_loc) statement }
-[@@deriving hash, map, sexp, of_sexp]
+type 'm expr_with = {expr: 'm expr_with expr; emeta: 'm}
 
+and typed_and_located =
+  { expr_type: unsizedtype
+  ; expr_loc: location_span sexp_opaque [@compare.ignore]
+  ; expr_adlevel: autodifftype }
+
+and expr_typed_located = typed_and_located expr_with [@@deriving sexp]
```
rybern: Do we need this alias? What if we renamed `typed_and_located` to `type_and_location`, and then write something like `type_and_location expr_with` in signatures, so it's more explicit what it is? (It's unfortunate that it doesn't read `expr_with type_and_location`, maybe there's a better phrasing, like renaming `expr_with` to `exprs` so we can write `type_and_location exprs`)

lib/Mir.ml
```diff
@@ -145,18 +145,23 @@ type ('e, 's) prog =
...
+
+type ('e, 'm) stmt_with = {stmt: ('e, ('e, 'm) stmt_with) statement; smeta: 'm}
```
rybern: Is it useful to be able to swap out the whole expression type, or do we only need to swap out the expression metadata type? If the latter, we could write something like
`
type ('e, 'm) stmt_with = {stmt: ('e expr_with, ('e, 'm) stmt_with) statement; smeta: 'm}
`
This was `'e` is expression metadata and `'m` is statement metadata.

rybern:
Hi Sean - sorry for the delay. I only came up with a couple of suggestions, this all looks good to me, although I'm not sure I understood some of the codegen (internal expressions?). Let me know if there's an area it would be helpful for me to comb over in detail.

lib/Mir.ml
seantalts: I thiiiink either one works? I guess that might help with some of the error messages.

lib/Mir.ml
seantalts: type_and_location with_expr? :P I haven't really programmed with this pattern at a large scale before, it may turn out we don't need the alias. Usually would help make refactoring easier; for example I just changed what stmt_loc referred to and didn't have to change much code using that.

seantalts:
@rybern no worries and thank you! I just pushed the changes you recommended, and I think it's probably better? deleted 13 lines, anyway 😎 haha.

rybern:
Looks great to me!

Issue #105, opened 2019-04-10
Refactor AST to use meta parameterization


seantalts:
Seems like all the same reasons apply as for the MIR example. See the #104 PR for details. 

Essentially we want to create a record parameterizing the AST over the metadata in the same way we did for the MIR.

PR #107, opened 2019-04-15
Refactor AST to use meta parameterization

enetsee:
Changes for #105


seantalts:
Tried to fix the merge conflicts; PTAL and make sure I didn't mess anything up!

VMatthijs:
Oh this is great! Very nice that you had time to make this refactor. Wanted to do it myself but have been so busy.

seantalts:
Yeah we can 😂 Just pushed a commit to remove it.

VMatthijs:
@seantalts ,  any idea what is happening with Jenkins here?

seantalts:
Tagging in @serban-nicusor-toptal

serban-nicusor-toptal:
Should be just fine now.

seantalts:
Thank you! Merged :D

lib/Ast.ml
```diff
@@ -61,30 +61,41 @@ type 'e expression =
...
 [@@deriving sexp, hash, compare, map]
 
+type 'm expr_with = {expr: 'm expr_with expression; emeta: 'm}
+[@@deriving sexp, compare, map, hash]
+
 (** Untyped expressions, which have location_spans as meta-data *)
-type untyped_expression =
-  { expr_untyped: untyped_expression expression
-  ; expr_untyped_loc: Mir.location_span sexp_opaque [@compare.ignore] }
+type located_meta = {loc: Mir.location_span sexp_opaque [@compare.ignore]}
```
seantalts: Hey, just looking at this - we don't actually need to use a record for located_meta, it's just a span. Can I remove this layer of indirection?

seantalts: Or if you guys think we should have them always be a record we'd want to change the MIR to match I think. Right now it's just defined like this: https://github.com/stan-dev/stanc3/blob/master/lib/Mir.ml#L165

enetsee: I think using the `location_span` directly makes sense given we are unlikely to attach more meta data at this point.

PR #110, opened 2019-04-29
Separate libs

enetsee:
This PR supercedes #99. 

The current code separates the various compiler components into their own libraries leaving the compiler unchanged.

Contained in Mir.ml is a sketch of what to do next. The idea is to specify module signatures for frontend, backend and optimization passes then write the compiler as a functor parameterized over those modules. 

The sketch also assumes that we have reified the current exception based approach in the frontend to use error variants for syntactic and semantic errors.

The full move to parameterizing the compiler obviously isn't necessary to merge the code as-is.

I'll be keeping track of the discussion here as much as possible whilst I'm away.


seantalts:
What do you think about moving the compiler module structures to a new file?

PR #114, opened 2019-05-08
Rename Mir package to Middle package

seantalts:


seantalts:
@VMatthijs This might also be annoying for your merge :(

PR #115, opened 2019-05-10
Changes to semantic check from dependence analysis

VMatthijs:
This slightly simplifies the semantic check. The biggest change is that `stan_math_returntype` gets a more simple type.


src/frontend/Semantic_check.ml
```diff
@@ -1206,7 +1238,10 @@ let rec semantic_check_statement cf (s : Ast.untyped_statement) :
             ^ ("'" ^ uid.name ^ "'")
             ^ " was found with the correct signature." )
       in
-      let cumulative_density_is_defined name argumenttypes =
+      let cumulative_density_is_defined name arguments =
+        let argumenttypes =
+          List.map ~f:(fun x -> (x.emeta.ad_level, x.emeta.type_)) arguments
```
seantalts: Can we make this a function and give it a name? I see it a lot.

PR #119, opened 2019-05-13
8 schools runs correctly!

seantalts:


seantalts:
I fooled myself again, it's not working yet - changing the name back.

seantalts:
Ok, this time I actually got it working! @VMatthijs you seem to have been right that the slowness was just due to poor fitting due (which ended up being due to not generating the jacobian__ adjustments for transforms yet)... !

Anyone want to help me design an end-to-end statistical test that runs across platforms? :P

bob-carpenter:
SBC would work, but it's slow.

seantalts:
> SBC would work, but it's slow.

And there's no agreed-upon numerical summaries yet... I was thinking of just implementing what Michael talks about in the README for this repo: https://github.com/stan-dev/stat_comp_benchmarks

seantalts:
Oh, and using that we'd have to set a gold standard with Stan 2 and judge against that. I think that much more limited form of acceptance testing is okay because we're trying to replicate Stan2 rather than create a new algorithm or model.

VMatthijs:
That's so cool! Congrats, @seantalts ! Will have a look at this.

bob-carpenter:
If we know the true answer you can do what Michael's suggesting in the benchmarks.  

Then all we need to do is figure out how long we need to run to reduce the standard error low enough to get a tolerable false positive rate on the test of (mu_est - mu_true) < (sigma_true / 4).  That probably won't be too bad---that's a pretty lenient test!

SBC can run without knowing the true answers.  Though after we run SBC, we'd have reason to trust a model's true answer from a long run.  To run SBC, we'd need to pick a statistical test for uniformity and set it at a low false-positive rate threshold. There's no more a best test for uniformity than there is a best sorting algorithm---it depends on what's being tested.  We can just choose one---it'll pick up big mistakes.  Subtle mistakes are hard without *a lot* of computation as the numerics get very tight.  The test Michael recommended won't pick up on more subtle mistakes, like introducing a 1% bias.

seantalts:
Yeah, creating a really good regression test suite is quite hard. But you've made me realize we can actually do something much simpler - just run Stan 2 against stan 3 every time. Then the numerical outputs should be pretty much identical, right?

src/frontend/Ast_to_Mir.ml
```diff
@@ -198,16 +210,16 @@ let constrainaction_fname c =
...
   ; dconstrain: constrainaction option
   ; dadlevel: autodifftype }
 
 let rec unsizedtype_to_string = function
   | UMatrix -> "matrix"
   | UVector -> "vector"
   | URowVector -> "row_vector"
-  | UReal -> "real"
-  | UInt -> "int"
+  | UReal -> "scalar"
```
VMatthijs: Is this what Stan currently says? I always find "scalar" such an ambiguous word. In my head it implies a contrast with a vector, but then you still don't quite know that it's continuous rather than discrete

seantalts: yep

bob-carpenter:
Only if nothing's changed that would affect the RNGs.  Otherwise, we'd still need to run Stan 2 a long time to get a stable estimate against which to test.

bob-carpenter:
In Stan, you also don't know if a continuous scalar is an autodiff variable or a double.

seantalts:
> Only if nothing's changed that would affect the RNGs.  Otherwise, we'd still need to run Stan 2 a long time to get a stable estimate against which to test.

Yep. I'm optimistic that these tests can be of excellent service not only for releasing stanc3 but also for ~98% of PRs in the future that don't touch RNGs - when they break, we'll have to break out the big guns manually (or script them at that time).

betanalpha:
Conventionally in mathematics a scalar refers to an element of
a field which excludes the integers.  In physics and other
disciplines it refers to a real number and distinguishes from
a collection of real numbers.  So I think the scalar nomenclature 
is fine here.

Bob’s point is more prescient — it would help to differentiate
between a real and a dual number (autodiff variable, point
in the cotangent bundle, point in phase space, etc, etc, etc).

betanalpha:
Initially a suite of regression tests should be sufficient, especially 
when the scope of the Stan3 language is largely the same as that
of the Stan2 language.  They should generate exactly the same 
C++ (at least initially, unless update are being convolved in with 
the new language) in which case on the same hardware/OS/toolkit
the results should be equivalent.

Otherwise verifying proper output qualitatively is really hard as
already discussed, and trying to make it a useful quantitative test 
is downright adversarial.

src/frontend/Ast_to_Mir.ml
```diff
@@ -10,6 +10,21 @@ let trans_fn_kind = function
...
+  | Some n -> n
+  | None -> raise_s [%message "No matching functions for" (name : string)]
+
+let%expect_test "distribution name mangling" =
+  print_endline (get_prob_fun_name "normal") ;
+  [%expect {| normal_log |}]
+
 let rec op_to_funapp op args =
   let argtypes =
     List.map ~f:(fun x -> (x.Ast.emeta.Ast.ad_level, x.emeta.type_)) args
```
VMatthijs: Shouldn't you be special casing on And and Or here to create EAnd and EOr?

src/frontend/Ast_to_Mir.ml
```diff
@@ -160,13 +171,16 @@ let mkfor bodyfn iteratee smeta =
...
-  | UArray _ -> mkfor (fun e -> for_scalar bodyfn e smeta) var smeta
-  | UFun _ | UMathLibraryFunction ->
-      raise_s [%message "Can't for over " (var : mtype_loc_ad with_expr)]
+let rec for_scalar st bodyfn var smeta =
+  match st with
+  | SInt | SReal -> bodyfn var
+  | SVector d | SRowVector d -> mkfor d bodyfn var smeta
+  | SMatrix (d1, d2) ->
+      mkfor
+        { expr= FunApp (StanLib, string_of_operator Times, [d1; d2])
```
VMatthijs: You have a single loop variable here, so I guess it's the responsibility of bodyfn to index correctly into the matrix using the single index?

src/frontend/Ast_to_Mir.ml
seantalts: Yep, that seems right. Good catch, thanks.

src/frontend/Ast_to_Mir.ml
```diff
@@ -255,18 +267,78 @@ let extract_constraint_args = function
...
+  | SVector _ | SRowVector _ | SMatrix _ -> UReal
+  | x -> remove_size x
+
+let internal_of_dread = function
+  | ReadParam -> FnReadParam
+  | ReadData -> FnReadData
+
+let assign_indexed vident smeta varfn var =
+  let indices =
+    match var.expr with Indexed (_, indices) -> indices | _ -> []
```
VMatthijs: Do you want to throw here if `var.expr` is not `Indexed (_. _)` or `Var _`?

src/frontend/Ast_to_Mir.ml
```diff
@@ -451,14 +457,19 @@ let rec trans_stmt declc (ts : Ast.typed_statement) =
...
         { loopvar= loop_variable.Ast.name
         ; lower= trans_expr lower_bound
         ; upper= trans_expr upper_bound
-        ; body= trans_single_stmt loop_body }
+        ; body }
       |> swrap
   | Ast.ForEach (loopvar, iteratee, body) ->
-      let newsym = Util.gensym () in
+      let newsym = gensym () in
       let wrap expr = {expr; emeta= {mloc; mtype= UInt; madlevel= DataOnly}} in
```
VMatthijs: I think the translation of the ForEach loop is still wrong.
It should probably first generate a fresh loop variable like you do.
Then it should inspect the type of the iteratee and distinguish between Eigen types and Arrays in the translation. It should calculate the correct bounds. And start the loop body with the declaration and definition of the old loop variable of the Foreach loop in terms of the new loop variable and the iteratee.
I think @bob-carpenter may have already written out what the code gen should look like.

src/frontend/Ast_to_Mir.ml
```diff
@@ -582,35 +637,37 @@ let trans_prog filename
...
+         { dread= Some ReadData
          ; dconstrain= Some Unconstrain
          ; dadlevel= DataOnly })
       parametersblock
   in
   { functions_block=
-      (* Should this be AutoDiffable for functions here?*)
       Option.value_map functionblock ~default:[] ~f:(fun fundefs ->
           List.map fundefs ~f:(fun fundef ->
               trans_fun_def
```
VMatthijs: Is it important that FunDef's get translated as AutoDiffable below here?

src/frontend/Ast_to_Mir.ml
```diff
@@ -520,14 +529,44 @@ let trans_fun_def declc (ts : Ast.typed_statement) =
...
+  | _ -> [{stmt= IfElse (cond, body, None); smeta= no_span}]
+
+let get_block block prog =
+  match block with
+  | Parameters -> prog.Ast.parametersblock
+  | TransformedParameters -> prog.transformedparametersblock
+  | GeneratedQuantities -> prog.generatedquantitiesblock
+  | Data -> prog.datablock
+
+let trans_prog filename p : typed_prog =
```
VMatthijs: I really like the structure of `trans_prog` now, with these `Check` and `Constrain` and `Read` actions. I wonder if it's worth @bob-carpenter having a look at just this function as a sanity check for the high-level structure of what needs to happen.

bob-carpenter:
I wrote out what I thought the code gen needed to look like for expressions and statements on the branch I was working on.

seantalts:
> I wrote out what I thought the code gen needed to look like for expressions and statements on the branch I was working on.

Didn't you just do expressions and we merged that in, right? Is there a Statement_gen.ml file lying around somewhere?

seantalts:
> I'm unable to follow the details however. It's all a bit scary. I presume we'll just have to have very good tests before we release this. Or you'd have to be fairly confident you'd got it right. Or someone with more expertise about how things currently work should look at this, like Bob or Mitzi. Or all of the above, ideally.

Definitely going to have good tests! And ideally make the code easier to follow :)

src/frontend/Ast_to_Mir.ml
seantalts: Functions should be generally AutoDiffable, though one day we can make use of the autodiff level on expressions in the optimization pass to do that more optimally ;)

src/frontend/Ast_to_Mir.ml
```diff
@@ -42,6 +42,12 @@ and trans_expr {Ast.expr; Ast.emeta} =
   and madlevel = emeta.ad_level in
   match expr with
   | Ast.Paren x -> trans_expr x
+  | BinOp (lhs, And, rhs) ->
+      { expr= EAnd (trans_expr lhs, trans_expr rhs)
+      ; emeta= {madlevel; mloc; mtype} }
+  | BinOp (lhs, Or, rhs) ->
+      { expr= EOr (trans_expr lhs, trans_expr rhs)
+      ; emeta= {madlevel; mloc; mtype} }
```
VMatthijs: Nice. Thanks for fixing!

seantalts:
Okay I'm going to merge and keep working on the automated Jenkins tests. You should be able to run the end-to-end tests locally with `cmdstan=/path/to/cmdstan dune runtest test/integration/code-gen`.

seantalts:
j/k got the end-to-end test finally passing on Jenkins :)

src/frontend/Ast_to_Mir.ml
seantalts: I don't think so, why do you ask?

seantalts:
@bob-carpenter let me know if you find the statement gen code somewhere, thanks!

bob-carpenter:
https://github.com/stan-dev/stanc3/tree/statement-generator-spec

includes both the expression and statement generation specs in the directory notes.

seantalts:
Ah gotcha. Thanks!

PR #123, opened 2019-05-15
minor doc fix

ode33:


VMatthijs:
@seantalts , any idea what the problem is with these performance tests?

ode33:
Is there something amiss?

seantalts:
You mean the end to end tests? It looks like opam removed old versions of Dune and Fmt today! I didn't realize they did that. We need to figure out how to pin versions...

seantalts:
Hmm, opam looks like it has records of that version - maybe the Docker image just needs to run opam update first.

seantalts:
@ode33 nothing you need to worry about - one of the Dockerfiles we use for testing is broken right now.

PR #124, opened 2019-05-16
Try updating opam after installing it

seantalts:


seantalts:
I think we might need to just have this in there temporarily for a while to update the Docker images...

PR #126, opened 2019-05-17
Error variants

enetsee:
This PR introduces error variants for semantic and syntactic errors. 

The main change is to `Semantic_check.ml` which now uses monadic validation to accumulate semantic errors rather than using exceptions. The top level function `semantic_check_program` now returns a `Result.t` type indicating success or failure.

The top-level parser function `parse_file` and `parse_string` have also been changed to emit a result type but the lexer can still generate exceptions and generates warnings as a side effect.

The `stanc` executable has also been changed to accommodate the changes.

The PR leaves us kind of halfway between a pure approach and a side-effecting one. In particular: 

- The lexer still generates exceptions on failure
- The lexer still uses side-effecting functions for deprecation warnings
- The semantic check uses reference cells to manage front-end options
- The semantic check uses side-effecting functions for modifying the symbol table

The third of these also means that we can't yet hide the frontend implementation using the signature in `Middle.Frontend_intf` since `stanc` directly sets mutable variables in frontend.


src/frontend/Errors.ml
```diff
@@ -5,30 +5,30 @@ open Middle
...
-exception SemanticError of (string * location_span)
+exception SemanticError of (string * Mir.location_span)
 
 (** Exception [FatalError [msg]] indicates an error that should never happen with message
     [msg]. *)
 exception FatalError of string
 
 (* A semantic error reported by the toplevel *)
-let semantic_error ~loc msg = raise (SemanticError (msg, loc))
+(* let semantic_error ~loc msg = raise (SemanticError (msg, loc)) *)
```
VMatthijs: Can this be deleted?

src/frontend/Errors.mli
```diff
@@ -18,8 +18,8 @@ exception SemanticError of (string * location_span)
     so we can trace their origin. *)
 exception FatalError of string
 
-val semantic_error : loc:location_span -> string -> 'a
-(** Throw a semantic error reported by the toplevel *)
+(* val semantic_error : loc:location_span -> string -> 'a
```
VMatthijs: Can this be deleted?

enetsee:
I've got a bug caused by not adding an id to the symbol table - I'll update when it's fixed.

src/frontend/Parse.ml
```diff
@@ -2,7 +2,15 @@
...
 open Core_kernel
-open Errors
+open Middle
+
+type syntax_error = Parsing of string * Mir.location_span
+
+let syntax_error_message = function Parsing (msg, _) -> msg
+let syntax_error_location = function Parsing (_, loc) -> loc
+
+(* TODO :render messages along with location! *)
```
VMatthijs: Can you please still do this?

src/frontend/Parse.ml
```diff
@@ -208,6 +221,8 @@ let%expect_test "parse indices, two different colons" =
   let ast =
     parse_string Parser.Incremental.program
       "model { matrix[5, 5] x; print(x[2 - 3 ? 3 : 4 : 2]); }"
+    |> Result.map_error ~f:render_syntax_error
+    |> Result.ok_or_failwith
```
VMatthijs: Does it make sense to make these handlers already part of the parsing function? I just keep seeing this combination repeated.

enetsee: Yes, the idea is to hide of this from stanc by using the `Frontend.S` module signature. I can't do that right now (reason outlined above) so tried to keep things as-is until we make the change.

src/frontend/Semantic_error.ml
```diff
@@ -0,0 +1,552 @@
...
+    | IllTypedBinaryOperator of operator * unsizedtype * unsizedtype
+    | IllTypedPrefixOperator of operator * unsizedtype
+    | IllTypedPostfixOperator of operator * unsizedtype
+    | NotIndexable of unsizedtype
+
+  (* let assignmentoperator_to_stan_math_fn = function
+    | Plus -> Some "assign_add"
+    | Minus -> Some "assign_subtract"
+    | Times -> Some "assign_multiply"
+    | Divide -> Some "assign_divide"
```
VMatthijs: What are these definitions doing here? Can they be deleted?

src/middle/Backend_intf.ml
```diff
@@ -0,0 +1,14 @@
+open Mir
+
+module type S = sig
```
VMatthijs: Is this at all related to the PR?

src/middle/Compiler.ml
```diff
@@ -1,124 +1,32 @@
...
+include Core_kernel
+
+module Make
+    (F : Frontend_intf.S)
+    (O : Optimization_intf.S)
+    (B : Backend_intf.S) : S with type frontend_error := F.frontend_error =
+struct
+  type compiler_opts =
+    { frontend_opts: F.frontend_opts
+    ; optimization_opts: O.optimization_opts
```
VMatthijs: Is this at all related to the PR?

src/middle/Compiler_intf.ml
```diff
@@ -0,0 +1,24 @@
...
+  type compiler_opts_error
+  type compiler_opts
+
+  val default_compiler_opts : compiler_opts
+
+  val compiler_opts_of_string :
+    string -> (compiler_opts, compiler_opts_error list) result
+
+  val compile_from_file :
+    opts:compiler_opts -> file:string -> (string, frontend_error) result
```
VMatthijs: Is this at all related to the PR? Also, just wondering: does it matter whether this is a .ml or .mli file?

src/middle/Compiler.ml
enetsee: No - I tidied it up before I realized I wasn't going to be able to use the signatures yet.

src/middle/Middle.ml
```diff
@@ -1,12 +1,37 @@
 include Mir
 open Core_kernel
+module Mir = Mir
```
VMatthijs: What does this do?

src/middle/Compiler_intf.ml
enetsee: Same as above regarding; it's in an .ml file since we want to expose the module signatures and functor from middle so that it can be used to either write new frontends/backends or construct the core compiler with them

src/middle/Middle.ml
enetsee: Before this change, Mir was being _included_ in Middle.  With this approach, Mir is exposed as a module  which I think is a little nicer since `Middle` contains more than just `Mir`.

As a consequence, you can now open `Middle` and not have all the `Mir` types in scope.

src/middle/Backend_intf.ml
enetsee: No - I tidied it up before I realized I wasn't going to be able to use the signatures yet.

src/stanc/stanc.ml
```diff
@@ -79,25 +79,43 @@ let add_file filename =
 (** ad directives from the given file. *)
 let use_file filename =
   let ast =
-    try Parse.parse_file Parser.Incremental.program filename
-    with Errors.SyntaxError err ->
-      Errors.report_syntax_error err ;
+    try 
+      match Parse.parse_file Parser.Incremental.program filename with
+      | Result.Ok ast -> ast
```
VMatthijs: So does it say OK even if there is a lexing error, for instance?

src/stanc/stanc.ml
```diff
@@ -79,25 +79,43 @@ let add_file filename =
...
+      | Result.Error err ->
+          let loc = Parse.syntax_error_location err
+          and msg = Parse.syntax_error_message err in
+          Errors.report_parsing_error (msg, loc) ;
+          exit 1
+    with Errors.SyntaxError err ->
```
VMatthijs: Is this only to catch the syntax errors that are not parsing errors? If so, should the parsing errors then be removed from the syntax errors? I feel like the syntax errors and their handling could be simplified further.

src/stanc/stanc.ml
enetsee: No, lexing errors still raise exceptions at the moment which is outside the context of validation.

src/stanc/stanc.ml
enetsee: Yes, exactly. As I mentioned in the first message, we're sort of halfway between a monadic and exception based approach right now.

enetsee: And I agree about simplifying further. Ideally we want no exceptions coming from `frontend` - keep the core of the compiler pure and do all the IO in the `stanc` wrapper.

src/middle/Validation_intf.ml
```diff
@@ -0,0 +1,42 @@
+module type Infix = sig
+  type 'a t
+
+  val ( >>= ) : 'a t -> ('a -> 'b t) -> 'b t
```
VMatthijs: There's part of me that really likes abstracting out this sort of validation monad because obviously that is what is going on with this sort of error accumulation. Another part of me wonders whether it's slightly overengineered and would make the semantic check harder to understand for new developers and harder to maintain.
Let me think about it a bit longer. I think I mostly like it.

src/frontend/Semantic_error.ml
```diff
@@ -0,0 +1,552 @@
...
+    | DuplicateArgNames ->
+        Fmt.pf ppf "All function arguments must have distinct identifiers."
+    | IncompatibleReturnType ->
+        Fmt.pf ppf
+          "Function bodies must contain a return statement of correct type in \
+           every branch."
+end
+
+type t =
+  | TypeError of location_span * TypeError.t
```
VMatthijs: This is super slick! :-) I should've done that immediately when I wrote it. The problem was that I had no idea what all the error conditions would be and how they should be organized when I started writing the semantic check.

src/middle/Validation_intf.ml
enetsee: Yeah, I get where you're coming from. 

For type-checking / semantic-checking, what we actually want is a monad transfomer stack with `Reader` for the options, `Validation` to accumulate errors and `State` to carry the environment around.

Clearly this is going to be quite intimidating for most developers and probably impenetrable for someone with out an FP background. 

On the plus side, I this the approach will make something like a language server easier and should make the experience nicer for users i.e. we can report syntactic and semantic errors together rather than just the first.

src/frontend/Semantic_check.ml
```diff
@@ -170,164 +104,67 @@ let probability_distribution_name_variants id =
...
-   |AnyReturnType, AnyReturnType ->
-      AnyReturnType
+      Validate.ok (Mir.ReturnType UReal)
+  | _, _ when rt1 = rt2 -> Validate.ok rt2
+  | _ -> Semantic_error.mismatched_return_types loc rt1 rt2 |> Validate.error
 
 let check_fresh_variable_basic id is_nullary_function =
-  (* No shadowing! *)
-  (* For some strange reason, Stan allows user declared identifiers that are
+  Validate.(
```
VMatthijs: Could you maybe explain what this line does? Validate is a functor, right? It's Validation.Make. So I would expect it to apply to be applied to a module which contains a single type. Here, it seems to be applied to a value of result type. Is that related to the dot after Validate? Could you help me parse this? Thanks!

enetsee: At the moment, `Validation` is written so that you construct it via `Validation.Make` passing in another module which has some type `t` which is the type of error you are going to accumulate.

In `semantic_check.ml`, I create a module `Validate` by passing `Semantic_error` into `Validation.Make` fixing the type of error at `Semantic_error.t`.

When I write `Validate.( ... )`, this `Validate` module is in scope inside the brackets. This is particularly helpful when using infix functions like `>>=`; without it I would have to qualify everything.

VMatthijs: Oh gotcha! So would this be equivalent to writing a 
` let open Validate` first?

enetsee: Yes - there are two ways of writing local open. As well as being more lightweight, I always think this syntax mirrors what I'm doing in my head.

VMatthijs: Cool, thanks!

src/middle/Validation_intf.ml
enetsee: With all that being said, I think we can actually make the API to the `SemanticCheck` monad quite nice so that beginners can almost treat it like a DSL.

VMatthijs: OK. I've thought about it more and don't see a way around it. I think this is more or less the most slick and least intimidating way of achieving what we want.

src/middle/Compiler_intf.ml
VMatthijs: Could you not do that if it's in a .mli file? I guess I still don't fully understand when you put signatures in a .mli file and when in a .ml file.

src/middle/Middle.ml
VMatthijs: Gotcha, thanks!

src/stanc/stanc.ml
VMatthijs: Agreed about throwing the exceptions as far to the outside as possible. That's the only way I can imagine accumulating more than one exception. We'd still have to think about how to continue parsing after an error, but presumably we could figure that out based on what Merlin does.

VMatthijs:
I think this is great. Just let me know when it passes all tests and I'll have another look!

src/middle/Mir_intf.ml
```diff
@@ -0,0 +1,200 @@
+open Core_kernel
```
seantalts: You mentioned this at the meeting but I had something totally different in mind. What is the idea behind this Mir_intf abstraction? We aren't going to have multiple MIRs supported...

enetsee: The idea here isn't to hide the MIR implementation but rather to have a `Mir` module exposed. 

The way it is at the moment means that all the MIR types are in scope as soon as you open `Middle`; since `Middle` now contains a bunch of other stuff I wanted to scope them.

Having the signature contained in `Mir_intf.mli` means I can include it in `Middle.mli` to get the nested module.

seantalts: I think if we want to partition off the stuff in Middle I would go the other way - put the e.g. pretty printer in its own module, etc. I think the most important part of the Middle (and its main API) is the MIR and that should be front and center.

seantalts: Also just one meta comment - it seems like this PR could have been broken into a few smaller ones to be easier to handle - for example, one could have just dealt with the changes to error variants and semantic check, another could have refactored the Middle, another could have done the refactorings necessary to get the different compiler ends into shape to fit into the compiler module interface you wrote. It's probably too late now for this PR (definitely don't want to make you do a ton of work to try to separate them out or redo them), but just for future PR harmony :)

enetsee: Yeah - I did a bunch of stuff to `Middle` thinking I had done enough with the frontend to get the abstraction working; turns out I hadn't. It won't be a lot of work to revert `Middle` and I agree that this PR is overreaching a little!

enetsee:
I've created a new file in middle `Mir_pretty_printer.ml` during the merge from master. This was to avoid a cycle between `Stan_math_signatures.ml` (which was moved to `Middle` on master) and `Middle.ml`.

enetsee:
@VMatthijs  @seantalts this is ready for review now :fireworks:

seantalts:
Okay cool! It might take me a little while, 3000 lines is quite a lot! Minor, but it seems like maybe we could just take out all the Compiler related stuff - esp because I'm seeing the _intf pattern that we're still discussing (if we can use the "module type of" pattern I think we should do that instead if it's equivalent).

src/frontend/Ast_to_Mir.ml
```diff
@@ -3,8 +3,10 @@ open Middle
 
 (* XXX fix exn *)
 let unwrap_return_exn = function
-  | Some (ReturnType ut) -> ut
-  | x -> raise_s [%message "Unexpected return type " (x : returntype option)]
+  | Some (Middle.ReturnType ut) -> ut
+  | x ->
+      raise_s
+        [%message "Unexpected return type " (x : Middle.returntype option)]
```
seantalts: We opened Middle at the top - we shouldn't need to change lines like these, probably an artifact from modularizing the Mir type.

src/frontend/Ast_to_Mir.ml
```diff
@@ -136,14 +138,48 @@ let trans_printables mloc (ps : Ast.typed_expression Ast.printable list) =
       | Ast.PExpr e -> trans_expr e)
     ps
 
+let unsizedtype_of_indexed ut
```
seantalts: What motivated not using the Semantic check function anymore?

src/frontend/Errors.ml
```diff
@@ -1,34 +1,30 @@
 (** Setup of our compiler errors *)
 
 open Core_kernel
-open Middle
```
seantalts: What's bad about opening Middle? I don't see any ambiguity and editors are quick to resolve to the right place with goto-definition.

src/frontend/Errors.ml
```diff
@@ -1,34 +1,30 @@
...
-(* A semantic error reported by the toplevel *)
-let semantic_error ~loc msg = raise (SemanticError (msg, loc))
-
 (* A fatal error reported by the toplevel *)
 let fatal_error ?(msg = "") _ =
   raise (FatalError ("This should never happen. Please file a bug. " ^ msg))
 
 (** Parse a string into a location *)
-let rec parse_location str =
+let rec parse_location str : Middle.location =
```
seantalts: Why add type signatures?

enetsee: I added when type inference was going a bit nuts; should be ok to remove

src/frontend/Errors.ml
enetsee: Force of habit - I almost never open modules at the top level. I'll open and remove the qualifications

src/frontend/Ast_to_Mir.ml
enetsee: During the refactor, the function in `Semantic_check` was changed to return an `unsizedtype Validate.t` since there is a possibility that the indexed type is not semantically valid. 

In`Ast_to_Mir` we are only ever working with semantically correct ASTs so we should always be able to return an `unsizedtype`. Any deviation from this is genuinely exceptional behavior so we should use exceptions here to handle that case.

Whilst I could have reused the function inside `Semantic_check`, converted the `Validate.t` to a result and then raised an exception in the `Error` case, it would have meant exposing the `Validate` module outside of the semantic check module and violating DRY seemed a little nicer.

An alternative would be to expose a function `unsizedtype_of_indexed_exn` from `Semantic_check` although it would be a little odd in terms of the minimal API that module exposes. 

Happy to do the latter if you think it makes more sense?

src/middle/Middle.mli
```diff
@@ -1,27 +1,46 @@
-include module type of Mir
 open Core_kernel
+include module type of Mir
+include module type of Mir_pretty_printer
+module Validation : Validation_intf.Validation
```
VMatthijs: Do we still need this `_intf` pattern here? Sean seemed to be able to make the `include module type of Foo` syntax work even in cases where there was a `Foo.mli` file. Perhaps it can't be made to work at all with functors?

src/frontend/Semantic_error.mli
```diff
@@ -0,0 +1,79 @@
+open Middle
+
+type t
+
```
VMatthijs: This splitting out of the Semantic_error conditions is really great. :-)

src/frontend/Ast_to_Mir.ml
VMatthijs: Is there somewhere neutral that we can put it so that both files have access to it? It seems unnecessary to duplicate that function, particularly because I remember that its logic is a bit subtle. This would be my only real requested change before this can be merged.

src/middle/Validation.ml
```diff
@@ -0,0 +1,55 @@
+include Validation_intf
```
VMatthijs: Actually, could you add a little bit of documentation to the Validation monad? A minimum would be a comment at the top to explain what it's supposed to do (e.g. accumulate errors).

src/frontend/Ast_to_Mir.ml
enetsee: Sure - I think the second option above is most sensible i.e. expose expose a function `unsizedtype_of_indexed_exn` which unsafely coverts the `unsizedtype Validate.t` into an `unsizedtype`.

src/middle/Middle.mli
enetsee: I'll take a proper look at Sean's approach and see if it can be done.

src/middle/Validation.ml
enetsee: Will do - good shout.

src/frontend/Parse.ml
enetsee: Done

src/frontend/Semantic_error.ml
enetsee: Done

VMatthijs:
Woohoo! Well done!

PR #132, opened 2019-05-21
Move stan math signatures to middle

VMatthijs:
This moves the Stan Math Signatures to the Middle. This is for two reasons:
1. the math signatures are part of the spec of the Stan language (see language manual) hence should be part of any implementation
2. the math signatures are super useful to have available for optimizations, so we can check that certain partial evaluations and vectorizations are in fact supported by the math library.

These are the required changes to the existing codebase to make analysis and optimization its own library.


VMatthijs:
@seantalts , could you maybe have a look at this? Also, any idea why Jenkins hasn't been working?

seantalts:
hm, why are the tests not happening...

PR #133, opened 2019-05-21
Move transformations to Middle

seantalts:
Should have noticed this before, but we need these in the MIR as well.

This PR also attempts to fix ReadParam generation to account for differences in declared vs. actual sizes of parameters.


VMatthijs:
Could you maybe explain a bit about why you need them in the MIR? They won't be on MIR declarations, will they?

seantalts:
Good point! I initially had a version that required that. I think it's still possible to not have them in the MIR and we should stick to that. My bad! Closing.

PR #135, opened 2019-05-23
Jenkins windows fix

VMatthijs:
`dune runtest` should be run without `-x windows`.


VMatthijs:
@seantalts , could you have a look? Do you still want the tests to run in verbose mode?

seantalts:
Thank you! Been using verbose for now just because we have no other way of recording which tests are run :/

PR #137, opened 2019-05-28
codegen - all stat_comp_benchmark models compile

seantalts:
Ran it with the performance suite like so:
```
git clone --recursive https://github.com/stan-dev/performance-tests-cmdstan
cd performance-tests-cmdstan
./compare-git-hashes stat_comp_benchmarks develop stanc3-dev develop develop
```

Looks like the following are working with similar-ish performance (and no locations right now):
```
('stat_comp_benchmarks/benchmarks/low_dim_corr_gauss/low_dim_corr_gauss.stan', 1.04)
('stat_comp_benchmarks/benchmarks/gp_pois_regr/gp_pois_regr.stan', 1.04)
('stat_comp_benchmarks/benchmarks/eight_schools/eight_schools.stan', 1.09)
('stat_comp_benchmarks/benchmarks/arK/arK.stan', 1.06)
('stat_comp_benchmarks/benchmarks/gp_regr/gp_regr.stan', 1.05)
('stat_comp_benchmarks/benchmarks/garch/garch.stan', 1.05)
('stat_comp_benchmarks/benchmarks/arma/arma.stan', 1.25)
```

Meaning the following need runtime work (I know a few areas that aren't right): low_dim_gauss_mix_collapse, low_dim_gauss_mix, irt_2pl, pkpd, and sir.


VMatthijs:
Woohoo! And they give the same results?

seantalts:
The ones that work give exactly the same results numerically :)

src/stan_math_backend/Expression_gen.ml
```diff
@@ -203,8 +203,19 @@ and gen_distribution_app f =
 (* assumes everything well formed from parser checks *)
```
VMatthijs: Question: how is the code gen currently organised? I see that the expression generation is in its own file. Does the other file only contain code gen for statements? In that case, would it make sense to call it Statement_gen?
Or better still, can we call them Expression_code_gen and Statement_code_gen?

src/stan_math_backend/Expression_gen.ml
```diff
@@ -203,8 +203,19 @@ and gen_distribution_app f =
...
 and gen_fun_app ppf f es =
```
VMatthijs: Also, does this file also include all of Bob's work on expression gen?

src/frontend/Ast_to_Mir.ml
```diff
@@ -352,12 +352,12 @@ let read_decl dread decl_id transform sizedtype smeta decl_var =
     | Indexed (_, indices) -> {var with expr= Indexed (readfn var, indices)}
     | _ -> readfn var
   in
-  let st =
+  let forl, st =
     match dread with
-    | ReadData -> sizedtype
-    | ReadParam -> param_size transform sizedtype
+    | ReadData -> (for_scalar, sizedtype)
```
VMatthijs: This is really interesting! So the reading of data just iterates over eigen-types, rather than doing something efficient using Eigen library functions?

bob-carpenter: Did you have an Eigen library function in mind?

src/frontend/Ast_to_Mir.ml
```diff
@@ -770,11 +763,15 @@ let%expect_test "read data" =
      (For (loopvar sym1__) (lower (Lit Int 1)) (upper (Lit Int 5))
       (body
        (Block
-        ((Assignment (mat ((Single (Var sym1__))))
-          (Indexed
-           (FunApp CompilerInternal FnReadData__
-            ((Lit Str mat) (Lit Str matrix) (Lit Int 10) (Lit Int 20)))
-           ((Single (Var sym1__)))))))))) |}]
+        ((For (loopvar sym2__) (lower (Lit Int 1))
```
VMatthijs: I see that indexing for loops in the MIR also starts at 1 (rather than 0, as C++ likes it). Is that intentional?

src/frontend/Ast_to_Mir.ml
```diff
@@ -770,11 +763,15 @@ let%expect_test "read data" =
...
+          (upper (FunApp StanLib Times__ ((Lit Int 10) (Lit Int 20))))
```
VMatthijs: I see here that you are indexing an Eigen matrix with a single index living in [1, n_rows * n_columns]. Does that work as a way to get the elements out?

src/frontend/Ast_to_Mir.ml
```diff
@@ -770,11 +763,15 @@ let%expect_test "read data" =
...
+          (body
+           (Block
+            ((Assignment (mat ((Single (Var sym1__)) (Single (Var sym2__))))
+              (Indexed
+               (FunApp CompilerInternal FnReadData__
+                ((Lit Str mat) (Lit Str matrix) (Lit Int 10) (Lit Int 20)))
+               ((Single (Var sym1__)) (Single (Var sym2__)))))))))))))) |}]
 
 let%expect_test "read param" =
   let m = mir_from_string "parameters { matrix<lower=0>[10, 20] mat[5]; }" in
```
VMatthijs: In this test, the way you are assigning to the array `mat` seems very inefficient, or am I wrong? It seems to me like you are doing something like
```
for (i in 1:n)
  x[i] = do_possibly_expensive_computation()[i]
```
Is that how you always assign to arrays? Or is this specific to the read_param situation.

src/frontend/Ast_to_Mir.ml
```diff
@@ -810,11 +807,11 @@ let%expect_test "gen quant" =
   print_s [%sexp (m.generate_quantities : stmt_loc list)] ;
   [%expect
     {|
-    ((IfElse (Var emit_generated_quantities__)
+    ((Decl (decl_adtype DataOnly) (decl_id mat)
+      (decl_type (SArray (SMatrix (Lit Int 10) (Lit Int 20)) (Lit Int 5))))
+     (IfElse (Var emit_generated_quantities__)
```
VMatthijs: Could you maybe remind me what this emit_generated_quantities__ flag is for again? Thanks!

src/frontend/Ast_to_Mir.ml
VMatthijs: Not really. What I am mostly surprised by is to see code like
```
for (i in 1:100)
  mat[i] = some_possibly_expensive_function_call()[i]
```

bob-carpenter:
> I see here that you are indexing an Eigen matrix with a single index living in [1, n_rows * n_columns]. Does that work as a way to get the elements out?


Yes, and it'll visit in column-major order as the data is stored.

bob-carpenter:
> -           (FunApp CompilerInternal FnReadData__
> -            ((Lit Str mat) (Lit Str matrix) (Lit Int 10) (Lit Int 20)))

I don't know what the source was here, but you can grab the size out of a matrix with .size() same way you can grab .rows() and .cols().

> for (i in 1:n)
>   x[i] = do_possibly_expensive_computation()[i]
> 
> Is that how you always assign to arrays? Or is this specific to the read_param situation.

What's the expensive computation?  Isn't it just indexing and maybe bounds checking?

bob-carpenter:
> Could you maybe remind me what this emit_generated_quantities__ flag is for again? Thanks!

The write_array method has a flag that controls whether the generated quantities are written or not.  There's also one for the transformed parameters.  The transformed parameters are a deterministic function of the parameters and data, and the generated quantities are a function of all that plus randomness.

VMatthijs:
> Could you maybe remind me what this emit_generated_quantities__ flag is for again? Thanks!
> The write_array method has a flag that controls whether the generated quantities are written or not. There's also one for the transformed parameters. The transformed parameters are a deterministic function of the parameters and data, and the generated quantities are a function of all that plus randomness.

Thanks!

VMatthijs:
> - (FunApp CompilerInternal FnReadData__ - ((Lit Str mat) (Lit Str matrix) (Lit Int 10) (Lit Int 20)))
> I don't know what the source was here, but you can grab the size out of a matrix with .size() same way you can grab .rows() and .cols().
> for (i in 1:n) x[i] = do_possibly_expensive_computation()[i] Is that how you always assign to arrays? Or is this specific to the read_param situation.
> What's the expensive computation? Isn't it just indexing and maybe bounds checking?

In this case, yes. In the case of read_params the same pattern is used. I guess the "possibly expensive computation" there is to read an array of parameters, after which only one entry is used. No idea if that counts as wasteful, but I wanted to double check.

src/stan_math_backend/Expression_gen.ml
seantalts: No, it contains all of the rest of the code gen. I'd pull out the statement gen into another file if you think that'd help. The expression_gen was kept separate as a sort of learning exercise for Bob, and I'm not unhappy with that split out, so it might make sense to go ahead and do the same for statements?

src/stan_math_backend/Expression_gen.ml
seantalts: Yep, that work is the direct predecessor of this file.

src/frontend/Ast_to_Mir.ml
seantalts: In this case we don't have that expensive call, but I think it's basically up to code gen to figure that out and do something like instruction selection here where it can see the entire for loop and do the right thing.

src/frontend/Ast_to_Mir.ml
seantalts: Yeah, right now I'm keeping the MIR indexed to 1 everywhere (or trying to). This seemed a little easier because we often want to print out the index in terms the user expects.

src/frontend/Ast_to_Mir.ml
seantalts: I'm not sure, indexing matrices at that level is broken right now though in general :D

seantalts: I mean with Eigen you can generally write `mat(i)` where `0 < i < (mat.rows() * mat.cols())`. But the code gen is broken.

src/frontend/Ast_to_Mir.ml
seantalts: General purpose array assignment isn't simplified at all in the MIR yet, so it does the same thing stan2 does (using the `stan::model::assign` function). And here we know that reading parameters isn't expensive, and I am tentatively okay with punting that to code-gen for the backends where there is a more efficient way to do this (but open to other thoughts, just seems secondary at the moment).

src/frontend/Ast_to_Mir.ml
seantalts: Like, in the code somewhere? Or just in the PR? If here, it's just a flag we allow to be passed to write_array (and maybe another method...):
```cpp
void stan_model::write_array(RNG& base_rng__,
                             std::vector<double>& params_r__,
                             std::vector<int>& params_i__,
                             std::vector<double>& vars__,
                             bool include_tparams__,
                             bool include_gqs__,
                             std::ostream* pstream__) const {
```

seantalts:
Woops. Went through and answered all of the comments in line only to see Bob had answered a bunch via email at the bottom. Sorry for the spam. I think all the questions are answered and issues are noted, so merging! Thank you both!

src/stan_math_backend/Expression_gen.ml
VMatthijs: Sure! Sounds good! That sort of organization might make it a bit easier to browse this beast.

bob-carpenter:
> On May 29, 2019, at 6:14 PM, seantalts <notifications@github.com> wrote:
> 
> @seantalts commented on this pull request.
> 
> In src/frontend/Ast_to_Mir.ml:
> 
> > @@ -770,11 +763,15 @@ let%expect_test "read data" =
>       (For (loopvar sym1__) (lower (Lit Int 1)) (upper (Lit Int 5))
>        (body
>         (Block
> -        ((Assignment (mat ((Single (Var sym1__))))
> -          (Indexed
> -           (FunApp CompilerInternal FnReadData__
> -            ((Lit Str mat) (Lit Str matrix) (Lit Int 10) (Lit Int 20)))
> -           ((Single (Var sym1__)))))))))) |}]
> +        ((For (loopvar sym2__) (lower (Lit Int 1))
> +          (upper (FunApp StanLib Times__ ((Lit Int 10) (Lit Int 20))))
> 
> I mean with Eigen you can generally write mat(i) where 0 < i < (mat.rows() * mat.cols()). But the code gen is broken.

Right, but the condition should be written as i < mat.size(), not with the explicit multiplication.

bob-carpenter:
Oh, do the email replies not show up inline?  I never see my own responses in email, but I thought that was just GitHub saving me from spamming myself.

seantalts:
Yeah, emailed responses seem to often have the relevant quote but are pasted at the bottom of the PR when accessed via web.

PR #138, opened 2019-05-29
Fix MIR pretty printing bugs (#116)

### LINKED ISSUE #116: Fix MIR pretty printer bugs
The MIR pretty printer currently makes braces { } part of if-, for-, while- and fundef- statements. At the same time, it doesn't print them for blocks. We need to print braces for the Block structure of the program instead and not as a part of these other statements.
The trick will be to still get sensible indentation.

There is another bug in the pretty printer which is more severe: the output var blocks are currently not consistently closed. This behaviour also seems to depend on the OS. For instance, the static binary on Alpine does close them, but the Ubuntu one does not. This makes the pretty printer hard to use for testing optimization passes.

enetsee:
This PR fixes #116; previously `if`, `for`, `foreach` and `while` statements always printed contained statements in curly braces whilst block did not indicate any scoping was happening. The PR reverses that so that braces are only printed if the body of one the statements is a block and blocks are printed with braces outside of those statements.

### DIFF
--- a/src/middle/Middle.ml
+++ b/src/middle/Middle.ml
@@ -107,19 +107,17 @@ let pp_statement pp_e pp_s ppf = function
   | Return (Some expr) -> Fmt.pf ppf {|%a %a;|} pp_keyword "return" pp_e expr
   | Return _ -> pp_keyword ppf "return;"
   | IfElse (pred, s_true, Some s_false) ->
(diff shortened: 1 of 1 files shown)


VMatthijs:
Running some tests for this now on the dependence analysis branch.

PR #140, opened 2019-05-30
Switch end-to-end tests to use performance suite and cmdstan `stanc3-dev` branch

seantalts:


seantalts:
Since every PR must have apparently two things in it, I also put in a fix for the low_dim_gauss_mix stat_comp_benchmarks model. Also fun: now you can see a list of which models we are testing and confirm working here: https://github.com/stan-dev/stanc3/pull/140/files#diff-a64e71122706823ee0cf96c71b0b78ba

Jenkinsfile
```diff
@@ -45,25 +45,22 @@ pipeline {
...
-                   git clone -j${env.PARALLEL} --recursive --branch develop https://github.com/stan-dev/cmdstan
-                   cd cmdstan && make -j${env.PARALLEL} build && cd ..
-               """
+          git clone --recursive --depth 50 https://github.com/stan-dev/performance-tests-cmdstan
+                   """
                 sh """
-                   eval \$(opam env)
-                   CMDSTAN="`pwd`/cmdstan" dune runtest test/integration/good/code-gen
+          cd performance-tests-cmdstan
+          STANC=\$(readlink -f ../bin/stanc) ./compare-git-hashes.sh "stat_comp_benchmarks --tests-file ../notes/working-models.txt" develop stanc3-dev develop develop
```
VMatthijs: I can't say I understand any of what this is doing, but I'm just going to trust that it is right.

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -562,27 +562,24 @@ let pp_log_prob ppf p =
...
-                  , assign
-                      { expr= FunApp (CompilerInternal, f, args @ [var "lp__"])
-                      ; emeta }
-                  , Some
-                      (assign {expr= FunApp (CompilerInternal, f, args); emeta})
-                  )
-            ; smeta }
-        | _ -> {stmt; smeta} )
-      p.log_prob
+  let rec add_jacobians {stmt; smeta} =
```
VMatthijs: good idea to give this a name!

Jenkinsfile
seantalts: That's okay with me, haha. I think if we were post-launch a code reviewer might just inspect the Jenkins output to be sure it was testing the right thing (in this case stan2 vs stanc3 for the list of models).

VMatthijs: I tried that but Jenkins is giving me a 404 when I try to look at the output.

seantalts:
Oops. @serban-nicusor-toptal migrated Jenkins master over to an EC2 machine last night, but looks like there are still some kinks.

VMatthijs:
Looks good. Good idea to make the Mac and Windows and Static tests only run on master by the way!

PR #143, opened 2019-05-31
Fix #142 - statements `dropped in semantic_check_block` 

### LINKED ISSUE #142: Bug introduced in semantic check
The recently merged PR https://github.com/stan-dev/stanc3/pull/126 introduced a bug in the semantic check. When checking the following model
```
      model {
        int x;
        for (i in 1:6) {
          print(x + 42);
          continue;
          x = 3;
        }
      }
```
the semantic check throws away the statement `x=3`.
If all is well, the AST pretty printer should be checking on all of our test models the invariant that the AST doesn't change under checking (that it only gets decorated). Clearly, this situation just wasn't a part of the test models.

The issue here is to fix this particular example and probably make sure something similar isn't happening with `break`. More generally though, I'm worried that more bugs may have been introduced in the semantic check. It was just such a big refactor. How do we convince ourselves that there haven't? (By the way, this PR for instance rolled back previous bug fixes to the MIR pretty printer. I've now reintroduced them again in the dependence-analysis branch which hopefully will be merged today.)

@enetsee , want to have a look at this?

enetsee:
During semantic checking of blocks, we drop statements after an escape statement when determining the return type but all statements should have bee included in the semantically checked block but only those used in determining return type were.

### DIFF
--- a/src/frontend/Semantic_check.ml
+++ b/src/frontend/Semantic_check.ml
@@ -1354,16 +1354,15 @@ and semantic_check_block ~loc ~cf stmts =
      do not count for the return type. 
   *)
   let validated_stmts =
-    List.map ~f:(semantic_check_statement cf) stmts
-    |> Validate.sequence
-    |> Validate.map ~f:list_until_escape
+    List.map ~f:(semantic_check_statement cf) stmts |> Validate.sequence
   in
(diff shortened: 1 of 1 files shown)


VMatthijs:
Looks good to me!

PR #144, opened 2019-05-31
Fix #63, which fixed low_dim_gauss_mix

### LINKED ISSUE #63: Add unnormalized distributions to the language
Based on discussions with @seantalts and @bob-carpenter :

For each distribution, e.g. `poisson_lpmf`, make the compiler accept the function name `poisson_unnormalized_lpmf` as well.
This should code gen to `poisson_lpmf<true>`.
This allows us to treat
` x ~ poisson(42);`
as syntactic sugar for
`target += poisson_unnormalized_lpmf(x| 42);`.

This is related to issue https://github.com/stan-dev/stan/issues/1299 .

Also make sure that algebraic optimizations like `bernoulli_lpmf(x|inv_logit(theta))->bernoulli_logit_lpmf(x|theta)` and automatic vectorization optimizations apply to these unnormalized distributions as well.

seantalts:
Thanks for the pointer @syclik :)

### DIFF
--- a/src/middle/Mir.ml
+++ b/src/middle/Mir.ml
@@ -200,3 +200,5 @@ type internal_fn =
   | FnPrint
   | FnReject
 [@@deriving sexp]
+
+let proportional_to_distribution_infix = "_propto"
--- a/src/stan_math_backend/Expression_gen.ml
+++ b/src/stan_math_backend/Expression_gen.ml
@@ -190,14 +190,23 @@ and read_data ut ppf es =
   pf ppf "context__.vals_%s(%a)" i_or_r pp_expr (List.hd_exn es)
 
 and gen_distribution_app f =
+  let sfx = "_unnormalized" in
   if
+    String.is_suffix f ~suffix:("_lpmf" ^ sfx)
+    || String.is_suffix f ~suffix:("_lpdf" ^ sfx)
+    || String.is_suffix f ~suffix:("_log" ^ sfx)
+  then
+    Some
+      (fun ppf ->
+        let f = String.chop_suffix_exn ~suffix:sfx f in
+        pf ppf "%s<propto__>(@[<hov>%a@])" f (list ~sep:comma pp_expr) )
+  else if
     String.is_suffix f ~suffix:"_lpmf"
     || String.is_suffix f ~suffix:"_lpdf"
     || String.is_suffix f ~suffix:"_log"
   then
     Some
-      (fun ppf ->
-        pf ppf "%s<propto__>(@[<hov>%a@])" f (list ~sep:comma pp_expr) )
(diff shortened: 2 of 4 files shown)


seantalts:
I feel like there's a better way to write this but blanking now.

src/middle/Mir.ml
```diff
@@ -200,3 +200,5 @@ type internal_fn =
   | FnPrint
   | FnReject
 [@@deriving sexp]
+
+let proportional_to_distribution_infix = "_propto"
```
VMatthijs: Is this the syntax we are now going for for the unnormalized distributions?

seantalts: So I needed a way when translating the AST to the MIR to mark a distribution as propto (Bob's preferred keyword). We thought it would be a good idea to just include that in the function name and allow users to use it too. It shows up in code gen because we have to remove it.

seantalts: Ah, re: the actual word choice - I don't care at all. Bob had a strong opinion in #63. You guys can battle it out there :P

VMatthijs: OK! So now this propto could be in function names form the MIR onwards?
I wonder if we should patch the type checking so it can also deal with it seeing that some optimizations will use the type checking on function names appearing in the MIR. Then it'll automatically also will become available in the language.

seantalts: Yeah, that's right. That sounds good to me, and happy to switch the actual word choice to whatever.

seantalts:
I'll merge for now, we can revisit word choice easily later.

src/stan_math_backend/Expression_gen.ml
```diff
@@ -190,14 +190,23 @@ and read_data ut ppf es =
...
 and gen_distribution_app f =
+  let ifx = Middle.proportional_to_distribution_infix in
   if
+    String.is_suffix f ~suffix:(ifx ^ "_lpmf")
+    || String.is_suffix f ~suffix:(ifx ^ "_lpdf")
+    || String.is_suffix f ~suffix:(ifx ^ "_log")
+  then
+    Some
+      (fun ppf ->
+        let f = String.substr_replace_first ~pattern:ifx ~with_:"" f in
```
VMatthijs: @seantalts , I just realized that this doesn't quite do the right thing in some edge cases. Could you maybe fix it? It could be that a function name has "_propto" in it earlier than in the suffix, technically. We don't want to replace the first, but we want to replace the last.

seantalts: Hmm, there is no `substr_replace_last`, and I think it would be very confusing to, at this point, create distributions with "_propto" in their name (possibly we should disallow this for UDFs?). I guess I can reverse it, replace first, reverse it again :P

PR #145, opened 2019-05-31
Fixing semantic check to accept _propto distributions

VMatthijs:
This addresses issue https://github.com/stan-dev/stanc3/issues/63 .

### DIFF
--- a/src/middle/Stan_math_signatures.ml
+++ b/src/middle/Stan_math_signatures.ml
@@ -7,8 +7,32 @@ open Mir
 (** The signatures hash table *)
 let stan_math_signatures = String.Table.create ()
 
+(** Name manging to get propto versions of distributions to also type check *)
+let remove_propto_infix ~suffix str =
+  Option.map
+    ~f:(fun x -> x ^ suffix)
+    (String.chop_suffix str
+       ~suffix:(proportional_to_distribution_infix ^ suffix))
+
+let propto_name_manging name =
+  Option.value
+    (remove_propto_infix name ~suffix:"_log")
+    ~default:
+      (Option.value
+         (remove_propto_infix name ~suffix:"_lpdf")
+         ~default:
+           (Option.value
+              (remove_propto_infix name ~suffix:"_lpmf")
+              ~default:name))
+
+let%expect_test "propto name mangling" =
+  let name = propto_name_manging "normal_propto_lpdf" in
+  print_s [%sexp (name : string)] ;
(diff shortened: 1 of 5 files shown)


seantalts:
What do you think of this:

```ocaml

(** Name manging to get propto versions of distributions to also type check *)
let remove_propto_infix suffix ~str =
  str
  |> String.chop_suffix ~suffix:(proportional_to_distribution_infix ^ suffix)
  |> Option.map ~f:(fun x -> x ^ suffix)

let propto_name_manging (name : string) =
  let suffices = ["_log"; "_lpdf"; "_lpmf"] in
  List.map ~f:(remove_propto_infix ~str:name) suffices @ [Some name]
  |> List.filter_opt |> List.hd_exn

let%expect_test "propto name mangling" =
  propto_name_manging "normal_propto_lpdf" |> print_string ;
  propto_name_manging "normal_lpdf" |> ( ^ ) "; " |> print_string ;
  propto_name_manging "normal" |> ( ^ ) "; " |> print_string ;
  [%expect {| normal_lpdf; normal_lpdf; normal |}]

```

seantalts:
Or like, any of the ideas contained therein. I wanted to see if I could figure out another way to write it, not sure this is a win. @enetsee may have other ideas. We can merge it for now and keep working at it :)

enetsee:
Doing this in OCaml using the standard library + Core is a bit ugly:

```ocaml
let propto_name_mangling name = 
  Option.(
    first_some 
      (first_some 
        (remove_propto_infix name ~suffix:"_log")
        (remove_propto_infix name ~suffix:"_lpdf")
      )
      (remove_propto_infix name ~suffix:"_lpmf")
    |> value ~default:name
  )
```
I would be inclined to define the flipped version of `first_some` (which is called `orElse` in F#): 

```ocaml
let propto_name_mangling2 name = 
  let or_else y x = Option.first_some x y in
  Option.(
    remove_propto_infix name ~suffix:"_log"
    |> or_else (remove_propto_infix name ~suffix:"_lpdf")
    |> or_else (remove_propto_infix name ~suffix:"_lpmf")
    |> value ~default:name
  )
```

seantalts:
That looks good to me, thank you! I'm gonna add or_else to some kind of Utils module inside Middle (for now).

VMatthijs:
That sounds like a good idea. I also notice that I wrote `name_manging` rather than `name_mangling`.

PR #147, opened 2019-06-01
Moving some stuff around, some tidying up based on PR comments

seantalts:
1) Made a Utils module in Middle to hold some utility functions that don't depend on the MIR or other Middle stuff. For now just populated with a few things but we can move things over as we see or create them.
2) replaced my ugly replace_last with something a little more robust based on Matthijs and Michael's code (though I got tired of typing out the distribution suffices and went with a list-based approach)
3) Pulled out Statement code gen into Statement_gen.ml


src/middle/Middle.ml
```diff
@@ -199,3 +167,9 @@ let operator_return_type_from_string op_name argtypes =
 
 let operator_return_type op =
   operator_return_type_from_string (string_of_operator op)
+
+let rec sexp_of_expr_typed_located {expr; _} =
```
VMatthijs: Is the point of these to not print all the metadata in the sexpressions?

seantalts: yep

VMatthijs:
Why are you deleting  test/integration/example-models/knitr/car-iar-poisson/nycTracts10/nycTracts10.RData ? Could we keep that?

seantalts:
Yeah, good point, we should probably keep that. Will try to get it added to the gitattributes before merging...

seantalts:
Oh, that file is actually garbage (a gzip). I'll ungzip it.

seantalts:
Or no, it's some other kind of "OGR" binary file... I'm not sure how to get Micky's git to interact nicely with us on that one... Maybe I'll just leave it for now and pray?

seantalts:
Oh SO weird, github is reporting that I deleted that file but I actually just changed the line endings.

PR #149, opened 2019-06-04
Fix some more models with code gen

seantalts:


seantalts:
I am so excited - all the stat comp benchmark models are passing now! 

@VMatthijs (and @enetsee) I changed the MIR pretty printer to print binary operators infix, hope this is okay with you guys! It changed a bunch of golds in the optimizer unit tests but I think you'll be pleased.

seantalts:
Damn, my local test suite was broken in a way that I hopefully [categorically fixed](https://github.com/stan-dev/performance-tests-cmdstan/commit/401d26dd77f5d8700ff76775f34f384986b116b8). This doesn't fix all the stat_comp_benchmark models after all. Will adjust the list...

VMatthijs:
Ah, still very close. I wonder why the IRT model isn't working. It seems so simple.

seantalts:
I agree, but haven't figured it out yet. Bernoulli logit is claiming its
arguments are different sizes...

src/middle/Mir_pretty_printer.ml
```diff
@@ -91,6 +91,11 @@ let rec pp_expr pp_e ppf = function
   | Var varname -> Fmt.string ppf varname
   | Lit (Str, str) -> Fmt.pf ppf "%S" str
   | Lit (_, str) -> Fmt.string ppf str
+  | FunApp (StanLib, name, [lhs; rhs])
+    when Option.is_some (Mir_utils.operator_of_string name) ->
+      Fmt.pf ppf "%a %a %a" pp_e lhs pp_operator
+        (Option.value_exn (Mir_utils.operator_of_string name))
```
VMatthijs: Could you maybe pretty print some extra parens here? Because the Ast_to_mir is removing parentheses, we might lose the correct associativity here if we don't.

src/stan_math_backend/Expression_gen.ml
```diff
@@ -175,6 +175,8 @@ and gen_misc_special_math_app f =
           else pf ppf "%s(@[<hov>%a@])" f (list ~sep:comma pp_expr) es )
   | f when f = string_of_internal_fn FnLength ->
       Some (fun ppf -> gen_fun_app ppf "stan::length")
+  | f when f = string_of_internal_fn FnNegInf ->
```
VMatthijs: Am I right in thinking that you are still missing
```
  | FnMakeArray
  | FnMakeRowVec
  | FnReadData
  | FnReadParam
  | FnWriteParam
  | FnConstrain
  | FnUnconstrain
  | FnCheck
  | FnPrint
  | FnReject
```
here?

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -472,7 +472,73 @@ using stan::math::lgamma;
 using stan::model::prob_grad;
 using namespace stan::math; |}
 
+let rec xform_readdata sizes s =
+  let get_single_size = function
```
VMatthijs: Shouldn't this sort of transformation happen in AST_to_mir?

It also seems a little bit hacky, to the untrained eye. Not that I have a better alternative to propose.

src/middle/Mir_pretty_printer.ml
seantalts: Sure thing!

src/stan_math_backend/Expression_gen.ml
seantalts: Sort of - you have found a code smell. There is a weird conceptual mismatch here where some of these functions will be taken care of in other places. But yeah there are also some missing. 

The key issue that maybe we should talk about at a meeting is that I was thinking that since `negative_infinity` is a StanLib function we should label FnNegInf as `StanLib` when we create it from the compiler. But perhaps the thing to actually do is just use the actual string `negative_infinity` and not create a `FnVariant` for it? OR to call it `CompilerInternal` and pretend that it doesn't have a StanLib equivalent. I'm not sure. What do you all think? I'm leaning towards one of those two options now - having these special variants sometimes labeled StanLib and sometimes CompilerInternal is not helpful.

VMatthijs: Ah, I see what you mean. So I guess all the CompilerInternal ones are handled elsewhere and this is just supposed to catch the cases of FnNegInf and FnLength where they are StanLib?

seantalts: Yeah :< Seems bad now that we're actually using it.

src/stan_math_backend/Stan_math_code_gen.ml
seantalts: We talked about this in the meeting - I think we should continue to think about where the line is for this stuff. I think this one in particular is super C++/StanMath dependent and it belongs here, but we might actually want to move more of Ast_to_Mir into the code gen side of things and spin off a separate MIR to MIR translation as part of each backend (that we could hopefully then re-run the optimizer on, or only optimize afterwards).

src/stan_math_backend/Expression_gen.ml
seantalts: I'll make a github issue to deal with this at some point in the future.

seantalts:
Added the parens, made an issue for the other one- PTAL

VMatthijs:
I just checked and I definitely parsed the array sizes the wrong way around. Thanks for fixing! It's right now.

src/stan_math_backend/Stan_math_code_gen.ml
VMatthijs: Sounds good!

seantalts:
Thanks! Tests passing; merging :)

Issue #112, opened 2019-05-01
Fn-kind is determined not quite correctly in semantic check


VMatthijs:
In the semantic check, the function kind (user defined, math library, compiler internal) is calculated in not quite the right way:

1. Conditional function applications also need a function kind
2. It is not enough to check whether a function is a Math library name, as nullary Math library functions may be overloaded in Stan (don't ask why...). It seems that what we need to do is check whether there is a user defined function in the symbol table with the right signature.'

enetsee:
I ran across a couple of issues that made me think that our current categorization of function kinds was lacking whilst making the changes to separate frontend/backend etc and @seantalts found several cases where what is current marked as `CompilerInternal` should actually be `StanLib`. 

> Conditional function applications also need a function kind

Could you give me an example of conditional function application?

> It is not enough to check whether a function is a Math library name, as nullary Math library functions may be overloaded in Stan (don't ask why...). It seems that what we need to do is check whether there is a user defined function in the symbol table with the right signature

I saw this just now whilst working on the error-variants branch.

I have to confess I'm still a bit confused on the details of exactly what each of our functions kinds is intended for and how we determine them. Could we set that design out here?

Alongside the immediate changes necessary for the initial release, Is it worth starting to set out what our intention is with some thought to how we might reduce the coupling with Stan Math?

seantalts:
> I have to confess I'm still a bit confused on the details of exactly what each of our functions kinds is intended for and how we determine them. Could we set that design out here?

If a user defines something with that name and type sig:
  `UserDefined`
Else if it refers to a function in the Stan library:
  `StanLib`
If it is neither but instead a construct of the MIR represented as a function application:
  `CompilerInternal`

Examples of the last include the functions for reading and writing parameters and data, constraining and unconstraining vars, and checking vars.

>Alongside the immediate changes necessary for the initial release, Is it worth starting to set out what our intention is with some thought to how we might reduce the coupling with Stan Math?

I think for the moment we've decided that the signatures supported by the Stan compiler now constitute the Stan standard library, which happens to have an implementation in the Stan Math library. I think at some point in the future it would be cool if someone could use stanc3 with their own stdlib definitions. I think I'd turn to you for ideas on how to do this - I might naively suggest making the Stan_math_signatures module provide an interface that the rest of the code can use and having that be pluggable at compile time such that you could compile a stanc3 with your own stdlib signatures.

PR #153, opened 2019-06-07
Fix fn kind 1

VMatthijs:
This fixes point 1. of issue https://github.com/stan-dev/stanc3/issues/112 . Moreover, it does a lot of DRYing.

I'd rather do point 2. in a separate PR as it's quite a different matter.

### DIFF
--- a/src/frontend/Semantic_check.ml
+++ b/src/frontend/Semantic_check.ml
@@ -266,13 +266,14 @@ let semantic_check_fn_normal ~loc id es =
     | Some (_, UFun (_, Void)) ->
         Semantic_error.returning_fn_expected_nonreturning_found loc id.name
         |> error
-    | Some (_, UFun (listed_tys, rt))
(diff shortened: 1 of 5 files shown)


src/frontend/Semantic_check.ml
```diff
@@ -725,6 +640,30 @@ and semantic_check_expression cf ({emeta; expr} : Ast.untyped_expression) :
...
+and semantic_check_funapp ~is_cond_dist id es cf emeta =
+  let name_check =
+    if is_cond_dist then semantic_check_conddist_name
+    else semantic_check_fn_conditioning
+  in
+  Validate.(
+    es
+    |> List.map ~f:(semantic_check_expression cf)
+    |> sequence
+    >>= fun ues ->
```
seantalts: :sunglasses: :sunglasses: :sunglasses:

VMatthijs:
Nice, thanks! I think I see one more simplification I can make. Will then merge it.

PR #155, opened 2019-06-09
[codegen] Fix irt_2pl

seantalts:


src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -472,59 +472,104 @@ using stan::math::lgamma;
...
+let add_pos_reset ({stmt; smeta} as s) =
+  match stmt with
+  | For {body; _}
+    when contains_fn (string_of_internal_fn FnReadData) false body ->
+      [{stmt= Assignment ((pos, []), loop_bottom); smeta}; s]
+  | _ -> [s]
+
+let rec use_pos_in_readdata {stmt; smeta} =
+  match stmt with
+  | Block
```
seantalts: This is so ugly! Not sure what to do about it.

enetsee: I'm not sure there is anything that _can_ be done; in F# & Haskell there is a neat feature called [Active Patterns](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/active-patterns) and [Pattern Synonyms](https://gitlab.haskell.org/ghc/ghc/wikis/pattern-synonyms) respectively that may be helpful but I think OCaml has no equivalent.

seantalts:
@VMatthijs if it's okay with you I'm just gonna start merging these in just because I've been doing all of this work serially and at this point it'd probably be hard to go back and just fix this branch (then I'd have to update all the other branches).

That said, I'm still hoping splitting it up makes it easier for y'all to comment and I can make another PR fixing the comments that come up in these PRs.

seantalts:
You can see when this merged to master that it fixed 2 models:
http://3.13.221.221:8080/blue/organizations/jenkins/stanc3/detail/master/339/tests

src/stan_math_backend/Stan_math_code_gen.ml
VMatthijs: Yep.. I know of no good alternative.

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -533,11 +604,19 @@ let escape_name str =
...
+    |> List.concat_map ~f:invert_read_fors
+    |> List.concat_map ~f:add_pos_reset
+    |> List.map ~f:use_pos_in_readdata
+  in
   let p =
     { p with
       prog_name= escape_name p.prog_name
-    ; prepare_data= List.map ~f:(xform_readdata []) p.prepare_data
-    ; transform_inits= List.map ~f:(xform_readdata []) p.transform_inits }
+    ; prepare_data= fix_data_reads p.prepare_data
```
VMatthijs: @seantalts , could you explain a bit more about why these need to be fixed rather than generating the correct code in AST_to_mir in the first place?

seantalts: It's the same thing we talked about in the meeting - the API to a Stan Math model class asks for data in the wrong order, so we have to invert the for loops in code gen in a C++ specific way. There are also a few other items that are even more C++ specific in there (using the `pos` variable, etc).

VMatthijs: Right! So I guess you're saying this is funny behaviour that is specific to this particular backend? That makes sense.

VMatthijs:
> You can see when this merged to master that it fixed 2 models:
> http://3.13.221.221:8080/blue/organizations/jenkins/stanc3/detail/master/339/tests

I see you're also testing on example models. Do we know that those models even perform well?

seantalts:
By and large they do not perform well. There is a list of ones that have been stable enough for regression testing that we currently use for Stan that is what I have stanc3 testing on master right now. And you can see the red section shrinking! :grin: http://jenkins.mc-stan.org:8080/job/stanc3/job/master/

There's a slightly larger list we don't test regularly because they don't take long enough to have stable timings that I will probably add to this suite once we nail all of these, I guess? And I'll also want to add Mitzi's mother model, some specific test cases, and any regressions we come across too.

src/stan_math_backend/Stan_math_code_gen.ml
seantalts: Yep. :/ And that raises the question of like, is the current thing where we translate a sized type into MIR for loops a good abstraction for the MIR at all (because other backends might not want that either). But so far it's workable so I'm leaving these more perfectionist philosophical questions off my plate until we launch (though others could take a crack at them).

VMatthijs: I think that makes a lot of sense. Once everything is in place, we should probably do a big refactor if we're feeling brave.

VMatthijs:
Gotcha! Love the graphs by the way.

seantalts:
:D The only bad thing is that the performance graphs are absolute but they should be relative to stan2. Just need to go in and make a script that will create a relative timings xml Jenkins can read and switch to that one.

PR #156, opened 2019-06-09
[codegen] simplify indexed C++ generation

seantalts:


seantalts:
[built on top of #155]

seantalts:
I'm not sure this one is a good idea to do now; I could just drop it. I think the `assign` and `rvalue` calls are safer.

VMatthijs:
It seems to me like @bob-carpenter  would be the right person to review this.

seantalts:
Don't worry about taking a look at this - it will be superceded by https://github.com/stan-dev/stanc3/issues/232. Closing.

PR #157, opened 2019-06-10
[codegen] Fix SIR stat_comp_benchmark model!

seantalts:
no longer based on #156

### DIFF
--- a/src/stan_math_backend/Expression_gen.ml
+++ b/src/stan_math_backend/Expression_gen.ml
@@ -34,9 +34,10 @@ let rec stantype_prim_str = function
   | UArray t -> stantype_prim_str t
   | _ -> "double"
 
-let local_scalar ut = function
-  | DataOnly -> stantype_prim_str ut
-  | AutoDiffable -> "local_scalar_t__"
+let local_scalar ut ad =
+  match (ut, ad) with
+  | _, DataOnly | UInt, AutoDiffable -> stantype_prim_str ut
+  | _, AutoDiffable -> "local_scalar_t__"
 
 (* stub *)
 let rec pp_return_type ppf = function
--- a/src/stan_math_backend/Expression_gen.ml
+++ b/src/stan_math_backend/Expression_gen.ml
@@ -83,18 +83,17 @@ let suffix_args f =
 let gen_extra_fun_args f = suffix_args f @ ["pstream__"]
 
 let rec pp_index ppf = function
-  | All -> pf ppf "stan::model::index_omni()"
-  | Single e -> pf ppf "stan::model::index_uni(%a)" pp_expr e
-  | Upfrom e -> pf ppf "stan::model::index_min(%a)" pp_expr e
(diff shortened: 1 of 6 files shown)


seantalts:
(again, still open to comments on this one!)

src/stan_math_backend/Expression_gen.ml
```diff
@@ -34,9 +34,10 @@ let rec stantype_prim_str = function
   | UArray t -> stantype_prim_str t
   | _ -> "double"
 
-let local_scalar ut = function
-  | DataOnly -> stantype_prim_str ut
-  | AutoDiffable -> "local_scalar_t__"
+let local_scalar ut ad =
+  match (ut, ad) with
+  | _, DataOnly | UInt, AutoDiffable -> stantype_prim_str ut
```
VMatthijs: What about `UArray UInt, AutoDiffable` etc here? Or do we only care about scalars? (Just want to double check.)

test/integration/good/code-gen/cpp.expected
```diff
@@ -38,28 +47,27 @@ class eight_schools_ncp_model : public prob_grad {
...
     sigma = std::vector<double>(J, 0);
     
     pos__ = 1;
     for (size_t sym1__ = 1; sym1__ <= J; ++sym1__) {
-      stan::model::assign(sigma, stan::model::cons_list(stan::model::index_uni(sym1__), stan::model::nil_index_list()), 
-                          context__.vals_r("sigma")[(pos__ - 1)], "assigning variable sigma[(sym1__ - 1)]");
-      stan::model::assign(pos__, stan::model::nil_index_list(), (pos__ + 1), "assigning variable pos__");
+      assign(sigma, cons_list(index_uni(sym1__), nil_index_list()), context__.vals_r("sigma")
+                                                                    [(pos__ - 1)], "assigning variable sigma[(sym1__ - 1)]");
+      assign(pos__, nil_index_list(), (stan::model::deep_copy(pos__) + 1), "assigning variable pos__");
```
VMatthijs: I remember that the old compiler would warn on making deep copies. Is that a runtime warning that the C++ currently also gives or would it be a compile time warning whenever `deep_copy` code is generated?

src/stan_math_backend/Expression_gen.ml
seantalts: Hmm, the way this is used it's intended to be called on a non-UArray, but it's not super obvious. I'll just add the recursive case to be safe.

test/integration/good/code-gen/cpp.expected
seantalts: And this is a particularly ugly place for a deepcopy, given that `pos__` is a compiler-generated integer lol. I can fix this particular one by checking the type to see that it's not an int. 

It's not a runtime warning - I think it probably needs to become stanc3-compile-time warning. Let's talk about it at the meeting and see how we should incorporate that.

PR #158, opened 2019-06-10
Fix hepatitisME model

seantalts:
Turns out using declarations to immediately generate checks doesn't work - you need to move the checks to the end of their respective blocks because they may be assigning the variables (and modifying them) at arbitrary locations in the block, and they need to be checked after that.


seantalts:
(based on #157 (based on #156 (based on #155)))

VMatthijs:
@seantalts , let me know when this is ready for review.

seantalts:
Ready! Sorry about that. Was rebasing things to get the indexing stuff out of there in case we didn't want to go with it.

PR #160, opened 2019-06-10
Fixing bug in FunKind determination

VMatthijs:
This fixes part 2 of issue https://github.com/stan-dev/stanc3/issues/112 .

### DIFF
--- a/src/frontend/Semantic_check.ml
+++ b/src/frontend/Semantic_check.ml
@@ -282,8 +282,7 @@ let semantic_check_fn_normal ~loc id es =
         |> ok
     | Some _ ->
         (* Check that Funaps are actually functions *)
-        Semantic_error.returning_fn_expected_nonreturning_found loc id.name
-        |> error
+        Semantic_error.returning_fn_expected_nonfn_found loc id.name |> error
     | None ->
         Semantic_error.returning_fn_expected_undeclaredident_found loc id.name
         |> error)
--- a/src/frontend/Semantic_check.ml
+++ b/src/frontend/Semantic_check.ml
@@ -305,14 +305,24 @@ let semantic_check_fn_stan_math ~loc id es =
       |> Semantic_error.illtyped_stanlib_fn_app loc id.name
       |> Validate.error
 
-let fn_kind_from_identifier id =
-  if is_stan_math_function_name id.name then StanLib else UserDefined
+let fn_kind_from_application id es =
+  (* We need to check an application here, rather than a mere name of the 
+   function because, technically, user defined functions can shadow
+   constants in StanLib. *)
+  if
+    stan_math_returntype id.name
+      (List.map ~f:(fun x -> (x.emeta.ad_level, x.emeta.type_)) es)
+    <> None
+    || Symbol_table.look vm id.name = None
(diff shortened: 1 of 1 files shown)


VMatthijs:
@seantalts , could you have a look at this?

src/frontend/Semantic_check.ml
```diff
@@ -305,14 +305,24 @@ let semantic_check_fn_stan_math ~loc id es =
...
-let fn_kind_from_identifier id =
-  if is_stan_math_function_name id.name then StanLib else UserDefined
+let fn_kind_from_application id es =
+  (* We need to check an application here, rather than a mere name of the 
+   function because, technically, user defined functions can shadow
+   constants in StanLib. *)
+  if
+    stan_math_returntype id.name
+      (List.map ~f:(fun x -> (x.emeta.ad_level, x.emeta.type_)) es)
+    <> None
```
seantalts: can UDFs shadow MathLib functions too? Or just constants? If functions too we'd need to change this logic

VMatthijs: Only constants, that is, function which take zero arguments.

seantalts: Hmm, so... is just the 2nd half of the `||` sufficient in that case? If there's no symbol at this point for it then it's probably StanLib right?

VMatthijs: Well spotted! Will change it now.

VMatthijs: Ah never mind. It's not that simple sadly. The counter example is for instance the program
```
parameters {
  real e;
}
transformed parameters {
  print(e());
}
```
Then, `e` in the `print`-statement needs to be inferred to have FnKind StanMath. However, your proposed formula would give it kind UserDefined.

VMatthijs:
@seantalts , is it OK, otherwise?

src/frontend/Semantic_check.ml
seantalts: yeah :( I guess you need to know the type of the symbol. Is that difficult to get? In my head the logic is basically the same as C++, where you assume user-defined stuff was defined after built-in stuff:

if name in symbol_table && symbol_table_type_of name == type_implied_by_funapp then UserDefined else (* we could check that it lives in StanLib or just say we're fucked anyway if it's not in there *) StanLib

seantalts: Basically saying "what would this resolve to at compile time for clang" because I think those are the actual semantics because we just spit out C++, haha.

VMatthijs: Yes, it's basically that logic. It's not at all hard to get the type of the symbol if it's in the symbol table. Would a different phrasing along those lines be more clear, do you think?

seantalts: I guess so, if only because to me the logic here seems like it will do what we need but is subtly different from looking the type up in the symbol table, if that makes sense? I'd kind of rather try to model the actual semantics if it's the same difficulty, but whatever's easier is better. just my $0.02.

VMatthijs: OK! That make sense. I'll submit a patch.

Issue #74, opened 2019-03-16
Compound assign declare not translated to MIR properly


VMatthijs:
I was trying some tests on the model below, and it appears like it's not being translated to the MIR correctly. In particular, the assignments seem to be thrown away. The model with `int i; i = 42;` does seem to work, so perhaps it's a problem with compound declare assign? 

```
      transformed data {
        int i = 42;
        int j = 2;
      }
      model {
        for (x in 1:i) {
          print(i + j);
        }
      }
```

PR #161, opened 2019-06-10
Add types to lhs

VMatthijs:
This PR fixes issues https://github.com/stan-dev/stanc3/issues/79 and https://github.com/stan-dev/stanc3/issues/74 .

### DIFF
--- a/src/frontend/Ast.ml
+++ b/src/frontend/Ast.ml
@@ -116,13 +116,15 @@ type 'e transformation =
   | Covariance
 [@@deriving sexp, compare, map, hash]
 
+type 'e lhs = {assign_identifier: identifier; assign_indices: 'e index list}
+[@@deriving sexp, hash, compare, map]
+
 (** Statement shapes, where we substitute untyped_expression and untyped_statement
     for 'e and 's respectively to get untyped_statement and typed_expression and
     typed_statement to get typed_statement    *)
 type ('e, 's) statement =
   | Assignment of
-      { assign_identifier: identifier
-      ; assign_indices: 'e index list
+      { assign_lhs: 'e lhs
       ; assign_op: assignmentoperator
       ; assign_rhs: 'e }
   | NRFunApp of fun_kind * identifier * 'e list
--- a/src/frontend/Ast.ml
+++ b/src/frontend/Ast.ml
@@ -116,7 +116,13 @@ type 'e transformation =
   | Covariance
 [@@deriving sexp, compare, map, hash]
 
-type 'e lhs = {assign_identifier: identifier; assign_indices: 'e index list}
+type ('e, 'm) lhs_with =
+  { assign_identifier: identifier
+  ; assign_indices: 'e index list
+  ; assign_meta: 'm }
+[@@deriving sexp, hash, compare, map]
+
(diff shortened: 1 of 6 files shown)


VMatthijs:
This depends on https://github.com/stan-dev/stanc3/pull/160 .

src/frontend/Ast.ml
```diff
@@ -116,13 +116,32 @@ type 'e transformation =
...
+
 (** Statement shapes, where we substitute untyped_expression and untyped_statement
     for 'e and 's respectively to get untyped_statement and typed_expression and
     typed_statement to get typed_statement    *)
-type ('e, 's) statement =
+type ('e, 's, 'l) statement =
   | Assignment of
-      { assign_identifier: identifier
-      ; assign_indices: 'e index list
+      { assign_lhs: 'l
```
VMatthijs: This would be more natural with higher-kinded polymorphism:
```
assign_lhs: 'e 'l
```
but OCaml doesn't support that.

VMatthijs:
@seantalts , could you maybe have a look at this one?

seantalts:
Will take a look, maybe in an hour or so

seantalts:
high level comment - is it possible to just use the expression metadata type here and say that an lhs gets one of those too? then we don't need the extra `'l` through everywhere, etc.

VMatthijs:
> high level comment - is it possible to just use the expression metadata type here and say that an lhs gets one of those too? then we don't need the extra `'l` through everywhere, etc.

That's what I initially wanted to do and I don't think that can be done easily. `'e` stands for expression, not for expression metadata. The alternative is to parameterize statement over expression metadata, but then the definition of statement would get more complicated (with a lot of repeated `expression` in it). This was the simplest pattern I could come up with.

seantalts:
Ah, damn. I wonder if we should have just gone with that instead of two-level types...

seantalts:
Ah I see, and you skipped the two-level type for lhs already too (which is what I was thinking of next, but you already beat me to it). Yeah makes sense!

seantalts:
I mean the other semi-ghetto option here would be to just figure out the type dynamically during Ast_to_Mir... How's that option sound now that you're deep into it?

VMatthijs:
> I mean the other option here would be to just figure out the type dynamically during Ast_to_Mir... How's that option sound now that you're deep into it?

I thought about that one too, as it'd be nice to keep the AST as simple as possible. The problem is that the recomputation of the type is a bit non-trivial. It would involve populating a symbol table and then calling `semantic_check_expression`. We can't just reuse the symbol table from the old semantic check as we might have local variables that have again gone out of scope. I suspect it's best to just save the type on the AST during semantic checking. That would also keep the interface between semantic_check and AST_to_MIR simpler.

seantalts:
Makes sense. Thanks!

seantalts:
Hey Matthijs, should the `id_type_` represent the type of the lhs var before any indexing and the `lhs_type_` post-indexing? I'm trying to use it now for simplifying indexing and I'm seeing arrays of matrices with both of those types set to `UMatrix` for example.

seantalts:
Oh wait, never mind - I think this is my fault somewhere else.

Issue #106, opened 2019-04-12
Replace `let _ =` with either a semicolon if `_` referred to `()` or adding a type signature


seantalts:
see "Returning unit" and "Defensive programming" https://opensource.janestreet.com/standards/ (the latter was clarified at the OCaml meetup to refer to cases where you are applying a function and might accidentally swallow a partially applied function instead of just the value it was supposed to return).

adamhaber:
I'll try to take this one.

seantalts:
I think they actually removed the section that detailed this issue. I think basically it's bad practice to use `let _ = ` because you are losing type information and can end up swallowing potential errors for longer. For example, if you say `let _ = f a in`  (where this is applying f to a, `f(a)` in Python) we think that `f` is supposed to return unit. But if you modify `f` to take two arguments, then `let _ = f a in` still works but now `_` is swallowing a partially applied function and f is never getting run (because of currying in OCaml). So we want to make a syntactic change (can be in just one file at first) such that wherever we see `let _ = <whatever> ` we change it to `<whatever> ;` because the semi-colon expects its left-hand side to be unit and throws an error otherwise.

adamhaber:
I'm trying to understand how to replace some of these. For example, in lexer.mll, there is:
```
                                if Stack.length include_stack = 1
                                then Parser.EOF
                                else
                                  let _ = (Stack.pop_exn include_stack) in
                                  let old_lexbuf =
                                    (Stack.top_exn include_stack) in
                                      token old_lexbuf }
```
Changing `let _ = (Stack.pop_exn include_stack) in` to `Stack.pop_exn include_stack ; ` resulted in:
```
File "projects/stanc3/src/frontend/lexer.mll", line 185, characters 34-61:
Error: This expression has type lexbuf but an expression was expected of type
         Parser.token
```

Also would appreciate help on how to replace `let _ = Debugging.ast_logger ast in` in stanc.ml.

seantalts:
Ah ha, so 
```ocaml
let _ = (Stack.pop_exn include_stack)
```
is actually discarding the value returned from Stack.pop_exn rather than just discarding unit. So the style guide here recommends we annotate the type of the value we're discarding:
```ocaml
let _ : lexbuf = Stack.pop_exn include_stack in
```
(the parens aren't necessary, but ocamlformat can't run on .mll files apparently).

[...]

PR #170, opened 2019-06-10
Location indexing

VMatthijs:
This PR implements all the functionality we need on the MIR to easily treat locations as integers in C++ code gen.


src/analysis_and_optimization/Mir_utils.ml
```diff
@@ -298,3 +298,29 @@ let rec update_expr_ad_levels autodiffable_variables e =
...
+    typed_prog_num * string Int.Table.t =
+  let module LocSp = struct
+    type t = location_span
+
+    let compare = compare_location_span
+    let hash = hash_location_span
+    let sexp_of_t = sexp_of_location_span
+  end in
+  let label_to_location = Int.Table.create () in
+  let location_to_label = Hashtbl.create (module LocSp) in
```
VMatthijs: @seantalts , this is the use of local state that I am talking about. How bad is this? Should I rewrite it using a purely functional traverse to enforce local data flow?

VMatthijs:
(This depends on https://github.com/stan-dev/stanc3/pull/161 )

src/analysis_and_optimization/Mir_utils.ml
seantalts: Seems fine given that you return an immutable map at the end. Code is reasonably easy for me to read.

seantalts:
I think this need to number locations will be fairly specific to C++ code gen, so I might be inclined to keep this in `backend` at first?

Separately, what do you think about offloading location printing to another module entirely where the API is very simple, something like `pp_globals ppf prog` and `pp_span ppf span`, where the globals one is called when we need to emit global variables in C++ and the span is called for each statement when we unwrap its metadata. Then, that module can build the hash table at emit globals time and just look up locations in it whenever it prints a span. This could end up being slightly slower, but I don't think it should be noticable. There's something kind of nice about the idea that the entire program doesn't have to be rewritten to deal with locations, but I guess as long as the rest of the code is written to be parametric over `smeta` then it doesn't matter. However, I think a bunch of code is likely not written parametrically over `smeta` right now because a bunch of it was written before that was born, so we'd have to go through and make those changes. If we go with the modifying the tree approach then the API would be those two functions + `alter_prog : prog -> (prog, Map)` and then the caller has to handle passing the Map along to all other functions (like `pp_generate_quantities` through to `trans_stmt` etc). 

Anyway, just trying to sort of sketch a few ideas and their potential pitfalls out with my familiarity with code gen. What do you think about that?

VMatthijs:
> Separately, what do you think about offloading location printing to another module entirely where the API is very simple, something like `pp_globals ppf prog` and `pp_span ppf span`, where the globals one is called when we need to emit global variables in C++ and the span is called for each statement when we unwrap its metadata. Then, that module can build the hash table at emit globals time and just look up locations in it whenever it prints a span.

Sure, that would work too. You'd just have to pass the hash table as an argument to pp_span or accept that there's some global state that pp_span has access to.

VMatthijs:
> There's something kind of nice about the idea that the entire program doesn't have to be rewritten to deal with locations, but I guess as long as the rest of the code is written to be parametric over `smeta` then it doesn't matter. However, I think a bunch of code is likely not written parametrically over `smeta` right now because a bunch of it was written before that was born, so we'd have to go through and make those changes.

Do you mean code generation code? Does it ever inspect smeta? The only change would be that smeta is an int rather than a location span. I had assumed it wouldn't break anything.

VMatthijs:
> If we go with the modifying the tree approach then the API would be those two functions + `alter_prog : prog -> (prog, Map)` and then the caller has to handle passing the Map along to all other functions (like `pp_generate_quantities` through to `trans_stmt` etc).

I imagined that you call alter_prog once, that the Map would not need to pass to anything except to the pp_globals, and then that each statement just updates the current error location to the integer in the meta data? Or do you mean that the Map would need to be passed to e.g. pp_generate_quantities as that would then call pp_globals? In my mind we either need to be passing around the Map or we need to have it as global state.

src/analysis_and_optimization/Mir_utils.ml
VMatthijs: Great! Thanks for looking at it. I just don't want to be religious about the purely functional style. I'd rather be pragmatic personally. Though I guess a point could be made for being consistent.

seantalts:
> Do you mean code generation code? Does it ever inspect smeta? The only change would be that smeta is an int rather than a location span. I had assumed it wouldn't break anything.

When generating new statements I previously had to just include a `no_span` or something `smeta`, so yeah some of that code would need to be updated to be parametric. It's doable.

> I imagined that you call alter_prog once, that the Map would not need to pass to anything except to the pp_globals, and then that each statement just updates the current error location to the integer in the meta data?

Ah yeah, the map just needs to get passed to `Locations.pp_globals` or whatever we call it, we wouldn't need it at the statement level if we transmuted the `prog` first. My bad.

Yeah, I think your thing makes sense. Just to try to paraphrase what I think the API would look like (pls correct where applicable):

Locations.mli
```
  type loc_t = int
  type state_t = (int, stmt_loc) Map 
  val prepare_prog : (mtype_loc_ad with_expr, stmt_loc) prog -> (mtype_loc_ad with_expr, loc_t) prog * state_t
  val pp_globals : Format.formatter -> typed_and_int_prog * state_t -> unit
  val pp_smeta : Format.formatter -> loc_t -> unit
```

VMatthijs:
Thanks for suggesting the interface, @seantalts ! That makes sense to me. It's easy for me to implement that, I think.

VMatthijs:
@seantalts , I think this is ready for review. One question is whether we want to add these update statement locations already to the MIR as statements. Then the optimizer can get rid of a lot of redundant ones. It could also get rid of a lot of redundant braces, which could make the code easier to read. On the other hand, it seems a bit funny to do this sort of backend specific transformation in the middle section.

VMatthijs:
I guess something that still needs to happen is to add these try catch blocks.

seantalts:
Yeah I can look at that. The infrastructure etc is all there, but it's commented out.

src/stan_math_backend/Statement_gen.ml
```diff
@@ -75,9 +69,34 @@ let trans_math_fn fname =
     value ~default:(fname, [])
       (bind (internal_fn_of_string fname) ~f:math_fn_translations))
 
+let rec block_wrapping {stmt; smeta} =
+  let in_block {stmt; smeta} =
+    { stmt=
+        ( match stmt with
+        | Block l -> Block l
+        | SList l -> Block l
```
seantalts: Aren't SLists different semantically from Blocks?

src/stan_math_backend/Statement_gen.ml
```diff
@@ -75,9 +69,34 @@ let trans_math_fn fname =
...
+  in
+  { stmt= block_wrapping_base (map_statement (fun x -> x) block_wrapping stmt)
+  ; smeta }
+
 let rec pp_statement (ppf : Format.formatter)
     ({stmt; smeta} : (mtype_loc_ad, 'a) stmt_with) =
+  let {stmt; smeta} = block_wrapping {stmt; smeta} in
+  (* Make sure that all statements are safely wrapped in a block in such a way that we can insert a location update before *)
   let pp_stmt_list = list ~sep:cut pp_statement in
+  let _ =
```
seantalts: The preferred Jane Street style for these is to use the semicolon `;` for functions that should return unit.

see https://github.com/stan-dev/stanc3/issues/106

src/stan_math_backend/Statement_gen.ml
```diff
@@ -75,9 +69,34 @@ let trans_math_fn fname =
...
+        map_statement (fun x -> x) in_block stmt
+    | _ -> stmt
+  in
+  { stmt= block_wrapping_base (map_statement (fun x -> x) block_wrapping stmt)
+  ; smeta }
+
 let rec pp_statement (ppf : Format.formatter)
     ({stmt; smeta} : (mtype_loc_ad, 'a) stmt_with) =
+  let {stmt; smeta} = block_wrapping {stmt; smeta} in
+  (* Make sure that all statements are safely wrapped in a block in such a way that we can insert a location update before *)
```
seantalts: Blocks have semantics - anything in that scope will be a local variable. I don't think we can add block structure. Why can't we always insert a location update?

src/stan_math_backend/Locations.ml
```diff
@@ -0,0 +1,60 @@
...
+  let module LocSp = struct
+    type t = location_span
+
+    let compare = compare_location_span
+    let hash = hash_location_span
+    let sexp_of_t = sexp_of_location_span
+  end in
+  let label_to_location = Int.Table.create () in
+  let location_to_label = Hashtbl.create (module LocSp) in
+  let _ = Hashtbl.add label_to_location ~key:no_span_num ~data:no_span in
```
seantalts: Semicolons

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -182,12 +177,19 @@ let pp_ctor ppf (p : typed_prog) =
...
+      (list ~sep:cut pp_statement_zeroing) ppf ls
+    in
     match s.stmt with
     | Decl {decl_id; decl_type; _} -> (
       match decl_type with
       | Sized st -> pp_set_size ppf (decl_id, st, DataOnly)
       | Unsized _ -> () )
-    | Block ls | SList ls -> (list ~sep:cut pp_statement_zeroing) ppf ls
+    | Block ls ->
+        pf ppf "{" ;
```
seantalts: I think pp_block below already adds the curly braces, no?

src/stan_math_backend/Locations.ml
VMatthijs: done.

src/stan_math_backend/Statement_gen.ml
VMatthijs: done.

src/stan_math_backend/Stan_math_code_gen.ml
VMatthijs: Yes, it does in some places. But we need this to also get braces around the try catch blocks.

src/stan_math_backend/Statement_gen.ml
VMatthijs: Yes, but we only call block_wrapping on individual statement, not on lists of statements. Then, we have that {S} and S have exactly the same semantics, as long as we have a single statement. The goal is only to be able to add a second statement updating the location anywhere where we have a single statement. For example, if we start with
```
if 2 >3 
  print(foo);
```
then, to transform it to the appropriate location update, we need to add a block:
```
if 2>3 {
  loc = 4;
  print(foo);
}
```

src/stan_math_backend/Statement_gen.ml
VMatthijs: See this example:
```
if 2 >3 
  print(foo);
```

VMatthijs:
Thanks! I'll revisit this once my PR to Stan is merged.

VMatthijs:
@seantalts , I think this is ready to be merged now. The update to Stan's rethrow_located did the trick. I've also made the changes you requested and added some comments to the code.

src/stan_math_backend/Stan_math_code_gen.ml
seantalts: Hmm I guess it makes sense. Seems slightly code smell that we're doing something for that here in the constructor for try catch rather than dealing with this in the try/catch level... Whatever, tests pass and we can deal with it later :)

src/stan_math_backend/Statement_gen.ml
```diff
@@ -75,9 +69,35 @@ let trans_math_fn fname =
...
+    | _ -> stmt
+  in
+  { stmt= block_wrapping_base (map_statement (fun x -> x) block_wrapping stmt)
+  ; smeta }
+
 let rec pp_statement (ppf : Format.formatter)
     ({stmt; smeta} : (mtype_loc_ad, 'a) stmt_with) =
+  let {stmt; smeta} = block_wrapping {stmt; smeta} in
+  (* Make sure that all statements are safely wrapped in a block in such a way that we can insert a location update before.
+     Note that this should not change the semantics as we are only inserting blocks around the whole body of a for-, while- or if-body.
```
seantalts: Excellent comment! We can merge this but I might suggest I rename the function to something like `block_wrap_stmt_bodies` or maybe `ensure_body_bock` - I think I failed to notice on first read through that it was just adding them for those 3 statement types and only if they were missing, which is key. My bad.

VMatthijs: Absolutely. Go ahead and rename it!

VMatthijs:
@seantalts , is there a way to check whether these location updates have slowed down the C++ much?

seantalts:
Jenkins is down but we should be able to see the performance change on the graphs for master.

PR #171, opened 2019-06-10
Fix litter.stan

seantalts:
Important point: A Mir Single index into a matrix returns a matrix row.

 That is all.


seantalts:
This is just reneging on the previous idea I had where we can just index into a matrix by iterating over the `.size()` (which is .rows() * .cols()) once I realized that apparently the way the C++ and Stan works is that a single integer indexing into a matrix returns a row from that matrix.

at some point we'll want to redo indexing to make it hella simple in the MIR...

Issue #41, opened 2019-02-01
Refactor pretty printer to use Format library


VMatthijs:

palmerlao:
Could I take a crack at this one?

seantalts:
Yeah, that'd be awesome! Let me or @VMatthijs know if you need any pointers
to get started.

VMatthijs:
Fantastic! I think the Mir pretty printer can be a good example of how to use the format library to get the line breaks and indentation right.

seantalts:
Fixed by @palmerlao in #180. Thanks again! Let us know if you need anything - happy to do a hangout to go over the way various modules work and are connected if that helps.

PR #180, opened 2019-06-18
Use Format and Fmt in Pretty_printing.ml

palmerlao:
This is my attempt at #41 .

For the most part, I tried to satisfy the golden tests.
I think the biggest difference is that `for`, `if`, and `else` followed by a single statement (as opposed to a block) now all start new lines with two more spaces. This is a short cut around handling cases along the lines of
```
for (k in 1 : K) for (j in 1 : J) {
  B[j, k] = xi[k] * B_raw[j, k];
}
```  
which would have been printed
```
for (k in 1 : K) for (j in 1 : J) {
                   B[j, k] = xi[k] * B_raw[j, k];
		 }
```
when using boxes in (what I believe is) the intuitive way.

Empty blocks also have trailing whitespace; I'm not very sure how to get around that.


src/frontend/Pretty_printing.ml
```diff
@@ -11,6 +11,38 @@ let indent_num = ref 1
 let begin_indent _ = indent_num := 1 + !indent_num
 let exit_indent _ = indent_num := -1 + !indent_num
 let tabs () = String.make (2 * !indent_num) ' '
```
VMatthijs: Are these three lines dead code?

VMatthijs:
This looks great to me.

One question: do you have any idea how to get the Fmt library to break lines in sensible places to stay within 80 characters per line? That would really be ideal, but I realize it may be a lot to ask for.

seantalts:
Re blocks: Let's try to use the way we print blocks here: https://github.com/stan-dev/stanc3/blob/master/src/middle/Mir_pretty_printer.ml#L163 

Re: line breaks: it should do something like that automatically if you're using an `hov` box. We should probably be using those in general. We can also use the `@[<hov>@]` syntax if you prefer vs. the `Fmt.pp_open_hov_box` calls.

seantalts:
Hey Palmer, do you have time to keep working on this? I think it's pretty close as-is and I'd be happy to pitch in and try to make the changes for line breaks and block printing if you're busy.

palmerlao:
Thanks for the nudge! I think I made some progress on making lines break before 80 characters, but based on `find . -type f -name '*.expected' -a ! -path '*bad*' | xargs grep '^.\{80\}' | grep -v auto-format | grep -v deprecated` probably have a bit more work to do especially around binary operations. 

I could use some clarification or assistance on the block printing issue.

seantalts:
Great, thanks! It's probably OK if some weird examples are over 80 characters for now (though I'll let Matthijs decide since he's reviewing this one).

For blocks, the thing that we have that seems to work is in that link and uses this `@;<1 2>` feature, described in [the doc](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Format.html) like so:

> `@;`: output a 'full' break hint as with print_break. The nspaces and offset parameters of the break hint may be optionally specified with the following syntax: the `<` character, followed by an integer nspaces value, then an integer offset, and a closing `>` character. If no parameters are provided, the good break defaults to a 'space' break hint.

before a vertical box:

```
  | Block stmts ->
Fmt.pf ppf {|{@;<1 2>@[<v>%a@]@;}|} Fmt.(list pp_s ~sep:Fmt.cut) stmts
```

VMatthijs:
Wow, this is starting to look really good! I'm so excited! :-)

src/frontend/Pretty_printing.ml
VMatthijs: Could you delete lines 10-13, as they are dead code?

VMatthijs:
I've just had a thorough look at the printed output and it looks great to me.

@palmerlao , what do you think still remains to be done on this?

@seantalts , I don't see that there is an issue with the blocks in the output. Is it just the style in which the code is written that you would like to be different?

My only request before this gets merged is that the dead code gets deleted.

seantalts:
Oh no, I thought there was an issue with it and that's why it came up and I was just posting how we did it elsewhere. If it works I'm happy.

src/frontend/Pretty_printing.ml
palmerlao: I believe this has been addressed in 75cf2c8

palmerlao:
@VMatthijs If you're happy with the formatting and only suggestion for the code was the dead code, I think it's done.

src/frontend/Pretty_printing.ml
```diff
@@ -24,330 +52,376 @@ let rec unwind_array_type = function
...
+  | Void -> Fmt.pf ppf "void"
+
+and pp_identifier ppf id = Fmt.pf ppf "%s" id.name
+
+and pp_operator ppf = function
+  | Middle.Plus | PPlus -> Fmt.pf ppf "+"
+  | Minus | PMinus -> Fmt.pf ppf "-"
+  | Times -> Fmt.pf ppf "*"
+  | Divide -> Fmt.pf ppf "/"
+  | Modulo -> Fmt.pf ppf "%%"
```
VMatthijs: This should be a single % presumably?

VMatthijs:
Hi @palmerlao , I've looked at the code and it looks great, with the one exception of the %% -> % change that I requested. (Or am I wrong about that one?)

I also performed the following check on your work on all of our 4000 test models:
parse program to AST (call ast1), pretty print, parse pretty printed output again to AST (call ast2). check that ast1=ast2.
Your code passes all checks so it seems great.

Will approve now and will merge once the change to % has been made. Thanks for the contribution! This is the sort of pretty printer we've wanted for years! :-)

src/frontend/Pretty_printing.ml
VMatthijs: I just realized I was wrong about this one. My bad!

VMatthijs:
Never mind about the %. Your code was right after all. Will merge now!

seantalts:
I love the new output, thank you! Cheers :)

PR #185, opened 2019-06-20
Add unconstrained types to the MIR

seantalts:


VMatthijs:
Just let me know when this is ready for review, OK?

seantalts:
I think it's ready - should probably get the MIR looked at ASAP and then merged even though I think it only fixes one model and it's not plugged all the way through. I also thought it might be fine if the MIR pretty printer ignored input and output variables - it's not a serialization format, it's a program pretty printer that we mostly use for tests of things in the Middle layer, none of which change those bits of a prog. But I'd be open to putting them back too but it was nice that it removed 400 lines of tests too.

VMatthijs:
Everything looks good, except I'm still not sure whether we should delete the pretty printers for the output vars etc. I would personally expect a pretty printer to convey most relevant information about the MIR.

seantalts:
Okay, I will add them back in for now. It was just annoying me and didn't seem important, but maybe the best answer is to fix the tests that don't use them to not print them.

seantalts:
want to take another look? Added back in the pretty printer for input and output vars to typed prog.

PR #186, opened 2019-06-20
Reverted back static dockerfile

serban-nicusor-toptal:


seantalts:
thanks! Merging to see if this fixes it...

seantalts:
Looks like it fixed it, thanks. We could even go to the alpine opam image one day but this is good

PR #189, opened 2019-06-20
Fix schools-3

seantalts:
Also, switched to emitting `rvalue` instead of `stan::model::rvalue`


seantalts:
Sorry, can't wait - just realized the `hepatitis.stan` failure was due to using a slightly old copy of the math library in the cmdstan `stanc3-dev` branch and now it passes, so merging this will drop model errors to 0 out of 87 tested models!

PR #192, opened 2019-06-22
Start generating writes for transform_inits

### LINKED ISSUE #178: Fix transform_inits method in code gen
It should be writing to the `params_r__` input vector but is currently not. The param reading and unconstraining might be working correctly already.

seantalts:
Fixes #178

### DIFF
--- a/src/frontend/Ast_to_Mir.ml
+++ b/src/frontend/Ast_to_Mir.ml
@@ -743,12 +743,9 @@ let trans_prog filename p : typed_prog =
     @ compiler_if "emit_generated_quantities__" gq_stmts
   in
   let transform_inits =
-    map
-      (trans_stmt
-         { dread= Some ReadData
-         ; dconstrain= Some Unconstrain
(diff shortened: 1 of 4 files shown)


VMatthijs:
This looks good, but has it been tested anywhere? Is it hard to do that?

seantalts:
I wrote to @bob-carpenter asking if there was a good way to do it. It might just be a manual thing for now.

bob-carpenter:
How can I help?

seantalts:
Oh, just the email about how to test this. Looks like I just need to add a test that sets initial values on both an old and a new Stan program and see that they get the same results. Unfortunately that doesn't fit into any of the existing test paradigms we have right now... Maybe I will just test it manually and leave it not tested automatically for now. It's not like we're testing this automatically with stanc2 anywhere end to end.

seantalts:
Tested this locally and it seemed to work, so merging for now.

PR #194, opened 2019-06-22
Make locations look more like stan2 location outputs for runtime errors

seantalts:


test/integration/included/stanc.expected
```diff
@@ -247,7 +247,7 @@ model {
 }
 
 
-Semantic error in file ../included/incl_stanc_helper_error_spread_over_files.stan, line 2, column 2 to file ../included/error_spread_over_files.stan, line 1, column 4, included from
+Semantic error in '../included/incl_stanc_helper_error_spread_over_files.stan', line 2, column 2 to file ../included/error_spread_over_files.stan, line 1, column 4, included from
```
VMatthijs: This is probably not the behaviour that we want.

src/middle/Middle.ml
```diff
@@ -32,16 +32,16 @@ let string_of_location_span loc_sp =
...
           ( if bl = el then
             sprintf "line %d, %s" bl
               ( if bc = ec then sprintf "column %d" bc
               else sprintf "columns %d-%d" bc ec )
           else sprintf "line %d, column %d to line %d, column %d" bl bc el ec
           )
       else
-        sprintf "file %s, line %d, column %d to file %s, line %d, column %d" bf
-          bl bc ef el ec
+        sprintf "'%s', line %d, column %d to file %s, line %d, column %d" bf bl
```
VMatthijs: probably you want to remove the file and add some quotes around the name here too

seantalts: Good idea! Does this address your other comment that "this is not the behavior we want?"

src/stan_math_backend/Locations.ml
```diff
@@ -46,8 +46,10 @@ let pp_globals ppf location_list =
   let location_list =
     List.map
       ~f:(fun x ->
-        if x = no_span then "Found before start of program."
-        else "At " ^ string_of_location_span x )
+        " ("
+        ^ ( if x = no_span then "found before start of program"
```
VMatthijs: Is this the behaviour that we would like? That means that we should probably never insert no_span ourselves? For instance, AST_to_mir and optimizations should never insert a no_span that ends up being the error location, because we wouldn't really expect those errors to correspond to being triggered before the start of the program.

seantalts: I don't think I changed any behavior here - this is what you had before just with parenthesis added around the outside to match the existing compiler's format, right? Or did I mess that up?

VMatthijs: Agreed. You didn't change the behaviour. I put that in, but I'm still not sure it's right.
Probably it's OK for now.

src/middle/Middle.ml
VMatthijs: It does!

seantalts:
That change ends up breaking a test in Errors.ml where it tries to round trip a location string and fails. But I'm not sure that our location emitting code is the same everywhere... How does this regex need to change for this? https://github.com/stan-dev/stanc3/blob/master/src/frontend/Errors.ml#L31

Do I need to change how locations are emitted in the parser too possibly?

src/stan_math_backend/Locations.ml
seantalts: I need to see some examples but off the cuff I'd say if there's a no_span we just don't emit any location updates at all.

VMatthijs:
Yes, I'd imagine something would also need to change in Preprocessor.ml, at least. This is because I couldn't figure out how to thread the locations through the parser in a more structured form, so I had to convert them to a string and to convert them back again.

src/stan_math_backend/Locations.ml
VMatthijs: That's actually a great idea.

seantalts:
Could you take another look? I propagated further and there are so many things that change! I think it's kind of all fine. 

I also spent a while banging my head against a bug in this and ended up simplifying the location span printing code a ton - it's a little less pretty in cases with includes but the code is much simpler. I think this is probably a good tradeoff.

src/stan_math_backend/Locations.ml
seantalts: Added it to this PR as a separate commit.

test/integration/included/stanc.expected
```diff
@@ -286,9 +283,9 @@ Invalid character found.
 
   $ ../../../../install/default/bin/stanc --include_paths="../included" --auto-format stanc_helper_with_bad_include_parse_error.stan
 
-Syntax error in file ../included/parse-err.stan, line 2, columns 2-9, included from
-file ../included/incl_stanc_helper_parse_error.stan, line 2, column 2, included from
-file stanc_helper_with_bad_include_parse_error.stan, line 2, column 0, parsing error:
+Syntax error in '../included/parse-err.stan', line 2, column 2, included from
+'../included/incl_stanc_helper_parse_error.stan', line 2, column 2, included from
+'stanc_helper_with_bad_include_parse_error.stan', line 2, column 0 to column 9, parsing error:
```
VMatthijs: This doesn't seem quite right.

I'm sorry by the way, I know this code is horrible with all the string mangling.

seantalts: I'm tempted to just leave off the 2nd location entirely if there are includes - it's super difficult as a user of this to parse the hierarchical include structure when two locations are involved. Would that work for you?

src/middle/Middle.ml
```diff
@@ -10,45 +10,31 @@ module Utils = Utils
...
+  (*
+  [%sexp (begin_loc : location)] |> Sexp.to_string_hum
+  |> print_endline ;
+ *)
+  sprintf "%s to %s"
+    (string_of_location begin_loc)
+    (string_of_location
+       ~print_file:(begin_loc.filename <> end_loc.filename)
+       ~print_line:(begin_loc.line_num <> end_loc.line_num)
+       ~print_incl:false end_loc)
```
VMatthijs: I really like this.

test/integration/included/stanc.expected
VMatthijs: Yes, I think that's probably the easiest way out here. When we have includes it's super hard to understand these error messages anyway.

seantalts:
Ok, I think I addressed everything. Thanks for the feedback, I think it's a lot better now than when I first posted it :) Merging!

Issue #196, opened 2019-06-24
Automatically generate dataset for testing Stan program, using the signature specified in the data-block


VMatthijs:

seantalts:
This is a super cool use case for the new compiler and should let us really expand the models we can run (albeit without any statistical information).

Issue #174, opened 2019-06-13
Automatically generate code for forward sampling from model


VMatthijs:
Using a dependence analysis, it should be easy to determine whether a Stan model represents a DAG. That should be a check that should be integrated with @rybern 's factor graph analysis.

If we determine that a Stan program is a DAG, then we can automatically generate code for forward sampling from the model, which is super useful for prior predictive checks and SBC. The idea is basically just to replace ~-statements (and appropriate target increments) with assignments of rng draws. The trick will be to try to transform to the code we want and to then type check before committing to make sure the appropriate rngs exist. Otherwise, we can throw an error suggesting that the user implement the needed rng and submit a PR to Stan Math.

One design decision will be what to do with covariates in the data block (which never appear to the left of a ~-statement). One natural thing to do would be to draw them from a sort of uniform distribution. It might be more useful though for some applications to draw them from a normal distribution. Another option is not to draw them at all but to leave them as IO calls. Perhaps we should offer multiple options to the users.

Note that this feature would be super useful for automatically generating synthetic datasets from models to improve our test coverage.

Voltemand:
But is this possible? See bentanalpha's comment [here](https://discourse.mc-stan.org/t/tool-to-auto-generate-model-diagrams/7361/7?u=voltemand). 

I'd be very to help with this - I've done a bit of thinking about simulating from dag's and auto generating stan code as part of some work on [greta](https://greta-stats.org/) but I'm not very familiar with either Stan's MIR or OCaml...

Best, 

Jeffrey

seantalts:
Stan can express models which would be hard to do this for automatically,
and it can express models for which it would be easy :) and for PPCs there
can be choices made even for DAG models around e.g. whether to create new
groups in a hierarchical model or just simulate within the existing groups.

Eventually we'll likely add a mechanism for labeling parts of the model for
the models where this is harder to do. And some will never be amenable as
they aren't generative. But we always encourage generative modeling and
many of the models people write are amenable.

Voltemand:
I am aware of those subtleties, I was sort of just being overly provocative; I was quite surprised by the assertion that it is impossible.

The way I (naively) think about it is that any sensible (read generative) model written in Stan can be converted to a DAG. 

Anyway, I love to help with all this - particularly if you are willing to accept contributions written by people not fluent in OCaml

seantalts:
[...]

PR #198, opened 2019-06-24
Data generation

VMatthijs:
This addresses issue https://github.com/stan-dev/stanc3/issues/196


VMatthijs:
@seantalts , this is ready for review I think.

It implements most of the functionality we want to be able to run models for which we don't have a dataset. The two mean features that are currently still missing are:
- properly generating a random value in case of the constrained matrices; currently I just generate the identity matrix in these cases as that's so much easier; it would be a significant amount of work to generate a random correlation matrix for example and I'm not sure it's worth it for the applications we have in mind;
- partially evaluating sizes and bounds; for example, while
```
data {
  int N;
  vector<N> x;
}
```
works,
```
data {
  int N;
  vector<N-42> x;
}
```
does not yet.

My sense is that the second feature is much more important to get datasets for models we care about than the first.
Note, however, that we can never do the second part perfectly. For example, it would be really hard to do 
```
data {
  int N;
  vector<foo(N)> x;
}
```
where foo is some math library function.

src/frontend/Data_generation.ml
```diff
@@ -0,0 +1,338 @@
...
+  in
+  String.concat ~sep:"\n" l
+
+(* ---- TESTS ---- *)
+
+let%expect_test "whole program data generation check" =
+  let open Parse in
+  let ast =
+    parse_string Parser.Incremental.program
+      "        data {\n\
```
seantalts: You can use multi-line strings with `{| some strings la la la |}`

VMatthijs: Ah good point. I still don't understand why the output is formatted so strangely though. When I run the dataset generation in the terminal it looks fine.

seantalts:
What is the output of this by the way?

One version of this would be to generate MIR (that gets translated to C++) that would generate the data when compiled, and then we could support anything. I think it could be worth making this complete rather than only working for the basic data types.

VMatthijs:
> One version of this would be to generate MIR (that gets translated to C++) that would generate the data when compiled, and then we could support anything.

I thought about that, but then we wouldn't be able to compare with the old compiler. This way, with an external dataset, we can actually do that. An alternative is to output a different Stan model which moves everything from the datablock to transformed data and uses rngs there. I guess the continuous types can even be moved to parameters. That is all assuming that the only thing we care about is being about to run models for the purposes of testing.

VMatthijs:
> What is the output of this by the way?

The output is an Rdata file that gets written to std::out if you pass a certain command line argument.

VMatthijs:
> I think it could be worth making this complete rather than only working for the basic data types.

In what sense? It works for all types, currently. The only restrictions are the two I outlined above. It doesn't do anything random for these constrained matrix types, but I'm not sure that's a problem. To fully solve the second issue, we'd have to take a different approach to this altogether. We'd be able to get most of the way there by just hooking up the partial evaluator from the optimizer though.

seantalts:
Oh I see, so it can still generate data it's just not random. I think that's fine, sorry I misunderstood. 

Re: c++ - I was imagining not even a separate model but just a C++ file that includes the Math library and compiles to a program that generates an RDump file. But I think if we have complete data type coverage with this PR then that's good enough for me. I'll take a look soon.

VMatthijs:
> Oh I see, so it can still generate data it's just not random. I think that's fine, sorry I misunderstood.
> 
> Re: c++ - I was imagining not even a separate model but just a C++ file that includes the Math library and compiles to a program that generates an RDump file. But I think if we have complete data type coverage with this PR then that's good enough for me. I'll take a look soon.

Right. I see what you mean. That would lead to a solution that would work in full generality with constraints and sizes involving arbitrary math library functions. It would be a lot more labour intensive to write though, so I wasn't sure it was worth it if we only care about this for internal development purposes for generating new datasets for testing models.

I was thinking that the project of automating prior and posterior predictive checks is closely related and that that will be user facing. I think it'd be worth investing more considerable effort into that project and actually generating the C++ to make it work at as high a level of generality as possible. That would probably not be finished for a while though.

This was mostly intended to be a quick fix to increase our test coverage for the new compiler. Perhaps it's worth talking about the use of this temporary solution though. For instance, is this still useful even if the datasets it generates are a bit silly. (I.e. are not generated from prior predictive checks taking into account the model, but rather are sort of randomly generated.)

seantalts:
That all sounds right to me. I think this will be helpful and great as-is for testing :) For the PPCs etc I think we'd need to actually generate the data from the implied data generating process as well and it'd only work for models where we can even tell which things are priors and which aren't so that'd be harder in that way too.

bob-carpenter:
This is getting a bit off topic, but a lot of things like prior and posterior predictive checks are pretty easy if you have a directed graphical model representation of the model and impossible in general if you don't.

VMatthijs:
> This is getting a bit off topic, but a lot of things like prior and posterior predictive checks are pretty easy if you have a directed graphical model representation of the model and impossible in general if you don't.

Agreed. So the idea would be to use a dependence analysis to figure out whether a model has a DAG /generative structure. See also this issue: https://github.com/stan-dev/stanc3/issues/174 .

This PR isn't meant to address that concern yet though. It's about the much  more short term concern of increasing test coverage for comparing stanc2 and stanc3 by generating (possibly silly) datasets for models which currently are lacking data to run the model with.

VMatthijs:
Update: I've implemented partial evaluation. This lets us also deal with data blocks like 
```
data {
  int N;
  vector[N-42] x;
}
```
now.

VMatthijs:
@seantalts , I think this is done. I've used it to generate data sets for all the models in the example-models/knitr folder. But I think it's probably best to check those in in a separate PR, right?

seantalts:
If you have them available feel free to check them in here! It'll be easy enough for me to tell the two apart. Thanks!

src/frontend/Debug_data_generation.ml
```diff
@@ -0,0 +1,557 @@
...
+  let data = Option.value ~default:[] s.datablock in
+  let l, _ =
+    List.fold data ~init:([], Map.Poly.empty) ~f:(fun (l, m) decl ->
+        let value = var_decl_gen_val m decl in
+        ( l @ [var_decl_id decl ^ " <- " ^ print_value_r value]
+        , Map.set m ~key:(var_decl_id decl) ~data:value ) )
+  in
+  String.concat ~sep:"\n" l
+
+(* ---- TESTS ---- *)
```
seantalts: Would you mind moving these to test/unit somewhere? Some of the output is quite long.

VMatthijs: Done.

VMatthijs:
@seantalts , I've added the auto-generated datasets. Do you agree with the file names?

Shall I auto-generate any others?

VMatthijs:
Also, thanks for reviewing!

seantalts:
Filenames look good to me. We can start with these for now, thank you!

PR #204, opened 2019-06-26
Moved building off to master agent and using docker registry for cache.

serban-nicusor-toptal:


README.md
```diff
@@ -8,7 +8,39 @@ We're currently able to successfully compile, link, and run [these models](https
 Stanc3 has 3 main src packages: `frontend`, `middle`, and `stan_math_backend`. The Middle contains the MIR and currently any types or functions used by the two ends.
 The entrypoint for the compiler is in `src/stanc/stanc.ml` which sequences the various components together.
 
-### Distinct Stanc Phases
+### Distinct Stanc Phasesvariables:
+  DIR_DEV: "C:\\home\\sites\\api"
```
seantalts: I think this went in the wrong place :P

serban-nicusor-toptal:
This PR should be stable and ready to merge.   

Docker Registry UI: https://registry-ui.mc-stan.org  
Docker Registry: https://registry.mc-stan.org

seantalts:
What does running our own registry open us up to from a security standpoint? Can we make sure it's closed off to anyone besides Jenkins machines?

serban-nicusor-toptal:
I took care of this.
There should be no security issues. I've proxied it through nginx, added
https certificates and username/password authentication.

seantalts:
Gotcha, cool. I looked over it more thoroughly and have two more questions:
1) How does this deal with a lack of changes to a Dockerfile and dependencies (via scripts)? It looks like it will just publish a new image with a new increment. It would be better to use a `when` condition or something to say only to execute the build phase if those files have changed, I think?
2) How will this deal with a pull request that changes the Dockerfile? In an ideal world, such a PR would use the cache from the docker image on `master` and build its own cached docker image. I think in this code it looks like it might replace the master docker image?

serban-nicusor-toptal:
I mostly took care of the base.
Setting up everything and doing basic versioning and cache.
Now it would be best to have all the possible scenarios on the table, like
yours above so I can improve the PR.
---
1. It will only watch the Dockerfile for changes. I'm open to hear what
other files may change or if we should add a parameter to trigger
build/push.
2. Yes, it will replace it as it is.

Should I do
```
if (*PR*)
    do not push image, just build and use.
```

seantalts:
> if (*PR*)    do not push image, just build and use.

Well, I think so - will it still be able to get the image from the registry if there are no differences to the Dockerfile?

> 1. It will only watch the Dockerfile for changes. I'm open to hear what other files may change or if we should add a parameter to trigger build/push.

Would it be easy enough to make it check the `scripts` folder as well for changes?

serban-nicusor-toptal:
Of course, if the image doesn't need to be be built for each PR we can just pull it from repository.
I'll look into `watching` the scripts folder, shouldn't be too hard.

seantalts:
Apparently "master" doesn't have executors anymore? Causing this build to get stuck, haha.

serban-nicusor-toptal:
Added one executor.  
About the above, docker watches by default the scripts folder because of the `COPY` step.  

```
Command "git merge 44cb9075c23d9bff38cc6bbfecc49d34a961fd55" returned status code 128:

stdout: 

stderr: error: Your local changes to the following files would be overwritten by merge:

	test/integration/example-models/Bayesian_Cognitive_Modeling/CaseStudies/HeuristicDecisionMaking/StopSearchData.RData

	test/integration/example-models/Bayesian_Cognitive_Modeling/CaseStudies/MPT/dataMPT.Rdata

Please commit your changes or stash them before you merge.
```

Argh, it runs into this now.

seantalts:
Hahaha, I hate that problem. I deleted those files from master so we should never deal with it again going forward, but for now you can hopefully git rm -f them in your branch and merge master in.

serban-nicusor-toptal:
Looks fine now: https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3/detail/PR-204/136/pipeline/

seantalts:

Issue #202, opened 2019-06-25
Extend partial evaluator with common functions on arrays and (row)vectors


VMatthijs:
It would be useful to add support for statically evaluating functions like `size`, `dims`, `max`, `min` when called on array, vector and row-vector expressions like
`{4, 2, 1}`
and
`[123, 2, 46, 3]`.

Such functions are used, for example, to determine sizes of containers in Rstanarm, so it would be useful to be able to statically evaluate them in order to
1. generate more efficient code for Rstanarm
2. auto-generate datasets for testing Rstanarm.

PR #212, opened 2019-06-28
Partially evaluate indexed array and row_vector expressions

VMatthijs:
This upgrades the partial evaluator with some cool capacities.

Basically, to evaluate `{24, 35, 46}[2]` to `35`.

### DIFF
--- a/src/analysis_and_optimization/Partial_evaluator.ml
+++ b/src/analysis_and_optimization/Partial_evaluator.ml
@@ -713,19 +713,25 @@ let rec eval_expr (e : Middle.expr_typed_located) =
         | e1', e2' -> EOr (e1', e2') )
       | Indexed (e, l) -> (
         (* TODO: deal with lists of indices *)
-        match (e.expr, l) with
-        | FunApp (t, f, elts), [i]
-          when f = string_of_internal_fn FnMakeArray
-               || f = string_of_internal_fn FnMakeRowVec -> (
-          match get_indices (List.length elts) i with
-          | None -> Indexed (e, l)
-          | Some indices ->
-              if List.length indices = 1 then
-                (List.nth_exn elts (List.hd_exn indices)).expr
-              else FunApp (t, f, List.map indices ~f:(List.nth_exn elts)) )
-        | _, _ -> Indexed (eval_expr e, List.map ~f:eval_idx l) ) ) }
+        match l with
+        | [] -> Indexed (e, l)
+        | [i] -> eval_indexed_expr e l i
+        | _ -> Indexed (e, l) ) ) }
 
 and eval_idx i = map_index eval_expr i
+
+and eval_indexed_expr e l i =
+  match e.expr with
+  | FunApp (t, f, elts)
+    when f = string_of_internal_fn FnMakeArray
+         || f = string_of_internal_fn FnMakeRowVec -> (
+    match get_indices (List.length elts) i with
+    | None -> Indexed (e, l)
+    | Some indices ->
+        if List.length indices = 1 then
+          (List.nth_exn elts (List.hd_exn indices)).expr
+        else FunApp (t, f, List.map indices ~f:(List.nth_exn elts)) )
+  | _ -> Indexed (e, l)
 
 let eval_stmt_base = map_statement eval_expr (fun x -> x)
 let eval_stmt = map_rec_stmt_loc eval_stmt_base
--- a/src/analysis_and_optimization/Partial_evaluator.ml
+++ b/src/analysis_and_optimization/Partial_evaluator.ml
@@ -679,7 +679,8 @@ let rec eval_expr (e : Middle.expr_typed_located) =
       | Indexed (e, l) -> (
         match (e.expr, l) with
         | FunApp (t, f, elts), i :: _
-          when f = string_of_internal_fn FnMakeArray -> (
+          when f = string_of_internal_fn FnMakeArray
+          || f = string_of_internal_fn FnMakeRowVec -> (
           match i with
           | All -> FunApp (t, f, elts)
           | Single e -> failwith "<case>"
--- a/src/analysis_and_optimization/Partial_evaluator.ml
+++ b/src/analysis_and_optimization/Partial_evaluator.ml
@@ -65,13 +65,48 @@ let apply_logical_operator_real (op : string) r1 r2 =
         | "Geq__" -> Bool.to_int (r1 >= r2)
         | s -> raise_s [%sexp (s : string)] ) )
 
+let get_indices length (indices : 'e index) =
+  let make_range bot top =
+    List.range ~start:`inclusive ~stop:`inclusive bot top
+  in
+  let get_int_value e =
+    match e.expr with
+    | Lit (Int, s) ->
+        let i = int_of_string s in
+        if i <= 0 || i > length then None else Some i
+    | _ -> None
+  in
+  let get_int_values l =
+    let l = List.map l ~f:get_int_value in
+    if List.exists l ~f:(fun x -> x = None) then None
+    else Some (List.map l ~f:(fun x -> Option.value_exn x))
+  in
+  let indices =
+    match indices with
+    | All -> Some (make_range 1 length)
+    | Single e -> get_int_values [e]
+    | Upfrom e ->
+        Option.map (get_int_value e) ~f:(fun bot -> make_range bot length)
+    | Downfrom e ->
+        Option.map (get_int_value e) ~f:(fun top -> make_range 1 top)
+    | Between (e1, e2) ->
+        Option.map2 (get_int_value e1) (get_int_value e2) ~f:(fun bot top ->
+            make_range bot top )
+    | MultiIndex e -> (
+      match e.expr with
+      | FunApp (_, f, l) when f = string_of_internal_fn FnMakeArray ->
+          get_int_values l
+      | _ -> None )
+  in
+  Option.map indices ~f:(List.map ~f:(fun x -> x - 1))
+
 let rec eval_expr (e : Middle.expr_typed_located) =
+  let e = {e with expr= map_expr eval_expr e.expr} in
   { e with
     expr=
       ( match e.expr with
       | Var _ | Lit (_, _) -> e.expr
       | FunApp (t, f, l) ->
-          let l = List.map ~f:eval_expr l in
           let get_fun_or_op_rt_opt name l' =
             let argument_types =
               List.map ~f:(fun x -> (x.emeta.madlevel, x.emeta.mtype)) l'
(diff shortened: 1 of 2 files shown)


VMatthijs:
This partly addresses the requests of https://github.com/stan-dev/stanc3/issues/202 .

src/analysis_and_optimization/Partial_evaluator.ml
```diff
@@ -677,9 +712,38 @@ let rec eval_expr (e : Middle.expr_typed_located) =
...
+                        if List.length indices = 1 then
+                          (List.nth_exn elts (List.hd_exn indices)).expr
+                        else
+                          FunApp (t, f, List.map indices ~f:(List.nth_exn elts))
+                    )
+                  | _ -> expr )
+                | i :: l' ->
+                    Indexed
+                      ( { e with
+                          expr= eval_indexed_expr_one_step (Indexed (e, [i]))
```
seantalts: Does this end up creating nested Indexed nodes?

src/analysis_and_optimization/Partial_evaluator.ml
```diff
@@ -677,9 +712,38 @@ let rec eval_expr (e : Middle.expr_typed_located) =
...
+                | [] -> e.expr
+                | [_] -> (
+                  match e.expr with
+                  | FunApp (t, f, elts)
+                    when f = string_of_internal_fn FnMakeArray
+                         || f = string_of_internal_fn FnMakeRowVec -> (
+                    match get_indices (List.length elts) (List.hd_exn l) with
+                    | None -> expr
+                    | Some indices ->
+                        if List.length indices = 1 then
```
seantalts: I think this might look nicer if included in the pattern matching:
```ocaml
| Some [index] -> (List.nth_exn elts index).expr
| Some indices -> FunApp (t, f, List.map indices ~f:(List.nth_exn elts))
```

src/analysis_and_optimization/Partial_evaluator.ml
```diff
@@ -677,9 +712,38 @@ let rec eval_expr (e : Middle.expr_typed_located) =
...
       | Indexed (e, l) ->
-          (* TODO: do something clever with array and matrix expressions here?
-  Note  that we could also constant fold array sizes if we keep those around on declarations. *)
-          Indexed (eval_expr e, List.map ~f:eval_idx l) ) }
+          let rec eval_indexed_expr expr =
+            let rec eval_indexed_expr_one_step expr =
+              match expr with
+              | Indexed (e, l) -> (
+                match l with
+                | [] -> e.expr
```
seantalts: Same here - can you include this pattern into the top level pattern matching? Would be nice not to have pattern matches so deep if possible - a lot to hold in your head at once when reading. here,
```ocaml
              | Indexed (e, []) -> e.expr
              | Indexed ({expr=FunApp (t, f, elts); _}, [index])
                         when f = string_of_internal_fn FnMakeArray
                         || f = string_of_internal_fn FnMakeRowVec -> 
 ....

```

src/analysis_and_optimization/Partial_evaluator.ml
```diff
@@ -65,13 +65,48 @@ let apply_logical_operator_real (op : string) r1 r2 =
...
+  let get_int_value e =
+    match e.expr with
+    | Lit (Int, s) ->
+        let i = int_of_string s in
+        if i <= 0 || i > length then None else Some i
+    | _ -> None
+  in
+  let get_int_values l =
+    let l = List.map l ~f:get_int_value in
+    if List.exists l ~f:(fun x -> x = None) then None
```
seantalts: would `~f:((=) None))` work?

src/analysis_and_optimization/Partial_evaluator.ml
```diff
@@ -65,13 +65,48 @@ let apply_logical_operator_real (op : string) r1 r2 =
...
+    else Some (List.map l ~f:(fun x -> Option.value_exn x))
```
seantalts: here `~f:Option.value_exn` should work I think

src/analysis_and_optimization/Partial_evaluator.ml
```diff
@@ -65,13 +65,48 @@ let apply_logical_operator_real (op : string) r1 r2 =
...
+    | Upfrom e ->
+        Option.map (get_int_value e) ~f:(fun bot -> make_range bot length)
+    | Downfrom e ->
+        Option.map (get_int_value e) ~f:(fun top -> make_range 1 top)
+    | Between (e1, e2) ->
+        Option.map2 (get_int_value e1) (get_int_value e2) ~f:(fun bot top ->
+            make_range bot top )
+    | MultiIndex e -> (
+      match e.expr with
+      | FunApp (_, f, l) when f = string_of_internal_fn FnMakeArray ->
```
seantalts: You could move this one up too:
```ocaml
    | MultiIndex {expr= FunApp (_, f, l); _} 
        when f = string_of_internal_fn FnMakeArray -> get_int_values l
```

serban-nicusor-toptal:
```
Read 455 sample input sentences and 455 error messages.
      ocamlc src/analysis_and_optimization/.analysis_and_optimization.objs/byte/analysis_and_optimization__Partial_evaluator.{cmi,cmo,cmt} (exit 2)
(cd _build/default && /root/.opam/4.07.0/bin/ocamlc.opt -w @a-4-29-40-41-42-44-45-48-58-59-60-40 -strict-sequence -strict-formats -short-paths -keep-locs -g -bin-annot -I src/analysis_and_optimization/.analysis_and_optimization.objs/byte -I /root/.opam/4.07.0/lib/base -I /root/.opam/4.07.0/lib/base/caml -I /root/.opam/4.07.0/lib/base/md5 -I /root/.opam/4.07.0/lib/base/shadow_stdlib -I /root/.opam/4.07.0/lib/bin_prot -I /root/.opam/4.07.0/lib/bin_prot/shape -I /root/.opam/4.07.0/lib/core_kernel -I /root/.opam/4.07.0/lib/core_kernel/base_for_tests -I /root/.opam/4.07.0/lib/fieldslib -I /root/.opam/4.07.0/lib/fmt -I /root/.opam/4.07.0/lib/jane-street-headers -I /root/.opam/4.07.0/lib/parsexp -I /root/.opam/4.07.0/lib/ppx_assert/runtime-lib -I /root/.opam/4.07.0/lib/ppx_bench/runtime-lib -I /root/.opam/4.07.0/lib/ppx_compare/runtime-lib -I /root/.opam/4.07.0/lib/ppx_deriving -I /root/.opam/4.07.0/lib/ppx_expect/collector -I /root/.opam/4.07.0/lib/ppx_expect/common -I /root/.opam/4.07.0/lib/ppx_expect/config -I /root/.opam/4.07.0/lib/ppx_hash/runtime-lib -I /root/.opam/4.07.0/lib/ppx_inline_test/config -I /root/.opam/4.07.0/lib/ppx_inline_test/runtime-lib -I /root/.opam/4.07.0/lib/ppx_sexp_conv/runtime-lib -I /root/.opam/4.07.0/lib/result -I /root/.opam/4.07.0/lib/seq -I /root/.opam/4.07.0/lib/sexplib -I /root/.opam/4.07.0/lib/sexplib0 -I /root/.opam/4.07.0/lib/splittable_random -I /root/.opam/4.07.0/lib/stdio -I /root/.opam/4.07.0/lib/stdlib-shims -I /root/.opam/4.07.0/lib/typerep -I /root/.opam/4.07.0/lib/variantslib -I src/middle/.middle.objs/byte -I src/middle/.middle.objs/native -no-alias-deps -opaque -open Analysis_and_optimization -o src/analysis_and_optimization/.analysis_and_optimization.objs/byte/analysis_and_optimization__Partial_evaluator.cmo -c -impl src/analysis_and_optimization/Partial_evaluator.pp.ml)
File "src/analysis_and_optimization/Partial_evaluator.ml", line 90, characters 6-14:
Error: This variant pattern is expected to have type 'e index
       The constructor Downfrom does not belong to type Middle.index
Post stage```

**Comment by seantalts** [comment]
Hey, I was thinking about this more w.r.t. eliminating all the special indexing types from MIR. I think what we should do is just have a phase that does some optimizations on the AST, including these optimizations. I think only a few will likely deserve to live there at first, but I think these and some other examples would make sense - like I'd like to collapse all Indexed nodes in the MIR to just have a string identifier as the object of indexing (rather than the current full expression) (if that's possible, still working it out).

Here are the ways that we could get rid of all nested Indexed types, I think. Here I'm using int literals for the examples but we can do these transformations with non-constant variable ints as well, and using 'x' to mean any other kind of index and 'arr' to be an int array.
* v[2:3][2] -> v[3]
* v[2][4] -> v[2, 4]
* v[:][x] -> v[x]
* v[x][:] -> v[x]
* v[2][2:3] -> v[2, 2:3]
* v[arr][2] -> (declare new_sym, fill with v[arr]); new_sym[2]
* v[2][arr] -> (declare new_sym, fill with v[2][arr])

**Comment by bob-carpenter** [comment]
> ... would you mind having a look at the list above and seeing if it makes sense to you? 

This one is illegal as the indexes are always read upwards (unlike R, which requires a lot of care on loop boundaries to not spin off into an infinite loop):

```
v[x][3:2] -> v[x][{3, 2}]
```

should be

```
v[x][3:2] -> throw indexing errror because 3:2 is empty
```

This one is subptimal

```
v[arr][2] -> (declare new_sym, fill with v[arr] via for loop); new_sym[2]
```

and can be simplified to

```
v [arr] [2] -> v[arr[2]]
```

The other way around (e.g., `v [2] [arr]`) is OK as is, assuming it's not part of a larger expression that lets it simplify, like `v [2] [arr] [3] -> v[2, arr[3]]`.

It makes me think there's some larger algebra there.  The way I think of all this  indexing is with only two kinds of indexes, multi-indexes and single indexes. So if`arr` is a multiindex so that for a single index `i`, we have

```
foo[arr][i] = foo[arr[i]]
```

When there are multiple indexes, it works the same way,

```
foo [arr1, ..., arrN] [i1, ..., iN] = foo [arr1[i1]] [arr[i2]] ... [arrN[iN]]
```

We only have

```
foo [i] [arr] = foo[i, arr]
foo [i] [j] = foo[i, j]
```

when `i` and `j` are single indexes and `arr` is a multi-index and `foo` has at least two indexing dimensions.  

With multi-indexes, it nests the other way

```
foo [idx] [i] = foo [idx[i]]
```

and then you can work on reducing `idx[i]` if possible.

**Comment by seantalts** [comment]
Whoa, that's really cool! And thanks for explaining your algebra and framework, it makes it so much clearer. I'm excited to implement these symbolic simplifications :)

**Comment by seantalts** [comment]
I'm not sure if this part of the algebra is right:
```
foo [arr1, ..., arrN] [i1, ..., iN] = foo [arr1[i1]] [arr[i2]] ... [arrN[iN]]
```
because foo[arr1] returns a collection of the same rank/dimension as foo. So foo[arr1, arr2] *also* has the same number of dimensions as foo, weirdly. So foo[arr1, arr2, i, j] assumes foo has at least 2 dimensions, of which only j indexes into the 2nd dimension. 

Here's an example with ranges that I think is equivalent and is easier to illustrate:
`foo[2:6, 2:4, 2, 3]` I think you can simplify step by step. First we grab 2-6 of foo. Then we grab 2-4 of that 1-6 section, and then the 2nd element of that. Then 3 doesn't interact with the multi indices at all, it just goes down a dimension. So I think these are all equal:
`foo[1:6, 2:4, 2, 3] = `foo[3:5, 2, 3]` = `foo[5, 3]`.

**Comment by seantalts** [comment]
I think this might be the form:
```
foo [arr1, ..., arrN] [i1, ..., iN] = foo [arr1[arr2[...arrN[i1]]][i2, ..., iN]
```

seantalts:
Oye, I think I'm really off here - stanc3 is assuming this model but Stan 2 doesn't use this model! Yikes. Never mind.

seantalts:

Issue #75, opened 2019-03-16
Remove Mir.Downfrom node


VMatthijs:
Observe that the Mir.Downfrom(e) node is semantically equivalent to Mir.Between(Lit(Int, 1), e), so we can remove that node. It was in the AST only because we wanted a 100% faithful representation of the syntax.

seantalts:
And UpFrom, right?

seantalts:
I think I just want to get rid of these entirely and replace them with For loops...

VMatthijs:
UpFrom is not so clear to me, as you'd need the length. I guess `Indexed(e', [UpFrom(e)])` would be something like `Indexed(e', [Between(e, length(e'))])`. Not quite as straightforward.

seantalts:
> UpFrom is not so clear to me, as you'd need the length. I guess Indexed(e', [UpFrom(e)]) would be something like Indexed(e', [Between(e, length(e'))]). Not quite as straightforward.

There is a FnLength if you wanted to remove UpFrom. I suspect it's a cheap operation and probably what we do under the hood anyway.

PR #214, opened 2019-06-28
Removing redundant downfrom index from MIR

VMatthijs:


VMatthijs:
This addresses issue https://github.com/stan-dev/stanc3/issues/75 . 

I wasn't sure whether it was a good idea to remove UpFrom, as I wasn't sure that the resulting code involving the call the length would be as efficient as what we are doing now.

seantalts:
Fair enough. I want to do a serious refactoring of MIR indices at some point post-launch to make them basically be as simple as Python and C++ indices are rather than the sort of weird Stan-specific stuff they are now (so we can deal with all the Stan semantics in Ast_to_Mir and focus on simpler semantics in Middle and target semantics in backends).

PR #220, opened 2019-07-04
integrate mother model

### LINKED ISSUE #188: Test that Mitzi's mother model gives the correct results
Also, extend the mother model with the missing data types (like unit vector), so these get tested as well.

seantalts:
Added a bunch of stuff to the mother model to make it have suitable output for light / simplified performance testing.
Added it to be tested always and got it compiling and running! Took a lot of changes but now I feel pretty confident that we have most of the language under test. :D 

Fixes #188


test/integration/good/code-gen/mother.stan
```diff
@@ -236,26 +236,39 @@ transformed data {
   simplex[N] td_simplex;
   simplex[N] td_1d_simplex[N];
   simplex[N] td_3d_simplex[N,M,K];
-  cholesky_factor_cov[5,4] td_cfcov_54;
+  cholesky_factor_cov[5,5] td_cfcov_54; // TODO: Change to 5,4
```
VMatthijs: Should this change still happen?

test/integration/good/code-gen/mother.stan
```diff
@@ -271,9 +284,9 @@ parameters {
   cholesky_factor_cov[3] p_cfcov_33_ar[K];
 }
 transformed parameters {
-  real<lower = 0> tp_real_1d_ar[N];
-  real<lower = 0> tp_real_3d_ar[N,M,K];
-  vector<upper=1>[N] tp_vec;
+  real<lower=0> tp_real_1d_ar[N];
+  real<lower=0> tp_real_3d_ar[N,M,K];
```
VMatthijs: should we also add in some lower, upper and offset,multipliers?

test/integration/good/code-gen/mother.stan
```diff
@@ -287,12 +300,59 @@ transformed parameters {
...
+
+  tp_cfcov_54 = p_cfcov_54;
+  tp_cfcov_33 = p_cfcov_33;
+  tp_cfcov_33_ar = p_cfcov_33_ar;
+
+  for (i in 1:2) {
+    for (j in 1:3) {
+      for (m in 1:4) {
+        for (n in 1:5) {
+          tp_ar_mat[m, n, i, j] = 0.4;}}}}
```
VMatthijs: What about also testing the other kinds of indexing like
`a[x: ]`, `a[:x]`, `a[x:y]`, `a[:, 34]`, `a[{13,2,1}, :]`?
We'd ideally want to test them on both the lhs and rhs.

seantalts:
Can you colate all of the other stuff you think should be added to the mother model and add it yourself in a separate PR?I think that would be a really valuable contribution but I don't think it's in the scope of this PR.

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -176,18 +194,16 @@ let pp_ctor ppf (p : Locations.typed_prog_num) =
       | ls -> Some ls )
     | _ -> None
   in
-  let rec pp_statement_zeroing ppf s =
-    let pp_statement_list_zeroing ppf ls =
-      (list ~sep:cut pp_statement_zeroing) ppf ls
+  let pp_statement_zeroing ppf ls =
```
VMatthijs: Could you explain the intention of this code (maybe in a comment)? I am a bit confused about what pp_statement_zeroing is supposed to do and how it should be different from pp_statement.
I guess it should not print any code for declarations and it should only generate the code for setting the sizes of declarations and otherwise it's the same as pp_statement. Where do you use this? (Just out of my own interest in understanding the code gen)

src/frontend/Ast_to_Mir.ml
```diff
@@ -539,17 +542,37 @@ let rec trans_stmt (declc : decl_context) (ts : Ast.typed_statement) =
       let wrap expr = {expr; emeta= {mloc; mtype= UInt; madlevel= DataOnly}} in
       let iteratee = trans_expr iteratee
       and indexing_var = wrap (Var newsym) in
+      let indices =
+        match iteratee.emeta.mtype with
+        | UMatrix ->
+            [ (Ast.Single (Ast.IntNumeral "1"), UInt)
+            ; (Ast.Single (Ast.IntNumeral "1"), UInt) ]
+        | _ -> [(Ast.Single (Ast.IntNumeral "1"), UInt)]
```
VMatthijs:
```let single_one = (Ast.Single (Ast.IntNumeral "1"), UInt) in
        match iteratee.emeta.mtype with
        | UMatrix ->
            [ single_one
            ; single_one ]
        | _ -> [single_one]```

src/stan_math_backend/Stan_math_code_gen.ml
seantalts: Yeah, it probably needs a better name now. It basically just elides the declarations at the top-level. I'll change it to something...

src/frontend/Ast_to_Mir.ml
seantalts: Sure

src/stan_math_backend/Stan_math_code_gen.ml
VMatthijs: Ah right. And should it not recurse into SList then? What if we have a bunch of SLists at the toplevel?

seantalts: I'm operating with the assumption that we're still going to be removing SList - I don't know what the semantics around it are at this point.

PR #226, opened 2019-07-15
Add metadata methods

seantalts:


enetsee:
Code looks good. Only one meta comment / question; would it be worth being consistent with and extending JSONSchema for the format? The convention for array looks pretty similar:

```json
{
  "type": "array",
  "items": {
    "type": "number"
  }
}
```
And this would play nicely if you wanted to spit out types before data, too:

```json
{
  "type": "array",
  "minItems": 2,
  "maxItems": 3
}
```

seantalts:
Oh cool! I hadn't seen that. I'll take a look and probably change it to
that, thanks!

seantalts:
Hmm, looked further - what would we get out of that? I'm not sure it's a huge win - we'd need to add vector and matrix with our own conventions for rows and columns anyway. And we wouldn't really benefit from any of the tools built around JSON Schemas - they look intent on validating that objects fit the schema, which we don't need to do. 

Also not looking forward to telling people why we call doubles/reals "number" 😂

enetsee:
> Also not looking forward to telling people why we call doubles/reals "number" 

LOL... That be good for next April 1? 

Yeah, it's immediately too far away to have much use I guess.

As I said, bikeshedding aside, all looks good to me.

enetsee:
(Not sure how I managed to close the PR)

seantalts:
ok, thanks!

PR #237, opened 2019-07-25
Optimization testing integration

VMatthijs:
This adds extra tests for the compiler optimizations and fixes some minor bugs.


VMatthijs:
@seantalts , could you maybe merge this? It only adds a bunch of extra tests and fixes some minor bugs in the optimizations.

VMatthijs:
@rybern , @seantalts , @enetsee , my new thinking is that we should roll out the optimizations one at a time. I think what was stopping me from getting this work done is that it seemed almost too big for one PR. I think it would be much more motivating (and safe!) to do one PR per optimization from now, ensuring the following:
- you've gone over all the code in the optimization and you've convinced yourself it is correct, taking particular care with all the possibilities for expressions (including compiler internal functions) to side effect
- you've gone over all the unit tests and added to them to convince yourself that the implementation does on the MIR what it should do
- the optimization is wired up
- you've gone over the stat-comp-benchmarks or other models that are being run on Jenkins and have made sure the optimization is stressed and is passing tests on those.

I think this'll will make it possible to start rolling out the optimizations sooner, without having to wait for one of us to have a huge chunk of time available all at the same time. Moreover, I think it'll ensure that we put more energy into ensuring correctness of each individual one, rather than rushing them in and being sloppy because of time pressure.

src/analysis_and_optimization/Partial_evaluator.ml
```diff
@@ -629,12 +674,12 @@ let rec eval_expr (e : Middle.expr_typed_located) =
...
             | op, [{expr= Lit (Int, i); _}] -> (
               match op with
-              | "PPlus_" | "PMinus__" | "PNot__" ->
+              | "PPlus__" | "PMinus__" | "PNot__" ->
                   apply_prefix_operator_int op (Int.of_string i)
               | _ -> FunApp (StanLib, op, l) )
             | op, [{expr= Lit (Real, r); _}] -> (
               match op with
-              | "PPlus_" | "PMinus__" ->
+              | "PPlus__" | "PMinus__" ->
```
seantalts: I think ideally we want to try to use the pattern Micky gave us, which is to match on any string and put a `when` guard checking which kind it is using `internal_fn_of_string` or `operator_of_string`.

test/unit/Optimize.ml
```diff
@@ -1295,6 +1295,340 @@ let%expect_test "unroll nested loop" =
...
+      prepare_data {
+
+      }
+
+      log_prob {
+        {
+          {
+            {
+              {
+                FnPrint__(1, 1, 1);
```
seantalts: Should these really all be so deeply nested in blocks?

Also these tests would be much more readable if they just printed the part under test, e.g. the log_prob here... taking a crack at that by updating the pretty printer to not print empty things and reduce the number of spare lines.

PR #238, opened 2019-07-25
Switch AST to possiblysized variable declaration types

seantalts:
This will be useful for two things:
1) eventually allowing unsized local variables
2) during desugaring generating new declarations that it seems impossible to determine the sizes of in general

### DIFF
--- a/src/frontend/Ast_to_Mir.ml
+++ b/src/frontend/Ast_to_Mir.ml
@@ -79,6 +79,7 @@ and trans_idx = function
 and trans_exprs = List.map ~f:trans_expr
 
 let trans_sizedtype = map_sizedtype trans_expr
+let trans_possiblysizedtype pst = map_possiblysizedtype trans_expr pst
 
 let neg_inf =
   { expr= FunApp (StanLib, string_of_internal_fn FnNegInf, [])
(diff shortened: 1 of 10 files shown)


src/frontend/Ast_to_Mir.ml
```diff
@@ -584,8 +584,14 @@ let rec trans_stmt (declc : decl_context) (ts : Ast.typed_statement) =
       raise_s
         [%message
           "Found function definition statement outside of function block"]
+  | Ast.VarDecl {decl_type= Unsized _; _} ->
+      raise_s [%message "Don't yet support unsized declarations"]
```
VMatthijs: Why do we not yet support unsized declarations at this point? We should be able to translate them to the MIR, right? Isn't it the code generation that's problematic?

seantalts: Yeah, it could be supported it just isn't in this PR. I figured whoever first starts emitting it will have to change it 😂 (figuring it'd be me in the simplify-indexing branch).

src/frontend/Pretty_printing.ml
```diff
@@ -210,29 +210,32 @@ and pp_transformation ppf = function
...
-            e2 )
-    | SArray _ -> (
-      match unwind_sized_array_type st with st, _ ->
-        (Fmt.const pp_sizedtype st, Fmt.nop) )
+and pp_transformed_type ppf (pst, trans) =
+  let pst =
+    match pst with
+    | Middle.Sized st ->
+        Middle.Sized (Fn.compose fst unwind_sized_array_type st)
+    | Unsized ut -> Middle.Unsized ut
```
VMatthijs: I think this is wrong. We seem to want to pull out the base type (discarding all array constructors here).

src/frontend/Ast_to_Mir.ml
seantalts: I'm working on adding it now tho... might as well.

src/frontend/Pretty_printing.ml
seantalts: I still don't get why this is written like this... seems like you'd want each pretty printer to be modular; e.g. unsizedtypes can always be emitted whole by pp_unsizedtype :P Oh well, I'll fix this locally for now.

seantalts:
Hey, fixed it. PTAL

VMatthijs:
Looks good to me now, except there are still some tests failing. Should the expects be updated?

seantalts:
oops, yes probably, will look into it. thanks!

PR #244, opened 2019-07-31
Simplify indexing v0

seantalts:
This is the first chunk that I got working. There are some new things in here:
1) AST->AST desugaring step, with some of @bob-carpenter's indexing algebra simplifications implemented
2) Added unsized types to MIR Assignments so we can eventually get rid of calls to `stan::model::assign`
3) Changed AST to use MIR fun_kind type for functions now that we can generate internal functions from the desugaring step.
4) Trying to replace all Matrix indexing with either calls to `row` or a new internal `FnMatrixElement` function, since this can't be expressed in C-style indexing easily.
5) Trying to allow actually emitting Unsized declarations from the MIR - for now this is done by explicitly adding calls to an internal `FnResizeToMatch` function that requires something of the same "top level" size on the rhs. So this should be called for each layer of a For loop assigning to an undeclared variable for the first time. Not sure how else to deal with it. These map to new boilerplate C++ functions called `resize_to_match`.
6) I also now keep track of which declaration identifiers are global or not and shortcircuit trans_decl if they are not global so that I don't attempt to generate e.g. parameter reads for them.
7) Added a new Frontend_utils file that just has an easy parsing function in it good for testing - not sure where this should live or if it exists.
8) New stanc flags for dumping the output of the various phases. 

TODO:
1) Need to code gen checks that `assign` and `rvalue` were performing
2) Further indexing algebraic desugaring simplifications
3) Eventually, change MIR Indexed (expr, expr index list) to Indexed (expr, expr) - ie all Indexeds will represent a possibly nested C style `arr[i]`.


seantalts:
I'm going to close this for now - I think I might reopen with just simplifications and not removing MultiIndexing - it's actually useful to have it in that format for TFP and Numpy as well.

seantalts:
@VMatthijs re: your question at the meeting today - I think one possible reason the extra desugaring step makes sense as AST->AST is because I was planning to remove all the extra indexing types from the MIR and needed to accomplish the desugaring as a series of passes. It was also nice to have access to the ForEach loop, and it was much easier to use the Semantic_check functionality for inference doing everything in the AST.

Relatedly, is it possibly true that any nested Indexed expressions could be expressed as a single flat Indexed? e.g. mat[5][arr, 2:3] -> mat[5, arr, 2:3].

VMatthijs:
> @VMatthijs re: your question at the meeting today - I think one possible reason the extra desugaring step makes sense as AST->AST is because I was planning to remove all the extra indexing types from the MIR and needed to accomplish the desugaring as a series of passes. It was also nice to have access to the ForEach loop, and it was much easier to use the Semantic_check functionality for inference doing everything in the AST.

Gotcha. Thanks!

VMatthijs:
> Relatedly, is it possibly true that any nested Indexed expressions could be expressed as a single flat Indexed? e.g. mat[5][arr, 2:3] -> mat[5, arr, 2:3].

I think that could be hard, but maybe it would owkr. For instance, you'd need to do things like mat[x:y][z] -> mat[x+z-1]
and
mat[arr][z]->mat[arr[z]].
Not sure how it would affect efficiency.

seantalts:
I think `mat[x:y][z]` is equivalent to `mat[x:y, z]`, no?

syclik:
Yes, it is.

seantalts:
Ayee, I think it actually isn't in Stan 2 - mat[x:y, z] selects rows x-z and column z. mat[x:y][z] selects row `x+z-1`! I have been really confusing myself over this the past two weeks, largely because I think stanc3 has the model where comma separated indices are equivalent to multiple index lists, but Stan 2 doesn't. This only exploded in my brain when I tried to index something with 3 dimensions with 4 indices - if they were equivalent, then `matrix[N, K] mat_arr[J];... mat_arr[2:5, 1:3, 1:2, 1, 3]` should be a valid (if silly) way to index a single element but you can have max `dim` elements in an index list.

syclik:
Ay! Sorry for the misinformation. As I hit send, I thought it could be a
little more complicated than that. And I do think what you described is
correct. (I could verify if you wanted.) I was using the scalar index logic
and that’s not what’s happening here.

Issue #241, opened 2019-07-27
"size(x)" parsed as CondDistApp in AST


seantalts:
```stan
transformed data {
  int x[1] = {1};
  print(size(x));
}}
```
goes to

```
(CondDistApp StanLib ((name size) (id_loc <opaque>))
           (((expr (Variable ((name x) (id_loc <opaque>))))
             (emeta ((loc <opaque>) (ad_level DataOnly) (type_ (UArray UInt)))))))
```

Matthijs I'm just going to assign these to you for you to triage as you see fit.

VMatthijs:
Oh, that's really confusing. Will look into it. I wonder if this is happening somewhere down the line from the parser, as the parser requires a `|` symbol to produce a ConDistApp.

VMatthijs:
Ah, it's a problem in the semantic check, apparently.

VMatthijs:
Fixed by https://github.com/stan-dev/stanc3/pull/245 .

PR #245, opened 2019-07-31
Add check to semantic check that original AST is not modified (only decorated)

VMatthijs:
This addresses issue https://github.com/stan-dev/stanc3/issues/241 .


src/frontend/Ast.ml
```diff
@@ -18,6 +18,8 @@ type 'e index =
 (** Front-end function kinds *)
 type fun_kind = StanLib | UserDefined [@@deriving compare, sexp, hash]
 
+let compare_fun_kind _ _ = 0
+
```
VMatthijs: This is a bit ugly. It's just a feature of our decision to add function kinds even to the unchecked AST. Really they should be part of the type that is only added with type checking.

seantalts: Is it possible to define this function at a scope closer to the test so that it's not always defined as this for all users of the AST? (Assuming some users might want to do actual comparisons of the function kinds, possibly).

VMatthijs: Good idea. I will try. Not sure if the ppx stuff works by picking up local definitions or if it only looks for global definitions. I'm afraid it may be the latter.

VMatthijs: Hmm. It's what I was afraid of. There is no way of moving this definition to local scope using ppx, as ppx only works in global scope.

I still think that we ultimately would want to move away from having the FnKinds as a part of the application. They should be part of the function type of the function being applied if anything. That way, they would also only be present in the typed ast and there would be no problem here. Currently, we are just setting them to fake values in the untyped ast and only to the correct value during type checking. That seems bad to me. At least, it should be an optional FnKind then if we are going this route, so we can set it to nothing in the untyped AST. Or we can just remove them in the untyped AST altogether.

What would your preference be, @seantalts ?

seantalts: Hmm, I could see it making sense to add the function kind only in the MIR. Do you remember why we didn't do that before? What would that look like?

VMatthijs: I guess it's added to the AST as well because the check that needs to happen to see the FnKind of an application is rather complicated and is performed anyway during type checking: it requires a lookup of the user defined functions that are in scope as they can sometimes shadow Stan Math functions.

seantalts: Gotcha. So it sounds like you'd like to keep it the way it is where we add the type during semantic checking? Would you then propose making it optional?

VMatthijs: OK. I've now gone with the most type safe route of using another two level pattern for the FunKinds. That allows me to remove the hacky compare function on FunKinds altogether. The route of optional FunKinds just seems not very type safe to me.

VMatthijs:
This is ready to be merged, from my point of view.

src/frontend/Ast.ml
seantalts: I think what you have now is not two-level actually, it's equivalent to Optional, right? Am I missing something? I think it might be cleaner as `option` because then you don't have to thread the `'f` type through everywhere...

seantalts: Oh wait - I guess the difference is that you can add the presence of the fun_kind to the type signature in an easy way with this parameter such that you never see code matching on Some or None, right? I think that is better, though I think we might be getting to the point where we need to start using full names for all of our tons of type parameters 😂

src/frontend/Semantic_check.ml
```diff
@@ -669,9 +669,10 @@ and semantic_check_funapp ~is_cond_dist id es cf emeta =
     ok
       { e with
         expr=
-          ( match e.expr with
-          | FunApp (fun_kind, id, ues) -> CondDistApp (fun_kind, id, ues)
-          | _ -> raise_s [%sexp ("This should never happen!" : string)] ) })
+          ( match (e.expr, is_cond_dist) with
+          | FunApp (fun_kind, id, ues), true -> CondDistApp (fun_kind, id, ues)
```
seantalts: Why does this step happen? Shouldn't we just keep the AST the same structurally? Isn't your check supposed to check for that?

src/frontend/Ast.ml
VMatthijs: This is different as it's forcing through the type system that the untyped AST basically always have None for the fun_kind and that the typed one always has Some _ for it.
The number of type parameters is definitely getting a bit large. Not sure I know great names for them though.

src/frontend/Semantic_check.ml
VMatthijs: Because the FunApp and ConDistApp use lots of the same checking functions. Those are currently set up to return a FunApp. The alternative is to thread the is_cond_dist argument through all of them and have them return ConDistApps when appropriate to start with, rather than having to correct for that at the end. Would that be preferable?

VMatthijs: I've gone with that alternative now. It also seems a bit cleaner to me.

seantalts: Ahh, okay. If this is where the error was happening before, it seems like a good refactoring to make it cleaner now. Will take another look, thanks!

PR #246, opened 2019-07-31
Checking invariant for pretty printing

VMatthijs:
This adds an invariant check to the pretty printing which will guarantee that any output produced by the pretty printer is correct.


src/frontend/Pretty_printing.ml
```diff
@@ -404,4 +404,15 @@ and pp_program ppf = function
       pp_opt_block ppf "generated quantities" bgq ;
       Format.pp_close_box ppf ()
 
-and pretty_print_program p = wrap_fmt pp_program p
+and pretty_print_program p =
+  let result = wrap_fmt pp_program p in
+  let check_correctness () =
+    let result_ast =
+      Errors.without_warnings
```
VMatthijs: This would be nicer if we rewrite the lexer to accumulate warnings purely functionally, rather than writing them to stderr immediately. That would make it a lot easier to handle them later if we want to discard them.

PR #248, opened 2019-08-01
Fix for index parsing

### LINKED ISSUE #240: m[5][2, 3] parsed as single Indexed expression in AST
Here's an example:
```
transformed data {
  matrix[3, 4] mat[5];
  print(mat[3][1, 2]);
}
```

AST:
```
(transformeddatablock
  ((((stmt
      (VarDecl
       (decl_type
        (Sized
         (SArray
          (SMatrix ((expr (IntNumeral 3)) (emeta ((loc <opaque>))))
           ((expr (IntNumeral 4)) (emeta ((loc <opaque>)))))
          ((expr (IntNumeral 5)) (emeta ((loc <opaque>)))))))
       (transformation Identity) (identifier ((name mat) (id_loc <opaque>)))
       (initial_value ()) (is_global true)))
     (smeta ((loc <opaque>))))
    ((stmt
      (Print
       ((PExpr
         ((expr
           (Indexed
            ((expr (Variable ((name mat) (id_loc <opaque>))))
             (emeta ((loc <opaque>))))
            ((Single ((expr (IntNumeral 3)) (emeta ((loc <opaque>)))))
             (Single ((expr (IntNumeral 1)) (emeta ((loc <opaque>)))))
             (Single ((expr (IntNumeral 2)) (emeta ((loc <opaque>))))))))
          (emeta ((loc <opaque>))))))))
     (smeta ((loc <opaque>)))))))

```

Should actually be parsed as a nested Indexed expression, `Indexed(Indexed(mat, [3]), [2, 3])`.

Matthijs how hard is this one? should we mark it good first issue?

seantalts:
Fix #240 in a hacky way that Matthijs can hopefully fix up.


VMatthijs:
OK, will look at this!

VMatthijs:
I'm a bit unsure about how to update the code generation for this. I believe there is a bug in stanc2 relating to this so I'm not sure this'll be easy to fix. It currently generates the same code for the following two:
```
model {
    real a[2, 3];
    a[:, 2] = {17};
}
```
and
```
model {
    real a[2, 3];
    a[:][2] = {17};
}
```
while I think they should have quite different behaviour. The first should pick out the second "column" and set it to {17} while the second should pick out the second "row" and set it to {17}. The resulting arrays `a` would end up being observably different.

Please correct me if I'm wrong!

VMatthijs:
I've filed a bug report here: https://github.com/stan-dev/stan/issues/2795 .

How shall we go ahead, @seantalts ? Are we happy just reproducing the (buggy) behaviour of stanc2 for now or do we immediately want to fix it? The latter would require some thinking about C++.

bob-carpenter:
@VMatthijs is right about behavior, as spelled out in https://github.com/stan-dev/stan/issues/2795

seantalts:
Hm, I would propose we disallow nested indexing on the lhs of assignment. What do you all think? This would mean this is illegal: `v[a][b] = 2;`. Do you all know if there are use cases for nested indexing in Stan in general? Even if so I can't think of any legitimate uses for lhs of assignment...

src/frontend/Ast_to_Mir.ml
```diff
@@ -472,14 +472,26 @@ let rec trans_stmt (declc : decl_context) (ts : Ast.typed_statement) =
...
+      let assign_identifier =
+        match lhs_base.Ast.lval with
+        | LVariable s -> s
+        | _ -> failwith "This should never happen."
+      in
+      let id_ad_level = lhs_base.Ast.lmeta.Ast.ad_level in
+      let id_type_ = lhs_base.Ast.lmeta.Ast.type_ in
+      let lhs_type_ = assign_lhs.Ast.lmeta.type_ in
+      let lhs_ad_level = assign_lhs.Ast.lmeta.ad_level in
+      let rec get_lhs_indices = function
```
seantalts: Your changes look awesome! My one comment would be that I think it'd just end up being cleaner to have `get_lhs_base` return the base identifier and the base identifier metadata separately rather than deconstructing it later. And you could probably have it return the indices at that time as well.

PR #249, opened 2019-08-02
Simplify indexing v0.1

seantalts:
This adds some desugaring steps that seem good and I think fixes some previous bugs. Adds a lot more indexing tests too. Depends on #248 and #250.

This is the first chunk that I got working. There are some new things in here:
1) AST->AST desugaring step, with some of @bob-carpenter's indexing algebra simplifications implemented
2) Added unsized types to MIR Assignments so we can eventually get rid of calls to `stan::model::assign`
6) I also now keep track of which declaration identifiers are global or not and shortcircuit trans_decl if they are not global so that I don't attempt to generate e.g. parameter reads for them.
7) Added a new Frontend_utils file that just has an easy parsing function in it good for testing - not sure where this should live or if it exists.
8) New stanc flags for dumping the output of the various phases. 
9) Adds FnResizeToMatch and accompanying C++ boilerplate functions; this was previously necessary when I was generating Unsized variable declarations but now I'm not doing that anymore. I'm not sure if we should keep them - I expect them to be helpful when @VMatthijs eventually gets optimization hooked up and it starts generating Unsized declarations in the MIr. But for now they're kind of vestigial - let me know if you want me to remove them from this PR. 

TODO:
1) Need to code gen checks that `assign` and `rvalue` were performing
2) possibly further indexing algebraic desugaring simplifications


seantalts:
Maybe desugar is now the wrong name for this step as well, now that it's just doing algebraic index simplifications. It could also now be moved to the optimization library in the MIR - tradeoff there is I guess just that you can't pretty-print the results back as a Stan program if we e.g. wanted this to be part of canonicalization. Maybe we should even be writing these as part of some AST smart constructors (cc @enetsee)?

seantalts:
Could also keep it as an AST->AST step but call it something like canonicalization / simplification - this would then be useful for the pretty printer as we can have an easy way to support pretty printing both with and without opinionated formatting / canonicalization. An example would be how ocamlformat and a bunch of other opinionated formatters actually remove unnecessary parentheses, etc. 

@VMatthijs Do you think you might have time to take a look at the code or just at the questions in the PR and comments?

seantalts:
ping @VMatthijs - have some high level questions here that would be helpful to get your thoughts on even if you don't have time for a full code review :)

VMatthijs:
Sorry it's taking me a while to get to this! Is it OK if I do a review tomorrow?

seantalts:
Sure, or feel free to just talk at a high level about how you think it should be organized and such. Up to you :) Thanks!

src/frontend/Pretty_printing.ml
```diff
@@ -379,33 +379,38 @@ and pp_block block_name ppf block_stmts =
...
+   *       result
+   *   in
+   *   if
+   *     compare_untyped_program
+   *       (untyped_program_of_typed_program p)
+   *       (Option.value_exn (Result.ok result_ast))
+   *     <> 0
+   *   then failwith "Pretty printing failed. Please file a bug."
+   * in
+   * check_correctness () ; *)
```
VMatthijs: Why is this commented out?

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -535,12 +535,43 @@ using stan::model::index_omni;
...
 using namespace stan::math; |}
 
+let pre_boilerplate =
+  {|#include <vector>
+#include <Eigen/Dense>
+
+template <typename T, typename S>
+std::vector<T> resize_to_match(std::vector<T>& dst, const std::vector<S>& src) {
+  dst.resize(src.size());
+  return dst;
```
VMatthijs: does the old compiler do something like this too?

src/middle/Mir.ml
```diff
@@ -112,7 +109,8 @@ type 's fun_def =
   ; fdloc: location_span sexp_opaque [@compare.ignore] }
 [@@deriving sexp, hash, map]
 
-type 'e lvalue = string * 'e index list [@@deriving sexp, hash, map, fold]
+type 'e lvalue = string * unsizedtype * 'e index list
+[@@deriving sexp, hash, map, fold]
```
VMatthijs: Right. I can see that it would be useful to have the type information here. Another option there is to go the full recursive route that I ended up using in the AST, where lvalues just have the same metadata as expressions, at each recursive nesting. That has the advantage that it is very easy to convert back and forth between expressions and lvalues.

On the other hand, based on the discussion here: https://github.com/stan-dev/stan/issues/2795 , it seems like that sort of recursive lvalue is something we might want to disallow anyway in the future?

Maybe we can discuss this on the phone?

src/frontend/Desugar.ml
```diff
@@ -0,0 +1,97 @@
...
+  | Indexed (obj, indices) ->
+      let rec remove_trailing_alls indices =
+        match List.rev indices with
+        | All :: tl -> remove_trailing_alls (List.rev tl)
+        | _ -> indices
+      in
+      Indexed (obj, remove_trailing_alls indices)
+  | e -> e
+
+(* TODO: same thing for Assignments *)
```
VMatthijs: It's very easy to also do this for assignments: there are functions for converting an lvalue back and forth to an expression.

src/frontend/Desugar.ml
```diff
@@ -0,0 +1,97 @@
...
+                     ( obj
+                     , inner_singles
+                       @ [ Single
+                             Frontend_utils.(
+                               binop (binop bot Plus single_e) Minus
+                                 (loop_bottom bot.emeta)) ]
+                       @ inner_tl )
+               ; emeta }
+             , outer_tl ))
+    | inner_singles, multis ->
```
VMatthijs: I think this would only match on multis = [], right? Don't you want a base case there?

src/frontend/Pretty_printing.ml
seantalts: Oh, it wasn't working before because I started using the pretty printer on both typed and untyped programs and I think the correctness check was failing because the conversion to untyped was changing the tree. I think it will probably work now that your PR is in, can try it out.

src/stan_math_backend/Stan_math_code_gen.ml
seantalts: No, these are new helpers for emitting unsized declarations. I used to need that in this PR but now we don't, but I figured it was helpful to have some code that works for that use case in there even if nothing is generating unsized declarations yet?

src/frontend/Desugar.ml
seantalts: You are right that it will only match on `[]` now (I was putting in the multi stuff one by one before, oops). When you say base case do you just mean replacing `multis` with `[]` to be clear?

seantalts:
Yeah I think it'd be useful to have a meeting about this. I can post it on discourse and we can find a time there? Or maybe we can discuss where to put this kind of simplification in the regular stanc3 meeting and then have a separate topic meeting about what to do about this stan 2 lhs indexing bug?

seantalts:
I think that might be the only benefit of putting it on the AST level, yeah. I'm not even sure it's a real benefit, haha. I'm happy to move it wherever.

src/frontend/Desugar.ml
VMatthijs: Should it throw though? I would expect the behaviour on [] to be either doing nothing or just getting rid of the redundant indexing. That is
`a[] -> a`.

seantalts: the guard on the outer match is ` when List.exists ~f:is_multi_index inner_indices`, so it only triggers if the 2nd list isn't empty. It's just a vestigial throw because we can't let OCaml know that we know that list isn't empty.

seantalts:
Just refactored it to use the MIR and be part of partial evaluation (which had the side effect of not hooking it up anymore to live code, but that's okay I guess). Do you wanna take a look, talk more about design, or should I just merge? Any of them are good with me.

seantalts:
I'm gonna go ahead and merge this, feel free to keep commenting and we can fix it with future PRs.

PR #251, opened 2019-08-07
Patterns (WIP)

enetsee:
As discussed, this is the branch with the MIR refactor. The structure is very similar to that of the demo I put together on the smaller language, in particular:

- I've factored out library code to 'common'
- The original 'pattern's are now in 'middle/Mir_pattern'
- 'Expr' and 'Stmt' have three submodules: 'Fixed', 'Typed'/'Location' and 'Labelled'
- 'Fixed' types expose a common interface defined by 'Common.Fix.S1' / 'Common.Fix.S2'  

Expressions and Statements are made by doing the following:

1) Define a pattern functor ('common/Pattern.S')
2) Create it's fixed point using 'Common.Fix.Make'
3) Specialize the fixed point to a particular type of meta data by defining a 'Common.Meta' and then 'Common.Specialize.Make' it

The recursion schemes are defined in 'Common.Recursive' and 'Common.Corecursive'; i have exposed specialised versions of each of them so that they are a bit easier to use for our most common use-case of term-rewriting.

Expressions and statements are also traversable; I created two helper functions for each `map_accum_left` and `map_accum_right` which specialize the `Traversable` to the `State` and `State.Right` applicative respectively. They have a very similar type signature to `fold` but allow you to simultaneously `map` the type too!

I haven't started on converting the AST or  the analysis and optimization library.

@seantalts @VMatthijs @rybern  it would be good to get your thoughts on this work so far.


seantalts:
Hey! Starting to take a look now. I think you've done a lot of impressive work in terms of factoring out commonly used patterns and figuring out the proper abstraction for them - kudos! I'm a bit nervous at the general size of the difference here - it looks like we're adding a 3000 line common module (which is pretty cool but it's so comprehensive that it feels like it could deserve its own open source project?), but ignoring those lines our net line count on frontend, middle, and the C++ backend are still 4000 additional lines with no change in behavior. I'm generally against putting too heavy an emphasis on the number of lines, but here it's kind of striking given that it's in the same language and I'm not counting the generalized abstractions that have already been factored out into `common`.

Is there anything we can do to slim the boilerplate footprint down? I know that eventually we can ppx derive traverse, but that doesn't seem like a huge factor. What about getting rid of a lot of the helper functions, or the modules for the distributions? I think those modules look like a great way to organize things but I'm just not sure if that organization serves to make even the rest of the project easier to read...

enetsee:
A fair chunk of the line count is coming from the .mli files I've added and functionality I've moved from 'analysis_and_optimization' but I agree that we can probably slim down the expression helpers too.

I can remove the distribution modules and handle the rewrites using the factored out helpers which should have quite a big impact!

enetsee:
Removing the distributions reduces it by ~1300 lines - are there any other candidates in the helpers that you think should go too?

enetsee:
I'm also accruing quite a few LoC by defining the pattern data-types twice, once in `Mir_pattern` and then again in each of the modules (this is done so you can refer to e.g. `Expr.Fixed.Pattern.t` rather than `Mir_pattern.expr`). I'd have to expose `Mir_pattern` if I removed these and it might be little jarring for the user but it would reduce line count quite a bit too I think.

enetsee:
Regarding `Common`, it's all stuff we were doing in an ad-hoc way but it definitely turned out to be more code than I anticipated.

My breakdown is as follows:

The 'type class' like modules and Monad/Applicative instances could definitely live elsewhere:
(Monoid, Functor, Bifunctor, Foldable, Bifoldable, Traversable, Bitraversable, State, Validation)

The recursion pattern and Pattern/Fix/Specialized stuff is pretty specific to our project - I've certainly never seen this type of representation (i.e. a parametrically polymorphic fixed point and a nested fixed point) anywhere else.

The monotone framework signatures are pretty generic and _may_ be useful as a standalone library.

enetsee:
@seantalts , the line count thing has been bothering me a little bit so I broke down the changes in the PR (below).

Basically, there is negligible change to `Frontend`, `Stan_math_backend` and `Stanc`. This is sensible since I haven't actually made any material changes which would use the functionality exposed by middle or common in any of these libraries - the changes are just to use the refactors types. I would anticipate that the loc in each of these will reduce when some of the code is switched to use the generic recursions.

`Common` accounts for 2773 lines, 1768 of which is implementation and 1005 is interface.

The stuff in `Common` that is definitely not project specific accounts for 719 lines. The stuff that could _possibly_ live in an external library is another 1912 lines (1185 for ml, 727 for mli). This leaves 142 lines which would probably shrink given we don't have to include everything here.


`Middle` accounts for 2981 lines, 2269 of which are implementation and 712 of interface. There is quite a bit of new functionality in these files 

There is repetition between `Mir_patterns` and `Expr` / `Stmt`. When combined with traversable related code this would be about 538 lines.

Only 165 lines (of 1268) of `Expr.ml` is used to define the pattern, fixed point, no-meta, typed and labelled implementations along with _all_ of the folds / traversables / recursion schemes. 

The remaining ~1000 lines are helpers and a partial evaluator. As well as smart constructors and simplifying rewriters moved from `analysis_and_optimization`, the helpers contain functions which use recursion schemes to e.g. find if a node contains a function application.  Similarly, only 172 lines (of 420) of `Stmt.ml` is used to define pattern, fixed point, no-meta, located and labelled implementations along with _all_ of the folds / traversables / recursion schemes.

I'm inclined to agree that the ratio of type definition to helper seems a little out of kilter.

I think the headlines and design points are:

1) A good chunk of `Common` should probably end up in another library/project altogether
2) A bunch of `Middle` will go when I have a ppx for traversable
3) A bunch of `Middle` could go if we doubled down on a module structure and deleted the repeated type defs in `Mir_pattern`
4) Removing everything other than type defs would leave `Expr` and `Stmt` at around 160-170 loc.

1-2  seem like longer term issues. 

For 3, I know you were keen to have all the patterns in one place for ease of browsing. I'm comfortable with this but it does introduce a bit of repitition.

For 4, I'm inclined to say we should move the bulk of the helpers back out of `Expr` and `Stmt` to 
keep them minimal.

Would you be happy with this?


**File**|**Sum of Impl Change**|**Sum of Intf Change**|**Sum of Total Change**
:-----:|:-----:|:-----:|:-----:
**_Common_**|**_1,768_**|**_1,005_**|**_2,773_**
Bifoldable|74|42|116
Bifunctor|5|5|10
Bitraversable|22|36|58
Common|36|36|72
Corecursive|179|110|289
Fix|185|67|252
Flowgraph|50|27|77
Foldable|46|24|70
Functor|5|5|10
Gensym|11|3|14
Helpers|7|5|12
Injectable|152|119|271
Label|16|16|32
Lattice|9|9|18
Meta|6|6|12
Monoid|38|17|55
Monotone\_framework|118|40|158
Pattern|27|27|54
Pretty |22|22|44
Projectable|202|157|359
Recursive|157|108|265
Specialized|58|20|78
State|170|27|197
Transfer\_function|26|21|47
Traversable|17|17|34
Validation|130|39|169
_Frontend_|_-13_|_17_|_4_
Ast|0|0|0
Ast\_to\_Mir|-13|-2|-15
Debug\_data\_generation|-2|0|-2
Errors|0|0|0
lexer|1|0|1
Parse|0|0|0
parser|0|0|0
Preprocessor|1|0|1
Pretty\_printing|0|0|0
Semantic\_check|0|0|0
Semantic\_error|0|19|19
**_Middle_**|**_2,269_**|**_712_**|**_2,981_**
Compiler|-1|0|-1
Expr|1,268|268|1,536
Fun\_kind|2|2|4
Int\_label|19|1|20
Internal\_fun|26|17|43
Location|96|10|106
Location\_span|49|10|59
Middle|-115|9|-106
Mir|-219|0|-219
Mir\_pattern|538|0|538
Mir\_pretty\_printer|0|0|0
Mir\_utils|-98|0|-98
Operator|36|29|65
Program|133|77|210
SizedType|44|33|77
Stan\_math|-2|0|-2
Stmt|420|188|608
Type|20|23|43
Type\_conversion|-38|0|-38
UnsizedType|93|47|140
Utils|-2|-2|-4
**_Stan\_math\_backend_**|**_37_**|**_9_**|**_46_**
Cpp\_Json|-1|0|-1
Expression\_gen|5|0|5
Locations|19|11|30
Stan\_math\_code\_gen|0|0|0
Statement\_gen|0|0|0
Transform\_Mir|14|-2|12
**_Stanc_**|**_0_**|**_0_**|**_0_**
stanc|0|0|0
Grand Total|4,061|1,743|5,804

seantalts:
> I'd have to expose Mir_pattern if I removed these and it might be little jarring for the user but it would reduce line count quite a bit too I think.

What's the difference in the user's code for each? Just trying to understand with examples how jarring it'd be. I do really like the idea of having a really simple AST and MIR definitions as those are our main hooks for anyone who wants to extend stanc3. Of course, we're already pretty far beyond that with our two-level types and 18000 type parameters. I honestly think it might just be worth explicitly writing out each type of tree (untyped AST, typed AST, typed MIR) unparameterized/fully concrete in separate modules - it might literally take less lines of code as well even if we have to write map and fold for each...

 > For 4, I'm inclined to say we should move the bulk of the helpers back out of Expr and Stmt to
keep them minimal.

When you say "back out of" do you mean back into optimization and analysis? If you're saying that these lines are essentially just moved and they aren't showing up as deleted because you haven't yet addressed the optimization and analysis package I think that's a solid argument for keeping them or at least analyzing the line count with that in mind. 

Separately, do we need all of the interface files? It seems like some of them are not necessarily narrowing what's presented in the `.ml` files - what are the reasons to keep an `.mli` file around in that case? I'm sure there are some but my understanding of OCaml mli files is pretty limited.

enetsee:
>> What's the difference in the user's code for each? Just trying to understand with examples how jarring it'd be. 

The best example I have is when you are writing a function directly on the underlying pattern; if we just alias the patterns, you have to prefix it with the `Mir_pattern` module rather than `Expr.Fixed.Pattern`. So you would have this:

```
let contains_fun_algebra ?kind ?name = function
  | _, Mir_pattern.FunApp (fun_kind, fun_name, args) ->
  ...
```

```
let contains_fun_algebra ?kind ?name = function
  | _, Expr.Fixed.Pattern.FunApp (fun_kind, fun_name, args) ->
  ...
```

Most of the time you should be using helper functions like `Expr.Fixed.pattern` of `Expr.Fixed.proj` which means you don't need to provide an explicit module prefix. I guess this would be fine as long as we explained it.

> I do really like the idea of having a really simple AST and MIR definitions as those are our main hooks for anyone who wants to extend stanc3. 

I can see both sides of this. For me, the structure actually became clearer after organizing them into modules and colocating only the types that needed to be (e.g. autodifftype, returntype and unsized type have to be together). That said, I'm not looking at the code with fresh eyes so my opinion is clearly not useful in judging what's best for a new contributor!

>I honestly think it might just be worth explicitly writing out each type of tree (untyped AST, typed AST, typed MIR) unparameterized/fully concrete in separate modules - it might literally take less lines of code as well even if we have to write map and fold for each...

I definitely agree that the fixed-point approach _and_ our use of parametric polymorphism for the meta-data introduces a big mental hurdle. 

The first thing I would say in its defence is that we have there are actually more cases that you list above (uniques labelled statements & expression for optimization, non-uniquely labelled 
statements for CPP code generation). 

Perhaps more importantly, using this approach to derive the various maps, folds, recursions and traversals actually makes adding a new expression or statement really easy and significantly less error prone. All you have to do right now is handle the new case in pretty printing and traversal (and the traversable will also go away at some point all being well).  

Finally, the refactored structure means that you can largely ignore exactly how any of this works since we have nice names for each specialized version of expressions, statements and programs. I know that some contributors will want to dig in

> When you say "back out of" do you mean back into optimization and analysis?

Pretty much; a lot of the helpers break out the monster case statment that currently sits in `Partial_evaluator` used to rewrite Stan Math function applications. 

There are also a bunch of other helpers (typically predicates on patterns or types) scattered throughout `Frontend` (Ast_to_Mir in particular) and `Stan_math_backend`.

enetsee:
>Now that I've written it down, I think I actually prefer it and I guess this would be fine as long as we explained it.

Hmm.. just tried this an you run into problems since the compiler no longer forces the alias and the original type to be the same.

One of the key goals of the refactor is provide standard interfaces for types which would not be possible if we did this.

enetsee:
@VMatthijs I've [implemented flowgraph construction](https://github.com/stan-dev/stanc3/blob/patterns/src/middle/Stmt_flowgraph.ml) based on the emails we exchanged a while back.

seantalts:
> The first thing I would say in its defence is that we have there are actually more cases that you list above (uniques labelled statements & expression for optimization, non-uniquely labelled
statements for CPP code generation).

That's fair. I was just reading [this SO link](https://stackoverflow.com/questions/49046135/adding-line-information-to-my-ast-in-ocaml) after seeing you say "I definitely agree that the fixed-point approach *and* our use of parametric polymorphism for the meta-data introduces a big mental hurdle" and I am now wondering why it isn't just parametric on meta-data? Is it just because then we'd have to write our own fold and map methods? If I had to cognitively weigh these two concepts I'd say the two-level types burden weighs something like 4 or 5 boilerplate tree recursions, haha. The accepted answer at that link explores a few alternatives and notably doesn't even mention two-level types at all as a solution, haha. 

I'm really sorry to bring this all up now (again! I know we had a few long conversations about this decision already). I'm just feeling a bit overwhelmed at the amount of code required to do this two-level pattern correctly, and how difficult it's going to be to try to onboard new folks. 

> One of the key goals of the refactor is provide standard interfaces for types which would not be possible if we did this.

Does that refer to the sort of splitting into modules of the current MIR types in MIddle along with their helper functions into separate modules so that you type `Middle.Expr.Fixed.fold` (?) instead of `Middle.fold_expr`?

Separately (related?), I'm not sure I see the benefit to splitting out types like `fun_kind` into their own modules. I've been working a lot in the backend before and after it was split into Expression_gen, Statement_gen, and Stan_math_code_gen and I've been finding it harder to navigate afterwards vs a single file. Every time I go to look something up, I have to think about what module it might be in. Maybe part of that is because the rules for what goes where aren't clear - like functionality for printing types live in expression gen because we need them there. Still, it seems like a waste of screen real estate to have to open fun_kind.ml or .mli in its own editor window :P

seantalts:
@rybern @VMatthijs definitely hope to hear from you two as well on this - I'll bring it up at the meeting tomorrow too.

Here is a simple little example of what I thought the "classic" technique might look like for us:

```ocaml
type ('a, 'm) with_meta = {obj: 'a; meta: 'm}

type 'm expr =
  | Bool of bool
  | And of 'm expr_with * 'm expr_with

and 'm expr_with = ('m expr, 'm) with_meta

type ('em, 'sm) statement =
  | Print of 'em expr_with list
  | IfThen of 'em expr_with * ('em, 'sm) stmt_with

and ('em, 'sm) stmt_with = (('em, 'sm) statement, 'sm) with_meta
```

enetsee:
> I'm really sorry to bring this all up now (again! I know we had a few long conversations about this decision already). I'm just feeling a bit overwhelmed at the amount of code required to do this two-level pattern correctly, and how difficult it's going to be to try to onboard new folks.

No worries; I get what you're saying and this clearly is a fiarly significant decision point so we should take our time and look at it from all sides. I've found the process of understanding this stuff it's own reward and I'm not going to be precious if we choose another approach!

> Does that refer to the sort of splitting into modules of the current MIR types in MIddle along with their helper functions into separate modules so that you type Middle.Expr.Fixed.fold (?) instead of Middle.fold_expr?

That is definitely part of it; this has been my instinctive approach to writing functional code for a while (hide your types implementation and colocate functions that operate on it) and I guess I've doubled down on it since seeing how others are doing things in ocaml e.g. https://blog.janestreet.com/core-principles-uniformity-of-interface/

For the ancilliary type in  MIR this 'standard interface' includes some the following:

- `Pretty` for pretty printing
- `Traversable` 
- `to_string` / `of_string_opt`
- `Comparator` and `Comparable` (from `Core_kernel`)

For our main `Fixed` types (`Expr.Fixed.t` and `Stmt.Fixed.t`) we have:

- `Foldable`
- `Pretty`
- `Recursive` and `Corecursive`
- `Traversable`

This means you only need have 9 functions to do _anything_ we are currently doing with one of these types:

- `transform_bottom_up`, `transform_top_down`, `transform_with_context` and `transform_partial` from `Recursive` / `Corecursive`
- `map_accum_left` and `map_accum_right` from `Traversable`
- `fold_left`, `fold_right`, `fold_map` from `Foldable`

For our main `Specialized` types (`Expr.Typed.t`, `Expr.Labelled.t`, `Stmt.Located.t` and `Stmt.Labelled.t`) we have:

- `Pretty` for pretty printing
- `Comparator` and `Comparable` (from `Core_kernel`)

Bear in mind that this would be exactly the same for the current AST if we were to continue down this route (and would be the same for the 'regression calculus' frontend I've been talking about).

In terms of `Fun_kind` the reason it is separated is because it is used in both `Expr` and `Stmt`. 

For other similar types (like `lit_type` and `'a index`) I have kept them with the 'parent' type if they are only used in that module (which is `Expr` for `index`).

enetsee:
```ocaml
type ('a, 'm) with_meta = {obj: 'a; meta: 'm}

type 'm expr =
  | Bool of bool
  | And of 'm expr_with * 'm expr_with

and 'm expr_with = ('m expr, 'm) with_meta
```

This is exactly the 'Separate IR' solution [here](http://blog.ezyang.com/2013/05/the-ast-typing-problem/) written with a custom product. Writing it with a tuple makes it even clearer: 

```ocaml
type expr = 
           | Bool of bool
           | And of expr * expr

type 'a expr_with =  expr * 'a
```

seantalts:
Do you think that seems simpler for new folks? I can't really tell. 

Wouldn't you have to pass the type parameter into expr as well, like this?
```ocaml
type 'a expr = 
           | Bool of bool
           | And of 'a expr * 'a expr

and 'a expr_with =  expr * 'a
```

seantalts:
PS thanks for the breakdown - I think I needed help really explicitly laying out what we're gaining from the refactor. I have to run for now but will carefully read through the code again with this in mind.

enetsee:
> Do you think that seems simpler for new folks? I can't really tell.

(First - I should say that my suggestion above isn't equivalent to what you have so I'm ignoring that in this response!)

I've had a think about this and for me there are four consideration when talking about the ease of use/understanding of the encoding of the intermediate representations:

1. Reading the definition;
2. Writing functions operating on the representation;
3. Adding/changing definitions;
4. Material describing the approach.

For consideration 1, I don't think this changes how hard it is to understand. I _think_ this is just a slight refactoring of the `Fixed` definition with the recursion pushed back inside the auxiliary type `expr_with` rather than inside your `with_meta`:

```ocaml
type ('a, 'm) with_meta = {pattern: 'a; meta: 'm}

let pattern {pattern;_} = pattern 
let meta {meta;_} = meta 

type 'e index =
  | All
  | Single of 'e
  | Upfrom of 'e
  | Between of 'e * 'e
  | MultiIndex of 'e
  
let index_bounds = function 
  | All -> [] 
  | Single e | Upfrom e | MultiIndex e -> [e]
  | Between (e1,e2) -> [e1;e2]

type litType = Int | Real | Str 
type fun_kind = StanLib | CompilerInternal | UserDefined

type 'm expr =
  | Var of string
  | Lit of litType * string
  | FunApp of fun_kind * string * 'm expr_with list
  | TernaryIf of 'm expr_with * 'm expr_with * 'm expr_with
  | EAnd of 'm expr_with * 'm expr_with
  | EOr of 'm expr_with * 'm expr_with
  | Indexed of 'm expr_with * 'm expr_with index list

and 'm expr_with = ('m expr, 'm) with_meta
```
For consideration 2, I wrote out a function which we originally had trouble defining this time using your suggested encoding :

```ocaml
let contains_fn ?kind ?name (expr : 'm expr_with) = 
  let rec aux accu expr = 
    match pattern expr with 
    | Var _ | Lit _ -> false || accu 
    | TernaryIf(pred,te,fe) -> aux  (aux (aux accu pred) te) fe
    | EOr(e1,e2) | EAnd(e1,e2) -> aux (aux accu e1) e2     
    | Indexed(e,idxs) ->  List.fold_left (fun accu idx -> List.fold_left aux accu @@ index_bounds idx)  (aux accu e) idxs
    | FunApp(fun_kind,fun_name,args) -> 
      let same_kind = match kind with Some k -> k = fun_kind | _ -> false 
      and same_name = match name with Some n -> n = fun_name | _ -> false 
      in
      (same_kind && same_name) || List.fold_left aux accu args
  in
  aux false expr
```

I think there are two points here.

First, this almost identical to how you would explicitly write this function with the `Fixed` type so there is no real gain in clarity here.   

Second, for newcomers, I think the most common issue will be correctly writing required recursion explicitly and correctly. Asking an inexperienced functional programmer to explicitly write a bottom-up or top-down transformation on the MIR would likely be a common source of error.

One major benefits of the `Fixed` encoding and the standard recursion patterns is that we get to separate the function on the pattern from the recursion. With the `Fixed` approach newcomers wouldn't (ever?) need to write a recursion explicitly.

Here is the same function on the `patterns` branch:

```ocaml
let any xs = List.exists (fun x -> x) xs

let contains_fun_helper ?kind ?name = function
  | _, Var _ | _, Lit _ -> false  
  | _, TernaryIf (e1, e2, e3) -> e1 || e2 || e3 
  | _, EAnd (e1, e2) | _, EOr (e1, e2) -> e1 || e2  
  | _, Indexed (e, idxs) ->
      e || any (compose any index_bounds) idxs
  | _ , FunApp (fun_kind, fun_name, args) ->
      let same_kind = match kind with Some k -> k = fun_kind | _ -> false 
      and same_name = match name with Some n -> n = fun_name | _ -> false 
      in
      (same_kind && same_name) || any  args

let contains_fun ?kind ?name expr =
  Expr.Fixed.cata (contains_fun_helper ?kind ?name) expr
```

Notice that there is _no_ recursion in the `contains_fun_helper` function; that is all dealt with by `Expr.Fixed.cata` which was *derived*  rather than having to be explicitly written. 

For consideration 3, there is no more work to do when defining a new constructor or changing one but updating functions over the new definition _would_ be harder for reasons outlined for point 2.

Finally, for consideration 4, I kind of like the fact that there is a well-known approach to doing this that we can point people to. 

I mentioned before that I am starting to write up the nuances of how we are using the `Fixed` approach; this coupled with the various blogs I read to get to grips with what's happening should be sufficient to get newcomers over the hump.

seantalts:
> (First - I should say that my suggestion above isn't equivalent to what you have so I'm ignoring that in this response!)

I didn't understand what you meant here - was the version with tuples that you wrote viable after all? Is that version what we're ignoring? 

Okay so I did a bad job here by bringing in this second topic of "should we even be using two-level types." Let's deal with that separately - I'll maybe make a discourse post with some examples for some newcomers to try to grok and see how crazy people find each one, or something. With the parameterized meta version of `contains_fn` above, you aren't using `fold_statement` anywhere, which I'd assume we'd define manually still in that version in order to use it.

But let's just assume we're going to stick with two-level types for now - sorry about that interlude. Okay. So now I want to analyze how easy the Stan compiler is to learn (and maintain), and for this we can kind of treat the Fixed library as a sort of external dependency and assume we won't have to maintain that (though that's probably not super true, haha). So in this world there are the 4 metrics that you proposed, plus I'd add that we need to think about the sheer volume of code as well. 

Here's the definition of `contains_fn` now:
```ocaml
let rec expr_contains_fn fname accum e =
  accum
  ||
  match e.expr with
  | FunApp (_, name, _) when name = fname -> true
  | x -> fold_expr (expr_contains_fn fname) accum x
```

I actually think this is nicer than the cata example, though I assume there is another version of it using the Fixed library that's better? 

Given that all computation can be expressed via fold and that usually `fold` itself is kind of a high barrier to entry for non-functional folks, I'm worried about asking folks to learn 9 new types of recursion and choose the correct one rather than learning about `fold` and reasoning from first principles how to create the function they want using that. I would say let's go with the 9 new types of recursion if that significantly reduced the total volume of code we had, but in general I'd like to keep to a minimum the number of new concepts folks who want to help out on the compiler need to know. I want to keep this project friendly to data scientists and C++ devs who want to help out as well as to compiler experts, so let's analyze this refactoring decision with that in mind I guess.

enetsee:
> I didn't understand what you meant here - was the version with tuples that you wrote viable after all? Is that version what we're ignoring?

Sorry, yes, I meant the version with tuples.

> Okay so I did a bad job here by bringing in this second topic of "should we even be using two-level types." Let's deal with that separately - I'll maybe make a discourse post with some examples for some newcomers to try to grok and see how crazy people find each one, or something.

Cool - it will be interesting to see what is difficult for people without an FP background. 

> I actually think this is nicer than the cata example, though I assume there is another version of it using the Fixed library that's better?

```ocaml
let contains_fun name = 
   exists ~pred:(function (_,FunApp(fun_name,_,_)) -> fun_name = name | _ -> false)
```
😜 

I would also say that the function you defined doesn't check for the presence of the named function inside the indices - this may well  be intentional for your use case but it does go to show that you have to be quite careful when writing these functions.

> I'm worried about asking folks to learn 9 new types of recursion and choose the correct one rather than learning about fold and reasoning from first principles how to create the function they want using that.

Hahaha - there are only 4 recursions and you can only write one of them in terms of `fold`! This was exactly the reason for all the helpers we had in `Analysis_and_optimization.Mir_helpers` which was the motivation for the refactor i.e. to make optimizations clearer.

>  I want to keep this project friendly to data scientists and C++ devs who want to help out as well as to compiler experts, so let's analyze this refactoring decision with that in mind I guess.

Makes total sense; there is balance to be struck between taking advantage of what OCaml offers and accessibility. I'll stop being 'that FP guy' now 😃

seantalts:
Is this digging into the partial evaluator at all? I'm starting to add the indexing stuff there and want to refactor some of the formatting stuff to make it easier to read.

seantalts:
Never mind, it's too much to do with this PR, haha. 

Separately I was wondering - does this branch take longer to compile for you?

seantalts:
> I would also say that the function you defined doesn't check for the presence of the named function inside the indices - this may well be intentional for your use case but it does go to show that you have to be quite careful when writing these functions.

This was concerning to me as I intended it to check them! Though in this specific case it doesn't matter if it does... but it seems like it would?

```ocaml
let%test "expr contains fn indexed obj" =
  {expr= Indexed (internal_funapp FnReadData [] (), []); emeta= ()}
  |> expr_contains_fn (string_of_internal_fn FnReadData) false

let%test "expr contains fn indexed index" =
  {expr= Indexed ({expr=Lit(Int, "1"); emeta=()},
                  [Single (internal_funapp FnReadData [] ())]); emeta= ()}
  |> expr_contains_fn (string_of_internal_fn FnReadData) false
```

These both pass.... next you're going to tell me that I was never really running the tests properly either

mind blown

enetsee:
> These both pass.... next you're going to tell me that I was never really running the tests properly either

I should have written 'inside the function arguments' rather than 'indices'. Try this one:

```ocaml
let%test "expr contains fn indexed index" =
  {expr=funapp "whatever" [internal_funapp FnReadData [] ()] () ; emeta= ()}
  |> expr_contains_fn (string_of_internal_fn FnReadData) false
```

src/frontend/Ast_to_Mir.ml
```diff
@@ -312,32 +300,45 @@ let param_size transform sizedtype =
...
+      shrink_eigen
+        (fun k -> Expr.(binop Typed.Meta.empty Plus k (k_choose_2 k)))
         sizedtype
-  | Covariance -> shrink_eigen (fun k -> binop k Plus (k_choose_2 k)) sizedtype
 
 let read_decl dread decl_id sizedtype smeta decl_var =
   let args =
     [ mkstring smeta decl_id
-    ; mkstring smeta (unsizedtype_to_string decl_var.emeta.mtype) ]
+    ; mkstring smeta (unsizedtype_to_string (Expr.Typed.type_of decl_var)) ]
```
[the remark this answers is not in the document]
enetsee: Yeah, I can see that record syntax might be nicer in some places; one place these helper functions are a win is where you want to write a function over multiple types of meta-data, e.g. in the partial evaluator.

src/frontend/Ast_to_Mir.ml
```diff
@@ -1,70 +1,59 @@
...
-  { expr= FunApp (StanLib, string_of_operator op, trans_exprs args)
-  ; emeta=
-      { mtype= operator_return_type op argtypes |> unwrap_return_exn
-      ; mloc= Ast.expr_loc_lub args
-      ; madlevel= Ast.expr_ad_lub args } }
+  let type_ = Stan_math.op_return_type op argtypes |> unwrap_return_exn
+  and loc = Ast.expr_loc_lub args
+  and adlevel = Ast.expr_ad_lub args in
+  let meta = Expr.Typed.Meta.create ~type_ ~loc ~adlevel () in
+  stanlib_fun meta (Operator.to_string op) @@ trans_exprs args
```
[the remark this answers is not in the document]
enetsee: The long-hand creation of the meta data is probably one of the ugliest things in the PR. I ended up hiding the implementation of the meta data type so I could get the `Specialized.Make` functor to work.

With my approach all specializations of a `Fixed.t` have the same signature:

```ocaml
module type S = sig
  type t [@@deriving compare, sexp]

  module Meta : Meta.S
  include Pretty.S with type t := t
  include Comparator.S with type t := t

  include
    Comparable.S
    with type t := t
     and type comparator_witness := comparator_witness
end
```

The functor handles making concrete functions for pretty printing, `sexp_of_t`, `t_of_sexp`, `compare` as well has creating a comparator so each specialized type has a `Map` and `Set` module. The existing optimization code made heavy use of `Set.Poly` and `Map.Poly` which is generally recommended against.

seantalts:
> Sorry if this came across as a criticism of the code, it wasn't intended as such!

Haha, no definitely not, I was just making a joke about my own inexperience in this area.

> The reason is because of the guard on the pattern

I was careful about that! :P 

Yeah I saw something like 2x as many objects to compile and it seemed like it was taking a lot longer, but maybe that's just the common library? I also didn't time it. I'll time it now.

seantalts:
compiling fresh goes from 14s to 19s for me for `dune clean; time dune build`

I'm just trying to be wary - compile time is a scant resource as well, because of how often every dev will be compiling and recompiling stuff. I haven't been tracking it that carefully but it's getting to the point where I'd like to start thinking about it a little bit.

enetsee:
If you pass '--trace-file whatever.trace' to dune you can load the file into chrome tracing (chrome://tracing).

It doesn't _look_ like any single bit of Common is driving this rather the combined effect of various unrelated parts (i.e. some is coming from monotone framework, some from fix, etc.) I can have a proper look at it tonight.

seantalts:
Whoa, that is so cool. I love dune more every day, haha.

src/middle/Expr.mli
```diff
@@ -0,0 +1,104 @@
...
+
+type litType = Mir_pattern.litType = Int | Real | Str
+[@@deriving sexp, hash, compare]
+
+type 'a index = 'a Mir_pattern.index =
+  | All
+  | Single of 'a
+  | Upfrom of 'a
+  | Between of 'a * 'a
+  | MultiIndex of 'a
```
seantalts: To try to help line counts, I'd propose just not having .mli files where you want to expose the entire types underneath. That's how Mir.ml works now - almost everything in there is (supposed to be) a type so it doesn't need  a separate .mli file.

src/middle/Mir_pattern.ml
```diff
@@ -0,0 +1,538 @@
...
+      ; output_vars = output_vars  
+      }
+    )
+    <*> functions_block
+    <*> input_vars
+    <*> prepare_data
+    <*> log_prob
+    <*> generate_quantities
+    <*> transform_inits
+    <*> output_vars
```
seantalts: This style is so so so much more verbose and hard to understand for me than the simple version of this would be.

enetsee: Hmm.. what did you have in mind - I'm not quite sure what the simple version is here. 

Note that each of these values is lifted into the applicative (i.e. `transform_inits` has type `'s list A.t`) OCaml doesn't have a `map7`. 

We also can't use `bind` since traverse is only defined on applicative rather than monad.

seantalts: I mean like a simple traverse implementation with signature something like `ref 'a -> ('b -> 'c) -> 'b -> 'c`.

seantalts: It also looks like we might be able to use `map` + `iter`: https://github.com/ocaml-ppx/ppx_deriving#plugins-iter-map-and-fold

enetsee: You could definitely do that for a traverse where the effect was state (in the same way as you could do a map with a ref cell rather than a fold) but what about where the applicative is something else e.g. validation? I'd also say when the effect in  traverse is state map_accum_left / right (defined in Program.ml) provide a clean interface.

seantalts:
I think it's just too confusing for me :( I've read at least 10 monad introductions and still have a ton of trouble following this code. I had another comment above, to try to recenter on the goals of the PR and figure out a smaller incremental change:

In the meeting we talked about how you really cared about two things. I forget the details but want to try to paraphrase them here and have you correct me so we can start from the same place. I remember agreeing with you about the two things but let's nail them down:

1) Refactoring the code such that we have a few standard recursion patterns (that may or may not be considered "recursion schemes") that we generalize and use whenever possible. Key example: exists for contains_fn.
2) I forget this one. Was it that you wanted to put the MIR and AST types into one module per type declaration and then have all of the helpers for that type live in that module a la https://blog.janestreet.com/core-principles-uniformity-of-interface/ ?

enetsee:
Yes, they were pretty much the things I was trying to achieve!

enetsee:
I can see that some of this stuff is going to be well out of people's comfort zones. 

On the one hand I can see that having e.g. mutable state would be more obvious to non FP Devs. 

On the other hand those approaches are pretty hard to generalize (e.g. in the suggested alternative to traverse with state, can you define a generic function where the mutation is local and has a polymorphic type?) and compose. 

My sense was this exact issue caused the multiplicity of ad-hoc folds and traversals we had.

That being said, I really don't want this to be a big thing so if that is better for the moment that's absolutely fine by me too!

As you say, let's set out the spirit of the PR and then work out what will work and the alternatives for what is a step to far.

src/common/Traversable.mli
```diff
@@ -0,0 +1,17 @@
+open Core_kernel
+
+module type S = sig
+  type 'a t
+
+  module A : Applicative.S
+
+  val traverse : 'a t -> f:('a -> 'b A.t) -> 'b t A.t
```
VMatthijs: @enetsee , is there a reason that you only implement the left-to-right traversal here? Aren't there many possible traversal orders? I see that Haskell does the same in its standard library though. I'm wondering if there's something I'm not understanding about traversals. For example, a traversal should specialize to a fold if you use a the applicative `A m _  = m` where `m` is a monoid. In that case, you at least have a left and a right fold, correct? Shouldn't you have something similar for traversals? Do you know that this traversal you implemented here is enough for our purposes?

enetsee:
My understanding is that the defined traversal is sufficient, see e.g https://www.cs.ox.ac.uk/jeremy.gibbons/publications/uitbaf.pdf

There is a special case of this technique in ‘State.ml’ where I define left and right associative applicative interfaces to state and use it to define a left and right ‘map accumulate ‘ given a traversable.

src/common/Traversable.mli
enetsee: My understanding is that the defined traversal is sufficient, see e.g https://www.cs.ox.ac.uk/jeremy.gibbons/publications/uitbaf.pdf

There is a special case of this technique in ‘State.ml’ where I define left and right associative applicative interfaces to state and use it to define a left and right ‘map accumulate ‘ given a traversable.

enetsee: I put an example of this here: https://sketch.sh/s/FJQ9AElOGhuW5t7uxROKHS/

enetsee:
Closing; initial refactor is now in #346

Issue #28, opened 2019-01-05
Investigate removing context_flags (and Symbol_table) from global state


seantalts:
Right now in Semantic_check.ml we have two global objects that get modified and passed around a lot: context_flags and the Symbol_table. I think the ways in which we use these typically involve modifying them for the execution of some function and then resetting them back to their original state after the function has executed. It looks like it might be less messy with these two just passed in to the function calls in question and having them return a new pair that we can update with if need be (seems rare). I'm not sure this is the case but we can use this issue to track an investigation into this possibility - maybe some highlighting some places where it's messy vs. clean in each approach and attempting to estimate the net effect of each.

VMatthijs:
Do you mean that we would rewrite the code in a purely functional style, i.e. passing around the Symbol_table and context_flags in a state-passing-style, rather than doing imperative updates on them? That's definitely a possibility. In that case we'd want to replace the hash table in the Symbol_table with a Map.

I originally chose an imperative style here precisely because I felt the data structures get modified relatively infrequently, so it seemed a bit much to always be passing them around, but that could've been the wrong decision. A functional style would certainly be more explicit, but also more wordy.

bob-carpenter:
> On Jan 11, 2019, at 7:34 AM, Matthijs Vákár <notifications@github.com> wrote:
> 
> Do you mean that we would rewrite the code in a purely functional style, i.e. passing around the Symbol_table and context_flags in a state-passing-style, rather than doing imperative updates on them? That's definitely a possibility. In that case we'd want to replace the hash table in the Symbol_table with a Map.
> 

I like this idea.  I find this kind of structure cleaner than a global
where feasible, especially in terms of the automatic cleanup.  But can
we make it performant enough?  We need random access to that symbol table.

> I originally chose an imperative style here precisely because I felt the data structures get modified relatively infrequently, so it seemed a bit much to always be passing them around, but that could've been the wrong decision. A functional style would certainly be more explicit, but also more wordy.

I wouldn't worry so much about wordiness.  All the arguments like this can
be encapsulated into a single bigger contextual object to cut down on
the number of arguments passed around.

seantalts:
[...]

PR #260, opened 2019-08-16
[WIP] Purify symbol table

palmerlao:
For now, this draft only contains modifications to `Symbol_table.{ml,mli}` to try to get some early feedback around if we want to continue down this path for #28 .

### DIFF
--- a/src/frontend/Symbol_table.ml
+++ b/src/frontend/Symbol_table.ml
@@ -37,6 +37,13 @@ let enter s str ty =
 
 let look s str = Map.find s.table str
 
+let debug s =
+  Debug.eprint "Symbol table keys:";
+  Map.keys s.table |> List.to_string ~f:(fun s -> s) |> Debug.eprint ;
+  Debug.eprint "Symbol table stack:";
+  s.stack |> List.to_string ~f:(fun s -> s) |> Debug.eprint ;
+  ()
+
 let begin_scope s =
   { s with scopedepth = s.scopedepth + 1; stack= sentinel_new_scope :: s.stack }
 
--- a/src/frontend/Symbol_table.ml
+++ b/src/frontend/Symbol_table.ml
@@ -62,17 +62,14 @@ let set_read_only s str =
 let get_read_only s str =
   match Map.find s.readonly str with Some () -> true | _ -> false
 
-let set_is_assigned s str = Hashtbl.remove s.isunassigned str
+let set_is_assigned s str =
+  { s with isunassigned= Map.remove s.isunassigned str }
 
 let set_is_unassigned s str =
-  let _ : [`Duplicate | `Ok] =
-    if Hashtbl.mem s.isunassigned str then `Ok
-    else Hashtbl.add s.isunassigned ~key:str ~data:()
-  in
-  ()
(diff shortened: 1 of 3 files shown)


src/frontend/Symbol_table.ml
```diff
@@ -5,75 +5,71 @@ open Core_kernel
...
-  ; scopedepth: int ref
-  ; readonly: (string, unit) Hashtbl.t
-  ; isunassigned: (string, unit) Hashtbl.t
-  ; globals: (string, unit) Hashtbl.t }
+  { table: (string, 'a, String.comparator_witness) Map.t
+  ; stack: string List.t
+  ; scopedepth: int
+  ; readonly: (string, unit, String.comparator_witness) Map.t
+  ; isunassigned: (string, unit, String.comparator_witness) Map.t
+  ; globals: (string, unit, String.comparator_witness) Map.t }
```
VMatthijs: `Map`s from `'a` to `unit` are are the same as `Set`s of `'a`. I was using this as a way of getting a Hashset which Core_kernel seemed to not supply.

VMatthijs:
@palmerlao , I added a minor comment, but it's looking really nice. :-)

src/frontend/Symbol_table.mli
```diff
@@ -2,32 +2,32 @@
 
 type 'a state
```
seantalts: I think this might be supposed to be called `t` if we follow the Jane Street style, but separately if we are doing it functionally do we even need a Symbol table module at all or should we just be using a map or set?

src/frontend/Symbol_table.ml
```diff
@@ -5,75 +5,71 @@ open Core_kernel
...
-  ; globals= String.Table.create () }
+let empty =
+  { table= Map.empty (module String)
+  ; stack= []
+  ; scopedepth= 0
+  ; readonly= Map.empty (module String)
+  ; isunassigned= Map.empty (module String)
+  ; globals= Map.empty (module String) }
+
+let sentinel_new_scope = "-sentinel-new-scope-"
```
seantalts: And this will go away with the functional approach too, right?

seantalts:
Basically seems like what I had in mind, modulo the couple of comments. Thanks for taking a crack at this!

src/frontend/Symbol_table.mli
VMatthijs: My sense is that we could probably go both ways.

The current approach where we use a combination of a map and a stack (and some sets) clearly works, in which case it makes sense to wrap things up in a module.
Alternatively, when working purely functionally people often use the call-stack to do the work of the stack data structure in the symbol table. However, then, it might still be useful to keep the sets around in this case.

Overall, my sense is that it might be useful to keep the Symbol table its own abstract type (module) as that gives you more flexibility regarding implementation and it puts more constraints in place to make sure people don't use arbitrary destructive updates on it.

src/frontend/Symbol_table.ml
VMatthijs: It could if you used the call stack instead of a separate stack inside the Symbol table module.

VMatthijs:
@palmerlao , let me know if you need any help with this!

seantalts:

PR #263, opened 2019-08-22
Sort type signatures in output.

seantalts:
This means the error messages won't change when the implementation of the signatures dictionary
changes. We also probably shouldn't rely on the ordering in the multi hash table, but not sure.


seantalts:
(I'm working on a major refactor of Stan_math_signatures to generate the signatures from a higher-level format that's useful for also generating the explicit template instantiations.

PR #266, opened 2019-08-24
Refactor easy+clean distributions into more structured form

seantalts:
Putting up an initial version to hopefully get some comments :) 

There really are a lot of special cases in here! But my thinking is we can possibly encode the semantics (e.g. "this is the CDF of this LPDF" or "this takes Stan's 'reals' type") and then list out the explicit exceptions to this rule in a way that 1) is way less code but also 2) is useful for other structured uses of the data, like generating explicit template instantiations 😅

### DIFF
--- a/src/middle/Stan_math_signatures.ml
+++ b/src/middle/Stan_math_signatures.ml
@@ -4,8 +4,143 @@ open Core_kernel
 open Type_conversion
 open Mir
 
+type dimensionality =
+  | DInt
+  | DReal
+  | DInts
+  | DReals
+  | DIntAndReals
+  | DRealAndVector
+  | DVectors
+  | DVector
+  | DMatrix
+
+type dist_fn = Lpmf | Lpdf | Rng | Cdf | Ccdf
+
+let vector_types_l = [UReal; UArray UReal; UVector; URowVector]
+let is_primitive = function UReal -> true | UInt -> true | _ -> false
+
 (** The signatures hash table *)
 let stan_math_signatures = String.Table.create ()
+
+let rng_return_type t lt =
+  if List.for_all ~f:is_primitive lt then t else UArray t
+
+let add_unqualified (name, rt, uqargts) =
+  Hashtbl.add_multi stan_math_signatures ~key:name
+    ~data:(rt, List.map ~f:(fun x -> (AutoDiffable, x)) uqargts)
+
+let rec expand_arg = function
+  | DInt -> [UInt]
+  | DReal -> [UReal]
+  | DReals -> [UReal; UArray UReal; UVector; URowVector]
+  | DInts -> [UInt; UArray UInt]
+  | DIntAndReals -> expand_arg DReals @ expand_arg DInts
+  | DVectors -> [UVector; UArray UVector; URowVector; UArray URowVector]
+  | DVector -> [UVector]
+  | DRealAndVector -> [UReal; UVector]
+  | DMatrix -> [UMatrix]
+
+(* XXX The correct word here isn't combination - what is it? *)
+let all_combinations xx =
+  List.fold_right xx ~init:[[]] ~f:(fun x accum ->
+      List.concat_map accum ~f:(fun acc ->
+          List.map ~f:(fun arg -> arg :: acc) x ) )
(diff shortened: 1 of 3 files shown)


seantalts:
I think this should be pretty uncontroversial but I'm a bit worried about some of the variable names and curious if anyone has ideas for making the code that loops over all the possible type signatures could be made cleaner :)

src/middle/Stan_math_signatures.ml
```diff
@@ -4,9 +4,242 @@ open Core_kernel
...
+  | DVector -> [UVector]
+  | DMatrix -> [UMatrix]
+  | DAllBase -> [UInt; UReal; URowVector; UVector; UMatrix]
+  | DDeepVectorized ->
+      List.(
+        concat_map (expand_arg DAllBase) ~f:(fun a ->
+            map (range 0 8) ~f:(fun i -> bare_array_type (a, i)) ))
+
+(* XXX The correct word here isn't combination - what is it? *)
+let all_combinations xx =
```
VMatthijs: I think combinations might be the right word. Or choices.

src/middle/Stan_math_signatures.ml
```diff
@@ -4,9 +4,242 @@ open Core_kernel
...
+(** All of the signatures that are added by hand, rather than the ones
+    added "declaratively" *)
+let manual_stan_math_signatures = String.Table.create ()
+
+let rec bare_array_type (t, i) =
+  match i with 0 -> t | j -> UArray (bare_array_type (t, j - 1))
+
+let rec expand_arg = function
+  | DReal -> [UReal]
+  | DReals -> [UReal; UArray UReal; UVector; URowVector]
```
VMatthijs: Is this not `vector_types_l`?

src/middle/Stan_math_signatures.ml
```diff
@@ -4,9 +4,242 @@ open Core_kernel
...
 let stan_math_signatures = String.Table.create ()
 
+(** All of the signatures that are added by hand, rather than the ones
+    added "declaratively" *)
+let manual_stan_math_signatures = String.Table.create ()
+
+let rec bare_array_type (t, i) =
+  match i with 0 -> t | j -> UArray (bare_array_type (t, j - 1))
+
+let rec expand_arg = function
```
VMatthijs: Move this definition up to be immediately after `dimensionality`?

VMatthijs:
I've looked at this for a bit now and I'm not sure what I think. It's great that it reduces the size of the code so much. On the other hand, this is not code that often needs to be refactored by developers, right? What happens with this code is that inexperienced people need to be able to add their new signatures to it easily. I wonder if the refactor would make this harder for them. Is there something that prompted this refactor, in particular that this is solving?

For correctness purposes, I would feel more assured if there were a test showing that the old and the new signatures end up being the same in the end.

seantalts:
Yes, this is essential for generating explicit template instantiations. I
also think we'd want to move to something more like this in the futureture
anyways- because now it's infinitely easier to add a new vectorised
function correctly, for instance. We can comment next to all the weird
cases where someone didn't add a function correctly (and they already kind
of stand out a lot more in this version). We'd also like to begin
associating functions together somehow, e.g. that the cdf is the integral
of the PDF, etc.

That said, this design might not be quite right! Open to ideas for other
ways of doing this. For example, maybe dimensionality should only deal with
vectorised functions? Or we should split the library into the explicit
different kinds of functions.

src/middle/Stan_math_signatures.ml
```diff
@@ -4,9 +4,242 @@ open Core_kernel
...
+  | DIntAndReals
+  | D1DeepReals
+  | DArrayAndVectors
+  | DVectors
+  | DVector
+  | DMatrix
+  | DDeepVectorized
+  | DAllBase
+
+type fkind = Lpmf | Lpdf | Rng | Cdf | Ccdf | Functional | DimPromoting
```
VMatthijs: I find the term `Functional` a bit confusing. Could we just call it `RealValued`?
Also could we document the intent of some of these in comments? Same for `dimensionality`?

PR #269, opened 2019-08-28
[WIP] Add Sparse Matrix Type

SteveBronder:
## Description

Add a sparse matrix type to the stanc3 compiler!

#### Good News:

The following program goes through the whole compiler and makes the .hpp!

```stan
data {
  int<lower=0> row_size;
  int<lower=0> col_size;
  sparse_matrix<lower=0>[row_size, col_size] x_matrix;
  vector[row_size] y_vector;
}
parameters {
  vector[col_size] beta;
}
model {
  y_vector ~ normal(x_matrix * beta, 1);
}
```

#### Bad News:

I did this in pretty much the jankiest way possible. There is a _lot_ of duplicated code, having an attribute style for sparse matrices would probably get rid of a lot of that. Sparse matrices are essentially treated the same as dense matrices besides a few points below.

Also hpp would surely not compile! You can see the hpp file in the gist [here](https://gist.github.com/SteveBronder/7f09f51ba17c671f33791ee054964876). I'll list out a couple places in particular that are bound to goof

- [ ] write_array:

It looks like we need a sparse equivalent to resize_match.

```cpp

template <typename T, typename S>
std::vector<T> resize_to_match(std::vector<T>& dst, const std::vector<S>& src) {
  dst.resize(src.size());
  return dst;
}
``` 

Do the arguments in [write_array](https://gist.github.com/SteveBronder/7f09f51ba17c671f33791ee054964876) also need to be SparseMatrices? I'm not as familiar with this part of Stan.

- [ ] Reading/processing in data is wrong

cpp below but it's easier to read from the pretty mir

```cpp
        for (size_t sym1__ = 1; sym1__ <= row_size; ++sym1__) {
          current_statement__ = 6;
          x_matrix[(sym1__ - 1)][(sym2__ - 1)] = context__.vals_r("x_matrix")[(pos__ - 1)];
          current_statement__ = 6;
          pos__ = (pos__ + 1);}} // Could add a space here
```

```stan
  for(sym1__ in 1:row_size) {
    for(sym2__ in 1:col_size) {
      x_matrix[sym1__, sym2__] = FnReadData__("x_matrix", "sparse_matrix", row_size, col_size)
                                 [sym1__, sym2__];
    }
  }
```

`x_matrix` needs to read in data from a set of tuples or via a `Map` function 

```cpp
matrix_x.setFromTriplets(TripletList.begin(), TripletList.end());
```

```cpp
Map<SparseMatrix<double>> matrix_x(rows, cols, nnz, outerIndex, innerIndices, values);
```

The triplets should not be that bad, the map may be a bit odder since I haven't sorted out the I/O for that yet.

[Sparse Matrix I/O Proposal Section](https://github.com/SteveBronder/design-docs/blob/spec/sparse-matrices/designs/0004-sparse-matrices.md#io)

Where do I change where that for loop is being made? I can tell it has something to do with `FnReadData` but I can't tell where. I can give this another look through tmrw

- [ ] Compressing the sparse matrix after each transformed section

Eigen sparse matrices come in an uncompressed form that allows for empty spaces between entries. Once we get through transformed data the matrix should be compressed for the rest of the program. Same for params (`matrix_x.compress()`). I

- [ ] Add nz_row and nz_col to bounds

idt we've decided on this yet but it's what Dan and Aki liked. The sparsity structure of the matrix will be defined by the user as

```stan
sparse_matrix<nz_col=nz_col_ind, nz_row=nz_row_ind>[N, M] A;
```

That's all I have for now, but so far it's been not too bad!! Things make sense!


SteveBronder:
Wugh, I goofed the stan signatures fill. I'll fix that up tmrw

SteveBronder:
Ty! I'll do the formatting tmrw

Yes I mostly started this up just to get an idea of what the internals look like and figure out what i do and dont understand. I'm going to make some comments tmrw trying to explain what I believe is happening in english. I think correcting the translation would both help me understand more and be a good starting point for docs

seantalts:
> Wugh, I goofed the stan signatures fill. I'll fix that up tmrw

I didn't see this PR and merged a big refactor of that file - it's now ~1200 fewer lines and under test :) Sorry about the conflict!

SteveBronder:
Okay at this point I'm confused about reading in data. I'm getting a pretty mir like

```
 for(sym1__ in 1:row_size) {
    for(sym2__ in 1:col_size) {
      x_matrix[sym2__, sym1__] = FnReadData__("x_matrix")[pos__];
    }
  }
```

But I really want this to be
```
x_matrix.fromTriplets(FnReadData__("x_matrix").begin(), FnReadData__("x_matrix").end());
```

I'm looking at [`data_read`](https://github.com/stan-dev/stanc3/blob/master/src/stan_math_backend/Transform_Mir.ml#L7) and get I can turn off increasing `pos__` by saying it's a adding `SSparseMatrix _` to the side of `is_scalar` that evaluates to true. But I can't figure out where in that code the actual pretty printing happens for generating the two loops. I also can't tell where `FnReadData__` is being pretty printed, Below I'm trying to parse out what's goin on, can you help me clear up the parts I've commented that I'm not getting?


```ocaml
(**
 * @param smeta A location span that (??Loop??)
 * @param (decl_id, st) Pair holding a string and a sized type
 * @return ...Data?
 *
 **)
let data_read smeta (decl_id, st) =
  let decl_var = (* Make the unsized type variable*)
    { expr= Var decl_id
    ; emeta= {mloc= smeta; mtype= remove_size st; madlevel= DataOnly} }
  in (* Dunno how stmt is on lhs and rhs / declared here*)
  let swrap stmt = {stmt; smeta} in
  let bodyfn var = 
    let pos_var = {expr= Var pos; emeta= internal_meta} in (*pos is just the iterator*)
    let readfnapp var =
      let f =
        internal_funapp FnReadData (* Defines the FnReadData__ *)
          [{var with expr= Lit (Str, decl_id)}]
          var.emeta
      in (*Unclear*)
      {expr= Indexed (f, [Single pos_var]); emeta= {var.emeta with mtype= UInt}}
    in
    let pos_increment = (*Do pos__ += 1 if this is not a scalar*)
      if is_scalar st then []
      else
        [Assignment ((pos, UInt, []), binop pos_var Plus (mir_int 1)) |> swrap]
    in
    SList (* Unclear:, Making a list and pumping it to swrap?*)
      ( assign_indexed (remove_size st) decl_id smeta readfnapp var
      :: pos_increment )
    |> swrap
  in
  let pos_reset = Assignment ((pos, UInt, []), loop_bottom) |> swrap in
  [pos_reset; for_scalar_inv st bodyfn decl_var smeta]
(*Unclear It's resetting pos__, idk what for_scalar_inv is, then a lot of stuff?**)
```

SteveBronder:
> Can you post the new API of the var_context that you want to generate C++ to use just to help me get some more context?


So from the [eigen docs](https://eigen.tuxfamily.org/dox/group__TutorialSparse.html) (Under "Filling a sparse matrix") for filling sparse matrices you can either fill a sparse matrix from a set of triplets 

```cpp
typedef Eigen::Triplet<double> T;
std::vector<T> tripletList;
tripletList.reserve(estimation_of_entries);
for(...)
{
  // ...
  tripletList.push_back(T(i,j,v_ij));
}
SparseMatrixType mat(rows,cols);
mat.setFromTriplets(tripletList.begin(), tripletList.end());
```
Or 

```cpp
1: SparseMatrix<double> mat(rows,cols);         // default is column major
2: mat.reserve(VectorXi::Constant(cols,6));
3: for each i,j such that v_ij != 0
4:   mat.insert(i,j) = v_ij;                    // alternative: mat.coeffRef(i,j) += v_ij;
5: mat.makeCompressed();                        // optional
```

But the second one has some catches

> The key ingredient here is the line 2 where we reserve room for 6 non-zeros per column. In many cases, the number of non-zeros per column or row can easily be known in advance. If it varies significantly for each inner vector, then it is possible to specify a reserve size for each inner vector by providing a vector object with an operator[](int j) returning the reserve size of the j-th inner vector (e.g., via a VectorXi or std::vector<int>). If only a rought estimate of the number of nonzeros per inner-vector can be obtained, it is highly recommended to overestimate it rather than the opposite. If this line is omitted, then the first insertion of a new element will reserve room for 2 elements per inner vector.
The line 4 performs a sorted insertion. In this example, the ideal case is when the j-th column is not full and contains non-zeros whose inner-indices are smaller than i. In this case, this operation boils down to trivial O(1) operation.

> The for loops happen in for_scalar_inv, which delegates some work to for_scalar. I claim you will still need to generate for loops for any arrays around your sparse matrices 

We can do two, it's just going to be v v slow for loading in data.

Right now I have the compiler emitting this part in the for loop, which is just the same as a standard matrix

```cpp
      pos__ = 1;
      for (size_t sym1__ = 1; sym1__ <= row_size; ++sym1__) {
        for (size_t sym2__ = 1; sym2__ <= col_size; ++sym2__) {
          assign(x_matrix,
                 cons_list(index_uni(sym2__),
                           cons_list(index_uni(sym1__), nil_index_list())),
                 context__.vals_r("x_matrix")[(pos__ - 1)],
                 "assigning variable x_matrix");
          pos__ = (pos__ + 1);
        }
      }
```

Where assign is in stan as

```cpp
template <typename T, typename U>
void assign(Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>& x,
            const cons_index_list<
                index_uni, cons_index_list<index_uni, nil_index_list> >& idxs,
            const U& y, const char* name = "ANON", int depth = 0) {
  int m = idxs.head_.n_;
  int n = idxs.tail_.head_.n_;
  math::check_range("matrix[uni,uni] assign range", name, x.rows(), m);
  math::check_range("matrix[uni,uni] assign range", name, x.cols(), n);
  x(m - 1, n - 1) = y;
}
```

So I think there's a couple ways to do this. One is something like this, if we have the matrix with something like

```stan
int<lower=0> nonzero_elements;
int<lower=0> nonzero_rows[nonzero_elements];
int<lower=0> nonzero_col[nonzero_elements];
sparse_matrix<lower=0>[nonzero_rows, nonzero_cols, row_size, col_size] x_matrix;
```

I'd like to output something like this.

```cpp
      pos__ = 1;
      nonzero_elements = context__.vals_i("nonzero_elements")[(pos__ - 1)];
      nonzero_rows = std::vector<int>(nonzero_elements, 0);

      pos__ = 1;
      for (size_t sym1__ = 1; sym1__ <= nonzero_elements; ++sym1__) {
        nonzero_rows[(sym1__ - 1)] =
            context__.vals_i("nonzero_rows")[(pos__ - 1)];
        pos__ = (pos__ + 1);
      }
      nonzero_col = std::vector<int>(nonzero_elements, 0);

      pos__ = 1;
      for (size_t sym1__ = 1; sym1__ <= nonzero_elements; ++sym1__) {
        nonzero_col[(sym1__ - 1)] =
            context__.vals_i("nonzero_col")[(pos__ - 1)];
        pos__ = (pos__ + 1);
      }
      x_matrix = Eigen::SparseMatrix<double>(row_size, col_size);
/**
 * Here we actually read these into a triplet
 */   // If this was a sparse matrix with dense matrix elements double would be an eigen type
      std::vector<Eigen::Triplet<double>> tripletList;
      x_matrix_triplet_list.reserve(nonzero_elements.size());
      pos__ = 1;
      for (size_t sym1__ = 1; sym1__ <= nonzero_elements.size(); ++sym1__) {
          assign(x_matrix_triplet_list, // index_uni's here are non-zero positions
                 cons_list(index_uni(nonzero_row[sym1__ - 1]), 
                           cons_list(index_uni(nonzero_col[sym1__ - 1]), nil_index_list())),
                 context__.vals_r("x_matrix_triplet_list")[(pos__ - 1)],
                 "assigning variable x_matrix");
          pos__ = (pos__ + 1);
        }
      }
    // Now at the end set the sparse matrix from the triplets list
    mat.setFromTriplets(x_matrix_triplet_list.begin(), x_matrix_triplet_list.end());
```

where assign would look something like

```cpp
template <typename T, typename U>
void assign(std::vector<T>& x,
            const cons_index_list<
                index_uni, cons_index_list<index_uni, nil_index_list> >& idxs,
            const U& y, const char* name = "ANON", int depth = 0) {
  int n = idxs.head_.n_;
  int m = idxs.tail_.head_.n_;
  x.push_back(T(n - 1, m - 1, y));
}
```

Do you think that's possible? I've been messing around with the code though I'm still having a hard go at figuring out how things jump around.

> (e.g. sparse_matrix[N, M, nzr, nzc] x_matrix[3, 4, 5], but that you should just add a new pattern match on Sparse Matrices to not emit any further loops but to emit an FnReadData (via that internal_funapp) of x_matrix. That happens in the backend's Transform_Mir stage and I think you should end up with something like,
for <any wrapping arrays> { x_matrix = FnReadData__('x_matrix') }

I'll give this a look. See if I can slap together something like the pattern above

> Then you can take a look at Expression_gen.ml where we write code to translate FnReadData into something meaningful in C++. At that point you can match on a funapp of FnReadData that's supposed to return a SparseMatrix like this:


| {expr=FunApp(CompilerInternal, f, args); emeta={mtype=USparseMatrix; _}} 
  when internal_fn_of_string f = Some FnReadData ->
  <do stuff>

> Does that help?

Sort of, though I find myself procrastinating and doing stupid things in stan math a good bit because this stuff is pretty confusing to me.

seantalts:
Code is often helpful but so are English words :P Let me try to paraphrase what I think your code is trying to do: 
1) read in some auxiliary stuff in a pretty normal way (`std::vector<int>`s for nonzero cols and rows)
2) add a new mechanism for reading in triplets of `(row_idx, col_idx, val)` given the nonzero rows and cols above and some data that is, I think, just a list of the vals without rows and cols attached.
3) pass that triplet list to Eigen to construct the `x_matrix`

Is that right? If so, here are some things to think about and possibly prototype in C++ before going back to OCaml:

1) What would reading an array of sparse matrices look like? What changes to make it doubly nested? I assume in that case the nonzero entries are enforced to all be the same, but I'm not sure how we're thinking about that.
2) We aren't going to add a Stan type for triplets just to support the reading in of sparse matrices (I think). Luckily, once you're in the backend and you're presented with MIR of the form above (something like`for <any wrapping arrays> { x_matrix[array_idx_1][array_idx_2...] = FnReadData__('x_matrix')[array_idx_1][array_idx_2...] }`), you can generate arbitrary C++ of your choosing whenever you see that. I suspect that whatever you choose will want to take place at that `FnReadData` level and not replace all of the for loops, but we could talk about replacing those too as early as `Transform_Mir.ml`. So no need to override `stan::model::assign` to deal with triplets :) What would your arbitrary C++ look like? How does that mesh with arbitrarily nested arrays of spares matrices? 

In my head, you'd end up replacing calls to `FnReadData` for sparse matrices with arbitrary code that reads out triplets into some auxiliary data structure (make sure its name ends with "__" so it doesn't clash with anything a Stan user could create) and then doing the Eigen call and assignment. 

And yeah, I think this is possible :) This stuff is super confusing especially when you have to think in such grandiose generality as the full Stan language permits. Often helps to work out as many cases as you can think of on paper first.

SteveBronder:
> Code is often helpful but so are English words :P Let me try to paraphrase what I think your code is trying to do:

Lol my b I updated the comment with some wording.

> What would reading an array of sparse matrices look like?

So like `sparse_matrix[yada, yada, yada, yada] int[10]`?

I'm not sure, you could have an assign specialized for an inner type that's an Eigen Sparse matrix.
It's not what we do now but maybe we could feed the `assign` the size of the inner sparse matrices along with a pointer to the first element in the array and have construction of the sparse matrix happen on the C++ side.

> What changes to make it doubly nested? 
>  I assume in that case the nonzero entries are enforced to all be the same, but I'm not sure how we're thinking about that.

Do you mean like `matrix[N, M] sparse_matrix[yada, yada, yada, yada]` like a sparse matrix with dense members? 
Do we want that? I have to think about that, it does seem like a nice pattern for sparse matrices with dense blocks!

Note for myself mostly, but something like

```cpp
std::vector<Eigen::Triplet<Eigen::Matrix<double, N, M>> tripletList;
```

and the loop to read stuff in looks like
```cpp
for (size_t sym1__ = 1; sym1__ <= nonzero_elements; ++sym1__) {
      assign(x_matrix_triplet_list, // index_uni's here are non-zero positions
          cons_list(index_uni(nonzero_row[sym1__ - 1]), // sparse nonzero rows 
                    cons_list(index_uni(nonzero_col[sym1__ - 1]), // sparse nonzero cols 
                     cons_list(index_uni(inner_x_matrix_rows[sym1__ - 1]), // inner eigen matrix rows
                       cons_list(index_uni(inner_x_matrix_cols[sym1__ - 1]), // inner eigen matrix cols
                        nil_index_list())))),
          context__.vals_r("x_matrix_triplet_list"),
          "assigning variable x_matrix");
}
```

And assign would be something like

```cpp
template <typename T, typename U> // Something here checking if T is triplet
void assign(std::vector<T>& x,
            const cons_index_list<
                index_uni, // Sparse nonzero row position 
                cons_index_list<index_uni, // Sparse nonzero col position 
                 cons_index_list<index_uni, // Eigen matrix rows 
                  cons_index_list<index_uni, // Eigen matrix cols
                   nil_index_list>>>>& idxs, // Eigen Matrix Columns
            const U& y, // First element of eigen matrix
            const char* name = "ANON", int depth = 0) {
  int sparse_n = idxs.head_.n_;
  int sparse_m = idxs.tail_.head_.n_;
  int inner_n = // stuff for getting number of eigen rows - 1
  int inner_m = // stuff for getting number of eigen cols - 1
  // Make the Eigen matrix
  T mat = Map<T> v(&y, inner_n * inner_m);
  x.push_back(T(n - 1, m - 1, mat));
}
```

bob-carpenter:
> On Sep 14, 2019, at 9:36 PM, Steve Bronder <notifications@github.com> wrote:
> 
> Code is often helpful but so are English words :P Let me try to paraphrase what I think your code is trying to do:
> 
> Lol my b I updated the comment with some wording.
> 
> What would reading an array of sparse matrices look like?
> 
> So like sparse_matrix[yada, yada, yada, yada] int[10]?
> 
> I'm not sure, you could have an assign specialized for an inner type that's an Eigen Sparse matrix.

Why is there a sparse matrix and an int?  I'm thinking something like

  matrix[3, 4] a[5, 6];

That's a 5 x  6 array of 3 x 4 matrices.  We'd want to do the same thing if matrix was a sparse matrix,

  sparse_matrix[...] a[5, 6];

which would produce a 5 x 6 array of sparse matrices, each of which would have to have the same sparsity pattern until we generalize further to raggedness.  Technically there could be a 5 x 6 array of sparsity patterns when we allow raggedness.

> It's not what we do now but maybe we could feed the assign the size of the inner sparse matrices along with a pointer to the first element in the array and have construction of the sparse matrix happen on the C++ side.
> 
> What changes to make it doubly nested?
> I assume in that case the nonzero entries are enforced to all be the same, but I'm not sure how we're thinking about that.

Yes, they'd have to be.  See above.

> 
> Do you mean like matrix[N, M] sparse_matrix[yada, yada, yada, yada] like a sparse matrix with dense members?
> Do we want that?

No!

For a sparse matrix, specifying it as triple (int[] ii, int[] jj, real[] y) is easiest on human readers;  So whatever that structure S that is in C++, we'd have a std::vector<S> for an array, vector<vector<S>> for a 2D array, etc.

Obviously, that triple can be coded with a tuple, integer arrays, and a real array (or vector).  But we don't have tuples yet.

seantalts:

PR #292, opened 2019-09-04
Optimization integration

rybern:
Small PR that adds a CLI option to run optimizations on compiled programs and a CLI option to print out optimized MIR.

The interface for the optimization function (Analysis_and_optimization.Optimize.optimization_suite) encapsulates the order of optimizations while letting callers turn individual optimizations on or off. This is for testing or adding fine-grained CLI flags.

Some more work needs to be done on the phase order and on testing optimizations themselves, but since this is all turned off by default, I don't think this PR breaks anything.


test/integration/good/compiler-optimizations/cpp.expected
```diff
@@ -915,7 +1470,7 @@ class optimizations_model : public model_base_crtp<optimizations_model> {
     
   std::string get_unconstrained_sizedtypes() const {
     stringstream s__;
-    s__ << "[{\"name\":\"theta\",\"type\":{\"name\":\"real\"},\"block\":\"parameters\"},{\"name\":\"phi\",\"type\":{\"name\":\"real\"},\"block\":\"parameters\"},{\"name\":\"x_matrix\",\"type\":{\"name\":\"matrix\",\"rows\":" << 3 << ",\"cols\":" << 2 << "},\"block\":\"parameters\"},{\"name\":\"x_vector\",\"type\":{\"name\":\"vector\",\"length\":" << 2 << "},\"block\":\"parameters\"},{\"name\":\"x_cov\",\"type\":{\"name\":\"vector\",\"length\":" << (2 + ((2 * (2 - 1)) / 2)) << "},\"block\":\"parameters\"}]";
+    s__ << "[{\"name\":\"theta\",\"type\":{\"name\":\"real\"},\"block\":\"parameters\"},{\"name\":\"phi\",\"type\":{\"name\":\"real\"},\"block\":\"parameters\"},{\"name\":\"x_matrix\",\"type\":{\"name\":\"matrix\",\"rows\":" << 3 << ",\"cols\":" << 2 << "},\"block\":\"parameters\"},{\"name\":\"x_vector\",\"type\":{\"name\":\"vector\",\"length\":" << 2 << "},\"block\":\"parameters\"},{\"name\":\"x_cov\",\"type\":{\"name\":\"vector\",\"length\":" << 3 << "},\"block\":\"parameters\"}]";
```
seantalts: it is suspicious that this line changed - do you know why?

test/integration/good/compiler-optimizations/cpp.expected
```diff
@@ -365,298 +311,768 @@ class optimizations_model : public model_base_crtp<optimizations_model> {
...
+          {
+            current_statement__ = 51;
+            lp_accum__.add(2);
+            sym25__ = 24;
+            break;
+          }}
+        while (sym25__) {
+          {
+            {
+              break;
```
seantalts: what is happening here??

seantalts:
Here's the two pages on the order of optimizations from that textbook for reference (our mileage may certainly vary):
[order of optimizations.pdf](https://github.com/stan-dev/stanc3/files/3579313/order.of.optimizations.pdf)

test/integration/good/compiler-optimizations/cpp.expected
rybern: Not sure. I'll see if putting dead code elimination at the end would fix this.

test/integration/good/compiler-optimizations/cpp.expected
rybern: I think it's just doing the partial evaluation (2 + ((2 * (2 - 1)) / 2)) = 3

seantalts:
ahhhh :)

rybern:
@seantalts  
I updated the phase ordering to follow the book you sent.

I wrote some notes as an Org file [here](https://github.com/stan-dev/stanc3/blob/optimization-integration/src/analysis_and_optimization/phase-ordering-notes.org) - is this a reasonable thing to do? I tried to include all my rationale and allow someone else can pick up where I left off, but I'm not sure if leaving Org note files is okay.

The optimizations are happening after the backend transformation. I didn't see a reason to split them before and after, since we don't use either of the optimizations that the book suggests should be applied to the highest level of representation. This could easily be changed, by moving the 'A' section into its own function.

seantalts:
Sweet! Is there a way to export .org files to something github can properly
display? Markdown Or html maybe?

rybern:
I think GitHub is exporting it to markdown, it's readable but not great because it doesn't collapse like emacs org-mode.

Do you like the html better? Attached here, but had to add a .txt go make GitHub happy
[phase-ordering-notes.html.txt](https://github.com/stan-dev/stanc3/files/3632518/phase-ordering-notes.html.txt)

The downside is that it's hard to edit without being in MD or Org mode.

rybern:
Also, build seems to be failing for a strange reason:
`fatal: unable to access 'https://github.com/stan-dev/example-models/': Could not resolve host: github.com`

seantalts:
Hopefully that's transient!

Having a nice doc on your thoughts here would be super readable even if it
ends up being more like a write-once blog post than a editable doc
reflecting the state of the code at all times. I couldn't load the HTML on
my phone as anything besides raw text; hopefully that format is better and
we can figure out how to display it.

seantalts:
Build worked now, must have been a network error.
image

There's still stuff like this in the HTML :/

PR #297, opened 2019-09-17
[WIP] Sparse Qs

SteveBronder:
I made a new PR because I had a lot of junk in the old one and was getting confused.

At this point it compiles but for the following stan program is still generating the wrong loop structure

```stan
data{
int row_size;
int col_size;
int inner_row_size;
int inner_col_size;
int nonzero_elements;
int nonzero_rows[nonzero_elements];
int nonzero_cols[nonzero_elements];

sparse_matrix[nonzero_rows,nonzero_cols,row_size,col_size] x_matrix;
}
```

For loop for x_matrix
```cpp
      x_matrix = Eigen::SparseMatrix<double>(row_size, col_size);
      pos__ = 1;
      for (size_t sym1__ = 1; sym1__ <= nonzero_rows; ++sym1__) {
        assign(x_matrix, cons_list(index_uni(sym1__), nil_index_list()),
               context__.vals_r("x_matrix")[(pos__ - 1)],
               "assigning variable x_matrix");
        pos__ = (pos__ + 1);
      }
```

At this point I've given up on the triplet thing. We can do a loop here but it's going to be very very slow. I want the above to look like this

```cpp
      // Need to check nonzero_rows and nonzero_cols is the same size
      for (size_t sym1__ = 1; sym1__ <= nonzero_rows.size(); ++sym1__) {
          assign(x_matrix,
                 cons_list(index_uni(nonzero_rows[sym1__]),
                           cons_list(index_uni(nonzero_cols[sym1__]),
                            nil_index_list())),
                 context__.vals_r("x_matrix")[(pos__ - 1)],
                 "assigning variable x_matrix");
        pos__ = (pos__ + 1);
      }
```

Then on the c++ side I can just use `x.coeffRef(n, m) = y`. But I have some Qs

1. Which parts of the code specifically generate the `[(pos__ - 1)]` on the `context__.vals_r`?
2. How can I add the `.size()` to the for loop?
3. Where is the %a in index_uni(%a) constructed? In particular how can I tell it that sym1__ needs to be wrapped by the first and second dimension for a sparse matrix?

The good news is that the parameters block seems to be doing the right thing

```cpp
      Eigen::SparseMatrix<double> alt_matrix;
      alt_matrix = Eigen::SparseMatrix<double>(row_size, col_size);
      alt_matrix =
          in__.sparse_matrix(nonzero_rows, nonzero_cols, row_size, col_size);
      for (size_t sym1__ = 1; sym1__ <= nonzero_rows; ++sym1__) {
        vars__.push_back(rvalue(alt_matrix,
                                cons_list(index_uni(sym1__), nil_index_list()),
                                "pretty printed e"));
      }
```

At the least I can handle all that stuff on the c++ side


src/middle/Middle.ml
```diff
@@ -143,8 +145,12 @@ let rec infer_type_of_indexed ut indices =
...
    |UMatrix, [Between _; Single _]
    |UMatrix, [MultiIndex _]
-   |UMatrix, [Single _] ->
-      UVector
+   |UMatrix, [Single _] -> UVector
+  | USparseMatrix, [All; Single _]
+   |USparseMatrix, [Upfrom _; Single _]
+   |USparseMatrix, [Between _; Single _]
+   |USparseMatrix, [MultiIndex _; MultiIndex _; All; All;]
+   |USparseMatrix, [Single _] -> UVector
```
SteveBronder: Can you explain the `index` types like `Single`? I can't really understand what they do or relate to?

src/frontend/Ast_to_Mir.ml
```diff
@@ -197,7 +197,7 @@ let param_size transform sizedtype =
   let rec shrink_eigen_mat f st =
     match st with
     | SArray (t, d) -> SArray (shrink_eigen_mat f t, d)
-    | SMatrix (d1, d2) -> SVector (f d1 d2)
+    | SMatrix (d1, d2) | SSparseMatrix (_, _, d1, d2) -> SVector (f d1 d2)
```
seantalts: I believe those two aren't the dimensions, those are the nonzero rows and columns, right?

You definitely don't want to do the same thing for sparse matrix that you do for matrix, even if you had the correct dimension.

src/middle/Middle.ml
seantalts: They correspond to these: https://mc-stan.org/docs/2_20/reference-manual/language-indexing-section.html

You can check out how it gets parsed in [the parser](https://mc-stan.org/docs/2_20/reference-manual/language-indexing-section.html)

Single - regular old C style index, a[1]
Upfrom - a[1:]
Between - a[1:5]
MultiIndex - indexing by an int array of indices, a[{1, 2, 5}]


That all is probably irrelevant as I'm pretty sure Dan Simpson wanted to disallow indexing of sparse matrices, right?

src/frontend/Ast_to_Mir.ml
SteveBronder: I just put those there because I have no idea what this function does

seantalts:
> At this point I've given up on the triplet thing. We can do a loop here but it's going to be very very slow.

Why change from that if it's going to be very very slow?

> Which parts of the code specifically generate the [(pos__ - 1)] on the context__.vals_r?

https://github.com/stan-dev/stanc3/blob/eb6893379652852ea0132ddef2c90b00914e3831/src/stan_math_backend/Transform_Mir.ml#L21

> How can I add the .size() to the for loop?

Rather than trying to outsource to the for_scalar function further, you should manually construct a For loop over either the sizedtype dimension of nonzero_rows (or cols) or just over the variable name with a `.size()` appended; former slightly preferred. Our translation from Stan's  "for each" loop does something fairly similar: https://github.com/stan-dev/stanc3/blob/master/src/frontend/Ast_to_Mir.ml#L443

> Where is the %a in index_uni(%a) constructed? In particular how can I tell it that sym1__ needs to be wrapped by the first and second dimension for a sparse matrix?

Check out the module doc for stan_math_code_gen: https://github.com/stan-dev/stanc3/blob/master/src/stan_math_backend/Stan_math_code_gen.ml#L3

The %a just means to compositionally call another pretty printer on some argument. As previously mentioned I suspect we won't need to get into this for what you're trying to do as sparse matrices won't be able to be indexed, but the line here:
```ocaml
  | Single e -> pf ppf "index_uni(%a)" pp_expr e
```

Means to print the string "index_uni()" with the result of calling pp_expr on the expression e. So this just fills the parenthesis with the correct index.

>       for (size_t sym1__ = 1; sym1__ <= nonzero_rows; ++sym1__) {

(in the parameters block) I think you probably want to iterate over items in nonzero_rows here.

src/frontend/Ast_to_Mir.ml
seantalts: In the spirit of teaching a person to fish, I'd highly recommend writing some expect tests and playing around with it. I think I showed you those on the call the other day, right? If not, you basically just write something like:
```ocaml
let%expect_test "some name" =
<code that does stuff and then prints it>
```
and then run `dune runtest src` (assuming its' in a `src` file) and it will spit out what it should print. it's like printf based unit testing and I promise it's way more useful than it sounds, haha.

The `shrink_eigen` function (which probably needs a better name) just applies a function to an inner nested eigen type in the case of a matrix transformation like `cholesky_cov`. If we're going with `sparse_matrix` as its own type, then you can't express things like a sparse cholesky_cov matrix so you can just raise an exception here.

src/middle/Middle.ml
```diff
@@ -283,5 +291,6 @@ let rec eigen_size (st : mtype_loc_ad with_expr sizedtype) =
   match st with
   | SArray (t, _) -> eigen_size t
   | SMatrix (d1, d2) -> [d1; d2]
+  | SSparseMatrix (d1, d2, d3, d4) -> [d1; d2; d3; d4]
```
seantalts: This doesn't make sense; I'd just return the matrix dimensions here.

src/middle/Middle.ml
```diff
@@ -204,6 +210,8 @@ let rec for_scalar st bodyfn var smeta =
   | SVector d | SRowVector d -> mkfor d bodyfn var smeta
   | SMatrix (d1, d2) ->
       mkfor d1 (fun e -> for_scalar (SRowVector d2) bodyfn e smeta) var smeta
+  | SSparseMatrix (d1, _, _, _) ->
+      mkfor d1 bodyfn var smeta
```
seantalts: Here is where I would generate a custom MIR For loop to loop over the size of the nonzero rows / cols.

src/frontend/Ast_to_Mir.ml
seantalts: Here's an example of an expect test

https://github.com/stan-dev/stanc3/blob/eb6893379652852ea0132ddef2c90b00914e3831/src/middle/Middle.ml#L228-L256

seantalts:
PS thanks for saving all these questions up and asking them all at once on a PR so it's easy to give line-based feedback and paste links :) I think this worked well from my perspective but we can have a call for some of the more intuitive parts soon if you like, hopefully I'm feeling better tomorrow.

src/middle/Middle.ml
SteveBronder: In `mkfor` it assums the loop is happening over a `Single` and I wasn't sure exactly which of these that lined up with. Thanks!

SteveBronder:
>  At this point I've given up on the triplet thing. We can do a loop here but it's going to be very very slow.

>> Why change from that if it's going to be very very slow?

Because I've been trying to write a for loop for 8 hours and the sunk cost is becoming real for about 20 lines of code

src/frontend/Ast_to_Mir.ml
SteveBronder: A better name for this may be `reshape_eigen` with note saying this is used in special functions to reshape special cases of eigen matrices.

src/middle/Middle.ml
SteveBronder: Putting these two comments together because they are on the same topic

> Rather than trying to outsource to the for_scalar function further, you should manually construct a For loop over either the sizedtype dimension of nonzero_rows (or cols) or just over the variable name with a .size() appended; former slightly preferred. Our translation from Stan's "for each" loop does something fairly similar,

I don't really get what you mean by manual here. Here's my understanding.

1. from data_read I have to end up calling `for_scalar_inv` on the size type 
2. `for_scalar_inv` has to return a record that looks like

```
{stmt= For {loopvar; lower; upper; body= {stmt; smeta}}; smeta}
```

i.e. a record with stmt being a statement with a location span and metadata with the other side being an unnamed location span.

Could I just bring `ForEach` in from the ast file, define it in the mir `statements` types where for is, and use that here? I think I'd have to define my first two dimensions (giving the nonzero columns) as some new 1D sized type, but it feels like that would give me what I want?

SteveBronder:
The %a just means to compositionally call another pretty printer on some argument. As previously mentioned I suspect we won't need to get into this for what you're trying to do as sparse matrices won't be able to be indexed, but the line here:

Yes I understand fmt, I've seen latex's hbox overfill statement too many times so the fmt stuff was oddly familiar

> stanc3/src/stan_math_backend/Expression_gen.ml
> Line 83 in eb68933
>  | Single e -> pf ppf "index_uni(%a)" pp_expr e 
> Means to print the string "index_uni()" with the result of calling pp_expr on the expression e. So this just fills the parenthesis with the correct index.

Yes I was should have made my question more direct. When calling pp_expr to get the result to put in %a for the inner access in the for loop. I don't understand what I should match on (and what I have to manipulate higher up) so that I can have it do the indexing on the nonzero rows and columns.

seantalts:
> Because I've been trying to write a for loop for 8 hours and the sunk cost is becoming real for about 20 lines of code

OCaml code is very dense! :P I totally feel your pain - this part is one of the most difficult parts to grasp, in my opinion. I had such a hard time debugging my reverse-engineering of the system from Stan 2. Try experimenting with each of the intermediate functions at the level of that function so that you can come to understand each one better, especially the `for_scalar` that you need to change. 

I think that the triplets way vs. the other way are going to both end up requiring you to understand the way reading works. Weird gotchas that I encountered when reverse engineering this for myself (that I should probably add as doc somewhere! it'd be great if we got a half page of notes out of this and put it in the source code or a README somewhere).

1) Parameters and Data are read in in a different order from each other; data use `for_scalar_inv` and params just use `for_eigen`. The latter means that it will read in an entire eigen data structure at a time because `reader.hpp` has methods for that but `var_context` (for data) does not. for_scalar_inv is `_inv` because to me it seemed inverted from both how parameters are and how I would have coded it). As you can see on the definition,

https://github.com/stan-dev/stanc3/blob/5fb913a41ca5b1de217b6c678e60d58d5d945dd3/src/middle/Middle.ml#L217-L224

it has special inverted structure for Arrays and Matrices such that we end up iterating over the dimensions "inside out." This is a great place to use an expect test to inspect what's going on. There's one already defined in Middle.ml, 
https://github.com/stan-dev/stanc3/blob/5fb913a41ca5b1de217b6c678e60d58d5d945dd3/src/middle/Middle.ml#L228-L256

But this one is written in terms of the MIR - which is a format you'll need to become familiar with over time. We can write the same test in terms of Stan and pretty printed MIR,
```ocaml
let%expect_test "data reads" =
  Frontend.Frontend_utils.typed_ast_of_string_exn {|
  data {
    matrix[2, 3] hi[4, 5];
  }
|}
  |> Frontend.Ast_to_Mir.trans_prog "model_name"
|> Transform_Mir.trans_prog |> Fmt.strf "%a" Pretty.pp_typed_prog |> print_endline
```

I put this in `test/unit/Stan_math_backend_tests.ml` because I wanted to have access to the entire parser and backend at the same time as this is more of an integration test. Now I run dune runtest test/unit,

```
hal ~/scm/stanc3 (master) $ dune runtest test/unit
     patdiff (internal) (exit 1)
(cd _build/default && /home/sean/.opam/default/bin/patdiff -keep-whitespace -location-style omake -unrefined test/unit/Stan_math_code_gen_tests.ml test/unit/Stan_math_code_gen_tests.ml.corrected)
------ test/unit/Stan_math_code_gen_tests.ml
++++++ test/unit/Stan_math_code_gen_tests.ml.corrected
File "test/unit/Stan_math_code_gen_tests.ml", line 63, characters 0-1:
 |    void
 |    operator()(const Eigen::Matrix<T0__, -1, -1>& x,
 |               const Eigen::Matrix<T1__, 1, -1>& y, std::ostream* pstream__)  const
 |    {
 |    return sars(x, y, pstream__);
 |    }
 |    }; |}]
 |
 |let%expect_test "data reads" =
 |  Frontend.Frontend_utils.typed_ast_of_string_exn {|
 |  data {
 |    matrix[2, 3] hi[4, 5];
 |  }
 ||}
 |  |> Frontend.Ast_to_Mir.trans_prog "model_name"
 ||> Transform_Mir.trans_prog |> Fmt.strf "%a" Pretty.pp_typed_prog |> print_endline ;
+|  [%expect {|
+|    input_vars {
+|      array[array[matrix[2, 3], 5], 4] hi;
+|    }
+|    prepare_data {
+|      data int pos__;
+|      pos__ = 1;
+|      data array[array[matrix[2, 3], 5], 4] hi;
+|      pos__ = 1;
+|      for(sym1__ in 1:3) {
+|        for(sym2__ in 1:2) {
+|          for(sym3__ in 1:5) {
+|            for(sym4__ in 1:4) {
+|              hi[sym4__, sym3__, sym2__, sym1__] = FnReadData__("hi")[pos__];
+|              pos__ = (pos__ + 1);
+|            }
+|          }
+|        }
+|      }
+|    }
+|    generate_quantities {
+|      if(PNot__(emit_transformed_parameters__ || emit_generated_quantities__)) {
+|        return;
+|      }
+|      if(PNot__(emit_generated_quantities__)) {
+|        return;
+|      } }
+|    transform_inits {
+|      data int pos__;
+|      pos__ = 1;
+|    } |}]
```

If you'd like to add these tests as part of your PR that would be really helpful for the next folks! I would also do it but I don't want to start modifying those files when I know you're in there doing a bunch of work. If you don't mind and would rather I add them I'm happy to go ahead.

2) There's something like a summary of where each type of reading and writing happens here,

https://github.com/stan-dev/stanc3/blob/5fb913a41ca5b1de217b6c678e60d58d5d945dd3/src/stan_math_backend/Transform_Mir.ml#L243-L261

gq and log_prob gets param-style reads of params; prepare_data (C++ constructor) gets data-style reads of data, and then very confusingly transform_inits gets *data* style reads of *params*. gq writes out params, transformed params, and generated quantities, while transform_inits just writes out params. Luckily all writing is the same.

seantalts:
> When calling pp_expr to get the result to put in %a for the inner access in the for loop. I don't understand what I should match on (and what I have to manipulate higher up) so that I can have it do the indexing on the nonzero rows and columns.

You'll probably want to generate correct MIR first. I recommend playing around just with the MIR part of it using expect tests like the one above until that looks like you want it to. A general design goal here is to have the ultimate string printing be very simple if it can be. Do you have a mock up of what you want there? I suspect any array level stuff will be the same as any other type, but for SSparseMatrices I think you want to generate a custom for loop that loops over the nonzero_rows and maybe even just calls that triplet function. That would mean you don't need to change any `Single` indices to switch them out for the nonzero_rows.

If you end up needing to change the code gen, you'll likely want to match on the entire assignment because that's the only way you can capture all of the information you'll need (I believe). Either that or the entire FnReadData function application. I wrote about the latter one here: https://github.com/stan-dev/stanc3/pull/269#issuecomment-530823480

SteveBronder:
> Parameters and Data are read in in a different order from each other; data use for_scalar_inv and params just use for_eigen. The latter means that it will read in an entire eigen data structure at a time because reader.hpp has methods for that but var_context (for data) does not. 

Is there a reason we don't just add that to the array_context that holds those? That function call has to do a lookup on a map every single call so it would probably be a lot faster to just do the call on the Stan side

Are there difficulties with that I'm not seeing?

I'm trying to plug in `for_eigen` to the bottom of `data_read` like
```
  match st with
  | SSparseMatrix _ -> [pos_reset; for_eigen st bodyfn decl_var smeta] 
  | _ -> [pos_reset; for_scalar_inv st bodyfn decl_var smeta]
```

But hacking at it for a minute there's some other stuff it looks like I need to clear up.

If I wrote those access methods could we eliminate a lot of this stuff for reading in data?

seantalts:
> That function call has to do a lookup on a map every single call so it would probably be a lot faster to just do the call on the Stan C++* side
That stuff will get easily optimized out once we hook up the optimizer, so the performance should end up being the same for both versions I believe.
 
I think @bob-carpenter can comment more on the design of the var_context; I'm not very familiar. One major consideration I can think of offhand (that I didn't think of last time we talked) is that when reading data in column-major order, it's all of the outer array indices that change the fastest. So you couldn't just outsource to a column-major `var_context.vals_r("some_matrix").matrix()` because it would need to be interspersed with the array indices for `some_matrix`.

I'm also not sure how much we can really eliminate for reading data even if we added eigen reader methods to the var_context...

seantalts:
Also take all of this with a grain of salt - I am slightly feverish :P

bob-carpenter:
The whole thing's a mess because we settled on last-index major (column major in matrices) I/O, but our internal array data structures are first-index major.  So we get hosed either way.

It'd be great if there were some way to fix that going forward to maintain data memory locality.  Luckily, it shouldn't be too bad if we're only doing things like big regressions where we just have a big matrix and vector.  Those will still be memory local.

seantalts:
> It'd be great if there were some way to fix that going forward to maintain data memory locality.  

Do you have ideas for this? Would it make sense to just change the IO (as I assume the row-major internal stuff is faster for most operations)?

bob-carpenter:
Yes, I think it'd make sense to change the I/O, but it won't make sense with the R dump stuff, as that's all in R's last-index major format.

SteveBronder:
@bob-carpenter am I right that `array_context` is the one that holds the big ol' array with all the data used in this block? After munging in there I think what we could do is something like below.

If we have a stan program now like


```stan
data{
    int N;
    int M;
    int inner_N;
    int inner_M;
    int inner_K;
    matrix[N, M] x_mat[inner_N,inner_M,inner_K];
}
```

Right now we get back a readin like 
```cpp
// setup the variable
      x_mat =
          std::vector<std::vector<std::vector<Eigen::Matrix<double, -1, -1>>>>(
              inner_M,
              std::vector<std::vector<Eigen::Matrix<double, -1, -1>>>(
                  inner_M, std::vector<Eigen::Matrix<double, -1, -1>>(
                               inner_K, Eigen::Matrix<double, -1, -1>(N, M))));
     // The readin block
      pos__ = 1;
      for (size_t sym1__ = 1; sym1__ <= M; ++sym1__) {
          for (size_t sym3__ = 1; sym3__ <= inner_K; ++sym3__) {
            for (size_t sym4__ = 1; sym4__ <= inner_M; ++sym4__) {
              for (size_t sym5__ = 1; sym5__ <= inner_M; ++sym5__) {
                assign(
                    x_mat,
                    cons_list(
                        index_uni(sym5__),
                        cons_list(
                            index_uni(sym4__),
                            cons_list(index_uni(sym3__),
                                      cons_list(index_uni(sym2__),
                                                cons_list(index_uni(sym1__),
                                                          nil_index_list()))))),
                    context__.vals_r("x_mat")[(pos__ - 1)],
                    "assigning variable x_mat");
                current_statement__ = 7;
                pos__ = (pos__ + 1);
              }
            }
          }
        }
      }
```

Which is just setting all the indices go off from left to right from our definition. What if we reworked the assignment to be something like
```
 context__.vals_r("x_mat", "assigning variable x_mat", x_mat, N, M, inner_m, inner_m, inner_K);
```

I'm getting a bit sleepy, but I think on the C++ side the signature would look like this

```
template <typename Var, typename... Args>
void vars_r(const char* name, const char* operation, Var& input_variable, Args... args) 
```

And as long as we know how to fill a vector and fill an eigen matrix from a contiguous array we can keep recursively calling those (ala Eigen::Map and clever indexing). It would really just be grabbing N * M of the array, filling the first eigen matrix up, then shifting the starting point by N * M until we hit the size of the first vector, then shifting it up again till we hit the size of the second vector, etc.

Does that make sense given the current data scheme on the C++ side?

src/middle/Middle.ml
seantalts: We don't want to add ForEach to the MIR just for this - IR design is a tricky process we want to take seriously. Plus, ForEach is not exactly what you want - you want to iterate over nonzero_rows indices so you can grab both that and the nonzero_cols elements together, right? A ForEach would give you the elements in nonzero_rows with no way to access its corresponding element in nonzero_cols.

But yes, by manually I mean you might have to return that record yourself. It's fine if `upper` is `internal_funapp FnLength [nonzero_rows_var] expr_meta`.

seantalts:
While we wait for Bob, let's assume for the sake of argument that that will work. Is that... simpler? In general I want to move as much computation as possible into stuff stanc3 has insight into, but I don't think it matters for reading data in. So my concerns are just around what's easier for folks to understand taken together as a system. C++ variadic template recursion does not seem at first brush to be simpler than OCaml recursion, but I could be convinced.

The compiler will still have to know how to do for loops over arbitrary declarations to check constraints, and inverted ones to write parameters back out to the interfaces, but that's fine. If we end up liking the C++ methods better, we may also want to add similar methods to the writer class so that we could get rid of the inverted for loops entirely from stanc3.

bob-carpenter:
> On Sep 18, 2019, at 12:31 AM, seantalts <notifications@github.com> wrote:
> 
> While we wait for Bob,

Is there something I'm supposed to give you for this?  I don't have any tricks up my sleeve to fix I/O.

seantalts:
Steve had a question about `array_context`: https://github.com/stan-dev/stanc3/pull/297#issuecomment-532418500

bob-carpenter:
> `context__.vals_r("x_mat", "assigning variable x_mat", x_mat, N, M, inner_m, inner_m, inner_K);`

> Does that make sense given the current data scheme on the C++ side?

Yes. 

But I thought the plan was to move functionality out of C++ functions and into generated code rather than the other way around.

SteveBronder:
> But I thought the plan was to move functionality out of C++ functions and into generated code rather than the other way around.

It's just a time thing to me. Doing this in the compiler is going to take me at least two weeks, doing it in C++ would probably take me a day or two. Just matters how we want to allocate time

seantalts:
Two weeks? It feels like you're almost there now that you know how data reading works... You just need to create one For loop manually :P Take a look at that link I posted earlier; there's code in the ForEach translation in Ast_to_Mir that does something kind of similar. It'd be awesome to teach you enough OCaml to be able to work on the code gen part of the compiler long term.

SteveBronder:
Maybe I'm giving myself too much buffer room and being a cranky geezer about learning things. I did get the ocaml debugger working in vscode so that should help a bit! If your having issues with that I can get on a call tmrw if you want help setting it up

seantalts:
Sweet! How is it?

I thought of another consideration - if we want to compile everything ahead of time, we won't be able to use variadic templated functions of arbitrary arity. So it's probably better if we code gen instead of writing C++ functions. I can take a crack at getting the for loop definition going for you and see if that gets you unstuck?

seantalts:
Getting flashbacks to working with you in your early days of C++ on the Math lib. It was a bit of a struggle, and I think definitely took more time to try to guide than it would have to do everything myself, but I think it was ultimately worth the investment for the project (and I think you said you thought so too). So I guess the question is whether you want to become a compiler writer or not. It's a hard job! Not many exist in the world. More would be better, but I understand the need to focus as well. The choice is yours :)

SteveBronder:
> Sweet! How is it?

It worked for like an hour then it broke lol

> I thought of another consideration - if we want to compile everything ahead of time, we won't be able to use variadic templated functions of arbitrary arity. So it's probably better if we code gen instead of writing C++ functions. 

> if we want to compile everything ahead of time, we won't be able to use variadic templated functions of arbitrary arity.

variadic templates are used all over the math library?

I can take a crack at getting the for loop definition going for you and see if that gets you unstuck?

As of now I have [this](https://github.com/stan-dev/stanc3/pull/297/files#diff-86315da5a493e62bf2264362775e532cR245)

```ocaml
let mkfortnite row_expr col_expr bodyfn iteratee smeta =
  let idx s = 
    Single {expr= Var s; emeta= {mtype= UInt; mloc = smeta; madlevel= DataOnly}}
  in
  let expr_idx s =
   Single {expr= s.expr; emeta= s.emeta}
  in
  let loopvar, reset = gensym_enter () in
  let lower = loop_bottom in
  let row_idx = add_int_index row_expr (idx loopvar) in
  let col_idx = add_int_index col_expr (idx loopvar) in
  let stmt = Block [bodyfn [(add_int_index iteratee (expr_idx row_idx)), (add_int_index iteratee (expr_idx col_idx))]] in
  reset ();
  {stmt= For {loopvar; lower; upper=row_expr; body= {stmt; smeta}}; smeta}
```

This doesn't work, my intent here is to just have one for loop that has a `Body` containing both the nonzero row and column indices I want to insert into. If you have time to help me out it would be v appreciated!

SteveBronder:
> Getting flashbacks to working with you in your early days of C++ on the Math lib. It was a bit of a struggle, and I think definitely took more time to try to guide than it would have to do everything myself, but I think it was ultimately worth the investment for the project (and I think you said you thought so too). So I guess the question is whether you want to become a compiler writer or not. It's a hard job! Not many exist in the world. More would be better, but I understand the need to focus as well. The choice is yours :)

I think there are a couple major differences between then and now. 

1. I could at least get things working in C++, even if badly, and we could iterate from there. Here that's not the case. I've been reading books but it's very different from going from C to C++. Teach a man to fish and all that, but if you leave someone to learn to fish for three weeks, they'll have starved out!

2. People like to complain about the novel that gcc can spew out, but I'd take that any day over Ocaml's compiler errors which are often equivalent to "No"

3. eod the math library is a lot simpler and there is a lot of documentation. The heavy use of recursion is also very different to me. Because I'm not that familiar with both this language and functional programming whenever something is not doc'd I have to look up what the code or type means and take a good guess at its intent. A good example of this is something like `with_expr`

```
type 'm with_expr = {expr: 'm with_expr expr; emeta: 'm}
[@@deriving compare, sexp, hash]
```

Which reading it is a record containing expr (aka itself) and emeta which is it's meta information. Like sure, I can read that and translate it, but it's an odd type and I don't purposefully understand it

4. The names here are often shortened in ways that are not clear to me and I have a hard time keeping track of what is short for which. 

I feel pretty good about my understanding of the lexer and parser, mir and down things start going ack

seantalts:
> variadic templates are used all over the math library?

Yes, but the Stan language doesn't support arbitrary arguments to those; the language spec has a limit to the number of args and nesting it expects from the Math library, which means we can compile it AOT for the Stan language. But the Stan language does support arbitrary nested arrays.

> `let mkfortnite`

I lol'd, haha.

> People like to complain about the novel that gcc can spew out, but I'd take that any day over Ocaml's compiler errors which are often equivalent to "No"

The errors that appear using Merlin (e.g. in any supported editor) are much nicer. Every now and then you get a "Syntax error" but you get used to that pretty quickly. Of the dozen languages I've used professionally I think the OCaml experience is actually really good, at least when you limit it to the subset we do. 

>  A good example of this is something like with_expr

This is the one place in the code base that I feel uncomfortable with. @VMatthijs, @rybern, and @enetsee are all more functionally oriented than I am and assure me this is a fairly reasonable thing. It's explicitly doc'd in the README and @enetsee's introduction that I think I might have linked earlier, but here it is:  https://gist.github.com/enetsee/b2152c99d9762d6bca8b0e1090aa6902#construction-of-intermediate-representations basically from there until the "Generalizing" heading. 

There's also a Lambda the Ultimate post describing both the problem these are trying to solve and the solution [here](http://lambda-the-ultimate.org/node/4170) and [here](http://lambda-the-ultimate.org/node/4170#comment-63836). 

This brings up an interesting question, which is "How hard would it be to learn a compiler code base in C++." We can actually try to answer that question if you want to see how hard it would be to add this to Stan 2 :p 

> The names here are often shortened in ways that are not clear to me and I have a hard time keeping track of what is short for which.

If you can ask your editor for the type of an identifier under your cursor quickly, this becomes super easy. I think the short names are often my fault; we can come up with longer ones or try to make them consistent if that helps. A lot of this was coded trying to rapidly replicate Stan 2.

seantalts:
If you want to do a code review and highlight names you think we should change please do! I'm not sure what the easiest way is, maybe just make a branch where you put comments in the code for each of them?

seantalts:
> This doesn't work, my intent here is to just have one for loop that has a Body containing both the nonzero row and column indices I want to insert into. If you have time to help me out it would be v appreciated!

One for loop over the nonzero rows right? I'll take a look.

seantalts:
Would you mind posting the C++ code you want to generate in the ideal case? Doesn't have to use "assign" or anything. Also do you have a test you're running that I can run?

seantalts:
I put up some code on your branch with an expect test showing what it's doing now. I'm not sure if this is what you're looking for or not, but I think it could be close?

SteveBronder:
Mamma Mia! Yes this is exactly what I wanted tysm! It would be nice to have a `sparse_matrix_name.reserve(stan::length(nonzero_rows))` before the loop but it's fine without it for now.

But yes this does exactly what I need to go do the read in stuff on the C++ side

```cpp
      pos__ = 1;
      for (size_t sym1__ = 1; sym1__ <= stan::length(nonzero_rows); ++sym1__) {
        assign(x_matrix,
               cons_list(index_uni(nonzero_cols[(sym1__ - 1)]),
                         cons_list(index_uni(nonzero_rows[(sym1__ - 1)]),
                                   nil_index_list())),
               context__.vals_r("x_matrix")[(pos__ - 1)],
               "assigning variable x_matrix");
        pos__ = (pos__ + 1);
      }
```

SteveBronder:
Reading over this right now but I'm pretty sure the following stan program is a bit nonsense math wise but I'm pretty sure all the stan functions it calls are sensible (with gist [here](https://gist.github.com/SteveBronder/bccbb4cafa5a01654801bc5a7e340fe2)) of hpp output.

```stan
data {
  int<lower = 0> row_size;
  int<lower = 0> col_size;
  int<lower = 0> nonzero_elements;
  int<lower = 0> nonzero_rows[nonzero_elements];
  int<lower = 0> nonzero_cols[nonzero_elements];
  sparse_matrix<lower = 0>[nonzero_rows, nonzero_cols, row_size, col_size] x_matrix;
   vector[row_size] y_vector;
}

parameters{
    vector[col_size] beta;
    sparse_matrix<lower=0,upper=10>[nonzero_rows, nonzero_cols, row_size, col_size] param_matrix;
}

model{
    y_vector ~ normal(param_matrix * x_matrix * beta,1);
}
```

SteveBronder:
I updated the hpp so you can ctrl+f for `/** SPARSE **/` to see the places I found the sparse matrices are used

seantalts:

PR #310, opened 2019-09-29
Small fix to warn-uninitialized

rybern:
Fixed issue with parameters not being included correctly in the scope of the model or generated quantities blocks. See changed test case.

### DIFF
--- a/src/analysis_and_optimization/Dependence_analysis.ml
+++ b/src/analysis_and_optimization/Dependence_analysis.ml
@@ -212,8 +212,11 @@ let mir_uninitialized_variables (mir : typed_prog) :
       (Set.Poly.of_list flag_variables)
       (Set.Poly.singleton "target")
   in
+  let parameters = Set.Poly.of_list (List.map ~f:fst
+    (List.filter ~f:(fun (_, {out_block; _}) -> out_block = Parameters) mir.output_vars))
+  in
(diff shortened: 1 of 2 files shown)


seantalts:
Thanks! I'm thinking about trying to turn this on and make them errors instead of initializing variables to quiet NaN. Any thoughts about that or the readiness here?

rybern:
Hmm. The way it currently works isn't great for arrays:
- If any element of the array is definitely initialized, it assumes the whole array is initialized. It's difficult in general to identify which elements have and haven't been initialized. I think the only way to guarantee element initialization is through some default or special initialization form.
- If array elements are only initialized inside a loop (as they often are), it will not think any have been initialized, since it can't in general prove that the loop ran at least once.

This makes me think we'll need NaN at least for array elements.

Side note, another thing I ran into:
```
transformed data {
  real<lower=0,upper=1> bounded;
}
```
This program throws throws an uninitialized variable `bounded` even though it's just declared, because the MIR includes the bounds checking code, which reads the variable value. Is it reasonable to throw an error here?

seantalts:
Ah yeah. In general that seems right re: arrays, though it'd be great to start translating stuff to vectorized forms and being able to analyze arrays.

re: uninitialized and unused variable - I think that's good. That changes a runtime error ("NaN is not between 0 and 1") into a compile time error ("bounded has not been initialized").

PR #303, opened 2019-09-23
Fixes #301 - add sigs for integrate_1d.

seantalts:
Also fixes stan-dev/stan/#2786

PR #313, opened 2019-09-30
Add ordered logistic and categorical logit GLM

rok-cesnovar:
This PR adds ordered logistic and categorical logit GLMs to the Stan language. I mostly just followed Sean's video on how to add a new function. These are just simpler than integrate_1d.

Stan Math backend links to the functions: 
https://github.com/stan-dev/math/blob/develop/stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp

https://github.com/stan-dev/math/blob/develop/stan/math/prim/mat/prob/categorical_logit_glm_lpmf.hpp


src/middle/Stan_math_signatures.ml
```diff
@@ -523,6 +523,14 @@ let () =
     ("block", ReturnType UMatrix, [UMatrix; UInt; UInt; UInt; UInt]) ;
   add_unqualified ("categorical_rng", ReturnType UInt, [UVector]) ;
   add_unqualified ("categorical_logit_rng", ReturnType UInt, [UVector]) ;
+  add_unqualified
+    ( "categorical_logit_glm_lpmf"
+    , ReturnType UReal
+    , [bare_array_type (UInt, 1); UMatrix; UVector; UMatrix] ) ;
```
rok-cesnovar: I was trying to add signatures for both int and std::vector<int>. This add just the latter right? Would DVInt be the right choice here? Or just make a duplicate?

src/middle/Stan_math_signatures.ml
```diff
@@ -1102,6 +1110,14 @@ let () =
     add_unqualified
       ("num_elements", ReturnType UInt, [bare_array_type (UVector, i)])
   done ;
+  add_unqualified
+    ( "ordered_logistic_glm_lpmf"
+    , ReturnType UReal
+    , [bare_array_type (UInt, 1); UMatrix; UVector; UVector] ) ;
```
rok-cesnovar: Same thing here. These are the support signatures: 

real ordered_logistic_glm_lpmf(int, matrix, vector, vector)
real ordered_logistic_glm_lpmf(int, row_vector, vector, vector)
real ordered_logistic_glm_lpmf(int[], matrix, vector, vector)
real ordered_logistic_glm_lpmf(int[], row_vector, vector, vector)

src/middle/Stan_math_signatures.ml
rok-cesnovar: Supported signatures should be: 
```
real categorical_logit_glm_lpmf(int, matrix, vector, matrix)
real categorical_logit_glm_lpmf(int, row_vector, vector, matrix)
real categorical_logit_glm_lpmf(int[], matrix, vector, matrix)
real categorical_logit_glm_lpmf(int[], row_vector, vector, matrix)
```

seantalts:
In the video I made a mistake and forgot to run the signature tests :((( Maybe I need to record doing that and reupload with that at the end or something... It should say in the description how to deal with that: 
To run all tests, you can use `dune runtest` without additional arguments. The specific test was one that prints out all the functions signatures here: https://github.com/stan-dev/stanc3/pull/303/files#diff-a81e7ff7cb8be61b752a9c826bcedda9

To just run that test and check it in, you can use:
```
dune runtest test/integration/signatures
dune promote
git commit test/integration/signatures ...
```

We can both double check that file to see that the signatures you thought you added were added :)

src/middle/Stan_math_signatures.ml
seantalts: How does it take both an it or an array of ints with the same last 3 arguments?

seantalts:
Oh wait, you did check that test in. But maybe you updated the code without updating that test?

rok-cesnovar:
Ah no, I manually added the signatures test. Rookie mistake 😊 

That does make much more sense now...

rok-cesnovar:
I will fix this up tomorrow, done for today. Thank you for the quick feedback!

src/middle/Stan_math_signatures.ml
t4c1: The signature with `int` and `row_vector` calculates stuff for a single instance. The signature with `int[]` and `matrix` does same for multiple instances. Other two signatures can broadcast arfuments - use same values of either first or second argument for all instances. 

More info in discussion that started here: https://github.com/stan-dev/math/pull/1242#issuecomment-495637043

rok-cesnovar:
@t4c1 do other CPU glms in Stan Math support broadcasting. If yes, it would be pretty trivial to add that in stanc3 now. It would be great to have that for the next release

From previous discussions I know at least @avehtari wanted this, but I am sure there are others also.

t4c1:
No. I was only asked this for the last two I was working on - that is these two that are being added in this PR. I never went back to implement this for others.

src/middle/Stan_math_signatures.ml
seantalts: That discussion says that the 2nd parameter should be a matrix only if y is an array,
> y can be a scalar OR array of size n,
> x can be a vector of size J OR n x J matrix (only if y is an array)

seantalts:
Oh, I didn't see before that the discussion continued for some time after that. If we allow broadcasting on both the y and the x but only w.r.t. each other then I think this looks right. I'll go ahead and merge, thank you both!

PR #314, opened 2019-10-01
Add optional use of OpenCL/matrix_cl overloads & add support for GPU GLMS

rok-cesnovar:
This is code from the pair programming session with Sean last week with the following additions: 
- merged all recent changes to master
- added all the currently implemented GLM signatures
- changed the opencl triggers map (see below)
- refactored the way we calculate when to use matrix_cl overloads
- add the `--use-opencl` flag (not sure if my way of coding that in would be considered "ocamlish")

## opencl_triggers map

This is the map of signatures that can be used directly with matrix_cl. For this PR this will only be GLMs, others will follow.

Each signature is assigned a set of two values: 
- list of indexes that mark the arguments that are `matrix_cl` in the signature; if the requirements are met, these arguments have to be moved to matrix_cl if not already there
- list of requirements: if any of these items in the list are met we trigger the use of opencl overload. For now we only have a single requirement, but I can imagine some signature having more than one

Each set in the requirements list has 2 elements:
- list of arguments that need to be already on the GPU in order to trigger (for now already on the GPU = is data)
- list of sets that mark if there are any type requirements: `(1,UMatrix)` means that argument 1 must be a matrix, signatures with row_vectors in argument 1 will not trigger

```ocaml
let opencl_triggers =
  String.Map.of_alist_exn
    [ ("normal_id_glm_lpdf",
        ([0;1],
          [ (* Array of conditions under which to move to OpenCL *)
            ([1],[]) (* Argument 1 is data *)
          ]
        )
      )
    ; ("bernoulli_logit_glm_lpmf",
        ([0;1], [([1],[])])
      )
    ; ("categorical_logit_glm_lpmf",
        ([0;1], [([1],[(1,UMatrix)])]) (* Additional requirement of argument 1 being a matrix *)
      )
```

## when to trigger

The logic goes as follows: 
1. for each requirement set (only one for now) check if arguments that must be on the GPU ( is data for now) are there; if not move to the next requirement set; if at the end of list return false
2. if the requirements are met for the data arguments we then check if the type requirements are met; if not go to 1. and check the next requirement set; if data types are ok, return true

For now we require all matrix_cl arguments to be data, because we dont support matrix_cl<rev> currently. Once we have that we will move to the less restrictive list for GLMs given [here](https://github.com/stan-dev/stanc3/pull/314#issuecomment-537393354)


src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -85,10 +85,6 @@ let pp_located_error ppf (pp_body_block, body, err_msg) =
 let pp_arg ppf (custom_scalar, (_, name, ut)) =
   pf ppf "const %a& %s" pp_unsizedtype_custom_scalar (custom_scalar, ut) name
 
-let to_indexed assignee idcs =
-  { expr= Indexed ({expr= Var assignee; emeta= internal_meta}, idcs)
-  ; emeta= internal_meta }
-
```
rok-cesnovar: These were removed due to not being used.

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -149,18 +145,6 @@ let pp_fun_def ppf {fdrt; fdname; fdargs; fdbody; _} =
...
-  | UReal -> "double"
-  | UArray t -> basetype t
-  | UMatrix -> "matrix_d"
-  | URowVector -> "row_vector_d"
-  | UVector -> "vector_d"
-  | x -> raise_s [%message "basetype not defined for " (x : unsizedtype)]
-
-let var_context_container st =
-  match basetype (remove_size st) with "int" -> "vals_i" | _ -> "vals_r"
-
```
rok-cesnovar: Same here.

src/analysis_and_optimization/Dependence_analysis.ml
```diff
@@ -212,11 +212,17 @@ let mir_uninitialized_variables (mir : typed_prog) :
...
       (Set.Poly.singleton "target")
   in
-  let parameters = Set.Poly.of_list (List.map ~f:fst
-    (List.filter ~f:(fun (_, {out_block; _}) -> out_block = Parameters) mir.output_vars))
+  let parameters =
+    Set.Poly.of_list
+      (List.map ~f:fst
+         (List.filter
+            ~f:(fun (_, {out_block; _}) -> out_block = Parameters)
+            mir.output_vars))
```
rok-cesnovar: `make format` did this. I did not touch this file otherwise.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -1,6 +1,49 @@
...
 open Middle
 
+let opencl_triggers =
+  String.Map.of_alist_exn
+    [ ("normal_id_glm_lpdf", [0; 1])
+    ; ("bernoulli_logit_glm_lpmf", [0; 1])
+    ; ("categorical_logit_glm_lpmf", [0; 1])
+    ; ("neg_binomial_2_log_glm_lpmf", [0; 1])
+    ; ("ordered_logistic_glm_lpmf", [0; 1])
+    ; ("poisson_log_glm_lpmf", [0; 1]) ]
```
rok-cesnovar: @t4c1 All these now require both matrix_cl arguments to be data and already on the GPU. Do you think we could be less conservative in any of them? Just double checking.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -240,24 +305,72 @@ let trans_prog (p : typed_prog) =
...
+      let data_var_idents = List.map ~f:fst p.input_vars in
+      let rec trans_stmt_to_opencl s =
+        { s with
+          stmt=
+            map_statement
+              (switch_expr_to_opencl data_var_idents)
+              trans_stmt_to_opencl s.stmt }
+      in
+      List.map stmts ~f:trans_stmt_to_opencl
+    else []
```
rok-cesnovar: Is this the OCaml way of doing this or is there a better way? Teach me :)

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -240,24 +305,72 @@ let trans_prog (p : typed_prog) =
...
     | _ -> false
   in
+  let translate_to_open_cl stmts =
+    if use_opencl then
+      let data_var_idents = List.map ~f:fst p.input_vars in
+      let rec trans_stmt_to_opencl s =
+        { s with
+          stmt=
+            map_statement
+              (switch_expr_to_opencl data_var_idents)
```
seantalts: This is where we say whether the var is available or not - it's only the data variables that we're sending.

I was actually thinking this might be okay for now. Some considerations:
1) This is only a bad encoding if this is done multiple times in a single `log_prob` call, right? In the simple GLM case it isn't really.
2) Once we hook up the optimizer, I would hope to see these "common subexpressions" eliminated and pulled up to a point in log_prob such that `to_matrix_cl(X)` is only called once. I think the optimizer is hooked up now so I'd be curious if it's doing that...
3) Eventually, hopefully our optimizer is smart enough to pull `to_matrix_cl` calls on data out from the log_prob into the prepare_data / constructor block, but for now I don't think that's implemented so I think what we're doing for data now makes sense

rok-cesnovar: For this model we should not be using matrix_cls
```
data {
  int<lower=1> k;
  int<lower=0> n;  
  vector[n] y;
}

parameters {
  vector[k] beta;
  real<lower=0> sigma;
  real alpha;
  matrix[n, k] X;
}

model {
  target += normal_id_glm_lpdf(y | X, alpha, beta, sigma);
}
```
But we still do, the above example `lp_accum__.add(normal_id_glm_lpdf<false>(y_opencl__, to_matrix_cl(X), alpha, beta, sigma));` is from this model. 

I think @t4c1 told me that if y was a parameter we would maybe be fine using a GPU, or the other way around, it depends on the GLM we use. But the speedups are not big in those cases.

seantalts: Ohh, I misunderstood the issue. So what are the conditions under which we should use normal_id_glm? Is it just if X (2nd argument) is data?

rok-cesnovar: I think X is mandatory to be data for all glms to move to opencl overload.

seantalts: And if 1st arg is already in opencl, we should switch to that version, right?

rok-cesnovar: Yes, we dont want to move it just for this function. We would be fine if the first version of this would use gpu glm only if x and y are both data. if that simplifies the problem. And we then go from there.

src/stan_math_backend/Transform_Mir.ml
t4c1: It is in most cases enough to have X already on GPU.

src/stan_math_backend/Transform_Mir.ml
t4c1: The decision rule for these 4 GLMs is simple: use opencl if and only if X is already on GPU.

For GLMs added [here](https://github.com/stan-dev/stanc3/pull/313#discussion_r330023209) we need different rules. `categorical_logit_glm_lpmf` should use opencl if X is a matrix (and not a row vector), regerdless of whether it is already on GPU or not. `ordered_logistic_glm_lpmf` should use opencl if X is a matrix and it is already on GPU. This rule will also be used for first four GLMs once broadcasting support for them is implemented.

rok-cesnovar: Based on Tadej's comment lets go with: "use matrix_cl overload if at least X (2nd arg) is data" (which also means its already on the GPU). I will try to see if I can make this work otherwise I will call for help.

rok-cesnovar: I am not sure I follow. 

These 2 could use GPU GLM in all cases (x data, y data, x & y data, no data)?

```
real categorical_logit_glm_lpmf(int, matrix, vector, matrix)
real categorical_logit_glm_lpmf(int[], matrix, vector, matrix)
```
What about these 2?
```
real categorical_logit_glm_lpmf(int, row_vector, vector, matrix)
real categorical_logit_glm_lpmf(int[], row_vector, vector, matrix)
```

t4c1: Actually we could also use opencl in this case if beta (last arguemnt) is already on GPU.

t4c1:
Rok asked me to summarize again which signatures require which conditions to use opencl.

`real normal_id_glm_lpdf(vector, matrix, real, vector, vector)`
`real normal_id_glm_lpdf(vector, matrix, vector, vector, vector)`
`real bernoulli_logit_glm_lpmf(int[], matrix, real, vector)`
`real bernoulli_logit_glm_lpmf(int[], matrix, vector, vector)`
`real poisson_log_glm_lpmf(int[], matrix, real, vector)`
`real poisson_log_glm_lpmf(int[], matrix, vector, vector)`
`real neg_binomial_2_log_glm_lpmf(int[], matrix, real, vector, real)`
`real neg_binomial_2_log_glm_lpmf(int[], matrix, vector, vector, real)`
`real ordered_logistic_log_glm_lpmf(int[], matrix, vector, vector)`
`real ordered_logistic_log_glm_lpmf(int, matrix, vector, vector)`
Use opencl if X (second argument - the matrix) is already on GPU.

`real ordered_logistic_log_glm_lpmf(int[], row_vector, vector, vector)`
`real ordered_logistic_log_glm_lpmf(int, row_vector, vector, vector)`
Never use opencl.

`real categorical_logit_glm_lpmf(int, matrix, vector, matrix)`
`real categorical_logit_glm_lpmf(int[], matrix, vector, matrix)`
Always use opencl.

`real categorical_logit_glm_lpmf(int, row_vector, vector, matrix)`
`real categorical_logit_glm_lpmf(int[], row_vector, vector, matrix)`
Use opencl if beta (the last argument) is already on GPU.

This should be all versions that are in stan now. Once signatures with broadcasting are added to remaining GLMs thew will either go into "Use opencl if X (second argument - the matrix) is already on GPU." if X is a matrix or "Never use opencl." if x is a row_vector.

rok-cesnovar:
> `real categorical_logit_glm_lpmf(int, row_vector, vector, matrix)`
 `real categorical_logit_glm_lpmf(int[], row_vector, vector, matrix)`
Use opencl if beta (the last argument) is already on GPU.

@t4c1 Just double checking if you are sure about this one, beta is an Eigen matrix, not `matrix_cl`, so this can never be true? This is the signature: 

```
categorical_logit_glm_lpmf(
    const matrix_cl<int>& y_cl, const matrix_cl<double>& x_cl,
    const Eigen::Matrix<T_alpha_scalar, Eigen::Dynamic, 1>& alpha,
    const Eigen::Matrix<T_beta_scalar, Eigen::Dynamic, Eigen::Dynamic>& beta) {
```

src/stan_math_backend/Transform_Mir.ml
seantalts: Doesn't this return an empty list if the flag isn't set? Which would eat all the statements in the pipeline right? I think this should return `stmts` in the `else` case probs?

seantalts:
Sounds like it might be time to "graduate" to the "required args" version Rok had written up on the thread. I think in this first version we were just assuming that if either of the opencl args were data we would wanna convert it.

Rok or Tadej do you wanna take a stab at adding that? Or want me to?

rok-cesnovar:
> Sounds like it might be time to "graduate"

Story of my life :) haha

> Rok or Tadej do you wanna take a stab at adding that? Or want me to?

I will work on this today and tomorrow before morning in New York (3 pm CET). If I fail to make anything useful happen, I will kindly ask for your assistance, if that is OK with you?

t4c1:
@rok right, we can't do that. Just Ignore this case, it would be quite rare anyway.

rok-cesnovar:
@seantalts I am making some progress, so hopefully I wont have to trouble you much. I will continue working on this through the weekend.

test/integration/good/code-gen/dune
```diff
@@ -18,4 +18,15 @@
...
- (action (diff mir.expected mir.output)))
\ No newline at end of file
+ (action (diff mir.expected mir.output)))
+
+ (rule
+ (targets cl.output)
+ (deps (package stanc) (:stanfiles  glm.stan))
+ (action
+  (with-stdout-to %{targets}
+   (run %{bin:run_bin_on_args} "%{bin:stanc} --print-cpp --use-opencl" %{stanfiles}))))
```
rok-cesnovar: @seantalts Can I add this rule? Or is there some other way of doing this?

seantalts: Looks good to me :)

rok-cesnovar: To add a bit more context. I made a glm.stan file that has all the possible signatures and different variations of when the first and second argument are data or parameters. This should be tested I guess (with --use-opencl).

rok-cesnovar:
In terms of logic of when to use matrix_cl in GLMs I think this is now really really close. The only thing missing is logic to not create matrix_cl's for data that is not used. We dont want to rely on dead code elimination.

The other thing missing is OCaml code optimization. I am sure I took some long roads where there are shorter roads.

rok-cesnovar:
This is now ready for review. I completely redid the top-level comment to reflect what is going on in more detail. 

The implementation is now as close as it can be to what @t4c1 suggested should be the requirements to move to matrix_cl signatures. The only one that we currently have a more strict requirement is 

> real categorical_logit_glm_lpmf(int, matrix, vector, matrix)
> real categorical_logit_glm_lpmf(int[], matrix, vector, matrix)
> Always use opencl.

We must still the requirement that arg 1 is data. We dont support matrix_cl<rev> currently and this would not calculate X adjoint values, so this will be added later.

The `test/integration/good/code-gen/glm.stan` model is used to test all possible signatures to see if they trigger correctly. There is also a data argument that is used with a glm that does not trigger. That one is used to check that we dont move unused data arguments to the GPU.

test/integration/good/code-gen/cpp.expected
```diff
@@ -457,6 +457,816 @@ class eight_schools_ncp_model : public model_base_crtp<eight_schools_ncp_model>
 
 typedef eight_schools_ncp_model_namespace::eight_schools_ncp_model stan_model;
 
+// Boilerplate
```
rok-cesnovar: Is it ok to add another glm model to the regular test suite?

Or should we just move all models that will be added to test --use-opencl to a separate folder and have:
-  cpp-no-cl.expected (to test we dont break anything if --use-opencl is not on) 
- cl.expected (--use-opencl on)

seantalts:
Awesome! I'll take a look tomorrow  :D

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -2,41 +2,89 @@ open Core_kernel
...
+    ; ("poisson_log_glm_lpmf",
+        ([0;1], [([1],[])])
+      ) 
+    ]
 let opencl_suffix = "_opencl__"
 let to_matrix_cl e = {e with expr= FunApp (StanLib, "to_matrix_cl", [e])}
 
-let rec switch_expr_to_opencl available_cl_vars e =
-  let is_avail = List.mem available_cl_vars ~equal:( = ) in
+let rec switch_expr_to_opencl _available_cl_vars e =
```
seantalts: Why this name change? I think usually the leading underscore means it won't be used, but we use it on the next line right?

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -2,41 +2,89 @@ open Core_kernel
...
+        | [] -> false
+        | hd :: tl -> if i=0 then check_if_data hd else nth_arg_data tl (i-1)
+      in
+      let rec req_met (data_arg, type_arg) = 
+        match data_arg with
+         | hd :: tl -> 
+          if nth_arg_data args hd then req_met (tl, type_arg)  else false
+        | [] -> data_types_match type_arg (*if the list is empty that means that all requirements are satisfied, move to matching data types *)
+      in
+      let rec triggers_match _md =
```
seantalts: I think here you could use `List.exists`

test/integration/good/code-gen/glm.stan
```diff
@@ -0,0 +1,62 @@
+data {
```
seantalts: This is silly, but can we name this `optimize_glm.stan` for now? The reason is very silly - I'd like for its output to be after the mother models' output in cpp.expected because I would like to be able to verify that the mother model's generated cpp hasn't been changed, and the globbing is happening in alphabetical order.

rok-cesnovar: That seems reasonable.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -1,6 +1,92 @@
...
+          [ (* Array of conditions under which to move to OpenCL *)
+            ([1],[]) (* Argument 1 is data *)
+          ]
+        )
+      )
+    ; ("bernoulli_logit_glm_lpmf",
+        ([0;1], [([1],[])])
+      )
+    ; ("categorical_logit_glm_lpmf",
+        ([0;1], [([-1],[(1,UMatrix)])]) (* Always use OpenCL, when argument 1 is a matrix *)
```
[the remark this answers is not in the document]
rok-cesnovar: This one was removed since, did you comment to a previous commit?

I wanted to use -1 as kind of saying [-1] as a wildcard (no data needed), but then figure I dont need it.

src/stan_math_backend/Statement_gen.ml
```diff
@@ -43,7 +43,7 @@ let rec integer_el_type = function
   | SArray (st, _) -> integer_el_type st
 
 let pp_decl ppf (vident, ut, adtype) =
-  let pp_type =
+  let pp_type =
```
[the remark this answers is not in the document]
seantalts: Possibly. For some reason Github loves to use cached versions of the page even when I refresh.

src/stan_math_backend/Transform_Mir.ml
rok-cesnovar: No reason, will change it back. 

I was trying to understand stuff and I find that for me the best way of learning what things do is commenting out lines and seeing what changes or brakes. And runtest fails if you have unused variables.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -2,41 +2,89 @@ open Core_kernel
...
+        | (i,t) :: tl -> 
+          if nth_arg_type args i t then data_types_match tl else false
+        | [] -> true (* No type requirements or end of list *)
+      in
+      let check_if_data arg = 
+        match arg.expr with 
+          | Var s when is_avail s -> true
+          | _ -> false
+      in
+      let rec nth_arg_data arg i =
```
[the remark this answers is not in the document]
rok-cesnovar: Oh lord... I am such a noob :blush: I kept wanting to do arg.nth. I knew I saw a nth in there but didnt try hard enough haha

rok-cesnovar:
@seantalts I saw you pushed to the rok-pairing-session branch on stanc3 repo. Just letting you know that these changes were done on my repo (I dont have push permissions here).

seantalts:
😬 oops. It's just fixing the merge conflicts.

rok-cesnovar:
No problem for me, you just wasted your time fixing merge conflicts on a stale branch :)

rok-cesnovar:
@seantalts  Thank you for the advices. I still need to move everything above the `when`, which is going to be a more significant refactor, the rest is done. I should have that done tmrw. Will request your re-review.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -1,6 +1,64 @@
...
+  let any_req_met args req_args =
+    List.exists ~f:(fun x -> req_met x args) req_args
+  in
+  let get_trigger f = Map.find_exn opencl_triggers f in
+  let first (a, _) = a in
+  let second (_, a) = a in
+  let cl_args f = first (get_trigger f) in
+  let req_args f = second (get_trigger f) in
+  let map_args args f =
+    if any_req_met args (req_args f) then
```
rok-cesnovar: I am almost certain there is an easier way to get (req_args f), same for cl_args.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -1,6 +1,64 @@
...
+  in
+  let check_type (i, t) args = (List.nth_exn args i).emeta.mtype = t in
+  let check_if_data ind args =
+    match (List.nth_exn args ind).expr with
+    | Var s when is_avail s -> true
+    | _ -> false
+  in
+  let req_met (data_arg, type_arg) args =
+    List.for_all ~f:(fun x -> check_if_data x args) data_arg
+    && List.for_all ~f:(fun x -> check_type x args) type_arg
```
rok-cesnovar: Is this the right way of doing it? I want to do a `for_all` with a function where the second argument is the same list for all calls (we need the entire lists for all elements in type_arg.

rok-cesnovar:
I would also appreciate any help on what would be the cause of the errors in the tests. It seems to fail the end-to-end tests that I am not sure how I could have effected at this point.

test/integration/good/compiler-optimizations/cpp.expected
```diff
@@ -363,6 +363,9 @@ class optimizations_model : public model_base_crtp<optimizations_model> {
       Eigen::Matrix<double, -1, 1> x_cov_in__;
       x_cov_in__ = Eigen::Matrix<double, -1, 1>(3);
       
+      Eigen::Matrix<double, -1, 1> x_cov_in__;
+      x_cov_in__ = Eigen::Matrix<double, -1, 1>(3);
```
seantalts: Looks like there are now twice as many declarations for this?

test/integration/good/code-gen/cpp.expected
```diff
@@ -162,10 +162,14 @@ class eight_schools_ncp_model : public model_base_crtp<eight_schools_ncp_model>
     try {
       local_scalar_t__ mu;
       
+      current_statement__ = 1;
+      mu = in__.scalar();
       current_statement__ = 1;
       mu = in__.scalar();
```
seantalts: For some reason on your branch stanc3 is emitting twice as many data reads as it should... Maybe a bad merge?

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -308,14 +388,59 @@ let trans_prog (p : typed_prog) =
...
   let gq =
     ( add_reads p.generate_quantities p.output_vars param_read
+    |> translate_to_open_cl
     |> constrain_in_params p.output_vars
     |> insert_before tparam_start param_writes
     |> insert_before gq_start tparam_writes )
     @ gq_writes
     |> add_fills
   in
+  let log_prob =
```
seantalts: Yeah, here's the bad merge - this log_prob is the one I think we / you moved to, but the one down on line 447 still exists. So it's adding reads twice.

rok-cesnovar: Argh. Let me look into that. Sorry :/

rok-cesnovar:
Should go green now and actually be ready for review. Sorry for the bad merge and wasting time...

seantalts:
No worries! Happy to help out and it’s pretty normal to need a second pair
of eyes especially with a new language. Will take a look tonight or
tomorrow morning. Excited to get this in for 2.21 :)

seantalts:
I thought I commented on this somewhere but can you explain the UMatrix thing? Why does it matter what type it is vs. how big it is? If someone is telling us they want GPUs on shouldn't we do that even if it's just a large row vector?

seantalts:
(if you already said stuff about this somewhere please page slap me with the link :D. Here is the line I'm talking about: `    ; ("ordered_logistic_glm_lpmf", ([0; 1], [([1], [(1, UMatrix)])]))`)

rok-cesnovar:
@t4c1 is the one suggesting this policy. He basically said that the algorithm will never be faster on a gpu with row_vector as the second argument. I know he explained to me this once, because I was wondering the same thing, but cant really give you the exact details. We can wait for Tadej to answer if you want the details.

But lets look at a somewhat similar example. We would always want to do a multiplication of a matrix with a matrix. But we would really not want to do a vector x row_vector multiplication (unless some seriously huge sizes were in play) if it involved data transfer off course.

rok-cesnovar:
This is the post with the signatures: https://github.com/stan-dev/stanc3/pull/314#issuecomment-537393354

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -28,35 +28,29 @@ let rec switch_expr_to_opencl available_cl_vars e =
...
     match (List.nth_exn args ind).expr with
     | Var s when is_avail s -> true
     | _ -> false
   in
-  let req_met (data_arg, type_arg) args =
-    List.for_all ~f:(fun x -> check_if_data x args) data_arg
-    && List.for_all ~f:(fun x -> check_type x args) type_arg
+  let req_met args (data_arg, type_arg) =
+    List.for_all ~f:(check_if_data args) data_arg
+    && List.for_all ~f:(check_type args) type_arg
```
rok-cesnovar: omg :) : (╯°□°）╯︵ ┻━┻)

That was it all along.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -28,35 +28,29 @@ let rec switch_expr_to_opencl available_cl_vars e =
...
-  let cl_args f = first (get_trigger f) in
-  let req_args f = second (get_trigger f) in
-  let map_args args f =
-    if any_req_met args (req_args f) then
-      List.mapi args ~f:(fun i a -> move_cl_args i a (cl_args f))
-    else args
+  let any_req_met args req_args = List.exists ~f:(req_met args) req_args in
+  let maybe_map_args args (cl_args, req_args) =
+    match any_req_met args req_args with
+    | true -> List.mapi args ~f:(move_cl_args cl_args)
```
rok-cesnovar: This one is supposed to be faster and also more ocamilsh right?

test/integration/good/dune
```diff
@@ -8,3 +8,15 @@
...
  (name runtest)
  (action (diff pretty.expected pretty.output)))
+
+
+(rule
+ (targets integrate_1d.output)
+ (deps (package stanc) (:stanfiles (glob_files */integrate_1d.stan)))
+ (action
+  (with-stdout-to %{targets}
+   (run %{bin:run_bin_on_args} "%{bin:stanc} --auto-format" %{stanfiles}))))
```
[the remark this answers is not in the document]
rok-cesnovar: Ah, I am an idiot. I tried to make an example for Charles on Discourse but then bailed because I think the folder tests is good enough. Will remove.

src/stan_math_backend/Transform_Mir.ml
seantalts: Like using pattern matching here? It just seemed a little clearer to me, but it's possible it's not. I don't feel strongly about it. I'm not sure if it's faster, haha.

rok-cesnovar: I think I remember it from somewhere, idk. But yeah I agree its nicer style. Thanks.

test/integration/good/dune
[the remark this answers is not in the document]
rok-cesnovar: Fixed.

rok-cesnovar:
Fixed the integrate_1d fiasco. Thank you for the style guides. I think I am going to come back to those commits a few time to get out of the bad habits I got experimenting myself. But I guess you have to fail in order to learn :)

I think the UMatrix should be there, I trust Tadej's judgment on the requirements there. We will revisit those when we add matrix_cl<rev> and all of the inputs of the GLMs will become matrix_cls. I expect things to change a lot then.

rok-cesnovar:
Will try and think of a way to rewrite the triggers map once I start adding more support for the rest of the GPU functions. Thanks again for all the lessons!!

seantalts:
No worries, I'm not even sure that's the right thing, I was just trying to brainstorm why it felt a little awkward (the "code smell" to me was probably just all the `_exn` function calls - some are unavoidable, but).

t4c1:
I guess we could also speed up the case of row vector x on GPU if either x or y was large enough. However this would require a special set of kernels for this case. 

I don't think if we could make one general kernel that would be optimal for all adge cases. At least it would be more complicated than current kernels. 

Overall I don't think it is worth the effort, especially since this case is alredy much faster than full matrix x with same number of attributes and instances.

PR #316, opened 2019-10-03
Fix compilation of special hof function argument orders

seantalts:
@rybern this should fix what you found in test/integration/good. I'll test the rest of the models too to make sure they compile. Thanks!


seantalts:
(as always, feel free to comment on these post-merge)

PR #317, opened 2019-10-03
TFP backend working on 8 schools

seantalts:
Obviously still a huge WIP but want to get it mainlined and it doesn't need to be done yet - it's part of a separate stan2tfp binary (we can work on the name).


seantalts:
@enetsee would you mind taking a gander at this at some point? Can be at a high level if you want, like giving suggestions for how to organize the new driver given the other drivers, or that kind of thing.

src/middle/Mir.ml
```diff
@@ -100,13 +100,13 @@ and 'e expr =
   | Indexed of 'e * 'e index list
 [@@deriving sexp, hash, map, compare, fold]
 
-type fun_arg_decl = (autodifftype * string * unsizedtype) list
+type fun_arg_decl = autodifftype * string * unsizedtype
```
enetsee: I agree with the change but is it making a separate PR?

src/middle/Mir_utils.ml
```diff
@@ -96,3 +96,68 @@ let string_of_operators =
...
+     functions, but in practice we may get that wrong from time to time and
+     this count keeps us from infinite loops.
+  *)
+  let count = ref 0 in
+  while Stdlib.( != ) !old_stmts !new_stmts && !count < 100 do
+    old_stmts := !new_stmts ;
+    new_stmts := f !old_stmts ;
+    count := !count + 1
+  done ;
+  !new_stmts
```
enetsee: What are we doing with this fix-point function? The comment suggests that it is being used to simulate a bottom up transform over statement but I can't reason about how that works. 

I'm slightly uncomfortable in applying a bounded fixed point iteration (which may not converge) to what should presumably be a deterministic transform of the MIR. Is there a reason we can't just do a bottom up transform?

src/middle/Mir_utils.ml
```diff
@@ -96,3 +96,68 @@ let string_of_operators =
...
+  let flatten_block s =
+    match s.stmt with
+    | SList ls | Block ls ->
+        if List.for_all ~f:(Fn.non is_decl) ls then ls else [s]
+    | _ -> [s]
+  in
+  let ellide_skip s = match s.stmt with Skip -> [] | _ -> [s] in
+  List.map stmts ~f:cleanup_stmt
+  |> List.concat_map ~f:flatten_block
+  |> List.concat_map ~f:ellide_skip
```
enetsee: Could this be put into analysis and optimization?

src/stanc/stanc.ml
```diff
@@ -156,38 +156,11 @@ let add_file filename =
...
-      | Result.Error err ->
-          let loc = Parse.syntax_error_location err
-          and msg = Parse.syntax_error_message err in
-          Errors.report_parsing_error (msg, loc) ;
-          exit 1
-    with Errors.SyntaxError err ->
-      Errors.report_syntax_error err ;
-      exit 1
-  in
+  let ast = Frontend_utils.get_ast_or_exit filename in
```
enetsee: Not to labor the point but I think we _should_ be doing IO here rather than in a util function in the frontend not least because it means for the 'composable compiler' thing to work it would have to be exposed by all frontends.

src/tfp_backend/Code_gen.ml
```diff
@@ -0,0 +1,231 @@
...
+|}
+
+let pp_prog ppf (p : typed_prog) =
+  pf ppf "%s@,@,%a@,class %s(tfd__.Distribution):@,@[<v 2>%a@]" imports
+    (list ~sep:cut pp_fundef) p.functions_block p.prog_name pp_methods p
+
+(* Major work to do:
+1. Work awareness of distributions and bijectors into the type system
+2. Have backends present an environment that the frontend and middle can use for type checking and optimization.
+*)
```
enetsee: This is sort of a meta-comment but I think there may be some value in looking at translating to a TFP DSL before generating raw python. It's pretty tricky to reason about generating strings vs. a principled translation to another data structure.

test/integration/good/code-gen/cpp.expected
```diff
@@ -176,30 +174,29 @@ class eight_schools_ncp_model : public model_base_crtp<eight_schools_ncp_model>
...
         lp_accum__.add(normal_log<propto__>(mu, 0, 5));
-        current_statement__ = 8;
-        lp_accum__.add(cauchy_log<propto__>(tau, 0, 5));
-        current_statement__ = 9;
+        current_statement__ = 6;
+        lp_accum__.add(normal_log<propto__>(tau, 0, 5));
+        current_statement__ = 7;
         lp_accum__.add(normal_log<propto__>(theta_tilde, 0, 1));
-        current_statement__ = 10;
+        current_statement__ = 8;
```
enetsee: Are these changes necessitated by the PR or unrelated?

src/frontend/Frontend_utils.ml
```diff
@@ -7,3 +7,32 @@ let typed_ast_of_string_exn s =
...
+    match Parse.parse_file Parser.Incremental.program filename with
+    | Result.Ok ast -> ast
+    | Result.Error err ->
+        let loc = Parse.syntax_error_location err
+        and msg = Parse.syntax_error_message err in
+        Errors.report_parsing_error (msg, loc) ;
+        exit 1
+  with Errors.SyntaxError err ->
+    Errors.report_syntax_error err ;
+    exit 1
```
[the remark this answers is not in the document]
seantalts: The frontend is pure, but where should these helpers live? They're used in 3 different drivers.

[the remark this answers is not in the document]
seantalts: (this is a real question - happy to put them wherever you think makes sense).

src/middle/Mir.ml
seantalts: Is this a big change that makes the rest of this PR hard to review?

src/middle/Mir_utils.ml
seantalts: Maybe we can now with the refactor! I'll try to use that functionality that you added.

src/middle/Mir_utils.ml
seantalts: Sure, could you point me to a specific file to put it in? Happy to organize it that way.

src/stanc/stanc.ml
seantalts: I'm not sure I get what you mean... the I/O IS being done here. The place that the function it calls lives is irrelevant to the modularity you're espousing (and that I agree with). Happy to put the helpers wherever, though I'm even less sure I understand what you mean now.

src/tfp_backend/Code_gen.ml
seantalts: We talked about this in a meeting - at some point you end up with strings and in many ways the MIR is that low level DSL (with a TFP specific Transform_mir step happening first). Leaving this for future innovation once we figure out what we want the MIR to solve.

test/integration/good/code-gen/cpp.expected
seantalts: The model had to change to fit properly.

src/middle/Mir_utils.ml
seantalts: It worked!

seantalts:
I think I've responded to all of the comments - changes:
* got rid of the iterative solution for clean_stmts, now it uses the rewrite_bottom_up function
* Moved cleanup_stmts to Mir_utils in analysis and optimization
* rewrote some other transformations to use rewrite_bottom_up.
* Added automated Jenkins tests for 8 schools tested end to end with TFP. 

I'm not sure what you had in mind for this kind of fronend-specific shared driver code. If we think about the number of possible pairings, we get into a combinatoric explosion. After thinking about this and messing around with a few options, I actually think we should leave it in the frontend. The reasoning:

1) Having code that performs I/O located in the src/frontend directory actually costs us nothing in terms of modularity / encapsulation etc. The compiler frontend still upholds all of the purity properties you mentioned on this PR.
2) As we've talked about before, there will be many combinations of different frontends, backends, and drivers. We don't want to make separate packages for e.g. driver utils that use the Stan frontend, the Stan math backend, and the javascript driver; another one that uses the Stan frontend, Stan Math backend, and stanc driver; another that uses Stan frontend, TFP, and a JS driver... etc.
3) As time goes on, we can evolve these things.

src/middle/Mir.ml
enetsee: Not now the Mir refactor is in!

seantalts:
Thanks! So for these IO helpers, you'd have some `driver_utils` (or something) directory and it would have e.g. Compiler.ml and these helpers, and then you could construct drivers in separate directories as they currently are pulling code as needed from the driver_utils folder, which would just collect all available shared driver code (and in turn, depend on every frontend and backend?)? The line you linked to is a good discussion point because we won't always be compiling from a file, so the signature there is not as generic as we need it to be (because we will have a javascript API, protobuf API, etc). So I'm having trouble imagining a Compiler.ml that includes IO that is 1) as abstract as we need it to be to cover all drivers that yet is 2) useful in that abstraction and 3) not a dependency cluster because it would pull code from the entire rest of the project even though each backend needs one much smaller subset of code. At some point it's better to just write a loose collection of helper functions and weave them together in a custom way.

PR #327, opened 2019-10-09
Fix location numbers - off by one error.

seantalts:
Also removes error messages and rethrows original exception which preserves it's type which is important for some Stan sampling code that relies on the type of exception to do different things.


seantalts:
Okay I think this should be the last(?) thing holding up stan-dev/cmdstan#737

PR #328, opened 2019-10-09
Optimization fixes

rybern:
* test/integration/good/compiler-optimizations/optimizations.stan now compiles with all optimizations enabled.
* I also added some CLI options that were very useful for debuging optimizations. You can now display the initial, transformed, or optimized MIR as an S-expression or pretty-printed.

There were four optimizations that didn't compile:

### Partial evalution
Some function types were being replaced by the StanLib type, resulting in them not being passed proper `pstream__` arguments.

### Expectation propagation
FnReadParam, FnWriteParam, and FnUnconstrain were all being propagated despite being effectual functions. Now, expressions that can have side effects are filtered out of expression propagation. 

### AD level optimization
The problem is when:
 * AD level optimization demotes constraint dummy variables (\_in__ variables) in log_prob to 'data'
 * When add_jacobians is applied to the constraint statement, one branch adds an lp__ argument to the constrain function call, which then requires that the dummy variable is autodiffable (see the Stan Math template signature for e.g. cov_matrix_constraint)

Possible solutions were:
 1. Don't demote the AD level unconstrained dummy variables to below the AD level of the original variable.
 2. Don't add autodiff-only constraint calls to non-autodiffable variable constraints.
 3. Add the constraint variant as a source of AD.

I implemented #⁠1 since #⁠2 seemed potential hazardous and #⁠3 seemed harder to implement. The implementation just adds a "\_in__" for each variable in the automatically_differentiable set.

### Lazy code motion
The issues here were just that the transformed code wasn't meeting the expectations of codegen, mostly to do with where to put blocks. I ended up changing codegen to be more robust, e.g. by adding a block if one is required.


seantalts:
The AD level optimization problem is interesting - I think even if the constraint isn't autodiffable, we need to adjust the LP, right? If that's true maybe option 4 is "adjust the Math library to allow for non-autodiffable constraints," right? That seems like the right thing long term, but I'm happy with the solution here for now.

seantalts:
re: Adding blocks - my philosophy has been that anything we can do in a MIR transform step we should, preferring to keep code gen as simple as possible. That's why there's an explicit step in Transform_Mir to `ensure_body_in_blocks`. Maybe we can just run that after optimization?

PR #345, opened 2019-10-15
Remove second scoped definition of `sexp_of_expr_typed_located` and `sexp_of_stmt_loc`

enetsee:
At the moment `sexp_of_expr_typed_located` and `sexp_of_stmt_loc` are both defined twice, once by the deriving annotation on the type definition and then, again, in `Middle.ml`. Since the implementations are different the pretty printed s-expression of the type differs depending on whether the implementations in `Middle.ml` are in scope.

This PR removes the second definition so we have consistent s-expressions and updates tests to this format.

### DIFF
--- a/src/middle/Middle.ml
+++ b/src/middle/Middle.ml
@@ -62,12 +62,6 @@ let operator_return_type_from_string op_name argtypes =
 let operator_return_type op =
   operator_return_type_from_string (string_of_operator op)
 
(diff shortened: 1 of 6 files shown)


enetsee:
@seantalts once you have reviewed, I'll merge back into the MIR refactor and should be able to get that submitted soon after.

PR #346, opened 2019-10-15
MIR refactor

enetsee:
This PR is a minimal version of #251 . 

This version has a flatter functor stack for creating fixed-point types and gets rid of a lot of functionality, most notably for recursion schemes.

The current PR only affects MIR since we agreed to move the AST across to this representation in a separate piece of work.

Analysis and optimization is also mostly untouched other than those changes necessary due to the new type. Again, changes to this code will fall under another PR.


seantalts:
Thanks for putting this up! Let's not merge until after 2.21 goes out since there's a feature freeze going on - I also have my work cut out for me fixing bugs until that happens so I probably won't have time to read this yet :/ But very much looking forward to it!

src/common/Fixed.ml
```diff
@@ -0,0 +1,236 @@
...
+
+  include Foldable.S2 with type ('a, 'b) t := ('a, 'b) t
+  include Pretty.S2 with type ('a, 'b) t := ('a, 'b) t
+
+  val fold_pattern :
+       f:('a * 'r1 First.Pattern.t -> 'r1)
+    -> g:('b * ('r1, 'r2) Pattern.t -> 'r2)
+    -> ('a, 'b) t
+    -> 'r2
+  (**
```
seantalts: Are all of these docstrings in the right place? Does this work with ocamldoc?

src/common/Fixed.ml
```diff
@@ -0,0 +1,236 @@
...
+    let t = f x in
+    {t with pattern= Pattern.map (rewrite_top_down ~f) t.pattern}
+end
+
+(** Nested fixed-point type where an element of the `Pattern` is itself
+a fixed-point type. We use this to represent statements which contain 
+expressions.
+*)
+module type S2 = sig
+  module First : S
```
seantalts: I see why you called Expr First earlier, but I'm not sure we need this level of abstraction... I don't think we're going to have any more two-level types in this compiler. All of them will involve expressions and statements. Are there any gains to be had if we assume that prediction is true?

src/common/Foldable.ml
```diff
@@ -0,0 +1,119 @@
...
+user must provide
+- `S` defines the _complete_ set of type and functions signatures that will be
+exposed by any module supporting the  `standard interface` associated with `S`
+- `Make` is a functor (i.e. a function from module to module) that takes the 
+user supplied `Basic` module and returns a completed module with the signature
+defined by `S`
+
+Where a standard interface is defined for difference arities of 
+type constructors, the convention is to append the arity to the names of the 
+signatures and functors e.g. `Basic2`, `S2` and `Make2`
```
seantalts: This is awesome doc! thank you :) IF there are links to related Jane Street doc that would be good to include too.

src/common/Fixed.ml
```diff
@@ -0,0 +1,236 @@
...
+
+(** The fixed-point of `Pattern.t` annotated with some meta-data *)
+module type S = sig
+  module Pattern : Pattern.S
+
+  type 'a t = {pattern: 'a t Pattern.t; meta: 'a}
+  [@@deriving compare, map, fold, hash, sexp]
+
+  val fix : 'a * 'a t Pattern.t -> 'a t
+  val unfix : 'a t -> 'a * 'a t Pattern.t
```
seantalts: Do we need this?

src/common/Fixed.ml
```diff
@@ -0,0 +1,236 @@
...
+  val pattern_of : 'a t -> 'a t Pattern.t
```
seantalts: Or this? Given we're exposing record types.

src/common/Foldable.ml
```diff
@@ -0,0 +1,119 @@
...
+      ~f:(fun x accu -> accu || pred_first x)
+      ~g:(fun x accu -> accu || pred_second x)
+      ~init x
+
+  let all ~pred_first ~pred_second ?(init = true) x =
+    fold_right
+      ~f:(fun x accu -> accu && pred_first x)
+      ~g:(fun x accu -> accu && pred_second x)
+      ~init x
+end
```
seantalts: I love this module :)

src/common/Foldable.ml
```diff
@@ -0,0 +1,119 @@
...
+module Make2 (X : Basic2) : S2 with type ('a, 'b) t := ('a, 'b) X.t = struct
+  let fold_left ~f ~g ~init x = X.fold f g init x
+
+  let fold_right ~f ~g ~init x =
+    let f' k x z = k @@ f x z and g' k x z = k @@ g x z in
+    fold_left ~f:f' ~g:g' ~init:Fn.id x init
+
+  let any ~pred_first ~pred_second ?(init = false) x =
+    fold_right
+      ~f:(fun x accu -> accu || pred_first x)
```
seantalts: Doesn't the accumulator come first in all fold functions?

src/frontend/Ast_to_Mir.ml
```diff
@@ -1,44 +1,54 @@
 open Core_kernel
+open Common
```
seantalts: I would suggest opening all of the relevant MIR types up top again - like UnsizedType etc. I'm not a big fan of fully qualified access being required at all call-sites.

src/middle/Expr.ml
```diff
@@ -0,0 +1,240 @@
...
+end
+
+(** Expressions with associated location and type *)
+module Typed = struct
+  module Meta = struct
+    type t =
+      { type_: UnsizedType.t
+      ; loc: Location_span.t sexp_opaque [@compare.ignore]
+      ; adlevel: UnsizedType.autodifftype }
+    [@@deriving compare, create, sexp, hash]
```
seantalts: Why use create instead of record syntax?

src/frontend/Ast_to_Mir.ml
```diff
@@ -1,44 +1,54 @@
...
+  and loc = emeta.loc
+  and adlevel = emeta.ad_level in
   match expr with
   | Ast.Paren x -> trans_expr x
   | BinOp (lhs, And, rhs) ->
-      { expr= EAnd (trans_expr lhs, trans_expr rhs)
-      ; emeta= {madlevel; mloc; mtype} }
+      Expr.
+        { Fixed.pattern= EAnd (trans_expr lhs, trans_expr rhs)
+        ; meta= Typed.Meta.create ~type_ ~adlevel ~loc () }
```
seantalts: Why not use the record syntax here?

src/frontend/Ast_to_Mir.ml
```diff
@@ -49,52 +59,55 @@ and trans_expr {Ast.expr; Ast.emeta} =
...
+              , Internal_fun.to_string FnMakeRowVec
               , trans_exprs eles )
         | Indexed (lhs, indices) ->
             Indexed (trans_expr lhs, List.map ~f:trans_idx indices)
         | Paren _ | BinOp _ | PrefixOp _ | PostfixOp _ ->
             raise_s [%message "Impossible!"]
-      in
-      {expr; emeta= {mtype; mloc; madlevel}}
+      and meta = Expr.Typed.Meta.create ~type_ ~adlevel ~loc () in
+      Expr.Fixed.fix (meta, pattern)
```
seantalts: Could this be also written as 
`Expr.Fixed.{meta={Typed.Meta.type_; adlevel; loc}; pattern}`?

seantalts:
PS I keep merging code gen fixes that give you merge conflicts. I'm happy to take responsibility for that and go through and resolve them before we end up merging.

src/analysis_and_optimization/Dataflow_utils.ml
```diff
@@ -18,36 +18,29 @@ let union_maps_left (m1 : ('a, 'b) Map.Poly.t) (m2 : ('a, 'b) Map.Poly.t) :
...
-  | IfElse (pred, then_s, else_s_opt) ->
-      let s', c = f state then_s in
-      Option.value_map else_s_opt
-        ~default:(s', IfElse (pred, c, None))
-        ~f:(fun else_s ->
-          let s'', c' = f state else_s in
-          (join s' s'', IfElse (pred, c, Some c')) )
-  | _ as s -> fwd_traverse_statement s ~init:state ~f
+let branching_traverse_statement stmt ~join ~init ~f =
+  Stmt.Fixed.Pattern.(
```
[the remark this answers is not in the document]
enetsee: The three times `IfElse` is used would need qualifying without a local open

src/middle/Stmt.ml
```diff
@@ -0,0 +1,317 @@
+open Core_kernel
+open Common
+open Helpers
+
+(** Fixed-point of statements *)
+module Fixed = struct
+  module First = Expr.Fixed
```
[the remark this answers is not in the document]
enetsee: Sure - did you have something in mind? It will be used by both `Ast.Stmt` and `Mir.Stmt` eventually but calling it `Expr` seemed a little odd given the context.

src/middle/Expr.ml
```diff
@@ -0,0 +1,240 @@
...
+
+    let empty =
+      create ~type_:UnsizedType.UInt ~adlevel:UnsizedType.DataOnly
+        ~loc:Location_span.empty ()
+
+    let adlevel {adlevel; _} = adlevel
```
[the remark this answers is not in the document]
enetsee: I think it would be much better to hide the implementation but we're kind of at a halfway house right now so both are exposed.

src/analysis_and_optimization/Dependence_analysis.ml
```diff
@@ -25,8 +25,9 @@ let reaching_defn_lookup (rds : reaching_defn Set.Poly.t) (var : vexpr) :
 
 let node_immediate_dependencies
     (statement_map :
-      (label, (expr_typed_located, label) statement * node_dep_info) Map.Poly.t)
-    (label : label) : label Set.Poly.t =
+      ( label
+      , (Expr.Typed.t, label) Stmt.Fixed.Pattern.t * node_dep_info )
```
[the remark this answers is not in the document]
enetsee: I'm not sure what you mean here, I haven't touched `Dependence_analysis` so there shouldn't be any refactors outside changing the names of the statement and expression types.

src/middle/Expr.ml
```diff
@@ -0,0 +1,240 @@
...
+    let loc {loc; _} = loc
+    let pp _ _ = ()
+    let with_type ty meta = {meta with type_= ty}
+  end
+
+  include Specialized.Make (Fixed) (Meta)
+
+  let type_of x = Meta.type_ @@ Fixed.meta_of x
+  let loc_of x = Meta.loc @@ Fixed.meta_of x
+  let adlevel_of x = Meta.adlevel @@ Fixed.meta_of x
```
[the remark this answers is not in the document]
enetsee: I find these are *very* useful! You mean of the variable `x`? Happy to change.

src/analysis_and_optimization/Dependence_analysis.ml
```diff
@@ -152,17 +161,17 @@ let all_labels
...
       ~f:(fun (VVar s, _) -> s)
-      (stmt_rhs_var_set (Map.Poly.find_exn flowgraph_to_mir label).stmtn)
+      (stmt_rhs_var_set (Map.Poly.find_exn flowgraph_to_mir label).pattern)
   in
   union_map labels ~f:label_vars
 
-let rec var_declarations (sw : ('e, 'm) stmt_with) : string Set.Poly.t =
-  match sw.stmt with
+let rec var_declarations (sw : ('e, 'm) Stmt.Fixed.t) : string Set.Poly.t =
+  match Stmt.Fixed.pattern_of sw with
```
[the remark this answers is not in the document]
enetsee: Again, I haven't touched this code other than making it compile. This PR was meant to make minimal changes and we explicitly decided not to refactor anything on analysis and optimization (particularly not dependence analysis which isn't actually used yet).

src/analysis_and_optimization/Factor_graph.ml
```diff
@@ -7,37 +7,34 @@ open Mir_utils
...
-    (statement_map :
-      (label, (expr_typed_located, label) statement * 'm) Map.Poly.t)
-    (label : label) : (label * factor) list =
+let rec extract_factors statement_map label =
   let stmt, _ = Map.Poly.find_exn statement_map label in
   let this_stmt =
     List.map (extract_factors_statement stmt) ~f:(fun x -> (label, x))
   in
-  fold_statement
+  Stmt.Fixed.Pattern.fold
```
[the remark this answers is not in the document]
enetsee: I'm hoping the naming discipline will also help people understand the difference between the pattern and the fixed point type!

src/analysis_and_optimization/Mir_utils.ml
```diff
@@ -20,57 +17,46 @@ let map_rec_expr_state
...
+let map_rec_stmt_loc_num flowgraph_to_mir f s =
   let rec map_rec_stmt_loc_num' (cur_node : int)
-      ({smetan; stmtn} : stmt_loc_num) : stmt_loc =
+      (stmt : Stmt.Located.Non_recursive.t) =
     let find_node i = Map.find_exn flowgraph_to_mir i in
     let recurse i = map_rec_stmt_loc_num' i (find_node i) in
-    {smeta= smetan; stmt= f cur_node (map_statement (fun x -> x) recurse stmtn)}
+    Stmt.Fixed.
+      { pattern=
+          f cur_node (Stmt.Fixed.Pattern.map (fun x -> x) recurse stmt.pattern)
```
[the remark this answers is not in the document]
enetsee: True; there are lots of `fun x -> x` throughout the codebase so we should probably clean them all up at some point.

src/analysis_and_optimization/Mir_utils.ml
```diff
@@ -20,57 +17,46 @@ let map_rec_expr_state
...
 let map_rec_state_stmt_loc_num
-    (flowgraph_to_mir : (int, stmt_loc_num) Map.Poly.t)
+    (flowgraph_to_mir : (int, Stmt.Located.Non_recursive.t) Map.Poly.t)
     (f :
          int
       -> 's
-      -> (expr_typed_located, stmt_loc) statement
-      -> (expr_typed_located, stmt_loc) statement * 's) (state : 's)
-    (s : stmt_loc_num) : stmt_loc * 's =
+      -> (Expr.Typed.t, Stmt.Located.t) Stmt.Fixed.Pattern.t
```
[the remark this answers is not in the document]
enetsee: No; it's the pattern of a `Stmt.Located` which doesn't have its own name.

src/analysis_and_optimization/Mir_utils.ml
```diff
@@ -81,86 +67,92 @@ let map_rec_state_stmt_loc_num
   let state = !cur_state in
   (stmt, state)
 
-let stmt_loc_of_stmt_loc_num
-    (flowgraph_to_mir : (int, stmt_loc_num) Map.Poly.t) (s : stmt_loc_num) =
+let stmt_loc_of_stmt_loc_num flowgraph_to_mir s =
+  (* (flowgraph_to_mir : (int, stmt_loc_num) Map.Poly.t) (s : stmt_loc_num) = *)
```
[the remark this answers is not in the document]
enetsee: Remnants of the refactor; I'll delete.

src/analysis_and_optimization/Mir_utils.ml
```diff
@@ -81,86 +67,92 @@ let map_rec_state_stmt_loc_num
...
-  | TargetPE _ as s -> (state, s)
-  | NRFunApp _ as s -> (state, s)
-  | Break as s -> (state, s)
-  | Continue as s -> (state, s)
-  | Return _ as s -> (state, s)
-  | Skip as s -> (state, s)
-  | Decl _ as s -> (state, s)
+let fwd_traverse_statement stmt ~init ~f =
+  Stmt.Fixed.Pattern.(
+    (* (stmt : ('e, 'a) statement) ~init:(state : 'f)
```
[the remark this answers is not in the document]
enetsee: As above.

src/common/Common.mli
```diff
@@ -0,0 +1,16 @@
+(* Type-class like signatures *)
```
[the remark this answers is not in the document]
enetsee: I believe it is best practice to have *.mli files for everything and especially for top-level files in a library. Maybe we should ask a question on ocaml discuss about this?

src/common/Fixed.ml
enetsee: I haven't used ocamldoc before so I don't know I'm afraid. I'll take a look at the docs for it. Are there examples elsewhere in the code base of how it should be done?

src/common/Fixed.ml
enetsee: I'm not sure what you mean here - are you suggesting changing the name or getting rid of the fixed point type for statements?

src/common/Foldable.ml
enetsee: Not that I know of; I spent ages reading `Core` code and asked a couple of questions on OCaml discuss to work out what was going on initially!

src/common/Fixed.ml
enetsee: As with the `Meta` helpers, I would much prefer to hide the implementation of the `Fixed.t` so both are exposed. From my point of view, they are much lighter weight than record syntax.

src/common/Fixed.ml
enetsee: As above; `pattern_of` is already used a lot in the refactor and I used `unfix` a lot in the previous full refactor.

src/common/Foldable.ml
enetsee: I have no idea why this isn't in Core.

src/common/Foldable.ml
enetsee: I've only ever seen the accumulator come first in F#s `fold_back`. In every other language (including OCaml) the library implementation has it this way around e.g. https://ocaml.janestreet.com/ocaml-core/latest/doc/base/Base/List/#val-fold_right

src/frontend/Ast_to_Mir.ml
enetsee: It is considered best practice to open modules as close to call site as possible. Further, since all modules have a type `t` this would mean you would still need to qualify the types.

src/middle/Expr.ml
enetsee: As above; I would favor hiding the implementation but we expose both for now.

src/frontend/Ast_to_Mir.ml
enetsee: Again, I think hiding implementation and exposing helpers for constructing expressions and statements would make the code a lot easier to read.

src/frontend/Ast_to_Mir.ml
enetsee: Almost, it would be `Expr.{  Typed.meta = Meta.{type_;adlevel;loc}; pattern }`.

enetsee:
> PS I keep merging code gen fixes that give you merge conflicts. I'm happy to take responsibility for that and go through and resolve them before we end up merging.

No worries, I know you were targeting a release!

That being said, it has taken me nearly 2 hours to fix the merge conflicts (since changes to middle affect all parts of the compiler) so it would be great to get this in before other PRs mean I have to do it all again 😢 .

I wasn't sure what you were suggesting about `Fixed.S2` in your comments; could you clarify is you mean changing the name of the `First` module to something else or getting rid of it entirely? If you meant getting rid of it, what would you want to do for AST (and any other IRs we end up with)?

src/common/Fixed.ml
enetsee: I've just checked and I think they conform to ocamldoc conventions. I think ocamlformat enforces this (it certainly issues a warning about ambiguous associations of comments).

src/middle/Expr.ml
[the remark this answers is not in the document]
enetsee: Updated to `expr` and `stmt`.

src/analysis_and_optimization/Dataflow_utils.ml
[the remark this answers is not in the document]
seantalts: Would you be open to switching to `let open Stmt.Fixed.Pattern in` for these? That way we save the indent.

enetsee:
>In Core_kernel, fold_right's function argument has the accumulator second, but their fold_left (and ppx_deriving's fold, which I think is a fold left) has it come first. Can we follow the convention that it comes first for left folds and second for right folds?

This is already what they are  (see [here](https://github.com/stan-dev/stanc3/blob/patterns-minimal/src/common/Foldable.ml#L36), and signatures below) or do I have my wires crossed? 

```ocaml
  val fold_left : f:('b -> 'a -> 'b) -> init:'b -> 'a t -> 'b
  (** Left associative fold of a data structure; this is the same as the 
  function derived from `[@@deriving fold]` but with labelled arguments *)

  val fold_right : f:('a -> 'b -> 'b) -> init:'b -> 'a t -> 'b
  (** Right associative fold of a data structure *)
```

enetsee:
> Was talking about First as a name. I think it makes sense but what do you think about using the name Expr or something concrete so people can follow along more easily? I don't feel strongly about this.

I think the current name `First` is appropriate and that `Expr` would be a little incongruous. Apologies but I'd really prefer not to bikeshed too much more on names at this point.

enetsee:
> I feel strongly that we shouldn't be printing empty blocks in the MIR pretty printer - as the person who works on that the most, it's a lot easier to test and debug the MIR when it's just printing out the actual information in it. And you know if something doesn't print that it's empty, so it's not like the pretty printer is in any sense "incomplete" without doing that.

Sure; happy to change.

enetsee:
> Would you consider switching some of the local opens that you've done in the Stmt.Fixed.( style to let open Stmt.Fixed in instead? That way we save indent and I think it's otherwise just as clear. I don't feel that strongly about this but in general would like us to take tiny steps towards saving screen real estate like this where it's a cheap win.

As I mentioned on the call, I'd rather not have to make further changes which amount to preference but can make the change if this is really a sticking point for the PR.

enetsee:
> Having both the record syntax and the record label helpers is the worst of both worlds, n my opinion :P

Agreed! 

> i I feel strongly that we should just go with one of those options. I think it would be the minimal change to just do records for now and evaluate going forwards what our needs are. 

I have made use of `pattern_of`, `type_of` and `adlevel_of` in _many_ places, but I suppose it is a minimal change from one perspective 😝. 

> Records are a good choice because they are the exact OCaml construct designed to allow you to name different parts of a data type and as such they define their own helper functions for getting and setting in an immutable way. We made a decision a while ago to use records for this rather than creating our own helper functions with named arguments. 

My perspective here is that it should be completely _irrelevant_ whether you implement the _fixed point type_ using a record, a single case union type or even non-recursively (like `stmt_loc_num`) - it is (literally!) an implementation detail, it's the _pattern_ that conveys meaning. 

By hiding the implementation we get a single consistent API for working with `Fixed` types. We are also free to change the implementation in the future without impacting every single part of the code base. 

In the current PR, a huge number of changes were made because the fields of the fixed point record are now consistently `pattern` and `meta` (rather than `smeta`, `emeta`, `expr` and `stmt`). The implementation has also leaked into our tests since we are using the s-expression representation to test for equality so the field names all had to be changed (although it is easy to do so with `dune promote`, it always makes me feel uneasy comparing string representations of anything!).

>  I don't think either syntax is inherently more readable (do you have some examples where you think named arguments and helper functions are more readable?) so if that's the case we should just go with whichever is fewer lines and easier on maintenance, which I think would be records.

There are several examples where you have used helpers in the `stan_math_backend` code so I'd ask you the same question, do you think those helper functions are more readable than a record literal?

For my part, I think the following is a pretty clear example of why we may want helpers:

original, from `Partial_evaluator`

```
...
            | ( "Times__"
              , [ { expr=
                      FunApp
                        ( StanLib
                        , "matrix_exp"
                        , [{expr= FunApp (StanLib, "Times__", [t; a]); _}] ); _
                  }
                ; b ] )
              when t.emeta.mtype = UInt || t.emeta.mtype = UReal ->
                FunApp (StanLib, "scale_matrix_exp_multiply", [t; a; b])
            | ( "Times__"
              , [ { expr=
                      FunApp
                        ( StanLib
                        , "matrix_exp"
                        , [{expr= FunApp (StanLib, "Times__", [a; t]); _}] ); _
                  }
                ; b ] )
              when t.emeta.mtype = UInt || t.emeta.mtype = UReal ->
                FunApp (StanLib, "scale_matrix_exp_multiply", [t; a; b])
            | "Times__", [{expr= FunApp (StanLib, "matrix_exp", [a]); _}; b] ->
                FunApp (StanLib, "matrix_exp_multiply", [a; b])
```

Alternative from original patterns PR:

```ocaml
let times_scale_matrix_exp_multiply ~type_of meta a b =
  match Fixed.(pattern2 a, pattern2 b) with
  | FunApp (StanLib, "matrix_exp", [FunApp (StanLib, "Times__", [x; y])]), _
    when type_of x = UInt || type_of x = UReal ->
      Some (stanlib_fun meta "scale_matrix_exp_multiply" [x; y; b])
  | _, FunApp (StanLib, "matrix_exp", [FunApp (StanLib, "Times__", [x; y])])
    when type_of x = UInt || type_of x = UReal ->
      Some (stanlib_fun meta "scale_matrix_exp_multiply" [x; y; a])
  | _ -> None
```
(Note that the helper function also means we can operate on `Expr.Numbered` as well as `Expr.Typed`.)

I agree that this is a bigger design issue and I'll remove the helpers for the PR, but it would be great if we could revisit this.

seantalts:
You're so right about the folds already following that convention, my bad! One thing I'm confused about but is probably fine - in your S2 folds, you have the two functions share an accumulator type, while the signatures for the ppx_deriving folds do not share one (due to operating on the two parts of the fixed point). Not sure the impact of that.

Re: helpers - I don't I just meant the ones that purely replicate record syntax - so that's ones like `adlevel` that given a meta data record return the `adlevel` of it. Whereas your other helpers like `type_of e` that dive into the expression's metadata record to return `type_` still seem like useful helpers. Does that clarify a little? Basically my perspective here is that all of the stuff we get for free from records it seems silly to duplicate, but anything that's even as complex as composing two record getters together could warrant a helper.

 > My perspective here is that it should be completely irrelevant whether you implement the fixed point type using a record, a single case union type or even non-recursively (like stmt_loc_num) - it is (literally!) an implementation detail, it's the pattern that conveys meaning.

> By hiding the implementation we get a single consistent API for working with Fixed types. We are also free to change the implementation in the future without impacting every single part of the code base.

I get the encapsulation mechanism but I have no idea when or why we would want to use it - when would we need to care and change implementations from record to tuple/variant/whatever implementation? That lack of imagination might be my own inexperience with large OCaml projects, but I'm also sympathetic to the idea that we might never need that facility and there's little value in doing extra work to ensure an unlikely future hypothetical.

Agreed that in this change when we changed all of the labels a lot of the code and tests changed. But that would be true again if we changed from "pattern" to "object" or whatever 3rd alternative naming convention.

seantalts:
That partial evaluator could be rewritten to be a lot more concise in a few different ways even in the current master branch. For example, I think it's probably more clear to actually encode the UReal | UInt thing in the pattern match itself like `| {expr=FunApp(_, "matrix_expr", [{emeta={mtype=UInt | UReal; _}; _}, y]; _}`. 

The code that you pasted replacing the original partial evaluator code doesn't seem semantically equivalent, is it? The original code checks that the scalar argument to "Times__" can appear in either position and I don't see how that's happening in your replacement...

seantalts:
Also I want to push back against the idea that discussing names is bike-shedding. In this instance I'm fine to go with the semantically-void "First" name here, but any decent code review should be looking at naming and highlighting anything where the names could be more clear. After all, [the only hard problems in computer science are naming and cache invalidation.](https://martinfowler.com/bliki/TwoHardThings.html). We're optimizing this code to be read for the first time by newcomers as much as possible.

I think code style is also important for a project, but it's critical to have that be well-defined. So for this project we're aiming to use ocamlformat and the [Jane Street style guide](https://opensource.janestreet.com/standards/). They prefer the explicitly scoped open in most cases, unless the module has a "standard and widely understood API" (like Monad_infix for e.g. your Validation module). Are you okay with agreeing to follow their style guide on this project? I can agree to try to avoid commenting on any minor style stuff that isn't present in that guide.

seantalts:
And JTBC I think following Jane Street's style guide here means keeping most (all?) of your scoped opens.

enetsee:
> Re: helpers - I just meant the ones that purely replicate record syntax - so that's ones like adlevel that given a meta data record return the adlevel of it. Whereas your other helpers like type_of e that dive into the expression's metadata record to return type_ still seem like useful helpers. Does that clarify a little? Basically my perspective here is that all of the stuff we get for free from records it seems silly to duplicate, but anything that's even as complex as composing two record getters together could warrant a helper.

Happy to remove the accessors for meta-data. These were originally exposed since the type of meta data was abstract.

enetsee:
> Also I want to push back against the idea that discussing names is bike-shedding. In this instance I'm fine to go with the semantically-void "First" name here, but any decent code review should be looking at naming and highlighting anything where the names could be more clear.

Absolutely; my point wasn't about naming in general but on the importance of changing the name of this type relative to me being unblocked on this. 

This type has been named `First` since the original `patterns` pull request and I just think at some point we have to pull the trigger on this.

enetsee:
> I think code style is also important for a project, but it's critical to have that be well-defined. So for this project we're aiming to use ocamlformat and the Jane Street style guide. They prefer your explicitly scoped open in most cases, unless the module has a "standard and widely understood API" (like Monad_infix for e.g. your Validation module). Are you okay with agreeing to follow their style guide on this project? I can agree to try to avoid commenting on any minor style stuff that isn't present in that guide.

I think this is a great idea not least because people joining the project should be familiar with the style if they have read any Jane Street code.

seantalts:
Omg you’re right again about fold. I think your emails are the first that I
see when I wake up and I need to give myself more time with caffeine before
hitting send! Sorry about that 😂

seantalts:
> This type has been named First since the original patterns pull request and I just think at some point we have to pull the trigger on this.

Then this is my fault - I hadn't been doing a complete review, figuring it was best to get high level details figured out first. That has the effect of focusing the conversation on the more important bits and saving some effort but also drawing out the process. Next time I can try to do a complete review each pass and see how that goes.

seantalts:
> Happy to remove the accessors for meta-data. These were originally exposed since the type of meta data was abstract.

I think the accessors for e.g. `pattern_of` would also qualify as duplicating what record syntax provides. Would you be okay taking those out too (at least for now, pending some larger discussion around why we would want to make these particular types abstract and hidden - to me those feels like another Java-ism from like 15 years ago that has since been replaced with immutable fields in the common wisdom).

enetsee:
Helpers removed and empty MIR blocks no longer printed. I think that's all the changes?

seantalts:
Did you double check the partial evaluator? I'll do another sweep as well - a bunch of that stuff isn't super well tested automatically.

enetsee:
> Did you double check the partial evaluator?

I have only changed type names within the `analysis_and_optimization` library; the example I posted was from the original `patterns` branch.

src/analysis_and_optimization/Dependence_analysis.ml
```diff
@@ -261,7 +261,9 @@ let log_prob_build_dep_info_map (mir : Program.Typed.t) :
     Stmt.Fixed.{meta= Location_span.empty; pattern= SList mir.log_prob}
   in
   let statement_map =
-    build_statement_map Stmt.Fixed.pattern_of Stmt.Fixed.pattern_of
+    build_statement_map
+      (fun Stmt.Fixed.({pattern; _}) -> pattern)
+      (fun Stmt.Fixed.({meta; _}) -> meta)
```
seantalts: Ah, this is an annoying thing about records :(

seantalts:
Ahh okay. Thanks for all this hard work! Merging :) I owe you several beers :P

enetsee:
Thank _you_! Now on to the AST...

PR #348, opened 2019-10-16
Deal with ints the same way Stan 2 does in function signatures

### LINKED ISSUE #343: generation of non-compilable C++
The attached Stan program is a legitimate Stan2 program which just compiles to C++ and then to a binary; running just fine with 2.20. With stanc3 a C++ code is generated which cannot be compiled.

This is a slightly modified code from the overpropagation to var issue, but this time focusing on a different aspect.

[oral_1cmt_mm_run_stan.txt](https://github.com/stan-dev/stanc3/files/3727641/oral_1cmt_mm_run_stan.txt)

In case you want to run the model, just take the data and seed from the issue #336

seantalts:
Assumes we'll never want to autodiff something declared as an integer, which might make sense. I wasn't able to see a decent quick way of enabling autodiff over ints while preserving the old behavior. Also doesn't fill ints with NaN now. 

Fixes #343

### DIFF
--- a/src/frontend/Ast_to_Mir.ml
+++ b/src/frontend/Ast_to_Mir.ml
@@ -457,53 +457,32 @@ let rec trans_stmt udf_names (declc : decl_context) (ts : Ast.typed_statement)
         ; body }
       |> swrap
   | Ast.ForEach (loopvar, iteratee, body) ->
-      let newsym = gensym () in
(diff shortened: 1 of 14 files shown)


bob-carpenter:
That's right---we can't autodiff through integer arguments because we need continuity for derivatives to be defined.

PR #349, opened 2019-10-17
 Semantic check warnings #336

### LINKED ISSUE #336: stanc3 checks `data` keywords and doesn't allow potential non-data to be passed in.
When calling the `integrate_ode` functions inside other functions, then this is prone to over-propagation from data to var for the data arguments. With the old compiler it was possible to sub-select from bigger data structures smaller ones and these then still stay as data (as it should be). The new compiler does not do that. The new compiler does "stronger" over-propagation from data to var which makes the programs fail to compile due to a syntax error.

Attached is a Stan program example which illustrates the problem.

The error message is:

```
Semantic error in '../stan_pkpd/oral_1cmt_mm/oral_1cmt_mm_run.stan', line 868, column 14 to column 99:
   866:      theta_tilde[P+1] = lref[1];
   867:      lref_tilde[1] = lref[2];
   868:      int_sol = integrate_ode(pk_1cmt_mm_lode, lref_tilde, 0, to_array_1d(Dt), theta_tilde, x_r, x_i);//, 1e-4, 1e-4, 1000);
                       ^
   869:      ka = theta[1];
   870:      for(i in 1:num_elements(Dt)) {

Ill-typed arguments supplied to function 'integrate_ode'. Available signatures:
((real, real[], real[], data real[], data int[]) => real[], real[], real, real[], real[], data real[], data int[]) => real[,]
Instead supplied arguments of incompatible type: (real, real[], real[], real[], int[]) => real[], real[], int, real[], real[], real[], int[].
```

I really think this should be fixed ... if possible before the 2.21 release.

[stanc3-overpropagate-error.zip](https://github.com/stan-dev/stanc3/files/3723633/stanc3-overpropagate-error.zip)

enetsee:
Fixes #336. @seantalts @wds15 @bob-carpenter - you may want to set out some additional tests for this based on the failing code.

### DIFF
--- a/src/frontend/Errors.ml
+++ b/src/frontend/Errors.ml
@@ -134,6 +134,11 @@ let report_semantic_error (message, loc_span) =
     (string_of_location_span loc_span) ;
   print_context_and_message message loc_span.begin_loc
 
+let report_semantic_warning (message, loc_span) =
+  Printf.eprintf "\nWarning: %s in %s:\n" "Semantic warning"
+    (string_of_location_span loc_span) ;
+  print_context_and_message message loc_span.begin_loc
+
 (** Switch to control whether warning messages should be printed to stderr (or discarded in case set to false) *)
 let print_warnings = ref true
 
--- a/src/frontend/Errors.mli
+++ b/src/frontend/Errors.mli
@@ -35,6 +35,8 @@ val report_parsing_error : string * location_span -> unit
 val report_semantic_error : string * location_span -> unit
 (** A semantic error message used when handling a SemanticError *)
 
+val report_semantic_warning : string * location_span -> unit
+
 val warn_deprecated : Lexing.position * string -> unit
 (** Warn that a language construct is deprecated *)
 
--- a/src/frontend/Frontend_utils.ml
+++ b/src/frontend/Frontend_utils.ml
@@ -4,6 +4,15 @@ let typed_ast_of_string_exn s =
   Parse.parse_string Parser.Incremental.program s
   |> Result.map_error ~f:Parse.render_syntax_error
   |> Result.ok_or_failwith |> Semantic_check.semantic_check_program
+  |> Result.map ~f:(fun (result, warnings) ->
+         ( result
+         , Fmt.(to_to_string @@ list ~sep:cut Semantic_warning.pp) warnings )
+     )
   |> Result.map_error
-       ~f:(Fmt.to_to_string (Fmt.list ~sep:Fmt.cut Semantic_error.pp))
+       ~f:
+         Fmt.(
+           to_to_string
+           @@ pair ~sep:cut
+                (list ~sep:cut Semantic_error.pp)
+                (list ~sep:cut Semantic_warning.pp))
   |> Result.ok_or_failwith
--- a/src/frontend/Semantic_check.ml
+++ b/src/frontend/Semantic_check.ml
@@ -8,7 +8,7 @@ open Symbol_table
 open Middle
 open Ast
 open Errors
-module Validate = Validation.Make (Semantic_error)
+module Validate = Validation.Make (Semantic_warning) (Semantic_error)
 
 (* There is a semantic checking function for each AST node that calls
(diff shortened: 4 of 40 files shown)


wds15:
Can I test this PR easily? I mean... is there a stanc3 binary I can grab from somewhere (for macOS or Linux)...

rok-cesnovar:
If you want I can send you the binary to your email or upload it somewhere. I just switch the branch and compile.

enetsee:
> Can I test this PR easily?

It would be good to have another example in the test suite. Would you be able to provide one the Stan models on which stanc3 was failing?

wds15:
No. It doesn't work yet - possibly I have been unclear or we have different notions.

The attached model should not error, but just compile (with a warning).

[oral_1cmt_mm_run_stan.txt](https://github.com/stan-dev/stanc3/files/3738904/oral_1cmt_mm_run_stan.txt)

Right now it errors at the integrate_ode call.

enetsee:
@wds15 - ok, I think we may have our wires crossed. This PR relaxes the requirement that arguments provided to user defined functions, not Stan math functions like `integrate_ode`. We will need to relax the requirement there too for this to work.

@seantalts I can do this but longer term we should really think if we want to allow this in the type and effect system. Maybe we should start grouping warnings for deprecations etc so users have fair warning if/when we turn this off?

enetsee:
I've just pushed a change which applies to stan math functions. I had to correct a bunch of models which were missing a data declaration so someone may want to take a quick look at them:

- test/integration/bad/ode/adams/bad_x_var_type.stan
- test/integration/bad/ode/adams/bad_x_var_type_adams_control.stan
- test/integration/bad/ode/adams/bad_x_var_type_control.stan
- test/integration/bad/ode/bad_x_var_type.stan
- test/integration/bad/ode/bad_x_var_type_bdf.stan
- test/integration/bad/ode/bad_x_var_type_bdf_control.stan
- test/integration/bad/ode/bad_x_var_type_rk45.stan
- test/integration/bad/ode/bad_x_var_type_rk45_control.stan

enetsee:
@wds15 I've included your model in the integration test set. It now compiles with the following warning:

> |Warning: Semantic warning in 'ode_sebastian.stan', line 868, column 14 to column 99:
|   -------------------------------------------------
|   866:      theta_tilde[P+1] = lref[1];
|   867:      lref_tilde[1] = lref[2];
|   868:      int_sol = integrate_ode(pk_1cmt_mm_lode, lref_tilde, 0, to_array_1d(Dt), theta_tilde, x_r, x_i);//, 1e-4, 1e-4, 1000);
|                       ^
|   869:      ka = theta[1];
|   870:      for(i in 1:num_elements(Dt)) {
|   -------------------------------------------------
|
|Warning: Argument to 'integrate_ode' has an incompatible autodiff level.

seantalts:
It looks like we're not going to get stanc3 into the next release; a small minority of people at the Stan meeting were really adamant that we not release something with known bugs (which doesn't make sense w.r.t. how we release the rest of Stan, but whatever). There was also a surprising lack of familiarity with how semantic versioning, which has been used on this project I believe since its inception, actually works. I think I could have done a better job explaining where we are with stanc3 relative to the old compiler, but the damage seems done for 2.21. Sorry about that.

Anyway, could you make the warning say which arguments are incorrect and notify them that this could cause the C++ compiler to fail? 

> @seantalts I can do this but longer term we should really think if we want to allow this in the type and effect system. Maybe we should start grouping warnings for deprecations etc so users have fair warning if/when we turn this off?

Which thing is "this" referring to, in "if we want to allow this?" We definitely have a bunch of deprecated behavior we want to be warning about and be ready to remove in Stan 3, so coming up with some facility to handle that seems very prudent.

seantalts:
That said, I really appreciate you jumping in to try to enable this use-case last second. Thanks @enetsee  :)

enetsee:
Closing; the work on validation is in #357 and I think we may want to think more about our approach to semantic warnings.

PR #350, opened 2019-10-19
Fix expression gen for ceil and atan2 functions

seantalts:


seantalts:
@bob-carpenter pretty curious about the atan2 one - you had it called out specifically as the testcase for a function that required the `stan::math::` prefix, but it doesn't compile with that. Just want to run this by you and make sure it makes sense to you.

PR #357, opened 2019-10-24
Validation warnings

enetsee:
This adds the ability to generate semantic warnings to the existing `Validation` applicative. Validation now results in either:

- success: a typed program and a list or semantic warnings; or
- failure : a non-empty list of semantic errors and a list of semantic warnings.

`Validation` also now uses `Core_kernels` `Monad` and `Applicative` definitions and abstracts over modules rather than types.


enetsee:
This is building and passing all tests locally and I can't figure out what's going on with jenkins. @seantalts can you tell why this is failing?

seantalts:
@serban-nicusor-toptal should be on it soon. The same error is preventing master from building releases, which is critical as we need that to be working in order to put stanc3 back on CmdStan develop for more testing before the next release.

src/common/NonEmptyList.ml
```diff
@@ -0,0 +1,20 @@
+open Core_kernel
+
+type 'a t = NonEmpty of 'a * 'a list
```
seantalts: To quote Alan Perlis, "It is better to have 100 functions operate on one data structure than to have 10 functions operate on 10 data structures."

Can we stick with list? I don't think the non-empty guarantee is worth writing another list library.

test/unit/Dataflow_utils.ml
```diff
@@ -420,7 +420,7 @@ let example4_program =
       }
       |}
   in
-  let mir = Ast_to_Mir.trans_prog "" (semantic_check_program ast) in
+  let mir = Ast_to_Mir.trans_prog "" (fst @@ semantic_check_program ast) in
```
seantalts: We should write a new function that exposes the additional API and leave the old one alone instead of adding `fst` to everything.

seantalts:
My friend pointed out that this is a good improvement over what we have and said I should suck it up. Can you write up some doc about how to use this stuff? Similar to the doc we need around the new MIR refactor, with some overlap around e.g. how functors work and naming conventions there but also could explain bits of the code that are frightening to look at. Here's an example bit of doc I put together for the format library recently: https://github.com/stan-dev/stanc3/wiki/Format---Fmt-code-generation-quickstart

I think we need the doc anyway (it can be in the code or on the wiki, or whatever), and it could be a good opportunity to dogfood the doc and see if I can learn how to use the Validate functors + applicatives enough to code review them :P Given that our target audience for this code is mostly C++ and Stan developers, a very minimal bar would be to make sure that I can understand it quickly as the stanc3 person who is least familiar with this stuff. Though given that I have years of functional programming under my belt already, I'm sure we'll need to amp up documentation efforts especially as we get more and more into pure category theory and less and less towards concepts that are familiar to C++ and Stan developers.

test/unit/Dataflow_utils.ml
enetsee: How about the tests to expect no warnings; semantic check always returns a (possibly empty) list of warnings with either success (a typed AST) or failure (a non-empty list of errors) so it makes sense to test that.

src/common/NonEmptyList.ml
enetsee: Non-empty list has always been used in validation, I just pulled it out to a separate module. `NonEmptyList` is also a private module of common so isn't exposed. That said, the API to `Validation` guarantees you can't get to an error state without at least one error and we always convert to list so it isn't adding a lot.

enetsee:
> how functors work and naming conventions there but also could explain bits of the code that are frightening to look at

I've got some doc's on naming conventions in `Foldable`; how about moving that that along with some stuff lifted from [real world ocaml](https://dev.realworldocaml.org/functors.html)?

> I'm sure we'll need to amp up documentation efforts especially as we get more and more into pure category theory and less and less towards concepts that are familiar to C++ and Stan developers.

You want to write a monad tutorial 😱 ? I recommended [this artical](https://fsharpforfunandprofit.com/posts/recipe-part2/) to Bob a while back which I think he found useful for understanding why monadic and applicative interfaces end up being so useful.

I invited Scott Wlaschin (the author of the site) to give a 3 day tutorial on functional programming to my team at work and most of the team were at least comfortable using these interfaces for standard things like validation etc after that. 

His advice for introducing this to teams is to start with the concrete and then show what they have in common so maybe prefacing the existing `Applicative` inline docs with some examples of using the API would be a good way to explain what's going on.

src/common/NonEmptyList.ml
enetsee: Removed `NonEmptyList`

test/unit/Dataflow_utils.ml
enetsee: Added explicit expect tests for warnings

src/common/Validation.ml
```diff
@@ -63,7 +58,6 @@ end
   let apply x ~f =
```
seantalts: (this comment isn't in quite in the right place, but doc above refers to `NonEmpty` list still).

Seems like there are a lot of ways you could encode the results of type checking. One way would be that we split Semantic_check into a type inference tree manipulation and a separate type checking phase that returns a possibly-empty list of `Either` Warnings or Errors; another might be to keep the phases intertwined but still use a single possibly empty list to represent warnings and error accumulation.

Can you tell me a little bit about how you thought about this design space?

src/common/Validation.ml
```diff
@@ -63,7 +58,6 @@ end
   let apply x ~f = 
     match (f,x) with
   ```
-
   `apply` 'unwraps' the function and the value and performs different actions 
   depending on whether an error has occured. Because we have two arguments each 
   of which can be in two states, we have to check for four conditions:
```
seantalts: It says here (and in the railway oriented programming blog post) that we generally want to exit early if we hit a failure. How does this pattern change to account for collecting all errors and warnings?

src/common/Validation.ml
```diff
@@ -110,58 +104,71 @@ end
...
-  let is_error = function Error _ -> true | _ -> false
-  let is_success = function Ok _ -> true | _ -> false
+  let warn ~warn x : 'a t = Result.Ok (x, [warn])
+  let is_error = function Result.Error _ -> true | _ -> false
+  let is_warning = function Result.Ok (_, _ :: _) -> true | _ -> false
+  let is_success = function Result.Ok (_, []) -> true | _ -> false
 
   let get_errors_opt = function
-    | Error (NonEmpty (x, xs)) -> Some (x :: xs)
+    | Result.Error (es, ws) -> Some (es, ws)
```
seantalts: What's the use of these custom accessor functions that stac the `option` type on here? A List can be empty and has its own library of functions for accessing various parts of it.

src/frontend/Semantic_check.ml
```diff
@@ -374,11 +374,21 @@ let semantic_check_binop loc op (le, re) =
...
+         ( result
+         , Fmt.(to_to_string @@ list ~sep:cut Semantic_warning.pp) warnings )
+     )
+  |> Result.map_error
+       ~f:
+         Fmt.(
+           to_to_string
+           @@ pair ~sep:cut
+                (list ~sep:cut Semantic_error.pp)
+                (list ~sep:cut Semantic_warning.pp))
```
seantalts: I think maybe the `|> fst` should go inside this `to_exn` function rather than at its two use sites?

src/frontend/Semantic_warning.ml
```diff
@@ -0,0 +1,18 @@
+open Middle
+
+type t =
+  | WarnAutodiffLevel of
```
seantalts: 👏

test/unit/Dataflow_utils.ml
```diff
@@ -119,7 +123,11 @@ let%expect_test "Loop passthrough" =
         }
       |}
   in
-  let mir = Ast_to_Mir.trans_prog "" (semantic_check_program ast) in
+  let typed_ast, warnings = semantic_check_program ast in
```
seantalts: Would you mind making a helper function to pull this out? Something very close to this stanza is repeated a lot in these test files.

src/common/Validation.ml
enetsee: I don't think this is needed at all; we just need `to_result` I think.

src/common/Validation.ml
enetsee: Type checking is inherently sequential e.g. we can't determine if function application is well typed until we have type checked each argument. We have to check types in sequence, from the bottom up.

Unless we do something a little bit fancy with error recovery, the type checker would stop as soon as it encountered the first error since it wouldn't have the necessary information to type check subsequent statements. So given e.g. a `Expr.Untyped.t` type checking would give us back a `(Expr.Typed.t,TypeError.t) Result.t`.

(If we _did_ implement error recovery, then we would always get a `Expr.Typed.t` but possibly a list of `TypeError.t`s.)

The point is that type checking will always need a monadic interface (whether implemented as a monad or not!) since it is sequential. Also, type checking moves us from `Untyped.t` to `Typed.t`, i.e. it decorates a piece of IR.

In contrast, semantic checking _should_ only capture the things we cannot check within the type system and may (currently does) depend on type information. At the moment (without an effect system) the things we can't verify with the type checker include things like illegal use of `TargetPlusEquals` outside the model block. 

The key thing here is that these semantic errors do not depend on one another - it is always illegal to place `TargetPlusEquals` in illegal places irrespective of other semantic errors. Semantic checking does not change the type of the intermediate representation.

Because one semantic check does not depend on a preceding one it only requires an applicative interface, so we can check for all errors in parallel. The semantic check should have the type `Expr.Typed.t -> (SemanticWarning.t list , (SemanticError.t * SemanticWarning.t) list) Result.t`.

If we were to introduce a simple effect system I think the only semantic errors we would check for would be 'identifier errors' and a couple of other things relating to duplicate names some of  which we could possibly do in the parser(?). 

Without an effect system, the additional errors would be around effects:

```ocaml
    | CannotAssignToReadOnly of string
    | CannotAssignToGlobal of string
    | ...
    | TargetPlusEqualsOutsideModelOrLogProb
```

(The first two reflect a function with an `Assign` effect and the last a `Target` effect. By checking different program blocks against sets of effects we can catch them as type errors.)

Either way I think separating type checking would be a good idea. Type systems can be hard enough to follow without mixing in unrelated logic.

src/common/Validation.ml
enetsee: > that we generally want to exit early if we hit a failure

That's the case with the _monadic_ interface to `Result.t` - as soon as we hit an error case we know that subsequent computations that depend on it cannot be `Ok`. You can see this from the type of `bind`:

```ocaml
val bind : `a result -> f:('a -> 'b result) -> 'b result
``` 
The second function argument (representing the continuation of a computation) takes an `'a` and the only way it can get is the first argument is `Ok` and contains an '`a'. If it's an `Error` the computation cannot proceed*.

However, ultimately we want `Validation` to expose only an `Applicative` interface (see previous reply). Importantly the implementation of `apply` for `Validation` is _not_ the same as `Result`. For a `Result.t` type we don't know what to do if we have two `Error`s but in `Validation.t` we know we can combine the data in the `Error` case. So, for `Validation`, given an `Applicative` interface, we still want to proceed with the other 'parallel' computations.

(* I have seen this short-circuiting implemented with exceptions in OCaml but more usually you just define `bind` such that it performs the redundant computation.)

src/common/Validation.ml
```diff
@@ -110,58 +104,71 @@ end
...
 
-  let append xs ys =
-    match (xs, ys) with
-    | NonEmpty (x, []), NonEmpty (y, []) -> NonEmpty (x, [y])
-    | NonEmpty (x, xs), NonEmpty (y, ys) -> NonEmpty (x, xs @ (y :: ys))
+    let apply f x =
+      match (f, x) with
+      | Result.Ok (f, ws), Result.Ok (x, vs) -> Result.Ok (f x, ws @ vs)
+      | Ok (_, ws), Error (e, vs) -> Error (e, ws @ vs)
+      | Error (e, ws), Ok (_, vs) -> Error (e, ws @ vs)
```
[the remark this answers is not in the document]
enetsee: Yes we can!

src/common/Validation.ml
seantalts: Thank you very much for writing this up! I had been under the impression that we could separate out a non-checking inference stage that returns a new tree, even if there were type errors. This could be a stretch goal one day as it might better support language servers as we could code inference such that it would essentially just use default logic to construct the tree - e.g. if two types don't line up, we leave those in the metadata incorrectly but then the parent expression appears well-typed from above. So inference is still monadic but perhaps checking becomes applicative and we can collect multiple type errors? Though maybe separating it out at that level would duplicate a ton of code. 

So I was wondering still about using basically `(Warning.t, Error.t) Either.t list` as the monad, rather than a custom one. Any thoughts about that?

seantalts: Sorry, need to think about the last question more... don't bother answering yet!

seantalts: Hmm okay I think I understand more and my design question translates to something like, why isn't `'a Validate.t = 'a * (Warning.t, Error.t) Either.t list)`? I see why, given that semantic checking is very entangled with inference now, that you need to always return a thing as well as the errors.  

Like not to get too Haskell-monad-transformers on you but we make a module that holds both some abstract type as well as some arbitrary monad and sort of does the obvious thing, piping the module's map to the underlying monad's map, etc. Does that make sense? 

Anyway I'd be happy to merge this once those three low level thingies are addressed, around pulling out some repeated functionality into a function and simplifying a bit

src/common/Validation.ml
seantalts: Gotcha. The fact that we both 1) need sequencing but 2) would like to not stop computation the second something in the sequence fails makes me wonder if we actually don't want either one of Monad or Applicative here. It seems like we basically want a Writer monad around a `(Warning.t * Error.t) list`, maybe?

seantalts:

Issue #342, opened 2019-10-14
Add warning about integer division


seantalts:
```
Info: integer division implicitly rounds to integer. Found int division: 2 * N / F
 Positive values rounded down, negative values rounded up or down in platform-dependent way.
```

adamhaber-atidot:
I'd be happy to try and tackle this if someone could provide initial guidance (where's the relevant bit of code, etc)

adamhaber:
I'd be happy to try and tackle this if someone could provide initial guidance (where's the relevant code, etc)

seantalts:
Hey! I'll have a lot more free time after the release on Friday but for now take a look at 
https://github.com/stan-dev/stanc3/blob/master/src/frontend/Semantic_check.ml

Especially the BinOp thing: https://github.com/stan-dev/stanc3/blob/master/src/frontend/Semantic_check.ml#L577

I hope @enetsee can explain the `liftA2 tuple2` part because don't know how that works. But basically if you can pattern match on a `Times` with an int and float (or float and int) and print a warning that should be okay.

Looks like you could fill in this: https://github.com/stan-dev/stanc3/blob/master/src/frontend/Semantic_check.ml#L577

adamhaber:
Just to be sure, you mean pattern match on `Divide` and not `Times`, right?

I'm trying to make something along these lines:
```ocaml
  | BinOp (e1, op, e2) -> 
    let le = semantic_check_expression cf e1
    and re = semantic_check_expression cf e2 
    and () = match (e1,op,e2) with
      | (_, Int), Divide, (_, Real) -> warn 
      | (_, Real), Divide, (_, Int) -> warn
      | _ -> () in
    Validate.(
      liftA2 tuple2 le re
      |> apply_const (semantic_check_operator op)
      >>= semantic_check_binop emeta.loc op)
```
I have two questions:
1. I'm still not able to get the types right. Both `e1,e2` are of type  `(location_meta, unit) expr_with`. How do I extract from this their numerical type? What I wrote above is just me sharing my failed attempt, it doesn't type-checks...
2. How "print a warning" should actually look like? I couldn't find the corresponding function... 

If this is an entirely wrong solution, let me know... :-)

enetsee:
Hey Adam,

> How "print a warning" should actually look like? I couldn't find the corresponding function...

Yeah, the mechanism to issue warnings with `Validate` isn't on master yet so this isn't possible at the moment.

[...]

PR #365, opened 2019-10-29
Add warning about integer division #342

adamhaber:
No idea why `test/unit/Optimize.ml` is changed. Is this a problem? `dune runtest` doesn't complain...


src/frontend/Semantic_check.ml
```diff
@@ -591,9 +591,22 @@ and semantic_check_expression cf ({emeta; expr} : Ast.untyped_expression) :
...
       let le = semantic_check_expression cf e1
-      and re = semantic_check_expression cf e2 in
+      and re = semantic_check_expression cf e2
+      and warn_int_division (x, y) =
+        match (x.emeta.type_, y.emeta.type_, op) with
+        | UInt, UReal, Divide | UReal, UInt, Divide ->
+            Fmt.pr
+              "Info: integer division implicitly rounds to integer. Found int \
+               division: %a / %a\n\
+              \      Positive values rounded down, negative values rounded up \
```
seantalts: May look nicer here to use OCaml's multi-line strings, which are anything put between `{|` and `|}`.

test/integration/bad/stanc.expected
```diff
@@ -1592,6 +1592,9 @@ Identifier expected after sized type in local (or model block) variable declarat
 
 
   $ ../../../../install/default/bin/stanc lp-error.stan
+Info: integer division implicitly rounds to integer. Found int division: 
+y ^ 2 / 2
+      Positive values rounded down, negative values rounded up or down in platform-dependent way. At 'lp-error.stan', line 5, column 16 to column 19
```
seantalts: Could we maybe move the location to the top line, e.g. "Found int division at %a:"

src/frontend/Semantic_check.ml
```diff
@@ -591,9 +591,22 @@ and semantic_check_expression cf ({emeta; expr} : Ast.untyped_expression) :
...
   | BinOp (e1, op, e2) ->
       let le = semantic_check_expression cf e1
-      and re = semantic_check_expression cf e2 in
+      and re = semantic_check_expression cf e2
+      and warn_int_division (x, y) =
+        match (x.emeta.type_, y.emeta.type_, op) with
+        | UInt, UReal, Divide | UReal, UInt, Divide ->
+            Fmt.pr
+              "Info: integer division implicitly rounds to integer. Found int \
+               division: %a / %a\n\
```
seantalts: I'd also put a solid line break before and after the code %a

test/integration/example-models/bugs_examples/vol1/dyes/pretty.expected
```diff
@@ -31,6 +31,13 @@ generated quantities {
   sigmasq_within <- 1 / tau_within;
 }
 
+Info: integer division implicitly rounds to integer. Found int division: 1 / sqrt(
+tau_between)
+      Positive values rounded down, negative values rounded up or down in platform-dependent way. At 'dyes.stan', line 21, column 19 to column 20Info: integer division implicitly rounds to integer. Found int division: 1 / sqrt(
+tau_within)
+      Positive values rounded down, negative values rounded up or down in platform-dependent way. At 'dyes.stan', line 22, column 18 to column 19Info: integer division implicitly rounds to integer. Found int division: 1 / tau_between
+      Positive values rounded down, negative values rounded up or down in platform-dependent way. At 'dyes.stan', line 39, column 21 to column 22Info: integer division implicitly rounds to integer. Found int division: 1 / tau_within
```
seantalts: Looks like we need a line break after the full message as well.

test/unit/Optimize.ml
```diff
@@ -2190,6 +2190,11 @@ model {
   Fmt.strf "@[<v>%a@]" Program.Typed.pp mir |> print_endline ;
   [%expect
     {|
+      Info: integer division implicitly rounds to integer. Found int division: 1 / 2.
```
seantalts: This is fine for now; an artifact of the fact that we're using the full parser in our unit tests here. This will go away when we switch to #357

adamhaber:
Thanks for the comments; it should look better now. Some expressions are broken to two lines for some reason that I don't understand, for example:
```
+|"Info: Found int division at 'validate_jacobian_warning_good.stan', line 57, column 2 to column 3:
+|1 / (
+|1 / z)
+|Positive values rounded down, negative values rounded up or down in platform-dependent way."
```

Also, in such cases:
```
+|"Info: Found int division at './/functions/mvmer_functions.stan', line 428, column 11, included from
+|'mvmer.stan', line 12, column 0:
+|mean_PPD / N
+|Positive values rounded down, negative values rounded up or down in platform-dependent way."
```
I assume this line break (in the first line) is because the line is too long?

seantalts:
Yeah, so when folks write pretty printers, they can (and should) insert "breaking spaces" fairly often at logical break points in their output so that if we need to go to a new line, we can. I think we're getting the weird behavior here either because we're 1) implicitly in a "vertical" box instead of the "horizontal-or-vertical" box we want to be in, or because 2) nothing that comes before on a very long line has a break in it. Notably, manual newlines are not counted as breaks but instead as another character worth one space, so I think we need to also go in and add breaks on the newlines in there. 

The second break in there seems ... fine? I could see arguments either way.

I'll take a quick pass over the formatter string and see if I can make those issues go away with the unfortunately large amount of experience I have working with it, then we can go over how those changes work (if I get something better working).

seantalts:
Tweaked it a little here:
https://github.com/stan-dev/stanc3/pull/365/commits/f7bfe3139e59336dbeb0055bea0c5d9936be2c5b#diff-5b95435ef954991e37fcffd291911d96L598-R603

Thought it looked kinda nice with the expression indented 2 lines, but don't feel strongly about that. Ended up switching away from the multi-line string because of how annoying it is mixed with the formatter, sorry about that.
Main thing was just adding a wrapping box and converting the multi-line string newlines into breaks.

seantalts:
I'll approve and you can decide which tweaks you like.

adamhaber:
This definitely looks much better, thanks!

PR #374, opened 2019-11-07
Add support for matrix_cl<int>

### LINKED ISSUE #344: Create a matrix_cl<int> for int and int[]
The current matrix_cl support in stanc3 always creates opencl variables as `matrix_cl<double>`. It should use `matrix_cl<int>` when the variable is an int scalar, int[] or int[ , ]. I missed this one when implementing the GPU GLM support. 

This should be fixed for the 2.21 release. I think it should be an easy fix, will try my best of solving it today.

rok-cesnovar:
Fixes #344 

int and int array will be represented with `matrix_cl<int>` as these are the only two "int" types that are used with GLMs.

Thanks for the help @seantalts!


src/stan_math_backend/Statement_gen.ml
```diff
@@ -48,7 +48,10 @@ let rec integer_el_type = function
 let pp_decl ppf (vident, ut, adtype) =
   let pp_type =
     if Transform_Mir.is_opencl_var vident then fun ppf _ ->
-      pf ppf "matrix_cl<double>"
+      match ut with
+      | UnsizedType.UInt ->  pf ppf "matrix_cl<int>"
+      | UArray UInt ->  pf ppf "matrix_cl<int>"
```
seantalts: Oops, this can actually be:

```ocaml
      match ut with
      | UnsizedType.UInt
      |UArray UInt ->  pf ppf "matrix_cl<int>"
```

rok-cesnovar: Fixed. Tnx

PR #375, opened 2019-11-07
Added 7 more bijectors

adamhaber:
Hi, 

1. The float helper function - I think it doesn't really matter eventually because we end up calling `tf.cast` anyway... but maybe it's still helpful
2. Testing - `dune runtest` doesn't really addresses these changes; I've implemented these transformations separately in TFP and made sure they indeed do as expected (bounds are as expected, etc). Hopefully I'll be able to add some TFP tests soon...
3. Offset and multiplier - I _think_ these are implemented in the right "direction" - in Stan Math their "forward" is `f(x) = mu + sigma * x`. Let me know if you think I get these wrong...


seantalts:
This looks great! Would you mind also adding a test model that uses these? It can be really simple but it would be good if we at least check that the right code is generated; running it is even better. You can see eight schools set up in `test/integration/tfp/`

adamhaber:
I saw the eight schools test but I'm not sure I understand what's what. I assume `eight_schools.py` was generated by stan2tfp? So this at least makes sure the code is generated... and then adding something like 
```python
    def test_eight_schools(self):
        target_dist = eight_schools_ncp_model(**eight_schools_data.data)
        mcmc_trace, _ = stan(target_dist)
        mu, tau, theta_tilde = [merge_chains(x) for x in mcmc_trace]

        self.assertAlmostEqual(significant_mean(mu), 4, delta=2)
        self.assertAlmostEqual(significant_mean(tau), 3, delta=2)
        self.assertAlmostEqual(significant_mean(theta_tilde), 0.08,
                               delta=0.1)
```
For the new model, where the numbers in the assert statements are from an equivalent PyStan run?

seantalts:
Yep, check out the `dune` file in that directory to see how eight schools is plumbed through. Should be able to copy and paste for a new model. And yeah, it's a bit hard to say exactly what numerical results we should test especially for totally made-up models, but let's hope it's near an equivalent PyStan run.

adamhaber:
I've tried to run this:
```stan
data {
         int<lower=0> N;
         int<lower=0,upper=1> y[N];
       }
       parameters {
         real<lower=0,upper=1> theta;
       }
       model {
         y ~ bernoulli(theta);
}
```
and got 
```
Uncaught exception:
  
  ("Not sure how to handle " bernoulli " yet!")
```
😄 
Currently only normal is supported, right?

Regarding the dune file (which I'm still not very good at reading)... This is what I have at the moment:
```
(rule
 (targets eight_schools.py.output)
 (deps (package stanc) (:stanfiles ../good/code-gen/eight_schools_ncp.stan))
 (action
  (with-stdout-to %{targets}
   (run %{bin:stan2tfp} %{stanfiles}))))

(alias
 (name runtest)
 (action (diff eight_schools.py eight_schools.py.output)))

 (rule
 (targets normal_lub.py.output)
 (deps (package stanc) (:stanfiles ../good/code-gen/normal_lub.stan))
 (action
  (with-stdout-to %{targets}
   (run %{bin:stan2tfp} %{stanfiles}))))

(alias
 (name runtest)
 (action (diff normal_lub.py normal_lub.py.output)))
```
If I understand this correctly, `normal_lub.py.output` is the generated file... but what is `normal_lub.py`?

seantalts:
Yep! It's just a skeleton end-to-end for eight schools, basically.

You can just include normal_lub.stan in the tfp directory if it's a new model. We can create a model that keeps track of all of the supported features so far :) 

normal_lub.py is the checked-in version that we want to diff against and keep updated. I think there may be some dune shorthand for what we're doing here but I haven't figured it out yet. That last line will diff the py file against the output file (which you can think of as essentially just being a temp file that you have to give a name to) and throw an error if they are different.

seantalts:
Can read more here: https://jbuilder.readthedocs.io/en/latest/tests.html#diffing-the-result

adamhaber:
Any idea on how to test offset/multiplier?

bob-carpenter:
Both the unconstrained parameters and constrained parameters are available through the model class.  So a one-parameter model with an affine transform should be testable that way.  It's a pain to put the models together and test, though.

adamhaber:
Regarding testing the new transformations - as far as I understand, writing such tests requires:
1. Compiling stan2tfp (with the new `code_gen.ml`) locally
2. Using stan2tfp to compile the .stan test file I want add (`normal_lub.stan`) into a .py file
3. Computing various quantities (eg, posterior means) of interests using something like Rstan and construct assert statements using those (which are added to `tests.py`). 
4. Adding other assert statements, specifically for the newly added transformations (for example, make sure bounded parameters are within bounds).

Please correct me if I got this wrong. If not - can you explain how do I compile stan2tfp locally? I don't see this in the makefile...

seantalts:
That's pretty close to right, but instead of making stan2tfp locally manually we'll have `dune` do this. The dune file you've been modifying causes `eight_schools.py` to be emitted by calling `stan2tfp.exe` on the eight_schools_ncp.stan file. It will always try to regenerate that file whenever `dune runtest` is run (very early in the build process) and fail if something that isn't checked in to git there is generated, showing the diff. The Python tests will then run using that auto generated eight_schools.py test; those are run later in the Jenkinsfile with a pure python + tf/tfp docker image. 

For the record, stan2tfp is built in `_build/default/src/stan2tfp/stan2tfp.exe` by `dune build` or `dune runtest`.

adamhaber:
I'm missing something, then... when I run `dune runtest` with the changed dune file, I get
```
File "test/integration/tfp/dune", line 19, characters 0-75:
19 | (alias
20 |  (name runtest)
21 |  (action (diff normal_lub.py normal_lub.py.output)))
Error: No rule found for test/integration/tfp/normal_lub.py
         run alias test/unit/runtest
```
When I remove these three lines, but keep 
```
 (rule
 (targets normal_lub.py.output)
 (deps (package stanc) (:stanfiles normal_lub.stan))
 (action
  (with-stdout-to %{targets}
   (run %{bin:stan2tfp} %{stanfiles}))))
```
dune runtest doesn't complain but no normal_lub.py is generated...

seantalts:
Ah right, just touch the file the first time; it will then show a full diff
and you can use “dune promote” to fill it in.

adamhaber:
Now normal_lub test should make sure upper/lower/upper-and-lower bounded variables are OK. 

From the fact that all these commits were OK on my laptop, I assume that `dune runtest` doesn't actually call `python3 test/integration/tfp/tests.py`? Is it possible to include it somehow?

seantalts:
Hm, it could be, but so far it hasn’t seemed like it was worth it given
that it’s a little annoying to integrate dune with external programs. Plus
many stanc3 devs may never install TF.

Open to recalculating that trade off but that’s my thought process so far.

adamhaber:
I'm trying to add tests for the Cholesky-related bijectors, but this requires that stan2tfp would be able to handle `lkj_corr_cholesky` which it currently doesn't... I've tried adding it myself:
```ocaml
let map_functions fname args =
  match fname with
  | "multi_normal_cholesky" -> ("MultivariateNormalTriL", args)
  | "lkj_corr_cholesky" -> ("CholeskyLKJ", args)
  | f when Operator.of_string_opt f |> Option.is_some -> (fname, args)
```
but the args are incorrect - TFP's `CholeskyLKJ` expects 2 args (dimension+concentration) while `lkj_corr_cholesky` expects only concentration. In addition, most of the stan models I've seen that contain `lkj_corr_cholesky ` also contain `diag_pre/post_multiply` which aren't implemented yet. 

If you have ideas for further tests that are currently implementable with stan2tfp I'd be happy to add them.

seantalts:
I guess there are two issues here:
1. Dealing with keyword arguments
2. Getting the dimension of a Stan expression

For 1, you should be able to use Python's `None` to align the arguments correctly positionally, even though it's more ideal in almost every way to specify them by name. 

For example,
```ocaml
let map_functions fname args =
  match fname with
  | "multi_normal_cholesky" -> ("MultivariateNormalTriL", args)
  | "lkj_corr_cholesky" -> ("CholeskyLKJ", {pattern=Var "None"; meta=Expr.Fixed.empty} :: args)
  | f when Operator.of_string_opt f |> Option.is_some -> (fname, args)
```

Would prepend `None` as the dimension argument. 

But that doesn't help us here, because we need to actually find the dimension. top level declarations (e.g. parameters, data, transformed *) have dimensions accessible on the AST. arbitrary expressions do not, but we might be able to just call `.shape` on the thing to be passed in?

adamhaber:
> ```ocaml 
> ("CholeskyLKJ", {pattern=Var "None"; meta=Expr.Fixed.empty} :: args)
> ```
This is great, I'll use this trick to implement all the `tfd`s that take either probs or logits, etc... Do you prefer this over adding 
```python
binomial = lambda n,p : tfd.Bonimial(n, probs=p)
binomial = lambda n,logit_p : tfd.Bonimial(n, logits=logit_p)
```
etc under the imports block?


> but we might be able to just call .shape on the thing to be passed in?

This works:
```python
x = tf.linalg.cholesky(tfd.LKJ(2,2).sample())
tfd.CholeskyLKJ(x.shape[0],2).log_prob(x)
```
I'm not really sure how to implement this in `Code_gen`, though... as far as I understand, this means that we need to add a case similar to   
```ocaml
  | FunApp (StanLib, f, obs :: dist_params)
    when String.is_prefix ~prefix:Transform_mir.dist_prefix f ->
      pf ppf "%a.log_prob(%a)" pp_call (f, pp_expr, dist_params) pp_expr obs
``` 
but append `obs`.shape to `dist_params`?

adamhaber:
Hi, 

I don't mean to bump this, I'm just not sure if there are any requested changed I should further implement... I can re-write the tests, but other than that, AFAIK, adding additional bijectors will require changes to AST?

seantalts:
Always feel free to bump things! Haha. I just finished a few weeks off and am returning this morning. We can merge this without requiring more bijectors, too! If we wanna go that route could you just fix the new merge conflicts in tests.py? Thanks! On to the further issues with other bijectors, 

> I tried this and got an "unbound record field pattern" error; I'm assuming from this that args should be a list of records? 

Since the MIR refactor we have renamed the records for both Stmt and Expr to be `pattern`, so now we have to explicitly say which is which. Some of the files have the relevant module opened but if not, we have to say e.g. `{Expr.Fixed.pattern=Var...; meta=...}`

> Do you prefer this over adding [new lambda definitions of the distributions with the Stan parameterization]

I think it's probably slightly preferred; the closer we can get to outputting generic actual TFP code the better, assuming we don't have to jump through crazy hoops to make it happen. That will make it easier for e.g. TFP folks to help us figure out how to emit more performant TFP code, for example.

> but append obs.shape to dist_params?

Something like that. Here we might prefer to just jump to string formatting, adding this case above the existing one,

```ocaml
  | FunApp (StanLib, f, obs :: dist_params)
    when f = Transform_mir.dist_prefix ^ "CholeskyLKJ"  ->
      pf ppf "%a.log_prob((%a).shape[0], %a)" pp_call (f, pp_expr, dist_params)
         pp_expr obs pp_expr obs
```

Maybe we'll end up with a set of these functions that take a shape first and we can refactor it that way to check that f is in that set as well.

seantalts:
Oh wait, I put the shape access in the wrong spot, which was much easier than the thing I'm about to suggest, which is just doing the call print inline with the argument added,

```ocaml
| FunApp (StanLib, f, obs :: dist_params)
    when f = Transform_mir.dist_prefix ^ "CholeskyLKJ"  ->
      pf ppf "%s(@[<hov>(%a).shape[0], %a@]).log_prob(%a)"
         f (list ~sep:comma pp_expr) dist_params pp_expr obs
```

adamhaber:
I think I got the types right, as well as the inline print statement (added an extra `pp_expr obs` after `f`); I can now compile `test_lkj.py` without errors. I've also updated Affine bijectors (which are soon-to-be deprecated) to a tfb.Scale and tfb.Shift.

However, I get a discrepancy in the log prob computation I don't understand. I generate cholesky factors of correlation matrices using 
```R
code <- "
data {
  int K;
}
parameters {
  cholesky_factor_corr[K] L;
}
model {
   L ~ lkj_corr_cholesky(2.0);
}
"
library(rstan)
data = list(K=5)
fit <- stan(model_code = code, data=data)
e<-as.data.frame(extract(fit))
write.csv(e, "test_lkj.csv", row.names=FALSE)
```
And compare the the TFP model's log prob to Stan's:
```python
m = test_lkj_model(5)
data = np.loadtxt("test_lkj_data.csv",skiprows=1,delimiter=',')

bijector = m.parameter_bijectors()[0]
stan_lps = []
tfp_lps = []

for d in data[:100]:
    datum = d[:-1].reshape(5,5).T
    stan_lps.append(d[-1])
    tfp_lps.append(m.log_prob_one_chain([datum]).numpy() - bijector.inverse_log_det_jacobian(datum,2).numpy())
``` 

The two lp arrays are highly correlated, but they're not even close to passing `np.allclose`:
```python
np.mean(np.array(tfp_lps)-np.array(stan_lps)),np.std(np.array(tfp_lps)-np.array(stan_lps))
```
>(0.8316613232249257, 1.0462720916260877)

The same "workflow" (comparing lps, subtracting `bijector.inverse_log_det_jacobian`, ...) works fine with other distributions. Any ideas?

seantalts:
very interesting! I'm not sure what might be causing numerical differences there. Might be worth making an issue on the TFP github and asking there too? 

I think we can leave this in for now while that gets sorted.

src/tfp_backend/Code_gen.ml
```diff
@@ -176,7 +180,18 @@ let pp_bijector ppf trans =
   let components =
     match trans with
     | Program.Identity -> []
-    | Lower lb -> [("Exp", []); ("AffineScalar", [lb])]
+    | Lower lb -> [("Exp", []); ("Shift", [lb])]
+    | Upper ub ->
+        [("Exp", []); ("Scale", [Expr.Helpers.float (-1.)]); ("Shift", [ub])]
+    | LowerUpper (lb, ub) ->
+        [ ("Sigmoid", [])
+        ; ("Scale", [Expr.Helpers.binop ub Operator.Minus lb])
```
seantalts: OCaml also has a "local open" syntax that can be helpful sometimes; not sure if it's worth it here yet you could make the top of this let be
```ocaml
let components = 
  let open Expr.Helpers in
  match trans with
...
```    
and that would give you access to `float` and `binop` here unqualified.

PR #381, opened 2019-11-09
Using bijectors to constrain initial values

adamhaber:
NUTS expects initial values to be in the constrained space; passing random(0,2) returns nans for constrained variables outside of this range (for example, <upper=0>).

### DIFF
--- a/test/integration/tfp/stan.py
+++ b/test/integration/tfp/stan.py
@@ -43,7 +43,7 @@ def stan(model, nchain=4, num_main_iters=1000, num_warmup_iters=1000):
   mcmc_trace, pkr = tfp.mcmc.sample_chain(
       num_results = num_main_iters,
       num_burnin_steps = num_warmup_iters,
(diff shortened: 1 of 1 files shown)


bob-carpenter:
The NUTS implementation in stan-dev/stan works entirely in the unconstrained space. The model class does all the constraining given unconstrained inputs.  That's how it is able to initialize as uniform(-2, 2) in all coordinates.  We would like to keep that initialization as it also makes sense given appropriate variable scaling with bijectors.

adamhaber:
Just to clarify, I was referring to TFP's NUTS implementation; it expect initial values that you can actually call `target_log_prob_fn` on. 

Changed the initialisation range to start from -2 to be more consistent with Stan.

bob-carpenter:
Stan's does from the outside interfaces, too.  But internally, it runs on the unconstrained scale.  Does TFP actually try to run NUTS on the constrained scale?

Issue #384, opened 2019-11-12
TFP backend - supporting more distributions


adamhaber:
Hi,

I'm interested in adding support for more distributions in the TFP backend. I'm trying to find the relevant part of the code to expand; I assume it's this:

```ocaml
let%expect_test "nested dist prefixes translated" =
  let open Expr.Fixed.Pattern in
  let e pattern = {Expr.Fixed.pattern; meta= ()} in
  let f =
    FunApp
      ( Fun_kind.StanLib
      , "normal_lpdf"
      , [FunApp (Fun_kind.StanLib, "normal_lpdf", []) |> e] )
    |> e |> translate_funapps
  in
  print_s [%sexp (f : unit Expr.Fixed.t)] ;
  [%expect
    {|
    ((pattern
      (FunApp StanLib tfd__.Normal
       (((pattern (FunApp StanLib tfd__.Normal ())) (meta ())))))
     (meta ())) |}]
```
My (uneducated) guess is that supporting more distributions means somehow replacing this with some sort of pattern matching, and that other than that it should be pretty similar? For example, replacing tfd__.Normal() with tfd__.Binomial() and normal_lpdf with binomial_lpmf?  

Any help on how to approach this would be appreciated... thanks!

seantalts:
That's an [expect test](https://blog.janestreet.com/testing-with-expectations/). Check out https://github.com/stan-dev/stanc3/blob/master/src/tfp_backend/Transform_mir.ml#L17

PR #388, opened 2019-11-17
Add more distributions #384

adamhaber:
Hi,

Added support for the following continuous distributions:
Unconstrained:
1.  Gumbel
2. Student T
3. Laplace

Positive:
1. Exponential
2. Chi square
3. Lognormal
4. Gamma
5. Inverse Gamma


src/tfp_backend/Transform_mir.ml
```diff
@@ -10,11 +10,21 @@ let remove_stan_dist_suffix s =
     ~f:(fun suffix -> String.chop_suffix ~suffix s)
   |> List.hd_exn
 
-let capitalize_fnames = String.Set.of_list ["normal"; "cauchy"]
+let capitalize_fnames =
+  String.Set.of_list ["normal"; "cauchy"; "gumbel"; "exponential"; "gamma"]
+
+(* Bernoulli, Binomial, NegativeBinomial should be mapped to TFP with either probs= or logits= flags *)
+(* Poisson should be mapped to TFP with either rate or log_rate *)
+(* | "lkj_corr_cholesky" -> ("CholeskyLKJ", args) *)
```
seantalts: Why is this one commented out?

adamhaber: `| "lkj_corr_cholesky" -> ("CholeskyLKJ", args)` is commented out because TFP's `CholeskyLKJ` takes two parameters (dim and concentration) whereas args only contains concentration. Perhaps there's some way to pull the dim from `cholesky_factor_corr[K] L` into `args`?

Regarding `Bernoulli`, `Binomial` etc - as far as I understand, we want to map `binomial_logit(N, alpha)` to `Binomial(N, logits=alpha)` and `binomial(N, theta)` to `Binomial(N, probs=theta)`. Here, as well, I wasn't sure how to do this so I commented it out for now...

adamhaber:
1. Sure, I'll rewrite the tests and skip the `run_mcmc` part. I think log_prob evaluation might be better, as it'll be easier to make sure differences in results are due to differences in distributions implementations/parameterisation and not VI/optimization algorithm differences. Makes sense?

2. The idea was to be consistent with Stan's function reference (roughly chapters 11-25); I thought it'll make it easier for us to keep track of what distributions are not supported yet.

src/tfp_backend/Transform_mir.ml
seantalts: One day we might like to have a different tree structure than MIR for printing Python, one that  supports keyword arguments and maybe inline function definitions. For now I think it might be okay to essentially use those as positional arguments by inserting `None` (or whatever the default is as appropriate.

seantalts:
> The idea was to be consistent with Stan's function reference (roughly chapters 11-25); I thought it'll make it easier for us to keep track of what distributions are not supported yet.

Oh I see. There's a lot of model code that's not really necessary for testing these non-statistical models - I was thrown off the scent a little when reading seeing lots of `normal` priors in both places. I might lightly suggest having each file contain the sort of minimal stuff under test, and testing each distribution in one place, if that makes sense. 

That can be a future improvement - this is great and I'm going to merge it. Thanks!

src/tfp_backend/Transform_mir.ml
adamhaber: One possible workaround is to define "helper" function in the generated .py file, something like
```python
binomial_logit = lambda n,a : tfd__.Binomial(n, logits=a)
binomial = lambda n,a : tfd__.Binomial(n, probs=a)
```
and then add
```ocaml
| "binomial_logit" -> ("binomial_logit", args)
| "binomial" -> ("binomial", args)
```
This will make the .py file uglier, but should allow us to extend the number of supported distributions without making significant changes to the compiler itself; and removing it once such changes will happen should be straightforward. What do think?

adamhaber:
Yes the tests are too busy, I agree. Just to make sure I understand - you suggest 
```stan
parameters {
    real y;
}
model {
    target += normal_lpdf(y | 0,1);
}
```
for each distribution, and then compare log probs?

seantalts:
Oh, I think grouping them up like you have now seems good. I just meant that we don't necessarily have to give their parameters priors and we don't have to test `normal` in the "unbounded" file (or in whichever categories it doesn't belong to in the reference manual). What you have written would work but I don't necessarily think we need a stand-alone `.stan` file for each distribution, so you could combine them.

PR #394, opened 2019-12-03
Number literal fixes

### LINKED ISSUE #389: underscores in numbers
Large number literals are more readable when they can have thousands separators.
e.g. 100_000 vs 100000.
Last year there was [a discussion on the forums](https://discourse.mc-stan.org/t/stan3-enable-underscores-in-numbers/6536) but it hasn't been implemented yet.

nhuurre:
Three changes
* Fixes #389 by changing the lexer to allow underscore separator in number literals. Underscores are kept in MIR but replaced with single quotes during C++ code generation. Underscores follow the same rules as Python so TFP backend does not need to do anything about them.
* The lexer no longer accepts invalid floating point numbers such as `.e-12` and even `.` (a lone decimal point).
* Integer literals with leading zeros were miscompiled because C++ compilers think they are octal (and Python just throws a SyntaxError). Correct Stan2 behaviour is now restored by stripping the zeros when creating MIR.


src/stan_math_backend/Expression_gen.ml
```diff
@@ -113,6 +113,7 @@ let is_user_dist s =
   ends_with_any user_dist_suffices s
   && not (ends_with_any ["_cdf_log"; "_ccdf_log"] s)
 
+let format_number = String.map ~f:(function '_' -> '\'' | c -> c)
```
seantalts: 1) I think it might be slightly better to remove this before we get to the MIR - partial evaluation and related optimizations I think rely on the numbers being valid OCaml numbers
2) This only works in C++14, and we technically need to support C++1y still. I'm not sure if [all of our supported](https://github.com/stan-dev/stan/wiki/Supported-C---Compilers-and-Language-Features) compilers support the `'`.

nhuurre:
OCaml is even more permissive than Python so these are all valid OCaml numbers. I thought digit separator was c++1y feature. cppreference.com linked from the compiler wiki page says it's supported since gcc 4.9 and clang 3.4.

But I tend to agree it's better to just remove the separators early rather than leaving it up to backend.

Oh, and I forgot Python won't accept leading zeros in floating point numbers either, so better clean that up too.

Issue #335, opened 2019-10-14
Not checking if _lpmf functions are only used within other _lp* functions and generating invalid C++


wds15:
The old compiler previously allowed functions containing "lp" to access target; the new compiler follows the manual and only allows functions ending in `_lp`, `_lpmf`, `_lpdf`, or `_log` to access them. Right now this is generating bad C++ code that gives a compile error complaining about `error: use of undeclared identifier 'lp__'` or `error: use of undeclared identifier 'lp_accum__'`.

# Example
This model runs just fine with 2.20, but the new stanc 2.21 compiler seems to create non-compiling C++ code.

fail log, code and data attached.

[stanc3-failure-oncobayes.zip](https://github.com/stan-dev/stanc3/files/3723392/stanc3-failure-oncobayes.zip)

seantalts:
This looks like it's failing because Stan 2 assumed that a function ending in `_lpmf_comp` was also a user-defined distribution, but as far as I know that's a bug (is that right, @bob-carpenter ?). You can see the spec here: https://mc-stan.org/docs/2_18/stan-users-guide/user-defined-probability-functions.html

wds15:
But stanc3 creates non compile-able c++ code. I would expect an error message from the parser at least.

seantalts:
Fair enough! This does seem like something the type checker should be catching.

wds15:
After fixing the problem with `_lpmf` I am now getting a C++ file which cannot be compiled:

```
../OncoBayes2/adapt_delta_tweak/ob2_model-stanc3.hpp:3989:68: note: insert an explicit cast to silence this issue
                                                                   g,
                                                                   ^
                                                                   static_cast<int>( )
../OncoBayes2/adapt_delta_tweak/ob2_model-stanc3.hpp:3990:68: error: non-constant-expression cannot be narrowed from type 'unsigned long' to 'int' in initializer list [-Wc++11-narrowing]
                                                                   (g + num_groups)}), nil_index_list()), "beta"),
                                                                   ^~~~~~~~~~~~~~~~
../OncoBayes2/adapt_delta_tweak/ob2_model-stanc3.hpp:3990:68: note: insert an explicit cast to silence this issue
                                                                   (g + num_groups)}), nil_index_list()), "beta"),
                                                                   ^~~~~~~~~~~~~~~~
                                                                   static_cast<int>( )
[...]

PR #397, opened 2019-12-06
remove  lp__ argument from _lpmf functions

nhuurre:
A quick fix for issue #335

See my comment on the issue for reasons why I chose this approach rather than the one suggested by the issue title.


seantalts:
I read the issue but I'm not sure what the approach here is - the legal argument you laid out on the issue might suggest we no longer allow all functions containing the characters `lp` to access `lp`, right? But the code here is all in the backend - I would have expected to see some changes to the front end to print pretty error messages when the user wants to do that but they aren't allowed to anymore. Could you write a few sentences explaining what the approach is? Thanks!

seantalts:
(PS I really enjoyed the name of this branch :P)

Issue #307, opened 2019-09-25
constraint functions can be shadowed by user-defined functions and variables


seantalts:
the C++ compiler fails to compile a model if a user defines something with the same name as one of the constraint functions, e.g. "lb_constrain"

PR #398, opened 2019-12-06
Add *_constrain and *_free functions from Stan Math

nhuurre:
Closes issue #307 because identifiers that shadow these will now cause the compiler to fail early with a clear error message.

The old Stan compiler did not expose these functions and in fact allowed their names to be used in a model without conflicts. While it would have been possible to recover the old behaviour I believe that exposing the functions is more reasonable.

Known bugs:
* `cholesky_corr_free` and `corr_matrix_free` are not autodiffable.
* If the first argument to `lub_constrain` is of type `int` then C++ compiler deduces a bad template specialization and the function will return nonsense.

Copyright holder: Niko Huurre

By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:

* Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
* Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)


seantalts:
Let me get these tests working again, sorry!

seantalts:
Hmm, this is not obvious though - do we actually want to expose these? @bob-carpenter what do you think? I could see it going either way, but if we were going to expose them we'd need to add all the doc to the manual as well. The names are a bit suboptimal too.

The other option I can think of is to appropriately namespace them in the compiled code - e.g. prepend `stan::math::` to their names.

seantalts:
In the absence of @bob-carpenter response, would you mind if we closed this for now and just namespaced these function calls instead for the time being? I'm not sure we want to go through with adding that many functions to the language all at once just to solve this bug in time for the release.

nhuurre:
No problem.

Issue #308, opened 2019-09-26
Inconsistent behaviour with simplex of dimension 0


sambrilleman:
#### Summary:
I have an edge case where a specific definition of a simplex with dimension zero causes R to crash (model `code3` below). Strangely, it only crashes for sampling, not for VB algorithms.

#### Description:
I discovered this when working on a development version of rstanarm. The reproducible example below is just a toy example. You could ask why I even need a simplex of dimension zero, and the truth is I probably don't. But some model specifications do not require the simplex and so simplex[0] was a method for dealing with that. However after finding this bug I have just replaced simplex[0] with simplex[1] and the model specification is still the same I think (since simplex[1] doesn't actually add any unconstrained parameters to the model). But it took me 1.5 days of debugging to find this bug because R just crashed rather than throwing any informative error message, hence why I thought it might be worth reporting it here to save other people a similar headache down the track. 

#### Reproducible Steps:
```
library(rstan)

zero <- 0

code1 <- 
"
data {
  int zero;
}
parameters {
  simplex[zero] junk[0];
  real y;
} 
model {
  y ~ normal(0, 1);
} 
"

code2 <- 
"
data {
  int zero;
}
parameters {
  simplex[zero] junk; 
  real y;
} 
model {
  y ~ normal(0, 1);
}
"

code3 <- 
"
data {
  int zero;
}
parameters {
  simplex[zero] junk[zero]; 
  real y;
} 
model {
  y ~ normal(0, 1);
} 
"

mod1 <- stan_model(model_code = code1)
mod2 <- stan_model(model_code = code2)
mod3 <- stan_model(model_code = code3)

vb1 <- vb(mod1, iter = 10) # works
vb2 <- vb(mod2, iter = 10) # throws error
vb3 <- vb(mod3, iter = 10) # works

fit1 <- sampling(mod1, iter = 10, chains = 1) # works
fit2 <- sampling(mod2, iter = 10, chains = 1) # throws error
fit3 <- sampling(mod3, iter = 10, chains = 1) # R crashes
```

#### RStan Version:
2.18.2 (same behaviour with 2.19 as well)

#### R Version:
3.6.1

#### Operating System:
Windows 10

sambrilleman:
> The behavior for me is not a crash, but a hang

Yeah, sorry, you are correct -- it hangs rather than crashes.

> so there seems to be something odd going on with the sizes being specified as data.

[...]

Issue #184, opened 2019-06-19
test that simplex and unit_vector have size > 0


bob-carpenter:
The `io::reader` class in Stan was not previously checking that `simplex` and `unit_vector` types were not declared with size zero.  There's a new PR fixing that, but my understanding is that stanc3 is not using the reader and writer classes. 

When a simplex or unit vector is declared to be of size zero, the model should throw `std::invalid_argument` to stop sampling.

PR #403, opened 2019-12-20
Validate array size

### LINKED ISSUE #401: check that array size is nonnegative
While looking into issue #184 I noticed that generated code _never_ checks if array or vector dimensions are negative.

This model throws `std::bad_alloc`
```
parameters {
  real x;
} model {
  real y[-1] = rep_array(0.0,-1);
  x ~ std_normal();
}
```
and this model goes into an infinite loop
```
parameters {
  real x[-1];
} model {
  x ~ std_normal();
}
```
The old compiler inserts `validate_non_negative_index()` calls where appropriate so that both of these throw `std::invalid_argument`.

Presumably this should be added in `stan_math_backend/Transform_mir.ml`. Pretty straighforward except  for one thing. `validate_non_negative` takes three arguments: the variable name, the size, and a string representing the size expression. It's not obvious where to get the last one because there's no MIR-to-Stan pretty printer. Two hackish ideas: either (1) SizedType gets an additional field that holds the original string, or (2) print the MIR expression as C++, which is probably close enough.

nhuurre:
Fixes #401 plus related issues #184 and #308 

EDIT:
Copyright holder: Niko Huurre

By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:

* Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
* Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)


bob-carpenter:
Before merging, we need a statement of copyright ownership and a BSD-3 license with CC-BY doc licensing.

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -602,6 +632,37 @@ std::vector<double> to_doubles__(std::initializer_list<double> x) {
 std::vector<stan::math::var> to_vars__(std::initializer_list<stan::math::var> x) {
   return x;
 }
+
+inline void validate_positive_index(const char* var_name, const char* expr,
```
seantalts: Would you mind explaining why these new functions were necessary?

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -395,6 +395,65 @@ let%expect_test "insert before" =
   [%sexp (l : int list)] |> print_s ;
   [%expect {| (1 2 3 4 5 999 6) |}]
 
+let rec add_validate_dims outvars stmts =
+  let params =
+    List.filter_map
+      ~f:(function
```
seantalts: Could you give this function a name? or the resulting `params` thing something indicating which params it should contain.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -395,6 +395,65 @@ let%expect_test "insert before" =
...
+    Option.value ~default:Internal_fun.FnValidateSize (Map.find params id)
+  in
+  let f = function
+    | Stmt.Fixed.({pattern= Decl {decl_id; decl_type= Sized st; _}; meta}) as
+      decl ->
+        let check fn x =
+          Stmt.Fixed.
+            { meta
+            ; pattern=
+                NRFunApp
```
seantalts: You can use the `internal_funapp` helper here to make this a little more readable, hopefully.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -395,6 +395,65 @@ let%expect_test "insert before" =
...
+                  ( CompilerInternal
+                  , Internal_fun.to_string fn
+                  , Expr.Helpers.
+                      [ str decl_id
+                      ; str (Fmt.strf "%a" Expression_gen.pp_expr x)
+                      ; x ] ) }
+        in
+        let rec map = function
```
seantalts: Could you give this a different name to indicate that it's not the same as a traditional `map`?

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -395,6 +395,65 @@ let%expect_test "insert before" =
...
+          | Program.UnitVector -> Some (decl_id, FnValidateSizeUnitVector)
+          | _ -> None )
+        | _ -> None)
+      outvars
+    |> String.Map.of_alist_exn
+  in
+  let fnvalidate id =
+    Option.value ~default:Internal_fun.FnValidateSize (Map.find params id)
+  in
+  let f = function
```
seantalts: Could you give this a name?

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -395,6 +395,65 @@ let%expect_test "insert before" =
...
+        map (SizedType.get_dims st) @ [decl]
+    | stmt -> [validate_dims_stmt stmt]
+  in
+  List.concat_map ~f stmts
+
+and validate_dims_stmt stmt =
+  Stmt.Fixed.
+    { stmt with
+      pattern=
+        ( match stmt.pattern with
```
seantalts: I think this would be more clearly written as 

```ocaml
let pattern = match stmt.pattern with
 | Stmt.Fixed.Pattern.Block s ->
￼            Stmt.Fixed.Pattern.Block (add_validate_dims [] s)
￼        | SList s -> SList (add_validate_dims [] s)
￼        | While (a, b) -> While (a, validate_dims_stmt b)
￼        | For f -> For {f with body= validate_dims_stmt f.body}
￼        | IfElse (p, t, e) ->
￼            IfElse
￼              ( p
￼              , validate_dims_stmt t
￼              , Option.map ~f:validate_dims_stmt e )
￼        | s -> s
in
Stmt.Fixed.{stmt with pattern}
```

src/stan_math_backend/Stan_math_code_gen.ml
nhuurre: They're just there for nicer error messages. Stan Math library has only `validate_non_negative_index`. For local variables we just need to check that the declared dimensions are nonnegative but for constrained parameters we must also ensure that the corresponding unconstrained type is valid. An N-dimensional simplex is represented internally by (N-1)-dimensional vector. Calling `validate_non_negative_index` with the calculated unconstrained size would give a confusing error message (the user asked for size 0 simplex so why does it complain about size -1 vector?).

The issue with unit vectors is more subtle. The constraining transform does not change the dimension. However, the transform is discontinuous which causes problems, especially in low dimensions, and the space of _one-dimensional_ unit vectors contains only two points: `[1.0]'` and `[-1.0]'`. HMC cannot possibly sample it efficiently.

src/stan_math_backend/Transform_Mir.ml
nhuurre: It should contain all params whose constraint requires extra validation (simplex, unit vector, cholesky factor). But filtering the other params out was just a premature optimization so I changed it to just include all params.

PR #406, opened 2019-12-22
added support for empty datablock

adamhaber:
At the moment, a Stan program without a data block creates an empty `__init__` function which raises an indentation error in `def log_prob_one_chain`. This adds a simple `pass;` statement to avoid this.

### DIFF
--- a/src/tfp_backend/Code_gen.ml
+++ b/src/tfp_backend/Code_gen.ml
@@ -97,9 +97,9 @@ let pp_init ppf p =
     pf ppf "self.%s = %a" name (pp_cast "") (name, st)
   in
   let ppbody ppf =
-    if not (List.is_empty p.Program.input_vars) then
-      (list ~sep:cut pp_save_data) ppf p.Program.input_vars
-    else pf ppf "pass;"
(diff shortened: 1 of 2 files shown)


src/tfp_backend/Code_gen.ml
```diff
@@ -96,7 +96,11 @@ let pp_init ppf p =
   let pp_save_data ppf (name, st) =
     pf ppf "self.%s = %a" name (pp_cast "") (name, st)
   in
-  let ppbody ppf = (list ~sep:cut pp_save_data) ppf p.Program.input_vars in
+  let ppbody ppf =
+    if not (List.is_empty p.Program.input_vars) then
```
seantalts: Thanks for the patch!

I think this might be more clear as a `match`, what do you think?

```ocaml
  match p.Program.input_vars with
    [] -> pf ppf "pass"
    _ -> (list ~sep:cut pp_save_data) ppf p.Program.input_vars
```

adamhaber: Sure, thanks!

seantalts:
Thanks!

PR #409, opened 2019-12-28
Add probs/logits distributions

adamhaber:
Added support for the following distributions:

- Beta
- Poisson (including log parameterisation)
- Binomial (including logit parameterisation)
- Bernoulli (including logit parameterisation)
- Von Mises
- Pareto

I've also refactored the tests folder to be less cluttered.

Regarding the match expressions in Transform_mir - is there a better way to extract the values from args? The _ case in all of them seems redundant, but AFAIK, since args for the different cases is of different lengths, I can't use something like `let y, p = args in`, right? 

Regarding additional distributions:

- Many distributions that are supported by Stan are currently not supported by TFP (beta binomial, hypergeometric, skewed normal, exponentially modified normal, etc).
- Some distributions are implemented with different parameterisation. For example, Stan's neg_binom(a,b) is NegativeBinomial(a, 1/(1+b)). 
- Some distributions require bijectors we currently don't have. For example, Wishart is identical in both Stan and TFP, but we haven't implemented support for covariance matrices yet. The simplex datatype is also not yet supported, and is needed for Stan's categorical distribution.
- Some distributions require different parameters in TFP. For example, multinomial in Stan expect a simplex vector theta, while in TFP it also expects `total_counts`.

EDIT: CI doesn't seem to find the samples files (*.npy containing Stan samples and lp__ per model). Is the new structure of test folder problematic, or can this be changed via CI settings?


nhuurre:
>is there a better way to extract the values from args?

You could match both function name and args at the same time
```ocaml
match (fname, args) with
| "binomial_logit", _ -> ("Binomial", args)
| "binomial", [y; n; p] ->
        ( "Binomial", [y; n; {Fixed.pattern= Var "None"; meta= Typed.Meta.empty}; p] )
```

>For example, Stan's `neg_binom(a,b)` is `NegativeBinomial(a, 1/(1+b))`.

There are some functions in `Expr.Hepers` that can be used to modify the argument expression. I think something like this should work:
```ocaml
| "neg_binom", [a;b] ->
        ( "NegativeBinomial",
        [a; Expr.Helpers.(binop (int 1) Divide (binop (int 1) Plus b))] )
```

Those file sample paths are relative to the current working directory (which may vary). To create a path relative to the test file you could use the modules `__file__` attribute.
```python
from os.path import dirname, join
stan_fit = np.load(join(dirname(__file__), "stan_samples/samples_cont_unbounded.npy"))
```

adamhaber:
Thanks! This is really helpful. 

Regarding the helper function - is there any way to put parenthesis in the denominator? 
```ocaml
| "neg_binomial", [y;a;b] -> ("NegativeBinomial",[y; a; Helpers.(binop (int 1) Divide (binop (int 1) Plus b))])
```
translates to
```python
tfd__.NegativeBinomial(nb_alpha, tf__.cast(1, tf__.float64) / tf__.cast(1, tf__.float64) + nb_beta).log_prob(y)
```
instead of 
```python
tfd__.NegativeBinomial(nb_alpha, tf__.cast(1, tf__.float64) / (tf__.cast(1, tf__.float64) + nb_beta)).log_prob(y)
```

seantalts:
I think for that we just need to fix the generator for operators to always
include parens (we could get smarter about when they’d be necessary but
probably no reason)

adamhaber:
I removed the negative binomial distribution; will add it (and other distributions that involve reparameterisation) when the generator fix is added.

seantalts:
Looks like there's a bunch of DS_STORE and other random files in the PR - would you mind removing them and possibly adding them to the .gitignore file?

seantalts:
Could you also write a little bit of doc (nothing more than a couple paragraphs) explaining the testing system you've created? I'm also wondering what the design space was like and how you came up with this one - not that I see anything wrong with it at all, just that I'm unfamiliar.

adamhaber:
> Looks like there's a bunch of DS_STORE and other random files in the PR - would you mind removing them and possibly adding them to the .gitignore file?

Sure, I'll remove them and the stanc.install file. Anything else?

> Could you also write a little bit of doc (nothing more than a couple paragraphs) explaining the testing system you've created? I'm also wondering what the design space was like and how you came up with this one - not that I see anything wrong with it at all, just that I'm unfamiliar.

Not sure I understand... can you elaborate? In general I've tried to be consistent with the discussion in #388 :

1. Test that lp computations are consistent between Stan and TFP
2. Run mcmc only when necessary (for example, to verify that bounded parameters are within bounds)
3. Group distributions according to the Stan manual.
4. Make Stan programs as small as possible (for example, only use a data block when needed).

seantalts:
I agree with all of your points summarizing :) I didn't realize how #1 was working before - would you mind adding doc for it to the code now? In particular I didn't realize that the way we were going to test log_prob was to save a full fit from Stan for each of these models - I was imagining using log_prob on some generated set of unconstrained parameter values that would be generated using the same algorithm for all models. I think if we can avoid committing entire Stan fits for every model we test under TFP that would be worth some significant effort (and I'm happy to help out with that or do all the work myself if that's helpful since it's my snafu! I realize I didn't understand the path we were headed down before with 388).

adamhaber:
No worries! :) So - just to be on the safe side - would this be better:
1. Add documentation (within `tests.py`?) elaborating on the different tests we're running.
2. Instead of calling `_compare_lp` which compares lp over an entire Stan fit, do:
```python
    def test_X(self):
        target_dist = test_X_model()
        self.assertTrue(np.allclose(target_dist.log_prob([x, y, z]),lp))  
```
where [x,y,z] is a random set of unconstrained parameter values, and lp is Stan's corresponding lp__?

seantalts:
> Sure, I'll remove them and the stanc.install file. Anything else?

Any files that don't look like they should be included - for example, `test/integration/tfp/.ipynb_checkpoints/Untitled-checkpoint.ipynb`. You can check the list of which files are currently included from the github interface or using a command like `git diff --stat origin/master`

> Instead of calling _compare_lp which compares lp over an entire Stan fit, do:

Hmm. I think we could formulate these as cram tests (aka gold tests, or expect tests in OCaml). Then we can have a way to automatically generate the correct values from Stan and an automated test of the TFP backend against those values. How does that sound? 

I don't want to be prescriptive here - I think some goals in maybe descending order of preference from my perspective would be:
1) Easy to understand what is being tested and what is going wrong when a test failure occurs
2) Easy to create new tests or update the tested values when something changes (though these are testing end-to-end stuff that isn't supposed to change except for bug fixes and major version number bumps)
3) Has a small footprint

So it's possible that the best solutions to 1 and 2 have a larger footprint and that's okay, we can just organize around it.

adamhaber:
> Any files that don't look like they should be included

Done, thanks.

> Hmm. I think we could formulate these as cram tests (aka gold tests, or expect tests in OCaml). Then we can have a way to automatically generate the correct values from Stan and an automated test of the TFP backend against those values. How does that sound?
I don't want to be prescriptive here - I think some goals in maybe descending order of preference from my perspective would be:

That's OK, I'm still very new to all this and I'm really interested to learn. :-) I agree about 1-3, and I really like 4 as a "guiding principle". I'm not sure, however, how to automatically generate the correct lp values from Stan. I did this by extracting lp__ from RStan, but I'm sure there's a better, more automated way.

adamhaber:
With #411 we can now support neg_binomial. Tested like the other distributions, for now.

seantalts:
> ... how to automatically generate the correct lp values from Stan. I did this by extracting lp__ from RStan, but I'm sure there's a better, more automated way.

What you did is really great and clever, it just has the one side effect of implying this kind of explosion of repo size over time (which might be okay?). The one thing I'd say was more importantly missing from that approach is that we'd want to check in the script for generating those files into the repo as well. 

Hmm... For the sake of forward progress, I think let's go ahead and merge this and I'll try to come up with another way to perform similar tests with less data checked in.

PR #417, opened 2020-01-04
Refactor GPU GLM conditions

rok-cesnovar:
Now that the broadcast GLMs are added we need to refactor the conditions. This was discussed in the [original PR](https://github.com/stan-dev/stanc3/pull/314) that added GPU GLMs. Quoting @t4c1:

> Once signatures with broadcasting are added to remaining GLMs they will either go into "Use opencl if X (second argument - the matrix) is already on GPU." if X is a matrix or "Never use opencl." if x is a row_vector.

So this does just that. Basically all GLMs have the same condition of when to switch to using OpenCL: second argument is a data matrix

### DIFF
--- a/test/integration/good/code-gen/cl.expected
+++ b/test/integration/good/code-gen/cl.expected
@@ -90,40 +90,56 @@ using namespace stan::math;
 
 static int current_statement__ = 0;
 static const std::vector<string> locations_array__ = {" (found before start of program)",
-                                                      " (in 'optimize_glm.stan', line 13, column 2 to column 20)",
-                                                      " (in 'optimize_glm.stan', line 14, column 2 to column 17)",
+                                                      " (in 'optimize_glm.stan', line 14, column 2 to column 20)",
                                                       " (in 'optimize_glm.stan', line 15, column 2 to column 17)",
-                                                      " (in 'optimize_glm.stan', line 16, column 2 to column 22)",
-                                                      " (in 'optimize_glm.stan', line 17, column 2 to column 13)",
-                                                      " (in 'optimize_glm.stan', line 18, column 2 to column 11)",
-                                                      " (in 'optimize_glm.stan', line 19, column 2 to column 19)",
-                                                      " (in 'optimize_glm.stan', line 20, column 2 to column 22)",
-                                                      " (in 'optimize_glm.stan', line 21, column 2 to column 23)",
-                                                      " (in 'optimize_glm.stan', line 25, column 2 to column 64)",
+                                                      " (in 'optimize_glm.stan', line 16, column 2 to column 17)",
+                                                      " (in 'optimize_glm.stan', line 17, column 2 to column 22)",
+                                                      " (in 'optimize_glm.stan', line 18, column 2 to column 13)",
+                                                      " (in 'optimize_glm.stan', line 19, column 2 to column 11)",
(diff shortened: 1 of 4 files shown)


test/integration/good/code-gen/cl.expected
```diff
@@ -146,7 +163,9 @@ class optimize_glm_model : public model_base_crtp<optimize_glm_model> {
   std::vector<int> y_vi_d;
   std::vector<int> y2_vi_d;
   int y_s_d;
+  double y_r_d;
   matrix_cl<double> X_d_opencl__;
+  matrix_cl<double> y_r_d_opencl__;
```
seantalts: interesting, maybe this is the wrong place for this question but do we really need to create a matrix to hold a single double? Can't our OpenCL kernels just take a double as one of the arguments? I'm guessing the answer is "yes we could do that but then we'd need to overload all of the kernels to do that, and that's hard for X reason" is that right?

rok-cesnovar: You are right. 

You either supply the kernel a cl::Buffer or a scalar. The cl::Buffer lives in the GPU memory (values need to be copied there beforehand) while the scalar is passed by value with the kernel call and the scalars lives in the CPU  memory. So its not only an overload, but you need to handle that variable completely different in this case. You also cant modify the value tho that is not a problem in this case. So that is the reason we are handling it this way.

Its far from optimal because we hold an entire object for a single value. If anyone has any suggestions on how we could improve this, they are more than welcome.

seantalts: > and the scalars lives in the CPU memory. 

Ahh, wait, does this mean the kernels would just have to send the double over to the GPU anyway separately? I was imagining the "store the double in its own cl::Buffer" approach would lead to increased latency because I thought that the double could have been sent over to the GPU *as part of* the "command" to run the kernel. If there's no performance or latency difference then this cl::Buffer approach seems optimal to me as you get more code re-use.

rok-cesnovar: No no, you are right. In this case it would get sent over as part of the command. That is why I said it lives in the CPU memory. Sorry for any confusion. 

We use this for kernels where we know an argument is going to always be a read-only scalar. For instance in `fill()` `val` is sent over as part of the "execute kernel" command:

`__kernel void fill(__global double* A, double val,`

Lets look at another example with rep_matrix, where B can a scalar, vector or row_vector. Here we always use cl::buffers and by extent matrix_cl which is essentially an object that holds the cl::buffer:
`__kernel void rep_matrix(__global double* A, __global double* B,`

If we wanted to optimize the case where B is a scalar to this:
`__kernel void rep_matrix(__global double* A, double B,`

we would also need to have a copy of the entire kernel since its no longer using pointers for B. Also in the case of where the scalar resides on the GPU already we would still need to use the first version. It would be slightly more efficient for this case tho.

seantalts: Gotcha. So the tradeoff is basically creating duplicates of the kernels just to change the function signature, and in this case all that buys you is one less to_opencl copy command during model construction, i.e. just once at the beginning of the program, and with the async stuff probably not adding any latency at all given it's also sending a much larger matrix over at the same time, right?

So probably not worth changing anything, right? Thanks for the discussion though!

rok-cesnovar: Yes, that is basically it. Return on investment is not worth it :) No problem, happy to discuss.

Issue #410, opened 2019-12-30
TFP backend - python keywords


adamhaber:
Python has 33 keywords (`import keyword; keyword.kwlist`). Some of these are also keywords in Stan (for, while, return, etc) but most of them aren't, which means users might use them as identifiers (eg, a parameter `lambda` in a Poisson distribution or a `class` vector) which could cause problems when generating TFP code. 

Do we want stan2tfp to check for these and warn/fail?

seantalts:
Good catch - what do you think about renaming them in the TFP backend instead? To something like `lambda_` or `user_lambda` ?

adamhaber:
Sure, `x_` sounds good.

I'll try to fix this; AFAIU, this means adding the appropriate check here:
```ocaml
let rec pp_expr ppf {Expr.Fixed.pattern; _} =
  match pattern with
  | Var ident -> string ppf ident
```
?

 I'll also add a note about that in the stan2tfp package docs.

seantalts:
I think it'd be nicer to do this in Transform_mir - any transformations we can express in the MIR I think we should do in that pre-printing transform stage. the pretty printing can get pretty hairy by itself and its nice to separate concerns if possible.

adamhaber:
Regarding adding a suffix - if, for some obscure reason, the Stan program contains two parameters `lambda` and `lambda_` (or any other keyword + suffix), this will be a problem. Do you think this is too much of a corner case, or should we check for this?

adamhaber:
Hi,

I've tried doing this in Transform_mir; I _think_ I'm covering all the relevant cases, and going over the MIR in the right way, but I'd appreciate an extra pair of eyes on this before submitting a PR. I'm currently not checking for the case described above (`lambda` and `lambda_`), but it shouldn't be very difficult to add, I think.

In short:

- Altered `translate_funapps` to append the suffixes for `Var`s and `Lit`s that are python keywords.
- Added a function `change_kwrds_stmts` that recursively adds the suffix for strings in `Decl`, `NRFunApp`, `Assignment` and `For`, and added it to the `trans_prog` pipe. 
- Added suffixes to the elements of `p.input_vars` and `p.output_vars`.

The full code is [here](https://github.com/adamhaber/stanc3/blob/add-suffix-to-python-kwrds-410/src/tfp_backend/Transform_mir.ml).

nhuurre:
> I'd appreciate an extra pair of eyes on this before submitting a PR.

[...]

PR #421, opened 2020-01-08
Add suffix to python kwrds 410

adamhaber:
- `await` and `async` are now in the keywords list.
- Code_gen had indentation problem with user defined functions, should be OK now.
- `Lit` removed from `translate_funapps_and_kwrds`
- User defined function properly handled (I hope) by checking all strings in `p.functions_block` and adding a case for `FunApp (UserDefined` in  `translate_funapps_and_kwrds`.

Following the discussion in #410 , the situation in which the Stan program contains both `lambda` and `lambda_` is not handled.


adamhaber:
Thanks for the helpful suggestion!

This adds an extra empty line between imports and class definition (in models without user defined function), so I've promote some of the earlier .py files.

adamhaber:
Removed WIP - I think this can be merged and we can handle the `lambda`/`lambda_` case later.

nhuurre:
Double underscore suffix is reserved in Stan but doesn't have special meaning in Python. You can rename `lambda -> lambda__` to avoid potential conflict with `lambda_`.

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

adamhaber:
Do you mean changing `let kwrds_suffix = "_"` to `let kwrds_suffix = "__"`? AFAIU, the problem with adding a suffix is quite general - whichever suffix we chose, if the original program has both `python_keyword` and `python_keyword+our_suffix` as identifiers, we're going to have a problem... right? OTOH, this may be too much of a corner case, and we can perhaps warn/fail if that's the case, and of course add this to the stan2tfp docs.

nhuurre:
Yes, I mean `let kwrds_suffix = "__"`
The original program cannot have any identifiers with `__` suffix because its reserved for internal use. (Try it.)
A warning may be necessary when renaming data or parameters because those change input/output behavour.
Sent with [ProtonMail](https://protonmail.com) Secure Email.

adamhaber:
> Yes, I mean `let kwrds_suffix = "__"`
The original program cannot have any identifiers with `__` suffix because its reserved for internal use. (Try it.)

Oh great, I didn't know that... thanks!

adamhaber:
Hmm, probably missing something very trivial, but for some reason the warning from [here](https://github.com/stan-dev/stanc3/blob/b8698864f00bf887f38686eeee162b00be2d6cd5/src/frontend/Semantic_check.ml#L599) is added to the generated TFP code...

adamhaber:
Pushed the new python_kwrds.py file, now including the warning. 

Perhaps that's a problem that should be handled at the level of the stan2tfp interface, and we're OK with the warning to be included in the .py file at the moment?

nhuurre:
This has been inactive for a while. If there's nothing more you want to add could you resolve the conflicting `dune` file and we can merge this?

The general problems with frontend warnings are beyond the scope of this. Including them in the `.py` is fine for now. (Although, we now have operator `%/%` for integer division. Tests models should use that unless explicitly testing the warning.)

The identifier warnings generated in `tfp_backend/Transform_mir.ml` are a different matter. I think either of the following options is fine:
* prefix the warning with `# ` so that it becomes a comment (rather than a syntax error) in the py file
* print it with `Fmt.epr`. It's the same as `Fmt.pr` but prints to the error stream instead of the output stream. That should show up in the console but not go to the output file.

adamhaber:
> This has been inactive for a while. If there's nothing more you want to add could you resolve the conflicting dune file and we can merge this?

Done.

> The general problems with frontend warnings are beyond the scope of this. Including them in the .py is fine for now.

Do such warnings always occur before the actual generated TFP code? If so, that makes fixing this at the interface level (like [stan2tfp](https://github.com/adamhaber/stan2tfp)) relatively easy. It's more of a workaround than a solution, though.   

> print it with Fmt.epr

Thanks, fixed it.

rok-cesnovar:
> it's almost as if you're the only one here who remembers to run `make format`

Guilty as charged :) 

Offtopic on this PR but would anyone oppose making `make format` a step in the Jenkins CI? Either: 
- fail if running make format produces any changes
- push make format changes automatically

Maybe the first one would be better for Git history?

Issue #423, opened 2020-01-10
[FR] Transformed Data Could Use OpenCL?


SteveBronder:
Can we move transformed data over to the OpenCL in the compiler?

The brms generated stan program below has a transformed data block of 

```cpp
transformed data {
  int Kc = K - 1;
  matrix[N, Kc] Xc;  // centered version of X without an intercept
  vector[Kc] means_X;  // column means of X before centering
  // matrices for QR decomposition
  matrix[N, Kc] XQ;
  matrix[Kc, Kc] XR;
  matrix[Kc, Kc] XR_inv;
  for (i in 2:K) {
    means_X[i - 1] = mean(X[, i]);
    Xc[, i - 1] = X[, i] - means_X[i - 1];
  }
  // compute and scale QR decomposition
  XQ = qr_thin_Q(Xc) * sqrt(N - 1);
  XR = qr_thin_R(Xc) / sqrt(N - 1);
  XR_inv = inverse(XR);
}
```
With a call to normal_id_glm_lpdf ala

```stan
parameters {
  vector[Kc] bQ;  // regression coefficients at QR scale
  // temporary intercept for centered predictors
  real Intercept;
  real<lower=0> sigma;  // residual SD
}
model {
  // priors including all constants
  target += student_t_lpdf(Intercept | 3, 0, 10);
  target += student_t_lpdf(sigma | 3, 0, 10)
    - 1 * student_t_lccdf(0 | 3, 0, 10);
  // likelihood including all constants
  if (!prior_only) {
    target += normal_id_glm_lpdf(Y | XQ, Intercept, bQ, sigma);
  }
}
```

If we can move XQ over to OpenCL then I think we would be hunky dory and a bunch of brms models could use the OpenCL stuff!

In the hpp file below you'll see that the compiler doesn't recognize transformed data as eligible for moving

Could the solution be just switching the [line here](https://github.com/stan-dev/stanc3/blob/2e79fb12ef08259b3cbfc8802ae3e22405396704/src/stan_math_backend/Transform_Mir.ml#L567) to be more than input_vars and also transformed data?

```ocaml
if !use_opencl then
      let data_var_idents = List.map ~f:fst p.input_vars in
      let switch_expr = switch_expr_to_opencl data_var_idents in
      let rec trans_stmt_to_opencl s =
        Stmt.Fixed.
          { s with
            pattern= Pattern.map switch_expr trans_stmt_to_opencl s.pattern }
      in
[...]

PR #424, opened 2020-01-10
OpenCL can now also use transformed parameters

SteveBronder:
Allows transformed data that is passed to an opencl capable function to be cached on the GPU

Licensed under BSD3

Do we need additional tests for this? Could add the brms model from #423

### DIFF
--- a/src/stan_math_backend/Transform_Mir.ml
+++ b/src/stan_math_backend/Transform_Mir.ml
@@ -564,7 +564,13 @@ let trans_prog (p : Program.Typed.t) =
   in
   let translate_to_open_cl stmts =
     if !use_opencl then
-      let data_var_idents = List.map ~f:fst p.input_vars in
+      let decl Stmt.Fixed.({pattern; _}) =
+      match pattern with
+        | Decl d ->
+          Some (d.decl_id)
+        | _ -> None
+      in
+      let data_var_idents = List.filter_map ~f:decl p.prepare_data in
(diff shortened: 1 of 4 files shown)


src/stan_math_backend/Transform_Mir.ml
```diff
@@ -564,7 +564,13 @@ let trans_prog (p : Program.Typed.t) =
...
   let translate_to_open_cl stmts =
     if !use_opencl then
-      let data_var_idents = List.map ~f:fst p.input_vars in
+      let decl Stmt.Fixed.({pattern; _}) =
+      match pattern with
+        | Decl d ->
+          Some (d.decl_id)
+        | _ -> None
+      in
+      let data_var_idents = List.filter_map ~f:decl p.prepare_data in
```
seantalts: Ah, I think you'll still need the input vars, no?

seantalts: (or wait, are those included in prepare data?)

SteveBronder: Oh my b I assumed prepare_data had all the data, lemme check this out

SteveBronder: Prepare data has both! I'll add that to the opencl test

SteveBronder:
Apologies this looks huge now I updated the OpenCL tests for GLM to also use transformed data. Think everything is hunky dory tho!

SteveBronder:
Cool beans ty!

Issue #425, opened 2020-01-12
TFP backend - Transformed data


adamhaber:
For the following Stan program:
```stan
data { 
    int<lower=0,upper=1> X; 
} 
transformed data {
    int X2 = X+1;
}
parameters { 
    real<lower=0,upper=1> p; 
} 
model { 
    X2 ~ bernoulli(p); 
} 
```
stan2tfp generates:
```python
class transformed_params_model(tfd__.Distribution):

  def __init__(self, X):
    self.X = X
     
  
  def log_prob_one_chain(self, params):
    target = 0
    X = self.X
    p = tf__.cast(params[0], tf__.float64)
    target += tf__.reduce_sum(tfd__.Bernoulli(None, p).log_prob(X2))
    return target
     
  def log_prob(self, params):
    return tf__.vectorized_map(self.log_prob_one_chain, params)
    
     
  def parameter_shapes(self, nchains__):
    X = self.X
    return [(nchains__, )]
     
  def parameter_bijectors(self):
    X = self.X
    return [tfb__.Chain([tfb__.Shift(tf__.cast(0, tf__.float64)),
                         tfb__.Scale(tf__.cast(1, tf__.float64) - tf__.cast(0, tf__.float64)),
                         tfb__.Sigmoid()])]
     
  def parameter_names(self):
    return ["p"]
     
model = transformed_params_model
```
in which X2 is not defined.

I assume we want to add `self.X2 = self.X + 1` in `__init__`, and then add `X2 = self.X2` everywhere we have `X = self.X`? If that's the case, I'll be happy to try and fix this (presumably by changing `pp_init`?).

nhuurre:
Sounds about right. In the C++ backend model class constructor executes the transformed data block. That corresponds to `__init__()` in Python.
`pp_init` reads data names from `p.input_vars`. The `transformed data` code is `p.prepare_data`.

brianwa84:
Fine for transformed data, but be careful with transformed parameters that
they end up in the scope of logprob, so that tf.GradientTape will properly
record ops for autodiff.

adamhaber:
Thanks @brianwa84 ! Regarding transformed parameters - defining them within the scope of `log_prob_one_chain` will cause their gradients to be recorded properly, but they won't be saved, right? 

@nhuurre - I've tried to fix this by changing `pp_init`'s:
```ocaml
  let ppbody ppf =
    match p.Program.input_vars with
    | [] -> pf ppf "pass"
[...]

PR #426, opened 2020-01-13
Add support for transformed data block

adamhaber:
Following the discussion in #425, this:

- Adds `self.X = ...` to the `__init__` function for X's from the transformed data block.
- Adds `X = self.X` to the `log_prob_one_chain` function for X's from the transformed data block.
- Adds some brms-style comments in the generated code (let me know if you think they're redundant).


adamhaber:
Promoted some earlier .py files, this should be fine now.

src/tfp_backend/Code_gen.ml
```diff
@@ -103,17 +103,35 @@ let pp_init ppf p =
...
+          (list ~sep:cut pp_prep_data_stmt)
+          p.Program.prepare_data
   in
   pp_method ppf "__init__" ("self" :: List.map ~f:fst p.input_vars) [] ppbody
 
 let pp_extract_data ppf p =
   let pp_data ppf (name, _) = pf ppf "%s = self.%s" name name in
   (list ~sep:cut pp_data) ppf p.Program.input_vars
 
+let pp_transformed_data ppf p =
```
seantalts: This is just printing out the transformed data var reassignments, right? Could we change the name to reflect that (and possibly share code with the above equivalent for data variables?

adamhaber: Renamed the functions to be more consistent with naming throughout the file, and added `pp_var_assignment` function that can is called from `pp_extract_data` and `pp_transformed_data`.

Issue #429, opened 2020-01-15
False positive on integer division warning


bgoodri:
When I do
```
./stanc.exe --allow_undefined --include_paths ~/stan-dev/rstanarm/src/stan_files/ --print-cpp ~/stan-dev/rstanarm/src/stan_files/mvmer.stan
```
I get an int division warning:
> Info: Found int division at '/home/ben/stan-dev/rstanarm/src/stan_files///functions/mvmer_functions.stan', line 422, column 11, included from
'/home/ben/stan-dev/rstanarm/src/stan_files/mvmer.stan', line 12, column 0:
  mean_PPD / N
Positive values rounded down, negative values rounded up or down in platform-dependent way.

The line in question is

https://github.com/stan-dev/rstanarm/blob/master/src/stan_files/functions/mvmer_functions.stan#L428

but `mean_PPD` is declared as `real` up above, so this is a double divided by int operation rather than integer division.

adamhaber:
This comes from `| UReal, UInt, Divide` here:

https://github.com/stan-dev/stanc3/blob/c63095d306f45b97364d985950b8cfd62509a7c9/src/frontend/Semantic_check.ml#L593-L605

which (probably) shouldn't be there. Should this be changed to `| UInt, UInt, Divide`?

seantalts:
Woops, yep. That might have been my bad in the initial spec.

seantalts:
This was fixed by #429.

PR #432, opened 2020-01-15
Fix 429 - removed real/int division, added int/int warnings

adamhaber:
Following #429 , now warns for int/real and int/int but not real/int.


bob-carpenter:
Only int/int should be flagged with a warning.

In C++, int/double and double/int are carried out with floating point arithmetic.

PR #433, opened 2020-01-17
bugfix 387 move context__.vals_r() calls out of for-loops

### LINKED ISSUE #387: context__.vals_r("Val") call in for loop is very very slow
The new compiler makes the matrices like

```cpp
      for (size_t sym1__ = 1; sym1__ <= K; ++sym1__) {
        current_statement__ = 13;
          current_statement__ = 13;
          assign(X,
                 cons_list(index_uni(sym2__),
                           cons_list(index_uni(sym1__), nil_index_list())),
                 context__.vals_r("X")[(pos__ - 1)], "assigning variable X");
          current_statement__ = 13;
          pos__ = (pos__ + 1);
        }
      }
```
On each iteration `context__.vals_r("X")`

1. constructs an std::string
2. does between 2 to 4 checks to see if the data exists somewhere
3. copies an std::vector<double> of the size of X which it then only grabs one value of before destroying the returned std::vector<double>.

If you have 100K observations, that's expensive. I manually moved this out of the loop in the generated .hpp and a model that took a day to load the data took 1 second.

```cpp
      {
        auto X_vals__ = context__.vals_r("X");
        for (size_t sym1__ = 1; sym1__ <= K; ++sym1__) {
          current_statement__ = 13;
            current_statement__ = 13;
            assign(X,
                   cons_list(index_uni(sym2__),
                             cons_list(index_uni(sym1__), nil_index_list())),
                   X_vals__[(pos__ - 1)], "assigning variable X");
            current_statement__ = 13;
            pos__ = (pos__ + 1);
          }
        }
      }
```

So we should do that

nhuurre:
Fixes #387

### DIFF
--- a/src/stan_math_backend/Stan_math_code_gen.ml
+++ b/src/stan_math_backend/Stan_math_code_gen.ml
@@ -238,8 +238,7 @@ let pp_ctor ppf p =
   let data_idents = List.map ~f:fst p.input_vars |> String.Set.of_list in
   let pp_stmt_topdecl_size_only ppf (Stmt.Fixed.({pattern; _}) as s) =
     match pattern with
-    | Decl {decl_id; decl_type; _}
-      when not (String.is_suffix decl_id ~suffix:"_flat__") -> (
+    | Decl {decl_id; decl_type; _} -> (
       match decl_type with
       | Sized st ->
           if Set.mem data_idents decl_id then pp_validate_data ppf (decl_id, st) ;
@@ -267,7 +266,7 @@ let pp_ctor ppf p =
 let pp_model_private ppf {Program.prepare_data; _} =
   let decl Stmt.Fixed.({pattern; _}) =
     match pattern with
-    | Decl d when not (String.is_suffix d.decl_id ~suffix:"_flat__") ->
+    | Decl d ->
         Some (d.decl_id, Type.to_unsized d.decl_type, UnsizedType.DataOnly)
     | _ -> None
   in
--- a/src/stan_math_backend/Stan_math_code_gen.ml
+++ b/src/stan_math_backend/Stan_math_code_gen.ml
@@ -238,7 +238,8 @@ let pp_ctor ppf p =
   let data_idents = List.map ~f:fst p.input_vars |> String.Set.of_list in
   let pp_stmt_topdecl_size_only ppf (Stmt.Fixed.({pattern; _}) as s) =
     match pattern with
-    | Decl {decl_id; decl_type; _} -> (
+    | Decl {decl_id; decl_type; _}
+      when not (String.is_suffix decl_id ~suffix:"_flat__") -> (
       match decl_type with
       | Sized st ->
           if Set.mem data_idents decl_id then pp_validate_data ppf (decl_id, st) ;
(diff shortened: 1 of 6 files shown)


rok-cesnovar:
Woohoo. Can confirm that this fixes the issue for the last mentioned model in the issue. Thanks @nhuurre

SteveBronder:
🙏 🙏 🙏  Feelin' blessed tysm! 🙏 🙏 🙏

SteveBronder:
Does this count as a bug fix or are we going to roll this out next release?

rok-cesnovar:
I think this should count as a bugfix.

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -266,7 +267,7 @@ let pp_ctor ppf p =
 let pp_model_private ppf {Program.prepare_data; _} =
   let decl Stmt.Fixed.({pattern; _}) =
     match pattern with
-    | Decl d ->
+    | Decl d when not (String.is_suffix d.decl_id ~suffix:"_flat__") ->
```
seantalts: This and the one above feels a bit hacky - can we instead generalize this to something like "compiler internal variables" and outsource to another function that checks if a declaration is compiler internal? Then we could treat `pos__` the same way, which would be fine, I think, and we can have many less instances of `"_flat__"` in the code.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -64,49 +64,72 @@ let rec switch_expr_to_opencl available_cl_vars (Expr.Fixed.({pattern; _}) as e)
...
+  | UInt | UReal ->
+      [ Assignment
+          ( (decl_id, unsized, [])
+          , { Expr.Fixed.pattern=
+                Indexed (readfnapp decl_var, [Single Expr.Helpers.loop_bottom])
+            ; meta= {decl_var.meta with type_= unsized} } )
+        |> swrap ]
+  | UArray UInt | UArray UReal ->
+      [Assignment ((decl_id, flat_type, []), readfnapp decl_var) |> swrap]
+  | _ ->
```
seantalts: I think it would be good to explicitly list the types we're matching on here because in the future we may add new types and we want the compiler to warn us about places where we'll need to make decisions.

src/stan_math_backend/Stan_math_code_gen.ml
nhuurre: Everything about the code generation around there is hacky. I'll wrap the flat__ array inside a block, that gets rid of this particular hack.

src/stan_math_backend/Transform_Mir.ml
nhuurre: Agreed. I guess UFun and UMathLibraryFunction shouldn't be allowed?

seantalts: probably not, yeah

src/stan_math_backend/Stan_math_code_gen.ml
seantalts: That's kind of subtle - is there anything wrong with something like `Decl d when not (compiler_internal d) ->`? (there may be, I didn't try it). Assuming compiler_internal basically checks if the variable name ends with `__`.

nhuurre: Not that I know of. But IMO it's the job of Transform_mir to move the toplevel decls to wherever they belong, the code printer should not try and figure it out. But that requires a bigger MIR refactor.

seantalts: That would require a Stan Math-specific LIR - something we may want in the future but doesn't currently exist. There are pros and cons to that approach - the major pro of abusing the MIR is that we can re-use analysis and optimization on code that is much closer to getting printed out.

PR #434, opened 2020-01-17
Pedantic mode

rybern:
NOTE: This post is outdated now, I posted a new version in the comments on Jan. 29

```
I've done a first pass at implementing a good number of the warnings for
pedantic mode. Bugs are likely, consider all this experimental.

All of them are behind a CLI flag --warn-pedantic. I'm not sure that's a
good long term solution. Some of them, like warning about
parameter-dependent control flow, should perhaps be on all the time.
There should probably be way to separate out those with high and low
false positive rates.

See test/unit/Pedantic_analysis.ml for examples and the actual error
messages.


1 Completed warnings
====================

1.1 Parameter-dependent control flow
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

1.1.1 Description
-----------------

  Control flow statements in the log_prob section should not depend in
  any way on the value of parameters, else they might introduce
  discontinuity.


1.1.2 Implementation notes
--------------------------

  Heavy use of dependence analysis. Iterates through all control flow
  statements, finds all the dependencies of their branching decision
  expressions, and checks that those have no parameter dependencies


1.2 Uniform distributions
~~~~~~~~~~~~~~~~~~~~~~~~~

1.2.1 Description
-----------------

  All uses of uniform() should be flagged. Just about all the examples
  we've ever seen are either superfluous (as they just add a constant to
  the log density) or mistaken (in that they should have been entered as
  bounds on parameters).


1.2.2 Implementation notes
--------------------------

  Search for uniform_propto_log, throw warning


1.3 Parameter defined but never used
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

1.3.1 Implementation notes
--------------------------

  Search all expressions in log_prob and function bodies for parameter
  references.

  I did not check if functions were of the target-modifying type, but
  IIUC that shouldn't be necessary to check since there's no shadowing,
  right?

  Note: I did not check if the usage would actually put a distribution
  on the parameter, I only check that it's mentioned. The former is
  probably what we actually want. TODO: I should actually build the
  factor graph and check that all parameters are in the factor graph.


1.4 Parameter on LHS of multiple twiddles
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

1.4.1 Implemenation notes
-------------------------

  Search program for twiddles (which only look like top-level TargetPE
  plus a distribution), look for duplicate LHS parameters

  Only catches multiple twiddles in the code, not execution, so does not
  e.g. catch twiddles within a loop.

  Does not handle array indexing at all, only string matches the
  parameters.


1.5 Parameter with /=1 priors
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

1.5.1 Description
-----------------

  Warn user if parameter has no priors or multiple priors Bruno
  Nicenboim suggested this on
  <https://github.com/stan-dev/stan/issues/2445>)


1.5.2 Implementation notes
--------------------------

  The definition of 'prior' seems tricky in Stan. I came up with a
  definition that makes sense to me.

  A likelihood is P(X|D,Y), a prior is P(X|Y), where Y are non-data
  variables. So the important feature seems to be the lack of dependence
  on data. But not 'dependence' in the programming sense, dependence in
  the probabilistic sense.

  We can use a factor graph to translate the idea to Stan. If we're
  wondering whether a neighboring factor F of a variable V is a prior,
  we should check whether F has any connection to the data that isn't
  intermediated by V. To do that, we can remove V from the graph and
  look for any path between F and the data using BFS.

  The results using this definition seem to match my intuition, but I'm
  betting others will have some thoughts.


1.6 Uninitialized variables
~~~~~~~~~~~~~~~~~~~~~~~

1.6.1 Implemenation notes
-------------------------

  I haven't worked on this for the PR, I just added it to the
  --warn-pedantic flag and relocated the code.

  It still does not handle array elements, that's another big TODO.


1.7 Large or small numbers
~~~~~~~~~~~~~~~~~~~~~~~~~~

1.7.1 Description
-----------------

  Andrew's suggested message: Warning message: "Try to make all your
   parameters scale free. You have a constant in your program that is
   less than 0.1 or more than 10 in absolute value on line **. This
   suggests that you might have parameters in your model that have not
   been scaled to roughly order 1. We suggest rescaling using a
   multiplier; see section *** of the manual for an example.


1.7.2 Implementation notes
--------------------------

  Look though all expressions for large numbers. I'm guessing there will
  be a lot of false positives, I'm wondering how best to narrow it down
  to the real issue.

   I also allowed 0 without a warning. CLOSED: [2020-01-11 Sat 11:35]


1.8 Parameters starting with "sigma"
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

1.8.1 Description
-----------------

  Any parameter whose name begins with "sigma" should have "lower=0" in
  its declaration; otherwise flag.


1.9 Parameter bounds
~~~~~~~~~~~~~~~~~~~~

  Parameter bounds of the form "lower=A, upper=B" should be flagged in
  all cases except A=0, B=1 and A=-1, B=1.


1.9.1 Implementation notes
--------------------------

  I was a little fuzzy on when bounds will be Ints vs. Reals. I ended up
  casting everything to float, which might backfire.


1.10 Positive-constrained distributions without corresponding bounds
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

1.10.1 Description
------------------

  Parameters with positive-constrained distribution (such as gamma or
  lognormal) and no corresponding constraint in the definition.


1.10.2 Implemenation notes
--------------------------

  I ended up with a reasonably general system for flagging questionable
  uses of distributions, so it should be easy to add more like this. See
  distribution_warning in Pedantic_analysis.ml

1.11 Invalid distribution parameters
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

1.11.1 Description
------------------

  gamma(A,B) or inv_gamma(A,B) should be flagged if A = B < 1. (The
  point is to catch those well-intentioned but poorly-performing
  attempts at improper priors.)


1.11.2 Implemenation notes
--------------------------

  Uses the same distribution warning system


2 General to-dos
================

2.1 More granular interface
~~~~~~~~~~~~~~~~~~~~~~~~~~~

  I think it might be cool if you could turn particular warnings on/off
  with CLI flags, haven't tried that yet


2.2 Some warnings would ideally be after optimizations
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  Some warnings look at literal values, so they would benefit from
  propagation and partial evaluation.

  I think you can still get the original location information fine with
  metadata, and there shouldn't be other pitfalls of using nonoriginal
  mir.


2.3 More distribution properties
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  I included just the ones that were given as examples, but it would be
  very easy to add more.


3 Possible warnings left to do
==============================

  * Cauchy as prior for variance
  * Code has no indentation.
  * Code has blank lines.
  * Vectorization without first argument
  * Index var used without indices in loop
  * Suspicious indentation
  * Check loop bounds when possible
  * Overpromoting
  * Missing Jacobian adjustment
```


bob-carpenter:
1.1  A simpler form of if/else would just check if the condition involves a `var` object rather than a `double`.

1.2  Flagging all uniform distribution uses will generate false positives.  In some cases the bounds are parameters and in some cases I believe we need the normalizing constants.  So I think we should flag cases where the variate is a parameter that isn't bounded.  Then the uniform() becomes redundant, but a lot of people have said they like the model block being a complete description and it's not that much of a performance hit as it's a constant after bounds checking.

1.3  Correct, no variable masking (I assume that's what shadowing is).

1.4  This is also going to have a lot of false positives.  I'm not sure what "does not handle array indexing at all" means.  What if we have 
```
a[0] ~ foo(theta);
for (n in 2:N)
  a[n] ~ bar(phi);
```
Does that count as `a` being on the LHS of two twiddles?  What about
```
for (n in 1:N) {
  a[n] ~ bar(phi);
  a[n] ~ foo(theta);
}
```
And what about the soft-centering we sometimes use:
```
alpha ~ normal(mu, sigma);
sum(alpha) ~ normal(0, epsilon);
```

1.5 This seems redundant with issue 1.4.   Also, there's no coherent definition of "prior" in Stan as it only requires the definition of a proper density.  The standard definition is `p(y, theta) = p(y | theta) * p(theta)` where `y` is data, `theta` parameters, and indeed the prior `p(theta)` does not depend on data.   This issue should be brought up with everyone, but particularly @andrewgelman.

1.6  Are undefined variables ones that are declared in a block other than data or parameters and then never defined?  This is an important one.

1.7  This large or small numbers thing is motivated by (a) default priors, and (b) adaptation.  It's not clear what is being done for large or small numbers.  For adaptation, the offset and multiplier transform makes up for (b).  So when we have
```
real<multiplier = 1e5> a;
...
a ~ normal(0, 1e5);
```
is totally OK as far as adaptation because the unconstrained representation of `a` will be on the unit scale.

1.8   What you're going to find is a gazillion people writing things like:
```
real<lower = 1e-5> sigma;
```
as a holdover from BUGS that had numerical issues when you get near zero.  I believe @andrewgelman is assuming people are following his and Jennifer Hill's notation in their regression book where you have `y ~ normal(mu_y, sigma_y)`.  But what you find is that our users are just as prone to write `y_sigma` as `sigma_y` and also prone to use other variables for scales.  But I don't see this is going to hurt anything.

1.9  Parameter bounds of [-1, 1] make sense for correlations.  And a lot of people convert survey data to a fixed continuous scale, like [1, 5].    Any cast should be to `double`, not `float`.  Anywhere a `real` is allowed an `int` is also allowed, but not vice-versa.

1.10 This one's great if we can find it.

1.11 is OK, but it's again going to be very scale-dependent and potentially yield false positives.

2.2 would also be great, but is of course, impossible to do in full generality statically because of undecidability.

3.  Individual comments:
* For no indents, extra blank lines, suspicious indedentation, I'd want to look at pure syntactic lint like missing spaces separately.  Ideally, we'd just have an auto-pretty-printer we could apply.  But I can see it making sense to also flag such programs.  We could also flag missing spaces around operators and alignment across lines.
* I don't know what "vectorization without first argument" means.  We run into problems when users make this mistake:
```
vector[5] x = ...;
for (n in 1:5)
  x ~ normal(0, 1);
```
It should be `x[n]` in the loop and what winds up happening is that we apply the same prior multiple times.
* Index var without indices used in loop isn't always an error, but most of the time it will be.
* How do we check loop bounds?  Against indexing?
* By overpromoting, is that something like defining a constant in the transformed parameter block?  Or a transformed parameter that could be a generated quantity?  These would be good to transform automatically if we could, but if not, flagging them is better than nothing.
* The missing Jacobian's been tricky because it's hard to tell if the operation applied to the LHS is linear.

rybern:
Thanks Bob. Note that most of these are collected from the Wiki and Discourse, so the motivation for each isn't necessarily coming from me.

1.1 I think the dependency approach is what we want, I'll try to convince you:

Let x be a parameter. This is clearly an issue:
   if (x > 1) ..

Still an issue:
   real y = x + 1;
   real z = y * 10;
   if (z > 1) ..

Not an issue:
   real y = 1;
   real z = y * 10;
   if (z > 1) ..

So it seems to me we have to search beyond the expression.

By the way, I think this applies as much to any control flow (I had it misnamed it in the post, fixed now).

1.2 It's currently only flagging cases where the first argument is a parameter (is that what variate means?).

I could see only flagging cases with unbounded parameters, but you will also miss cases where the bounds don't match, and it's not necessarily possible to check if they do match. Single line implementation change though, if we do want only unbounded.

1.4 Yeah, there's a lot more work to do here. None of your examples will be flagged, I don't catch twiddles more complex than `var ~` at all in an effort to be conservative. Eventually, I would say that the first example wouldn't be flagged and the second one would, and the third one probably shouldn't be (as it's cumbersome to express otherwise).

1.5 This isn't exactly like 1.4 in that it isn't looking for twiddles, it's looking for factors. I agree that there should be a discussion around the goal of this warning and around the definition of prior. 

1.6 Undefined should read uninitialized (updated my post). It's looking for variables that are read somewhere on the RHS before being written to.

1.7 This will be trickier to write. Should numbers only ever be flagged when they're literals in a distribution?

1.8 I agree that rigid string matching isn't ideal. Perhaps a better approach would be to use the same method as 1.11 and check bounds of any parameter used in the variance position of a normal distribution, regardless of name.

I could also only flag parameters with missing or negative lower bounds, rather than 0.

1.9 Yeah, I'm not sure what bounds should be flagged.
The cast is to a double precision number, they're just called floats in OCaml.

1.10 I don't see why we can't have a big table of distribution properties and flag whenever we can prove that they're violated. The positive constraint is currently only done for gamma and lognormal, I (or someone) should sit down and complete the full list.

3. I honestly haven't thought too much about these yet, I've just collected the ideas, so I can't clarify much.

andrewgelman:
Hi, see quick comments below on the items where I have thoughts/expertise.
Andrew

> On Jan 20, 2020, at 2:42 PM, Bob Carpenter <notifications@github.com> wrote:
> 1.2 Flagging all uniform distribution uses will generate false positives. In some cases the bounds are parameters and in some cases I believe we need the normalizing constants. So I think we should flag cases where the variate is a parameter that isn't bounded. Then the uniform() becomes redundant, but a lot of people have said they like the model block being a complete description and it's not that much of a performance hit as it's a constant after bounds checking.
> 
OK, there are a few possibilities here of interest.  First, what if a parameter is bounded and has a bounded uniform prior.  For example we have a correlation parameter bounded between -1 and 1, and we give it a uniform(-1, 1) prior.  As Bob said, this is fine, and maybe some people like the redundancy.  Second, what if we have a bounded parameter with a uniform prior but for parameterization reasons we don't want it given the hard constraint.  So, for example, we have a parameter with a uniform(-1,1) prior but in the declaration we don't bound it, so that way we won't get the logit transformation.  This might make sense, or it might not, but it could create computational issues.  So I don't mind flagging it.  Third is when the uniform is done in a mathematically valid but statistically inappropriate way.  For example I recall seeing a model in pharma or astro (I can't recall which) where some parameter was given a uniform(-2, 5) prior.  (This was the log of some physical parameter.)  I can't believe this would be a big deal.  The fourth setting I can think of is when people give a prior such as uniform(0,1000) as a way of setting a weak prior. This might not be horrible, but I think it's bad practice.

In short, I think it's ok to flag the uniform priors, at least most of the time, in that I suspect that most of the "false positives" are actually true positives!  But I'd be ok with Bob's suggestion--only flag the uniform prior when it does not correspond to the bounds in the parameter declaration--as I guess this would catch most of it.

> 1.8 What you're going to find is a gazillion people writing things like:
> 
> real<lower = 1e-5> sigma;
> as a holdover from BUGS that had numerical issues when you get near zero. I believe @andrewgelman <https://github.com/andrewgelman> is assuming people are following his and Jennifer Hill's notation in their regression book where you have y ~ normal(mu_y, sigma_y). But what you find is that our users are just as prone to write y_sigma as sigma_y and also prone to use other variables for scales. But I don't see this is going to hurt anything.
> 
If people are writing real<lower=1e-5> sigma; then I do think this should be flagged!

Also, sure, people can write "y_sigma" or "sgima" or all sorts of things.  We shouldn't imply to people that pedantic mode is a magic code cleaner; it's just trying to help.  Actually, this suggests to me that in the pedantic mode doc, we should give examples of obvious errors that are _not_ caught by pedentic mode.  This could help dispel overconfidence.
> 1.9 Parameter bounds of [-1, 1] make sense for correlations.
> 
I agree that <lower=-1,upper=1> can make sense.
> And a lot of people convert survey data to a fixed continuous scale, like [1, 5].
> 
 <lower=-1,upper=5> can make sense for _data_, but I think it's very rare that it would make sense for _parameters_.

bob-carpenter:
> On Jan 20, 2020, at 4:24 PM, rybern <notifications@github.com> wrote:
> 
> Thanks Bob. Note that most of these are collected from the Wiki and Discourse, so the motivation for each isn't necessarily coming from me.
> 
> 1.1 I think the dependency approach is what we want, I'll try to convince you:

You don't need to convince me.  I was just suggesting a much simpler
alternative that isn't as precise.


> 1.2 It's currently only flagging cases where the first argument is a parameter (is that what variate means?).

Usually in foo(y | theta1, ..., thetaN), the variable y is the variable
being given a distribution and theta1, ..., thetaN are parameters.  


> 1.6 Undefined should read uninitialized (updated my post). It's looking for variables that are read somewhere on the RHS before being written to.

I think "undefined" is OK, because the variables get declared, but never
get defined.  They'll technically be initialized in some blocks to NaN.


> 1.7 This will be trickier to write. Should numbers only ever be flagged when they're literals in a distribution?

That'd be one way to reduce false positives.

> 
> 1.8 I agree that rigid string matching isn't ideal. Perhaps a better approach would be to use the same method as 1.11 and check bounds of any parameter used in the variance position of a normal distribution, regardless of name.

That'd be great if it's possible

> 
> I could also only flag parameters with missing or negative lower bounds, rather than 0.

That's probably also useful other than the correlation parameter case which is naturally [-1, 1] bounds.


> 1.9 Yeah, I'm not sure what bounds should be flagged.
> The cast is to a double precision number, they're just called floats in OCaml.

OK.  C++ uses "float" for single precision and "double" for double precision and "long double" for quadruple precision.


> 1.10 I don't see why we can't have a big table of distribution properties and flag whenever we can prove that they're violated. The positive constraint is currently only done for gamma and lognormal, I (or someone) should sit down and complete the full list.

The constraints on parameters are all listed in the manual.  Most of the distributions have constrained parameters.  For instance, poisson(lambda) requires lambda to be positive, student_t(dof, mu, sigma) requires dof and sigma to be positive, exp(lambda) requires lambda to be positive, etc. etc.

> 
> 	• I honestly haven't thought too much about these yet, I've just collected the ideas, so I can't clarify much.

bob-carpenter:
> OK, there are a few possibilities here of interest. First, what if a parameter is bounded and has a bounded uniform prior. For example we have a correlation parameter bounded between -1 and 1, and we give it a uniform(-1, 1) prior. As Bob said, this is fine, and maybe some people like the redundancy. 

Yes, several people have told me that when I suggested getting rid of uniform(...).


> Second, what if we have a bounded parameter with a uniform prior but for parameterization reasons we don't want it given the hard constraint. So, for example, we have a parameter with a uniform(-1,1) prior but in the declaration we don't bound it, so that way we won't get the logit transformation. This might make sense, or it might not, but it could create computational issues. So I don't mind flagging it. 

This will almost certainly create computational issues if HMC otherwise takes it beyond a boundary.  


> Third is when the uniform is done in a mathematically valid but statistically inappropriate way. For example I recall seeing a model in pharma or astro (I can't recall which) where some parameter was given a uniform(-2, 5) prior. (This was the log of some physical parameter.) I can't believe this would be a big deal. 

To me, these are the same category as inv_gamma(1e-4, 1e-4) or
normal(0, 1e3).


> The fourth setting I can think of is when people give a prior such as uniform(0,1000) as a way of setting a weak prior. This might not be horrible, but I think it's bad practice.

I don't think we'll be able to separate this from (3) w/o inferring
appropriate scales.


> In short, I think it's ok to flag the uniform priors, at least most of the time, in that I suspect that most of the "false positives" are actually true positives! 

I'm OK flagging this stuff, too.  Ideally, we'd have a way to
add a NOLINT line like we can for cpplint if the person wants to
tell the pedantic parser to shut up.

> But I'd be ok with Bob's suggestion--only flag the uniform prior when it does not correspond to the bounds in the parameter declaration--as I guess this would catch most of it.

That's much harder to catch.


> > 1.8 What you're going to find is a gazillion people writing things like:
> > 
> > real<lower = 1e-5> sigma;
> > as a holdover from BUGS that had numerical issues when you get near zero. I believe @andrewgelman <https://github.com/andrewgelman> is assuming people are following his and Jennifer Hill's notation in their regression book where you have y ~ normal(mu_y, sigma_y). But what you find is that our users are just as prone to write y_sigma as sigma_y and also prone to use other variables for scales. But I don't see this is going to hurt anything.
> > 
> If people are writing real<lower=1e-5> sigma; then I do think this should be flagged!

Agreed.  And speaking to my earlier point, I just saw a practice talk with the equivalent of y_sigma.  So maybe it should flag "sigma" anywhere in the string.


> Also, sure, people can write "y_sigma" or "sgima" or all sorts of things. We shouldn't imply to people that pedantic mode is a magic code cleaner; it's just trying to help. Actually, this suggests to me that in the pedantic mode doc, we should give examples of obvious errors that are _not_ caught by pedentic mode. This could help dispel overconfidence.

Sure.


> > 1.9 Parameter bounds of [-1, 1] make sense for correlations.
> > 
> I agree that <lower=-1,upper=1> can make sense.
> > And a lot of people convert survey data to a fixed continuous scale, like [1, 5].
> > 
> <lower=-1,upper=5> can make sense for _data_, but I think it's very rare that it would make sense for _parameters_.

We can probably make that distinction.  What I've seen on the forums is users having a mean constrained to <lower = 1, upper = 5> if the survey question was.

andrewgelman:
Hi, see responses below. 
Andrew

> On Jan 20, 2020, at 8:40 PM, Bob Carpenter <notifications@github.com> wrote:
> 
> 
> > Third is when the uniform is done in a mathematically valid but statistically inappropriate way. For example I recall seeing a model in pharma or astro (I can't recall which) where some parameter was given a uniform(-2, 5) prior. (This was the log of some physical parameter.) I can't believe this would be a big deal. 
> 
> To me, these are the same category as inv_gamma(1e-4, 1e-4) or
> normal(0, 1e3).

I think inv_gamma(1e-4, 1e-4) is worse because it's a really strong prior that people think is very weak.
I think normal(0, 1e3) is not as bad as uniform(-1e3, 1e3).  The uniform is worse because it doesn't partially pool toward 0.
As discussed in the priors wiki, I think that if someone was going do use uniform(-2, 5), it would be better to do something like normal(1.5, 1.75).

> 
> > The fourth setting I can think of is when people give a prior such as uniform(0,1000) as a way of setting a weak prior. This might not be horrible, but I think it's bad practice.
> 
> I don't think we'll be able to separate this from (3) w/o inferring
> appropriate scales.

It's just about impossible for me to imagine a real setting where someone would want to do uniform(0, 1000).  If for example this is a scale parameter in a setting where the number 1000 is meaningful, I still wouldn't think that uniform is a good idea.

> > In short, I think it's ok to flag the uniform priors, at least most of the time, in that I suspect that most of the "false positives" are actually true positives! 
> 
> I'm OK flagging this stuff, too. Ideally, we'd have a way to
> add a NOLINT line like we can for cpplint if the person wants to
> tell the pedantic parser to shut up.

Yes, I agree that NOLINT should be an option; otherwise the parser could be annoying.

> > > And a lot of people convert survey data to a fixed continuous scale, like [1, 5].
> > > 
> > <lower=-1,upper=5> can make sense for _data_, but I think it's very rare that it would make sense for _parameters_.
> 
> We can probably make that distinction. What I've seen on the forums is users having a mean constrained to <lower = 1, upper = 5> if the survey question was.
> 

I can imagine someone doing that.  But I think it's a bad idea.

rybern:
I pushed an update:

- Scale parameters are now flagged by usage (currently only second parameter of `normal`), instead of matching by variable name, and the warning points to the usage site
- Uniform distributions now warn only if the variate bounds certainly differ from the distribution bounds
- Unused parameters are now flagged if they do not contribute to the value of target, rather than if they appear in the program at all
- Large/small numbers are only flagged if they appear as distribution arguments

bob-carpenter:
The unused parameter thing's super cool---that requires dataflow through everything.

You should coordinate with Andrew more than me on utility and false-positive/false-negative tradeoffs (or whatever Andrew wants to call erroneous and missing warnings). For instance, we might flag all uniform() distributions on efficiency grounds and reduce false negatives while increasing false positives.

rybern:
Another update:
I went through almost all of the distributions (missing some of the matrix distributions) in the Function Reference and added checks for parameter bounds and literals for the variate and argument positions. Crux of the code is [here](https://github.com/stan-dev/stanc3/pull/434/files#diff-862c4d435a0969296f57767c6147a563R133).

This should catch all variate parameters, scale parameters, shape parameters, degrees of freedom, etc, which are either insufficiently constrained parameters or are constants out of range.

I could do this for data variables (and thus discrete distributions like Bernoulli) if I could figure out how to access bounds for data variables, which seems like it will take some work.

I'll see if I can find a good way to test all this better, I currently only have tests for a fraction of them.

rybern:
I've done a lot more work on this since I first submitted the pull
request, so I'm posting a full updated description here along with tags for
pieces that are different from the original. I thought I'd leave the original up to
preserve the discussion.

I think this is just about ready to merge as a sort of alpha release.
Very happy to have feedback! @andrewgelman @bob-carpenter or anyone else.

```
I folded these original sections into the 'Distribution warnings' section:
* Uniform distributions
* Parameters starting with "sigma"
* Positive-constrained distributions without corresponding bounds
* Invalid distribution parameters


1 Completed warnings
====================

1.1 [Updated] Distribution warnings
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

1.1.1 Argument and variate constraint warnings
----------------------------------------------

  There is a warning for each constrained argument of each built-in
  distribution, based on the information from the Functions Reference.
  These include for example inclusive/exclusive upper and lower bounds,
  covariance matrices, cholesky correlation matrices, simplexes, etc.

  An exception is discrete distributions. I can't yet check the bounds
  of discrete variables or data variables. That'll be a future update.

  An argument constraint is checked for consistency against the
  parameter declaration or literal value (or what becomes a literal
  value after partial evaluation). For example, if a parameter is used
  as a scale parameter and is constrained to be lower=1, no warning is
  generated, but if it were constrained lower=-1, a warning is
  generated.

  Warning messages try to be as descriptive as possible, including
  English descriptions of the argument role (e.g. "a scale parameter")
  and the constraint (e.g. "constrained to be positive"), as well as the
  distribution name, variable name and location.

  Here's an example message pulled from a test in
  test/unit/Pedantic_mode.ml:
  
   Warning at 'string', line 84, column 17 to column 22:
     A chi_square distribution has parameter unb_p as degrees of freedom
     (argument 1), but unb_p is not constrained to be positive.

  This language could probably be improved if anyone wants to reformat
  it.

  Speaking of tests, all of the warnings have at least one test in the
  above mentioned file. There will likely still be bugs if I
  misinterpreted the Function Reference.


1.1.2 Special distribution warnings
-----------------------------------

* 1.1.2.1 Uniform distribution

  Warn on any use when the variate parameter's bound constraint doesn't
  match the uniform bounds


* 1.1.2.2 (Inverse) Gamma distribution

  Warn when arguments indicate that it might be a poor attempt at an
  improper prior


* 1.1.2.3 lkj_corr distribution

  Warn on use to suggest using Cholesky variant


1.2 [Updated] Parameter defined but never used
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  I now build a factor graph and check that there are no declared
  parameters missing from the factor graph. This should effectively
  check if any factors don't contribute (even indirectly) to the target
  value.


1.3 [Updated] Large or small numbers
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  Update: Only checking numbers which are used as arguments to built-in
  distributions.


1.3.1 Description
-----------------

  Andrew's suggested message: Warning message: "Try to make all your
   parameters scale free. You have a constant in your program that is
   less than 0.1 or more than 10 in absolute value on line **. This
   suggests that you might have parameters in your model that have not
   been scaled to roughly order 1. We suggest rescaling using a
   multiplier; see section *** of the manual for an example.


1.3.2 Implementation notes
--------------------------

  Look though all expressions for large numbers. I'm guessing there will
  be a lot of false positives, I'm wondering how best to narrow it down
  to the real issue.

  I also allowed 0 without a warning.


1.4 if/else in trans. params or model blocks
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

1.4.1 Description
-----------------

  Control flow statements in the log_prob section should not depend in
  any way on the value of parameters, else they might introduce
  discontinuity.


1.4.2 Implementation notes
--------------------------

  Heavy use of dependence analysis. Iterates through all control flow
  statements, finds all the dependencies of their branching decision
  expressions, and checks that those have no parameter dependencies


1.5 Parameter on LHS of multiple twiddles
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

1.5.1 Implemenation notes
-------------------------

  Search program for twiddles (which only look like top-level TargetPE
  plus a distribution), look for duplicate LHS parameters

  Only catches multiple twiddles in the code, not execution, so does not
  e.g. catch twiddles within a loop.

  Does not handle array indexing at all, only string matches the
  parameters.


1.6 Parameter with /=1 priors
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

1.6.1 Description
-----------------

  Warn user if parameter has no priors or multiple priors Bruno
  Nicenboim suggested this on
  <https://github.com/stan-dev/stan/issues/2445>)


1.6.2 Implementation notes
--------------------------

  The definition of 'prior' seems tricky in Stan. I came up with a
  definition that makes sense to me.

  A likelihood is P(X|D,Y), a prior is P(X|Y), where Y are non-data
  variables. So the important feature seems to be the lack of dependence
  on data. But not 'dependence' in the programming sense, dependence in
  the probabilistic sense.

  We can use a factor graph to translate the idea to Stan. If we're
  wondering whether a neighboring factor F of a variable V is a prior,
  we should check whether F has any connection to the data that isn't
  intermediated by V. To do that, we can remove V from the graph and
  look for any path between F and the data using BFS.

  The results using this definition seem to match my intuition, but I'm
  betting others will have some thoughts.


1.7 Undefined variables
~~~~~~~~~~~~~~~~~~~~~~~

1.7.1 Implemenation notes
-------------------------

  I haven't worked on this for the PR, I just added it to the
  --warn-pedantic flag and relocated the code.

  It still does not handle array elements, that's another big TODO.


1.8 Parameter bounds
~~~~~~~~~~~~~~~~~~~~

  Parameter bounds of the form "lower=A, upper=B" should be flagged in
  all cases except A=0, B=1 and A=-1, B=1.


1.8.1 Implementation notes
--------------------------

  I was a little fuzzy on when bounds will be Ints vs. Reals. I ended up
  casting everything to float, which might backfire.


2 General to-dos (some now completed)
=====================================

2.1 [Completed] Some warnings would ideally be after optimizations
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  Constant propagation and partial evaluation are now run before
  warnings are generated.


2.2 [Completed] More distribution properties
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  I've included constraint warnings from all built-in distributions, as
  well as some special distribution-specific ones.


2.3 More granular interface
~~~~~~~~~~~~~~~~~~~~~~~~~~~

  I think it might be cool if you could turn particular warnings on/off
  with CLI flags, haven't tried that yet


3 [Updated] To-do in future pull requests
=========================================

3.1 Improve location span printing
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  Currently location spans (including in warning messages) are a tad
  verbose in my opinion


3.2 Handle array elements in dependency analysis
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  This is potentially challenging


3.3 Figure out how to persist data variable constraints into the MIR
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  When I can do this, I also catch more issues with discrete
  distributions


3.4 Possible additional warnings types
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  I haven't thought about these much yet. I went for high value per
  effort ones first. E.g., I think the indentation ones might require a
  lot of lexer/parser work.
  * Cauchy as prior for variance
  * Code has no indentation.
  * Code has blank lines.
  * Vectorization without first argument
  * Index var used without indices in loop
  * Suspicious indentation
  * Check loop bounds when possible
  * Overpromoting
  * Missing Jacobian adjustment
```

bob-carpenter:
Is there an easy way to test this on examples I make up?

rybern:
Right now you'd have to check out my branch, compile it, and compile a model with --warn-pedantic on.

mcol:
See also this discussion https://github.com/stan-dev/stan/issues/1918.

rok-cesnovar:
@rybern if you have access to a mac to build the stanc executable we could make a cmdstan branch that would be easier to test this pedantic mode with. I can provide linux/win executables. I think that way testing this will get more traction.

rybern:
Thanks Rok, that's a good idea - unfortunately do not have access to a mac, I'll ask around

rybern:
@rok-cesnovar @bbbales2 @mitzimorris 
Following up on the meeting discussion, it sounds like a reasonable solution:
- I should fork CmdStan and add precompiled binaries, then publish a CmdStanR one-liner for my fork
- We should also work on the installation scripts to lower friction for people to checkout and build Stanc
Are those reasonable?

mitzimorris:
yes, reasonable, given the use case.

mitzimorris:
@serban-nicusor-toptal how are the nightly builds generated?
can these scripts be made to generate a build from any branch?
we need an easy way to generate builds for all platforms; also a more elegant solution as to how to include these builds with CmdStan.

mitzimorris:
Jenkinsfile logic to build releases:  https://github.com/stan-dev/stanc3/blob/ea5bea00b6282a9816c5a646f03fafed275b9221/Jenkinsfile#L158

bbbales2:
> then publish a CmdStanR one-liner for my fork

Yeah double check the error messages show up and the install from branch works (https://mc-stan.org/cmdstanr/reference/install_cmdstan.html)

> We should also work on the installation scripts to lower friction for people to checkout and build Stanc

I would like it if I could

```
apt install all-the-caml-stuff
git clone
build_compiler
```

To take the load of rok building me custom cmdstans.

serban-nicusor-toptal:
@mitzimorris Yes.
We can generate a build from any branch.
The release itself is done using [ghr](https://github.com/tcnksm/ghr) based on tags in Jenkinsfile [here](https://github.com/stan-dev/stanc3/blob/ea5bea00b6282a9816c5a646f03fafed275b9221/Jenkinsfile#L247)

mitzimorris:
@serban-nicusor-toptal - please help me connect the dots on what needs to be done:

> The release itself is done using ghr based on tags in Jenkinsfile here

given branch `rybern::pedantic`,  how does one run Jenkins file stage "Release tag and publish binaries"?
what is the URL to the published binaries?

who has perms to do this?  Rok?

rok-cesnovar:
Currently that stage is only run on merge to master, that builds the nightly build and the files are accesible in the releases tab under assests.

We discussed this awhile ago with Nic. What we would need to do is have something like the Custom job in math where someone with permission could input a branch and that would produce binaries.

I dont have any additional permissions. The binaries I built were all done locally.

rok-cesnovar:
Example for a build https://github.com/stan-dev/stanc3/releases/tag/nightly

serban-nicusor-toptal:
@mitzimorris I now understand, I thought that is it's possible. If there's a need for this I will create a separated job that anyone with access to Jenkins can run and download the binaries afterwards.

mitzimorris:
> If there's a need for this I will create a separate job that anyone with access to Jenkins can run and download the binaries afterward.

yes, there's a need for this.  please create a job - does it need a name?  "stanc3 test binaries" or something like that?

serban-nicusor-toptal:
Okay on it, any other requirements besides using a specific branch/commit ?

rybern:
Great. Ideally we would also be able to specify a fork (by owner) in addition to branch.

serban-nicusor-toptal:
Hey @mitzimorris @rybern 

You now have the following [Jenkins job](https://jenkins.mc-stan.org/job/stanc3-test-binaries/) that you can use to build stanc3 binaries.  
I've wrote a small readme [here](https://github.com/stan-dev/ci-scripts/blob/refactor-jenkins-docs/docs/jobs/stanc3-test-binaries.md) that may help. Please give it a try and let me know if there are any issues or if there should be any changes.
The PR for it in stanc3 can be found [here](https://github.com/stan-dev/stanc3/pull/486)

Thanks.

rok-cesnovar:
Fantastic, works like a charm.

If https://github.com/stan-dev/cmdstan/pull/843 will be merged anyone can test this by adding 
```
STANC3_TEST_BIN_URL=https://jenkins.mc-stan.org/job/stanc3-test-binaries/31/artifact`
```
to the make/local.

mitzimorris:
@serban-nicusor-toptal - problems - when I run the stanc3-test-binaries job, and specify a build parameter, per build instruction:

> git_branch - Please specify a git branch ( develop ), git hash ( aace72b6ccecbb750431c46f418879b325416c7d ), pull request ( PR-123 ), pull request from fork ( PR-123 )

with parameter "PR-434" - (this PR) - checking the git hash of the branch, it's building from master, not from `rybern:pedantic` which has git hash 
8d7ab03bab73af90b9be1c1f388f6d413e4be306

latest attempt:  https://jenkins.mc-stan.org/job/stanc3-test-binaries/34/

serban-nicusor-toptal:
@mitzimorris Seems I missed out a function on the PR, fixed now.
Please see https://jenkins.mc-stan.org/job/stanc3-test-binaries/38/  for correct results.
Thanks!

mitzimorris:
> If [stan-dev/cmdstan#843](https://github.com/stan-dev/cmdstan/pull/843) will be merged anyone can test this by adding

merged this this morning - @rok-cesnovar are you going to make a branch for Andrew for testing?  let me know if you need anything from me.

rok-cesnovar:
So this is now accesible via cmdstan on the `pedantic_mode` branch. 

If you want to use cmdstanr: 

You need to clone the cmdstan with thata specific branch. If you already have a cmdstan installation you can point to that using `set_cmdstan_path()`.
You can also use `install_cmdstan(repo_clone = TRUE, repo_branch = "pedantic_mode", cores = 4)` to install it using cmdstanr.
This might take a few minutes.

Then you can try it out using this simple script

```R
library("cmdstanr")

#setwd("~/Desktop/test_scripts/pedantic_mode")
#set_cmdstan_path("~/Desktop/cmdstan/")

#use the example model with an unused sigma
file <- file.path(cmdstan_path(), "examples", "bernoulli", "bernoulli.stan")

#forcing recompiling for eaiser testing
mod <- cmdstan_model(file, force_recompile = TRUE)
```

I introduced a "pedantic" error in the example bernoulli model (an unused variable) so you can see if it works. You should see
```
Compiling Stan program...
Warning:
  The parameter sigma was declared but does not participate in the model.
```

VMatthijs:
@rybern , is this using your dependence analysis (and therefore indirectly my reaching definitions analysis/monotone framework)? Is that all tested and working now? This is very cool! :-)

rybern:
@VMatthijs Yes it is using dependency analysis in a few places, and so it does use reaching definitions. I do consider reaching definitions to be tested and working, although not confident in the whole suite - would like to make an automated testing framework.

VMatthijs:
That's really cool, @rybern !

rybern:
Hey @serban-nicusor-toptal - what can I do to get new builds on Jenkins? Does that need to be triggered manually? Not very familiar with the setup. Thanks!

serban-nicusor-toptal:
Hey, if you want to trigger a new build the easiest way is to just push a new commit.
Else, you can go to the pr builds page [here](https://jenkins.mc-stan.org/job/stanc3/view/change-requests/job/PR-434/) and press `Build with Parameters` on the left.

rybern:
Is there a different method for building such that I can point at a compiled binary from a CmdStan branch with STANC3_TEST_BIN_URL? We have a pedantic_mode branch on CmdStan pointing at  https://jenkins.mc-stan.org/job/stanc3-test-binaries/38/artifact, which looks like it was built separately from the testing setup, is that right?

serban-nicusor-toptal:
Yes, please take a look [here](https://github.com/stan-dev/ci-scripts/blob/refactor-jenkins-docs/docs/jobs/stanc3-test-binaries.md)

rok-cesnovar:
@serban-nicusor-toptal It would probably be worth to have a really spet-by-step guide on how you do this for someone that has little experience with Jenkins.

The docs you linked are great, but require some baseline knowledge of our CI.

For example for building the stanc3 binaries you have to: 

1. Go to https://jenkins.mc-stan.org/ and click log in (I think anyone with permissions on the repo can log in).
2. Click stanc3-test-binaries from the list of jobs
3. Click `Build with Parameters` and replace master with the PR label (PR-434) or branch name.
4. You will see a new build under Build History in the left side with the number X (its currently 60). Once it stops blinking and it goes blue the build finished succesfully.
If you see grey or red then it failed and you need to investigate.
5. Add `STANC3_TEST_BIN_URL=https://jenkins.mc-stan.org/job/stanc3-test-binaries/X/artifact` to make/local in cmdstan and run `rm bin/stanc` and `make build` and you are good to go.

@rybern I updated the cmdstan branch with X=60.

serban-nicusor-toptal:
@rok-cesnovar Updated, thanks!

rybern:
That's great, thanks @serban-nicusor-toptal and @rok-cesnovar!

mcol:
Note a typo: "Augment %s" should be "Argument %s".

rybern:
Thanks! Fixed.

rybern:
See the discussion on Discourse here: https://discourse.mc-stan.org/t/pedantic-mode-is-ready-for-your-testing-and-feedback/13893/30?u=rybern

And the proposed documentation here: https://github.com/rybern/pedantic-mode-doc

rok-cesnovar:
@rybern can you merge the recent master in when you have the chance. Based on the feedback on the thread and here this is ready for merge.

rybern:
Thanks Rok, will do later today.

rybern:
@rok-cesnovar All set

PR #451, opened 2020-01-31
Parallel reduce_sum

rok-cesnovar:
This is a draft and a very brute force of making a prototype for parallel reduce_sum.

Any and all comments are welcome. 

The main things is that I now duplicate the functors, the duplicate has the pstream as the third argument.

The other thing is supporting functions with variadic arguments.

```
// f is the reducer function
real f(int start, int end, sliced_type sub_slice_container, ...);

// parallel reduce
real reduce_sum(real(int, int, sliced_type, ...) f,
                int grainsize, sliced_type large_container, ...);
```
... can be 0 to N arguments of any type.
For now I just brute force my way to support 0 to 3 arguments of all base types and their 1D arrays.


rok-cesnovar:
This model compiles in cmdstan for now: 

```
functions {
  real my_func(int start, int end, real[] y_slice, real mu, real sigma) {
    return normal_lpdf(y_slice | mu, sigma);
  }
}

parameters {
  real a[5];
}

model {
  target += reduce_sum(my_func, a, 1, 0.0, 1.0);
}
```

rok-cesnovar:
v0.1 stanc3 with parallel_reduce binary for linux if anyone wants to test
[stanc.zip](https://github.com/stan-dev/stanc3/files/4141013/stanc.zip)

bbbales2:
OOoooooon it!

bbbales2:
I got:

```
./stanc: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by ./stanc)
```

When I tried to run the binary. How hard is it to build stanc?

rok-cesnovar:
It does take some installing of all the required ocaml, dune, etc. But after you have that its a piece of cake. Weird that it doesnt run. I will check how Jenkins builds it for the release and we can try again.

rok-cesnovar:
Please try this. I think this should run. 
[stanc.zip](https://github.com/stan-dev/stanc3/files/4141112/stanc.zip)

rok-cesnovar:
I goofed the original functor arguments, it doesnt hurt the your example with normal. Cleaned that up now.

The hierachiral reduce models still doesnt compile in c++ then.

[stanc_parallel_reduce_sum_v0.2.zip](https://github.com/stan-dev/stanc3/files/4141173/stanc_parallel_reduce_sum_v0.2.zip)

p.s.: dont get your hopes up too much. Learning on the job :)

bbbales2:
Yeah I got a basic normal running. Those errors you posted look like the result of me not defining the count_var_impl_ correctly -- I hit those in the ODE stuff and can fix em' easy.

bbbales2:
Alright I think it's working for a basic normal example for me :D. I'll see if I can get the poisson thing to compile.

Yo Bob said the print function would have the variadic argument stuff in the stanc3 compiler. Is that something you can copy-paste over to this?

bbbales2:
parallel_reduce_test_model compiles for me.

bbbales2:
I posted some code over here: https://gist.github.com/rok-cesnovar/832f9752fc2a8ba7ae7b3da3eb5f2743

You might be able to add something like ```require_not_t<is_var<T>>...,``` accumulate_adjoints, count_var_impl_, and save_varis_ and make the error you're getting go away. I don't know if my compiler is right to not complain or yours is right to complain (vs. the C++ standard).

bbbales2:
I got the same errors you did on gcc and added the require_not_ts and it worked.

rok-cesnovar:
> Yo Bob said the print function would have the variadic argument stuff in the stanc3 compiler. Is that something you can copy-paste over to this?

Yeah, tried looking at that, but print isnt implemented as the rest of them so not trivial for me. Might be for the real gurus :)

> parallel_reduce_test_model compiles for me.

Nice!

> I posted some code over here:

Thanks, saw the comment yeah. I am using g++ 8.3. Just switched to clang and that works and that is fine with me :)

bbbales2:
> Thanks, saw the comment yeah. I am using g++ 8.3. Just switched to clang and that works and that is fine with me :)

I'll push the changes to math.

The example model is running for me and I'm getting different results for STAN_NUM_THREADS=8 and STAN_NUM_THREADS=1 so debug time wooo!

bbbales2:
Oh wait. It generates data randomly. Lemme generate the data in R. I might be lying.

bbbales2:
Alright with generated data Rhat = 1 so they're doing the same thing. I'm gonna go fix the indexing bug with start/end.

rok-cesnovar:
Thanks Ben!

bbbales2:
No thank youuuuuu

bbbales2:
> I'm gonna go fix the indexing bug with start/end.

Sebastian pointed out this is something we should handle with the compiler. So I did this then undid it :D.

rok-cesnovar:
I am not sure what you mean by start/end indexing bug. Can you point me to a discussion on this somewhere or give an example?

nhuurre:
The indexing bug means the first slice is called with start index 0 rather than 1.
I think the only discussion is https://github.com/stan-dev/math/pull/1616#issuecomment-580937478

rok-cesnovar:
Thank you Niko. If you have any suggestions on how to improve the ocaml code , seeing as you are more of a guru than me, I am all ears :) we can also wait for this to actually be confirmed to work with this brute force approach.

src/stan_math_backend/Expression_gen.ml
```diff
@@ -138,9 +138,14 @@ let fn_renames =
    hash map of metadata available for each expression that we could put something like this in.
 *)
 let map_rect_counter = ref 0
+
 let functor_suffix = "_functor__"
-let functor_reduce_sum_suffix = "_functor__reduce_sum__"
+let reduce_sum_functor_suffix = "_reduce_sum_functor__"
```
nhuurre: This causes a naming conflict the user defines both `foo()` and `foo_reduce_sum()` functions (not very likely, i know).
Could the suffix be simply `"_rsfunctor__"`?

rok-cesnovar: Good point. Thanks! Pushing the change.

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -169,6 +169,22 @@ let pp_fun_def ppf Program.({fdrt; fdname; fdargs; fdbody; _}) =
...
+    pp_returntype ppf fdargs fdrt ;
+    let first_three_args =
+      List.mapi ~f:(fun i x -> if i < 3 then x else "") args
+    in
+    let rest_args = List.mapi ~f:(fun i x -> if i >= 3 then x else "") args in
+    let arg_strs =
+      mk_extra_args prefix_extra_templates prefix_extra_args
+      @ List.filter ~f:(fun x -> x <> "") first_three_args
+      @ ["std::ostream* pstream__"]
+      @ List.filter ~f:(fun x -> x <> "") rest_args
```
rok-cesnovar: I am sure there is a nicer way to get the first three args and the rest as opposed to this foolish stuff. It works but not proud.

nhuurre: I think `let first_three, rest = List.split_n args 3` should do it.

rok-cesnovar: Nice! Thanks a lot!

rok-cesnovar:
Besides styling the Ocaml and making it look nicer there are 3 open issues here: 

- fix the indexing bug
- only create the "rsfunctor" if the underlying function is used in a reduce_sum call. I more or less have an idea on how to do that (accumulate a list of functions used in reduce_sums and use that before deciding whether to print the rsfunctor or not) but that is a final touch

- variadic arguments for reduce_sum. Basically I would like to add the following

```
add_qualified
          ( "reduce_sum"
          , ReturnType UReal
          , [ ( AutoDiffable
              , UFun
                  ( [ (DataOnly, UInt); (DataOnly, UInt)
                    ; (AutoDiffable, allowed_slice_types i)
                    ; _ ]
                  , ReturnType UReal ) )
            ; (AutoDiffable, allowed_slice_types i)
            ; (DataOnly, UInt)
            ; _ ] )
```
That will probably have to be solved some other way. I am more lost here.

rok-cesnovar:
Should the functor than call the function with

```cpp
struct my_func_3_rsfunctor__ {
template <typename T2__, typename T3__, typename T4__, typename T5__>
typename boost::math::tools::promote_args<T2__, T3__, T4__,
T5__>::type
operator()(const int& start, const int& end,
           const std::vector<T2__>& y_slice, std::ostream* pstream__,
           const T3__& mu, const T4__& sigma, const T5__& dummy)  const 
{
return my_func_3(start+1, end+1, y_slice, mu, sigma, dummy, pstream__);
}
};
```
or am I in the wrong? If this is correct that is done now.

nhuurre:
Looks good to me.

rok-cesnovar:
With this off by one fix, I have to implement the second point "only create the "rsfunctor" if the underlying function is used in a reduce_sum call". Otherwise there will be compilation issues if there are any other functions present in the model (those without start, end parameters).

wds15:
argh... I was hoping this is easier to handle on the Stan parser side this +1 thing...

rok-cesnovar:
Its not difficult. We only want to add the reduce sum functor if the function is actually used in a reduce sum anyways. This is how it should be done and this fix forcing us to do it is sooner rather then later is just fine.

Its more me experiencing growing pains learning stanc and Ocaml in the process. It would be a 5 min job for the real gurus :)

rok-cesnovar:
@wds15 I should have time to finish this up today. 

Just to double check, the reducer function must always be in the form

```stan
real f(int start, int end, sliced_type sub_slice_container, ...);
```

If the first 2 arguments are not ints and `sliced_type` is not one of the allowed types this should throw a semantic error. Or are there any exceptions?

It might also be smart to throw an error if the types in the first `...` doesnt match the types in the second `...`

```stan
// parallel reduce
real reduce_sum(real(int, int, sliced_type, ...) f,
                int grainsize, sliced_type large_container, ...);
```

So the TODO is now: 
- only create the "rsfunctor__" if the underlying function is used in a reduce_sum call.
- semantic check that the reducer function has the correct arguments
- make it an actual variadic function

wds15:
This all sounds fine to me what you propose.

So you are becoming an OCaml guru?

rok-cesnovar:
> So you are becoming an OCaml guru?

haha, I am still very much far way from that. I am interested in knowing more, partly because I I have a bunch of potentially interesting ideas I would like to implement and some require stanc support and partly because I want to support efforts like this.

rok-cesnovar:
I should be able to implement everything needed here, apart from maybe the variadic stuff. I imagine that will require some handhelding.

bbbales2:
It'd be really helpful to me to have a bunch more arguments here, but I see that we're sort of at the limit of brute forcing it.

Looking at the code, could we do something like what the vectorized functions are doing:

```
let math_sigs =
  [ ([UnaryVectorized], "acos", [DDeepVectorized])
...
```

The optional arguments here can be any valid Stan type, right? That's what the 'DDeepVectorized' argument is doing?

And so if I was brute forcing this I'd like to add to math_sigs something like (I'm assuming the thing to the left of the name has to do with return type and the things on the right are arguments):

```
[ ([UReal], "reduce_sum", [DDeepVectorized]),
  ([UReal], "reduce_sum", [DDeepVectorized; DDeepVectorized]),
  ([UReal], "reduce_sum", [DDeepVectorized; DDeepVectorized; DDeepVectorized]),
  ...
```

And like maybe just write a loop for that? So we can have up to some large, fixed number of arguments? So not quite variadic but we have a bunch more stuff.

rok-cesnovar:
Thanks to the additions by @rybern this is now extended to functions with 7 arguments, where the arguments are still base types (scalar, vectors, row vectors, matrices) or 1D arrays of the base types.

I think in the end we will need to modify stanc3 in a way that we dont need to add all possible signatures for these. But for the current testing period and at least until things get merged in Stan Math, this should do.

I am adding a few more cleanups today and then this should be the only thing left to do.

rok-cesnovar:
New Linux binaries that support 5 arguments
[stanc_parallel_reduce_sum_v0.3.zip](https://github.com/stan-dev/stanc3/files/4171145/stanc_parallel_reduce_sum_v0.3.zip)

The combinatoric explosion of signatures is definitely going to prevent us from doing anything above 5 with this approach. We currently support: int, real, vector, row_vector, matrix and their 1D arrays. Which means 10 types and that means M*10^N signatures for N arguments and M supported sliced types.

rybern:
Yeah. The problem is that we have to enumerate signatures at all. As far as I can tell, these are the ways we use the signatures:

1. Getting a return type from a name+signature
2. Checking if a name is present
3. Printing out all signatures or signatures for a given name, for error messages and documentation

1 and 2 would be easy and efficient to write as a function rather than a hash table.
3 requires some of enumeration, but in cases like reduce_sum, it's neither feasible to enumerate nor desirable to print all the cases. That would produce a really nasty error message.

So one solution would be to replace the current data structure:
`table from name to (list of (return_type * signature))`
with:
`table from name to ((signature -> return_type option) * documentation string)`

This still allows all of the use cases, but avoids combinatoric explosion.
We could write a helper function to go from our current list format to the new type, which would generate the doc string by enumeration like before, and use it in cases with only a couple signatures.

A different solution would be to replace the type of each argument in a signature:
`(autodifftype * type)`
with a set of allowed types:
`(autodifftype * type) Set.Poly.t`
or let it be some wildcard symbol, if we only ever care about some specific sets of types:
``[`Type of (autodifftype * type) | `Any]``

 I think that would work too and would be a smaller change.

rok-cesnovar:
> or let it be some wildcard symbol, if we only ever care about some specific sets of types:
> `` [`Type of (autodifftype * type) | `Any] ``

This one seems like the easiest solution to me. In any case this will probably be a significant refactor. Should we split that to a separate PR that would be done before this one?

rok-cesnovar:
@rybern would what I did now be considered bad? I basically added another UnsizedType: Any 
and the signatures are now listed as

```
real reduce_sum((data int, data int, int[]) => real, int[], int)
real reduce_sum((data int, data int, int[], *) => real, int[], int, *)
real reduce_sum((data int, data int, int[], *, *) => real, int[], int, *, *)
real reduce_sum((data int, data int, int[], *, *, *) => real, int[], int, *,
                  *, *)
```
... and so on up to 8 * and for all different allowed slice types. The test model that uses up to 8 arguments works and compiles with the Stan Math branch. But I am guessing I am missing something, this seemed to easy.

rybern:
One issue with adding the Any directly to the UnsizedType is that Any doesn't necessarily make sense anywhere an UnsizedType makes sense, for example it doesn't really make sense to have an Any as a user-defined function argument type, or having an Any in code gen. I think we could get away with just making a new type to extend UnsizedType and keep it within Stan_math_signatures.ml.

A limitation of the wildcard idea (which didn't occur to me before) is that in signature 
``real reduce_sum((data int, data int, int[], *) => real, int[], int, *)``
is a little too lax, because it doesn't require that the two *s are the same type (they do need to be the same, right?). We could:
1. Add an identifier to Any to indicate which types should match. These are type variables. So if instead of `Any` we had `TypeVar "A"` then we can express the correct type signature:
``real reduce_sum((data int, data int, int[], A) => real, int[], int, A)``
2. Ignore that problem and potentially not catch some type errors
3. Switch away from wildcards, maybe trying the function idea

I think #1 would be reasonable, that's how typed functional languages (including OCaml) do it, and we can keep it pretty contained within `stan_math_return_type` and the printing functions. It could also be extended to include other sets of types, like "B is any slice type". 

What do you think?

rok-cesnovar:
> it doesn't really make sense to have an Any as a user-defined function argument type, or having an Any in code gen. I think we could get away with just making a new type to extend UnsizedType and keep it within Stan_math_signatures.ml.

Agreed. Thanks.

> is a little too lax, because it doesn't require that the two *s are the same type

Oh yeah, I forgot about that.

I agree that 1 would be the best way of going forward. 
Something like 
real reduce_sum((data int, data int, int[], T1, T2) => real, int[], int, T1, T2)
does seem reasonable and hopefully, the error message will be understandable.

seantalts:
`print` has custom typechecking here: https://github.com/stan-dev/stanc3/blob/master/src/frontend/Semantic_check.ml#L778 it's not actually listed in Stan_math_signatures. We could go that route for this function, or just do like Scala does for tuples and define this for up to 22 arguments 😂 basically what @bbbales2 was saying [here](https://github.com/stan-dev/stanc3/pull/451#issuecomment-582523855).

re: @rybern's "So one solution would be to replace the current data structure [with one that also has a doc string]" - I would suggest just adding another data structure with the same key that contains the doc string if we want to override the default method for generating such a doc string.

The type parameter idea seems pretty good on its face, at least for the long term. But I would just like to say that for now we could also totally just add custom typechecking for this function like we have for `print` and `reject` and do a separate refactor to allow for variadic type signatures.

rybern:
Yeah, easy solution until a different PR sounds smart to me. Adding type variables would be some substantial code (especially if avoiding changes to UnsizedType).

rok-cesnovar:
> We could go that route for this function, or just do like Scala does for tuples and define this for up to 22 arguments

Yeah, we talked about this at Stancon last year, I remember the 22 :smile: I think going something like 22 would be fine. @bbbales2 , @wds15 Does that cover most if not all cases?

I will take a crack at this today and then this is probably ready and we need to polish and get the Math PR merged.

seantalts:
In our case going with 22 still sounded like a kind of combinatoric explosion in that hash table though because we store the explicit extension of all possible combinations there. I would just go with custom typechecking function a la print for now and then at some point refactor the type system probably along the lines I think @rybern was saying with sets of types in each argument slot.

rok-cesnovar:
Sounds like a plan. Thanks for the pointer to the print semantic check. Not sure how I missed that.

seantalts:
I'm not opposed to doing refactors first either, I just think it'd be easier to test and review if it's separate from the feature if that makes sense. So if you preferred to implement reduce_sum only with a changed type system, just change the system first in one PR and then do implement reduce_sum in the second if you don't mind :P

bbbales2:
I don’t think this should be merged into the main stanc until math development is finished. I’m happy with a separate branch and binaries I have to download.

We need this language stuff to evaluate the math stuff, really. But we also need to finalize the math stuff before we finalize this.

22 args is good :D.

rok-cesnovar:
> So if you preferred to implement reduce_sum only with a changed type system

No no, this way with a custom check function is much more accessible to my level of understanding of stanc3 so I am all for it.

> I don’t think this should be merged into the main stanc until math development is finished. I’m happy with a separate branch and binaries I have to download.

+1. I will make sure to put a call to reduce_sum in the mother model and motherHOF model so that this PR wont go green until downstream PR(s) are merged.

wds15:
The answer to all things is 42 - like always... but I can live with 22 for sure.

@bbbales2 are u still not convinced this all should go in? I am really happy with it. With the 22 args I should be able to test out another of my favourite models.

Will do that once we have a binary with 22 args.

@rok-cesnovar thanks so much!

rok-cesnovar:
Here is the latest version as of commit https://github.com/stan-dev/stanc3/pull/451/commits/e166c5b7ad4d15a081cf2ac57099b58a0f42212e

This should let you test whichever function you like @bbbales2 @wds15 
No limit to the number of arguments but also no semantic check on reduce_sum. This should let you do the performance tests and I can add the various checks that need to happen until you get Stan Math up to speed.

The models compile with the Stan Math branch so you should be good. But do let know if you see anything weird.

[stanc_parallel_reduce_sum_v0.4.zip](https://github.com/stan-dev/stanc3/files/4182951/stanc_parallel_reduce_sum_v0.4.zip)

rok-cesnovar:
One thing that needs to be investigated and also discussed is what happens if users name their functions `reduce_sum`. I am almost sure this would be problematic in the current code and the question is what should be the behavior here: 
a) restrict the names so that reduce_sum is not allowed as a user function
b) if reduce_sum is found in the functions block we then use that functions and not this parallel_reduce?

Is there a general policy for this kind of stuff?

wds15:
Thanks a lot.

Reduce_sum is a Stan language thing and as such the user must not use this name as a user defined function. That’s what we have been always doing, I think.

Let’s ping @bob_carpenter ....

rok-cesnovar:
Yeah, I am almost certain that is not allowed but just want to make sure. Its the first time I am changing anything in the semantic check portions of the compiler, so want to double check. 

There is this comment in the code somewhere that makes me think we do restrict that: 
> Stan allows user declared identifiers that are not of nullary function types to clash with nullary library functions. No other name clashes are tolerated.

cc @bob-carpenter (making sure Bob gets notified, you had a typo in the name)

bob-carpenter:
@rok:  I get so many notifications they're basically noise.  But I try to at least scan through everything.

As @wds15 points out somewhere else, we need to warn users better about changes like reserving a word that breaks backward compatibility.

What we really need to do is fix the way code's generated for functions by the transpiler so that we can support overloading.  What happens now is that each definition generates a struct as a functor and that struct's name clashes if there's overloads.  What we need to do is pack all the operators into a single struct and it should work OK.

bbbales2:
> are u still not convinced this all should go in?

I have a couple specific models I want to test with this. Might as well test now before we put in the work to finalize everything in case there are surprises.

> This should let you test whichever function you like

Awesome! I'll test this today.

bbbales2:
@rok-cesnovar, @rybern, @seantalts this is super convenient and is working for me for now! Thanks!

rok-cesnovar:
:dancer:

bbbales2:
This fails if I call reduce_sum twice with the same arguments. Something like:

```
target += reduce_sum(nbme, y_subset, grainsize,
                       X_subset, b,
                       sigma_state, a_state, state,
                       sigma_specialty, a_specialty, specialty,
                       sigma_state_specialty, a_state_specialty) +
    reduce_sum(nbme, y_subset, grainsize,
                       X_subset, b,
                       sigma_state, a_state, state,
                       sigma_specialty, a_specialty, specialty,
                       sigma_state_specialty, a_state_specialty);
```

The error I get from the C++ compiler starts with:
```
clang++ -g -std=c++1y -D_REENTRANT -Wno-sign-compare     -DSTAN_THREADS -I stan/lib/stan_math/lib/tbb_2019_U8/include -O3 -I src -I stan/src -I lib/rapidjson_1.1.0/ -I stan/lib/stan_math/ -I stan/lib/stan_math/lib/eigen_3.3.3 -I stan/lib/stan_math/lib/boost_1.72.0 -I stan/lib/stan_math/lib/sundials_5.1.0/include    -DBOOST_DISABLE_ASSERTS      -c -include-pch stan/src/stan/model/model_header.hpp.gch -x c++ -o nbme6.o nbme6.hpp
nbme6.hpp:574:23: error: unknown type name 'nbme_rsfunctor__'; did you mean 'nbme_functor__'?
          (reduce_sum<nbme_rsfunctor__>(y_subset, grainsize, pstream__,
                      ^~~~~~~~~~~~~~~~
                      nbme_functor__
nbme6.hpp:196:8: note: 'nbme_functor__' declared here
struct nbme_functor__ {
       ^
nbme6.hpp:578:24: error: unknown type name 'nbme_rsfunctor__'; did you mean 'nbme_functor__'?
            reduce_sum<nbme_rsfunctor__>(y_subset, grainsize, pstream__,
                       ^~~~~~~~~~~~~~~~
                       nbme_functor__
nbme6.hpp:196:8: note: 'nbme_functor__' declared here
struct nbme_functor__ {
...
```

rok-cesnovar:
I will take a look at this tomorrow. It looks like the rsfunctor is not getting generated (rafunctor is generated for user defined functions used in any reduce sum call) in this case. Can you check if there are nbme_functor and nbme_rsfunctor present?

rybern:
By the way - I've played a bit with a type variable implementation that would work for this and the new ODE interface, but I think it makes sense to hold off until we get them stable

rok-cesnovar:
Fantastic! 

I now managed to fix the bug Ben mentioned [above](https://github.com/stan-dev/stanc3/pull/451#issuecomment-590531785) and will try to also add all the checks. Then it should be close to being stable. I have no idea how close the ODE stuff is tho.

rok-cesnovar:
This now does a full semantic check: 
- checks that the supplied function is in the form `fun(int, int, sliced_type, ... )`
- checks that arg2 of reduce_sum(f, arg2, arg3, ...) is a sliced_type
- checks that arg3 is an int
- checks that the types in ... in fun and reduce sum match in types

I also added a few "bad" models to show the errors. 

The error is currently always "Illegal use of reduce_sum!". Its simple to extend that but I didn't bother with that for now if we decide to switch to the type variable implementation.

[linux binary](https://jenkins.mc-stan.org/job/stanc3-test-binaries/32/artifact/bin/linux-stanc)
[mac binary](https://jenkins.mc-stan.org/job/stanc3-test-binaries/32/artifact/bin/mac-stanc)
[win binary](https://jenkins.mc-stan.org/job/stanc3-test-binaries/32/artifact/bin/windows-stanc)

SteveBronder:
@rok-cesnovar we are in motion to get reduce_sum into stan math (if everything goes well sometimes in the next two weeks). What is left here to move this from a WIP to review and merge?

rok-cesnovar:
This is more or less ready. The only small detail missing is probably that all compiler errors with reduce_sum now return "Illegal use of reduce_sum!" (see https://github.com/stan-dev/stanc3/pull/451/files#diff-f44a209e74bb10d5dabc5ec56d4fd7b7).

Otherwise this needs the Math update and a review.

wds15:
Just a heads up: we are going to have two flavors of reduce sum. A default one which does not guarantee the same execution order and an additional static one which will always calculate things in the same order. I hope things get into math soonish and having two flavors (two functions ) is easy to deal with.

rok-cesnovar:
What will be the difference in the use? different function name, argument? If you give me the details I can add it tomorrow.

bbbales2:
> reduce_sum_static...the

It's somebody's bed time!

rok-cesnovar:
Ok, this is now ready I believe. 

The displayed error comes in two forms. First one I am for now calling the basic error is displayed when there is a mismatch in the non-variadic args. The basic args are:
`(int, int, sliced_type) => real, sliced_type, int`
If any of those are not correct the user would see:

```
Ill-typed arguments supplied to function 'reduce_sum'. Available arguments:
(int, int, real[], ...) => real, real[], int, ...
(int, int, int[], ...) => real, int[], int, ...
(int, int, matrix[], ...) => real, matrix[], int, ...
(int, int, vector[], ...) => real, vector[], int, ...
(int, int, row_vector[], ...) => real, row_vector[], int, ...
(int, int, real[,], ...) => real, real[,], int, ...

Instead supplied arguments of incompatible type: (int, int, real[], real, real) => real, real, int, real
```

If the supplied arguments to the reduce_sum function dont match the function argument types then its a more specific error:

```
Ill-typed arguments supplied to function 'reduce_sum_static'. Expected arguments:
(int, int, real[], real, real) => real, real[], int, real, real
Instead supplied arguments of incompatible type: (int, int, real[], real, real) => real, real[], int, real, int
```

rok-cesnovar:
This also now has the reduce_sum_static.

I will fixup the cmdstan branch tomorrow as I am still awaiting the binaries to be built: https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3-test-binaries/detail/stanc3-test-binaries/51/pipeline

@rybern @nhuurre @seantalts If any of you have time to look at this I would very much appreciate. I know that it does what it should and we have for now decided to not go with type variable implementation. I added a bunch of models that fail to compile and a few good examples.

wds15:
stan-math PR for reduce_sum just went in. Happy testing.

bbbales2:
Alright I went through and tested this a bunch today.

Looks really good. Two problems:

1. We should support calling reduce_sum recursively. I think at one point in Math we tested it, but not sure if we do not. If we don't support it, we should make it not compile.

2. We should support arrays of anything as the first argument, not just ```int[]```, ```real[]```, ```vector[]```, ```row_vector[]```, ```matrix[]```. If this is a problem, we can roll back the docs to say we just support these things specifically. This covers the important use cases.

I uploaded some test files over here: https://discourse.mc-stan.org/t/stanc3-parallel-reduce-sum/12895/21 cuz github doesn't like .stans.

But this looks really good! I tried to test the crap outta this. If necessary, we could absolute just say recursive reduce_sum isn't supported and roll with what we have (again I don't even think this is explicitly tested in math now, looks like this functor isn't used: https://github.com/stan-dev/math/blob/develop/test/unit/math/prim/functor/reduce_sum_util.hpp#L33, though it should work).

wds15:
Recursive calls of reduce sum have been tested in math and we do support it. It would be good to support that with the interface as well.

The thing with the arbitrary data types should also be possible as it is really a key strength to support things flexibly.

Cutting back in the interface the functionality would hurt a bit, but is an option.

...don’t get me wrong...this is really cool to have this already now! Must be a pain to write / be something totally new given the flexible arguments allowed.

rok-cesnovar:
The types are really easy to add. That takes me 2 seconds, just add to list of supported types in https://github.com/stan-dev/stanc3/pull/451/files#diff-d93e45d6d02ba542534135574b9f183bR85. Can you point me to the docs where all the types are listed? Or is it just all arrays of arbitrary dimensionality?

When writing that I took this doc:
```
So these are admissible for the sliced_type:

real[num_jobs]
int[num_jobs]
real[num_jobs,nested_dimension]
matrix[rows, cols][num_jobs]
vector[rows][num_jobs]
row_vector[cols][num_jobs]
```
Sorry if I missed any discussion on this. 

The recursive call should also not be too hard to fix. But definitely a bit more than the listed of supported types.

wds15:
The argument which we slice must be an array of whatever type. So in c++ terms it is std vector of T - and T is arbitrary. For the shared arguments the admissible type is T - so it’s anything the user wants.

Recursive calls are great to support...they make things more flexible.

rok-cesnovar:
> So in c++ terms it is std vector of T - and T is arbitrary.

Got it, thanks. Fix incoming.

> Recursive calls are great to support...they make things more flexible.

I think this should be doable fairly easy.

> Must be a pain to write / be something totally new given the flexible arguments allowed.

Not at all. I like new challenges :)

rok-cesnovar:
@bbbales2 huge thanks for the test models!! They are really thorough!

I now added support for recursive calls and more sliced types for up to 7 dimensions. Is 7 enough? All the rest signatures are done to 7 dimension like: `int[,,,,,,]`.

I updated the cmdstan branch, but its no longer required to use the branch. You can use the develop version and add 

`STANC3_TEST_BIN_URL = https://jenkins.mc-stan.org/job/stanc3-test-binaries/53/artifact`

to your make/local

rok-cesnovar:
Also m1 and m2 compile and run on develop cmdstan, m3 compiles so we covered everything there.

seantalts:
What happens with recursive calls w.r.t. the worker threads and tasks? Does Intel TBB handle that in a performant way? I had some prior experience with the Java Fork/Join framework where recursive usage did NOT perform well and I thought it was kind of endemic to parallelism, but maybe Intel did something smart about that situation?

bbbales2:
> What happens with recursive calls w.r.t. the worker threads and tasks?

I don't think it's something we'd ever want to do (maybe there's an advantage I'm not thinking of though), I just think it should compile without error.

seantalts:
If it's not something we'd ever want to do and it has worse performance than one might think, it might be a good candidate for exclusion now at least with the first release (with a good error message). We can always go back and make recursive calls work correctly without breaking backwards compatibility; we can't go do the reverse.

bbbales2:
@rok-cesnovar is that the right stanc3 binary? I'm getting these errors when I go to build:

```
Semantic error in '/tmp/RtmpIfrdHf/model-16f440fc3a41.stan', line 4, column 13 to column 38:
     2:    real g(int start, int end, real[] y_slice) {
     3:      if(size(y_slice) > 1) {
     4:        return reduce_sum(g, y_slice, 1);
                      ^
     5:      } else {
     6:        return normal_lpdf(y_slice | 0, 1);

A returning function was expected but an undeclared identifier 'reduce_sum' was supplied.
```

I tried setting the make/local flag and just re-installing from the repo.

bbbales2:
> If it's not something we'd ever want to do and it has worse performance than one might think, it might be a good candidate for exclusion now at least with the first release (with a good error message).

I'm neutral here, I guess. I'd like it to behave like a normal function since as far as the language goes it's just another function.

> We can always go back and make recursive calls work correctly without breaking backwards compatibility; we can't go do the reverse.

Well I don't really know what the future is for all the parallelism stuff or what the exact technical snags are gonna be.

I see the advantages of holding our cards, but at the same time, I just assume if we screwed up anything too badly here, we'll just mark this function deprecated, leave a placeholder implementation in place, build the new thing, and move on.

bbbales2:
But yeah, this function was all about evaluating a bunch of likelihood terms in parallel. I don't know of a situation where you'd program that recursively.

But like, maybe that's an argument that people won't end up using this recursively. It's not filling a need. And of course if they do find a reason to use it recursively, then maybe we need to rethink things anyway and figure out what it is they're doing lol.

wds15:
For me the recursive stuff would be very desirable to have - simply for the purpose of being able to compose my programs flexibly.

The tbb should still work sensibly well provided that the inner call still has some decent work to do.

So I would strongly prefer to allow recursion...I do have code which would use this facility already now.

seantalts:
All good points. I guess my guiding principles are something like:
1. If we literally don't have a single use-case in mind, probably leave it out.
2. If there is a use-case for something but the performance is way worse than it "should" be, leave it out.

A single use-case with decent performance would nullify both of these in my view. Then it goes to more stylistic design arguments, which I don't have any strong principles for, haha.

seantalts:
Et voila, Sebastian nullifies my arguments :) @wds15 it works well for that code, right? And wouldn't work better if you manually flattened that recursive structure out?

rok-cesnovar:
@bbbales2 sorry, jenkins restart the binary build job and built master instead.

Use STANC3_TEST_BIN_URL = https://jenkins.mc-stan.org/job/stanc3-test-binaries/54/artifact

bbbales2:
> I do have code which would use this facility already now.

Nay, this I can hardly believe :D. Bring forth your reducer so that we might judge it! (and what Sean said, write it in a flat loop)

> restart the binary build job and built master instead.

How rude

wds15:
Ok... what I have is code which uses nested `map_rect` .... which  I would turn into `reduce_sum` calls which are nested, since these `map_rect` calls actually do a `reduce_sum`.

So you see, there are programs which will be easily written using nested calls to `reduce_sum`.

Also, I suspect that not-so-big work chunks are better run through the `reduce_sum` thing, because it does nested AD on the spot and this keeps the AD tree smaller when compared to building a big tree - this is good news for better use of CPU caches.

Moreover, it would be odd to be allowed to nest a `map_rect` into a `reduce_sum` while two `reduce_sum`s can't be nested.

I am not advocating to go for nesting to gain speed - but for the purpose of writing programs which I can stick together as I want to. Also: Any other Stan construct can usually be nested, so why would we make an exception here? I understood that @rok-cesnovar has solved the parser issue already, no?

FYI: The application with the nested stuff is population pharmacokinetic model. You have subjects for which you compute drug concentrations over time. The outer reduce is for the subjects (summing the log-lik terms), the inner one comes from representing multiple dosing events in the form of linear superposition (meaning summation of many dose terms). There you have it - two `reduce_sum`s would do the job just fine.

wds15:
@seantalts we have a meeting in 20min on the matter of `reduce_sum`... drop us a mail if you wanna join.

bbbales2:
Well meeting is here: meet.google.com/pbn-quei-yfy . Rok feel free to come too. We'll probably be discussing the final doc or whatever to do.

bbbales2:
@rok-cesnovar yo just did testing, all seems to work fine!

@seantalts if you're the one reviewing this code, as a user, it's doing what I want.

These error messages might be a bit verbose but whatevs, they do the job:
```
Ill-typed arguments supplied to function 'reduce_sum'. Available arguments:
(int, int, real[], ...) => real, real[], int, ...
(int, int, int[], ...) => real, int[], int, ...
(int, int, matrix[], ...) => real, matrix[], int, ...
(int, int, vector[], ...) => real, vector[], int, ...
(int, int, row_vector[], ...) => real, row_vector[], int, ...
(int, int, real[,], ...) => real, real[,], int, ...
(int, int, int[,], ...) => real, int[,], int, ...
(int, int, matrix[,], ...) => real, matrix[,], int, ...
(int, int, vector[,], ...) => real, vector[,], int, ...
(int, int, row_vector[,], ...) => real, row_vector[,], int, ...
(int, int, real[,,], ...) => real, real[,,], int, ...
(int, int, int[,,], ...) => real, int[,,], int, ...
(int, int, matrix[,,], ...) => real, matrix[,,], int, ...
(int, int, vector[,,], ...) => real, vector[,,], int, ...
(int, int, row_vector[,,], ...) => real, row_vector[,,], int, ...
(int, int, real[,,,], ...) => real, real[,,,], int, ...
(int, int, int[,,,], ...) => real, int[,,,], int, ...
(int, int, matrix[,,,], ...) => real, matrix[,,,], int, ...
(int, int, vector[,,,], ...) => real, vector[,,,], int, ...
(int, int, row_vector[,,,], ...) => real, row_vector[,,,], int, ...
(int, int, real[,,,,], ...) => real, real[,,,,], int, ...
(int, int, int[,,,,], ...) => real, int[,,,,], int, ...
(int, int, matrix[,,,,], ...) => real, matrix[,,,,], int, ...
(int, int, vector[,,,,], ...) => real, vector[,,,,], int, ...
(int, int, row_vector[,,,,], ...) => real, row_vector[,,,,], int, ...
(int, int, real[,,,,,], ...) => real, real[,,,,,], int, ...
(int, int, int[,,,,,], ...) => real, int[,,,,,], int, ...
(int, int, matrix[,,,,,], ...) => real, matrix[,,,,,], int, ...
(int, int, vector[,,,,,], ...) => real, vector[,,,,,], int, ...
(int, int, row_vector[,,,,,], ...) => real, row_vector[,,,,,], int, ...
(int, int, real[,,,,,,], ...) => real, real[,,,,,,], int, ...
(int, int, int[,,,,,,], ...) => real, int[,,,,,,], int, ...
(int, int, matrix[,,,,,,], ...) => real, matrix[,,,,,,], int, ...
(int, int, vector[,,,,,,], ...) => real, vector[,,,,,,], int, ...
(int, int, row_vector[,,,,,,], ...) => real, row_vector[,,,,,,], int, ...
```

seantalts:
I can’t make it due to work but I trust you all :) let us know what you
decide. Either way this is very exciting!

rok-cesnovar:
I will be a bit late. Hope to make it before it ends. At the store rn and forgot that is not a 3 min thing nowadays:)

bbbales2:
Am I in the right meeting @wds15? Nobody else here

bbbales2:
@rok-cesnovar I made a couple slight changes to the test models.

For m1 I added recursion on a shared argument:
```
functions {
  real g(int start, int end, real[] y_slice) {
    if(size(y_slice) > 1) {
      return reduce_sum(g, y_slice, 1);
    } else {
      return normal_lpdf(y_slice | 0, 1);
    }
  }
  real h(int start, int end, real[] y_slice, real[] a) {
    if(size(a) > 1) {
      return reduce_sum(h, y_slice, 1, a[start:end]);
    } else {
      return normal_lpdf(a | 0, 1);
    }
  }
}

transformed data {
  int N = 100;
}

parameters {
  real y1[N];
  real y2[N];
}

model {
  target += reduce_sum(g, y1, 1);
  target += reduce_sum(h, y2, 1, y2);
}
```

For m3 I added recursion on the big everything function:
```
  real s(int start, int end, real[] y_slice,
         int a, real b, vector c, row_vector d, matrix e,
         int[] f, real[] g, vector[] h, row_vector[] i, matrix[] j,
         int[,] k, real[,] l, vector[,] m, row_vector[,] n, matrix[,] o,
         int[,,] p, real[,,] q) {
    return reduce_sum(s, y_slice, 1,
                      a, b, c, d, e,
                      f, g, h, i, j,
                      k, l, m, n, o,
                      p, q);
  }
```

rok-cesnovar:
I guess I am too late, it wont let me in. Let me know if you decided anything. 

We can change the message to something like

```
Ill-typed arguments supplied to function 'reduce_sum'. Available arguments:
(int, int, T[], ...) => real, T[], int, ...
(int, int, T[,], ...) => real, T[,], int, ...
(int, int, T[,,], ...) => real, T[,,], int, ...
(int, int, T[,,,], ...) => real, T[,,,], int, ...
(int, int, T[,,,,], ...) => real, T[,,,,], int, ...
(int, int, T[,,,,,], ...) => real, T[,,,,,], int, ...
(int, int, T[,,,,,,], ...) => real, T[,,,,,,], int, ...
(int, int, T[,,,,,,,], ...) => real, T[,,,,,,,], int, ...
where T is int, real, vector, row_vector or matrix.
```
Would that be better?

bbbales2:
> I guess I am too late, it wont let me in

We ended up going to a different hangout anyway :/

> We can change the message to something like

I like it! Maybe add 'Where T is any one of int, real, vector, row_vector, matrix'.

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -657,12 +680,47 @@ let pp_register_map_rect_functors ppf p =
     (list ~sep:cut pp_register_functor)
     (List.mapi ~f:(fun i f -> (i + 1, f)) functors)
 
+let fun_used_in_reduce_sum p =
+  let rec find_functors_expr accum Expr.Fixed.({pattern; _}) =
+    match pattern with
+    | FunApp (StanLib, "reduce_sum", {pattern= Var f; _} :: _) -> f :: accum
+    | FunApp (StanLib, _, fk :: _) -> find_functors_expr accum fk
+    | _ -> accum
```
nhuurre, line 688: This needs to be recursive for all patterns, not just the first arg of functions. Otherwise expressions like this will miss the second function:
```stan
target += reduce_sum(f, ...) + reduce_sum(g, ...);
```
I realize this all is copied from the above `map_rect`registration code. That code is equally broken; as far as I know `map_rect` doesn't even need the registration except when using MPI so I guess no one noticed.
The correct way to fold over an expression is, e.g. how `Transform_mir`'s `collect_vars_expr` does it. That function also illustrates the next point: `accum` should be a `String.Set.t` instead of a list. We don't care about the order and don't need duplicates.

src/frontend/Semantic_check.ml
```diff
@@ -314,6 +314,47 @@ let semantic_check_fn_stan_math ~is_cond_dist ~loc id es =
       |> Semantic_error.illtyped_stanlib_fn_app loc id.name
       |> Validate.error
 
+let semantic_check_reduce_sum ~is_cond_dist ~loc id es =
+  let args_match a b =
+    if List.length a = List.length b then
+      List.for_all2_exn ~f:(fun (_, x) y -> x = y.emeta.type_) a b
```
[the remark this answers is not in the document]
rok-cesnovar: Good call. Will use that.

src/frontend/Semantic_check.ml
```diff
@@ -314,6 +314,47 @@ let semantic_check_fn_stan_math ~is_cond_dist ~loc id es =
...
+  | { emeta=
+        { type_=
+            UnsizedType.UFun
+              ((_, st) :: (_, e) :: sliced_arg_fun :: fun_args, ReturnType rt); _
+        }; _ }
+    :: sliced :: grainsize :: args ->
+      let (_, sliced_arg_fun_type) = sliced_arg_fun in
+      if
+        st = UInt && e = UInt
+        && grainsize.emeta.type_ = UInt
```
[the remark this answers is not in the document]
rok-cesnovar: Yeah, I rewrote this piece a few times, had `(_, UInt) :: (_, UInt) ::` before. Good call. Thanks for the trick!

src/frontend/Semantic_check.ml
```diff
@@ -331,9 +372,11 @@ let fn_kind_from_application id es =
     corresponding semantic check
 *)
 let semantic_check_fn ~is_cond_dist ~loc id es =
-  match fn_kind_from_application id es with
-  | StanLib -> semantic_check_fn_stan_math ~is_cond_dist ~loc id es
-  | UserDefined -> semantic_check_fn_normal ~is_cond_dist ~loc id es
+  match (id.name, fn_kind_from_application id es) with
+  | "reduce_sum", _ | "reduce_sum_static", _ ->
+      semantic_check_reduce_sum ~is_cond_dist ~loc id es
+  | _, StanLib -> semantic_check_fn_stan_math ~is_cond_dist ~loc id es
```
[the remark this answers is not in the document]
rok-cesnovar: Yeah, I tried with stuff like below but it didnt work. So I brute forced it. Will try again.

```Ocaml
  match fn_kind_from_application id es with
  | StanLib when String.equal id.name "reduce_sum" || String.equal id.name "reduce_sum_static" -> 
    semantic_check_reduce_sum ~is_cond_dist ~loc id es
  | StanLib ->
    semantic_check_fn_stan_math ~is_cond_dist ~loc id es
  | UserDefined -> semantic_check_fn_normal ~is_cond_dist ~loc id es
```

src/stan_math_backend/Expression_gen.ml
```diff
@@ -300,6 +308,14 @@ and gen_fun_app ppf fname es =
...
         , f :: y0 :: t0 :: ts :: theta :: x :: x_int :: tl ) ->
           (fname, f :: y0 :: t0 :: ts :: theta :: x :: x_int :: msgs :: tl)
+      | ( true
+        , "reduce_sum"
+        , {pattern= FunApp (_, f, _); _} :: grainsize :: container :: tl ) ->
+          (strf "%s<%s>" fname f, grainsize :: container :: msgs :: tl)
+      | ( true
+        , "reduce_sum_static"
+        , {pattern= FunApp (_, f, _); _} :: grainsize :: container :: tl ) ->
+          (strf "%s<%s>" fname f, grainsize :: container :: msgs :: tl)
```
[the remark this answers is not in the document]
rok-cesnovar: Oh wow, thanks!

rok-cesnovar:
Thanks @nhuurre for reviewing. Will push the fixes tomorrow morning.

src/frontend/Semantic_check.ml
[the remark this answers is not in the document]
rok-cesnovar: Switched to using

```Ocaml
let arg_match (x_ad, x_t) y =
    UnsizedType.check_of_same_type_mod_conv "" x_t y.emeta.type_
    && UnsizedType.autodifftype_can_convert x_ad y.emeta.ad_level
  in
```

src/frontend/Semantic_check.ml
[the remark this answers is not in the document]
rok-cesnovar: Move the if to the when guard.

src/stan_math_backend/Stan_math_code_gen.ml
rok-cesnovar: Are you sure here? I added a test case in test/integration/good/code-gen/reduce_sum_m3.stan 
`  real t1a = reduce_sum(f1, y1, 1) + reduce_sum(f1a, y1, 1);`

where f1a is not used anywhere else and it creates the reduce sum functor and the model compiles. 

I was aware of this issue  thats why the find functors for reduce sum is:
```ocaml
let rec find_functors_expr accum Expr.Fixed.({pattern; _}) =
    match pattern with
    | FunApp (StanLib, "reduce_sum", {pattern= Var f; _} :: _) -> f :: accum
    | FunApp (StanLib, _, fk :: _) -> find_functors_expr accum fk
    | _ -> accum
```

while map_rects is
```ocaml
let find_functors_expr accum Expr.Fixed.({pattern; _}) =
    match pattern with
    | FunApp (StanLib, "map_rect", {pattern= Var f; _} :: _) -> f :: accum
    | _ -> accum
  in
```
Maybe I am misunderstanding something?

rok-cesnovar:
@nhuurre Thanks again! I think I resolved all the comments apart from the one about functors where I have a question. 

@bbbales2 I updated the test model with your changes. The updated error message is now:

```
Ill-typed arguments supplied to function 'reduce_sum'. Available arguments:
(int, int, T[,], ...) => real, int, T[,], ...
(int, int, T[,,], ...) => real, int, T[,,], ...
(int, int, T[,,,], ...) => real, int, T[,,,], ...
(int, int, T[,,,,], ...) => real, int, T[,,,,], ...
(int, int, T[,,,,,], ...) => real, int, T[,,,,,], ...
(int, int, T[,,,,,,], ...) => real, int, T[,,,,,,], ...
(int, int, T[,,,,,,,], ...) => real, int, T[,,,,,,,], ...
Where T is any one of int, real, vector, row_vector or matrix.
Instead supplied arguments of incompatible type: (int, real, real[], real, real) => real, real[], int, real
```

src/stan_math_backend/Stan_math_code_gen.ml
nhuurre: That doesn't make sense to me. The pattern for `reduce_sum(f1, ...) + reduce_sum(f1a,...)` is something like `FunApp (StanLib, "add", {...{Var "f1"}} :: {...{Var "f1a"}} :: [])`. Skipping the tail--as the recursive branch does--should hide "f1a". It didn't compile when I tried it.

`f1a_functor__` is [here](https://github.com/bstatcomp/stanc3/blob/5697210202b693f7a283a770eb36508952d03363/test/integration/good/code-gen/cpp.expected#L16753-L16761) but where is `f1a_rsfunctor__`?

Jenkins says all tests passed and I think that involves compiling all good test models so...I don't even know?

rok-cesnovar: Let me check again. This one does seem weird.

rybern: Ideally, the fold would recurse to all expression in the program and you wouldn't have to do it manually. I'm not sure what the behavior of Program.fold is. I ended up defining my own program fold in the pedantic mode branch that does recurse over all expressions and statements.

If the fold doesn't recurse over all expressions, why are we only looking for reduce_sum inside of StanLib function arguments? Couldn't it appear anywhere?

src/frontend/Semantic_check.ml
```diff
@@ -314,6 +316,47 @@ let semantic_check_fn_stan_math ~is_cond_dist ~loc id es =
...
+    |> Validate.error
+  in
+  match es with
+  | { emeta=
+        { type_=
+            UnsizedType.UFun
+              ( (_, UInt)
+                :: (_, UInt)
+                   :: ((_, sliced_arg_fun_type) as sliced_arg_fun) :: fun_args
+              , ReturnType UReal ); _ }; _ }
```
rybern, line 341: I wonder if the concrete types here are too strict - for example, do we want to allow functions that return Int, since they can be converted? In any case, too strict is probably fine.

src/middle/Stan_math_signatures.ml
```diff
@@ -251,13 +263,16 @@ let stan_math_returntype name args =
...
   in
-  if List.length filteredmatches = 0 then None
-    (* Return the least return type in case there are multiple options (due to implicit UInt-UReal conversion), where UInt<UReal *)
-  else
-    Some
-      (List.hd_exn
-         (List.sort ~compare:UnsizedType.compare_returntype
-            (List.map ~f:fst filteredmatches)))
+  match name with
+  | "reduce_sum" | "reduce_sum_static" -> Some (UnsizedType.ReturnType UReal)
```
rybern, line 267: Maybe we can pull these string constants out, since they're littered throughout the code? I think repeating string literals throughout the codebase is somewhat error-prone.

src/frontend/Semantic_check.ml
rok-cesnovar: Yeah, here I was debating with myself whether to allow int. I dont have an opinion. @wds15 or @bbbales2 do you guys have one?

src/middle/Stan_math_signatures.ml
rok-cesnovar: Good call, will do that. This will go away once we figure out the type variable thing.

src/frontend/Semantic_check.ml
```diff
@@ -314,6 +316,47 @@ let semantic_check_fn_stan_math ~is_cond_dist ~loc id es =
...
       |> Validate.error
 
+let semantic_check_reduce_sum ~is_cond_dist ~loc id es =
+  let arg_match (x_ad, x_t) y =
+    UnsizedType.check_of_same_type_mod_conv "" x_t y.emeta.type_
+    && UnsizedType.autodifftype_can_convert x_ad y.emeta.ad_level
+  in
+  let args_match a b =
+    if List.length a = List.length b then List.for_all2_exn ~f:arg_match a b
+    else false
```
rybern, line 326: could this if-then-else be an &&?

src/stan_math_backend/Stan_math_code_gen.ml
rok-cesnovar: Figured this out. Going to try the set thing also.

rok-cesnovar: Bailed on the sets thing. Will to a refactor with sets together with the fix to the map_rect function after this PR is merged.

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -657,12 +680,53 @@ let pp_register_map_rect_functors ppf p =
     (list ~sep:cut pp_register_functor)
     (List.mapi ~f:(fun i f -> (i + 1, f)) functors)
 
+let fun_used_in_reduce_sum p =
+  let rec find_functors_expr accum Expr.Fixed.({pattern; _}) =
+    match pattern with
+    | FunApp (StanLib, "reduce_sum", {pattern= Var f; _} :: _) -> f :: accum
+    | FunApp (StanLib, _, fk) -> 
+    List.concat [List.concat (List.map ~f:(fun i -> (find_functors_expr String.Set.empty i)) fk);accum]
+    | _ -> accum
```
nhuurre, line 689: Doesn't look like you "bailed on the sets thing", there's still a String.Set.empty here. More importantly, I don't think you don't need this FunApp branch. The last wildcard case should be
```ocaml
    | x -> Expr.Fixed.Pattern.fold find_functors_expr accum x
```
and covers everything.

rok-cesnovar: Fantastic. Thanks for all the tips! I redid this to sets now.

rok-cesnovar:
I think I resolved all the open comments now, apart from the question on whether we want to be strict on the user defined function return type or not. 

I will leave that one to @wds15 and @bbales2 to decide.

src/frontend/Semantic_check.ml
wds15: Hmm...never thought about this. Let’s check if we tested that in math and decide based on that? I really need to look that up.

bbbales2: I think only real for the function return type.

bbbales2: We'd never autodiff through something that returned an int, for instance.

wds15: Yeah...and we only documented that the function returns real. I am fine with that restriction.

rok-cesnovar: Cool cool. Will resolve this then.

wds15:
Just one thing: The error message says as a first line:

`(int, int, T[,], ...) => real, int, T[,], ...`

, but what about the signature

`(int, int, T[], ...) => real, int, T[], ...`

? Did I miss that? I mean, it's possible to instantiate this thing with an array of a scalar ( `std::vector<var>`), for example.

I am really looking forward to checkout `develop` from cmdstan and then have this shiny new `reduce_sum` available.

BTW... for the future we could think about handling sums of `reduce_sum` as

`reduce_sum() + reduce_sum()`

in an efficient way. That would be rat to do at some point. I got this thought when looking at the review comments here.

rok-cesnovar:
> The error message says as a first line

Ah, it was an off-by-one error in printing the error message (vector of 1 dimension means zero commas not 1 :) ). Fixed now. Thanks for pointing that out.

Once this passes I will start another build of the binaries.

rok-cesnovar:
Place 

STANC3_TEST_BIN_URL = https://jenkins.mc-stan.org/job/stanc3-test-binaries/54/artifact

in your make/local on cmdstan develop to test the latest binaries.

wds15:
For me this looks all great now.

wds15:
Just ran again one of my test models on this - speeds up nicely.

rok-cesnovar:
The deadline to get this in for 2.23 is end of monday anywhere on earth. Does anyone have any more comments or requests? Would love to see this get in. But obviously not gonna pressure if anyone objects. Thanks!

src/stan_math_backend/Expression_gen.ml
```diff
@@ -300,6 +308,10 @@ and gen_fun_app ppf fname es =
         , "integrate_ode_rk45"
         , f :: y0 :: t0 :: ts :: theta :: x :: x_int :: tl ) ->
           (fname, f :: y0 :: t0 :: ts :: theta :: x :: x_int :: msgs :: tl)
+      | ( true
+        , x
+        , {pattern= FunApp (_, f, _); _} :: grainsize :: container :: tl ) when List.mem ~equal:(String.equal) Stan_math_signatures.reduce_sum_functions x->
```
nhuurre, line 313: This could use some line breaks. It's kind of hard to review code that doesn't fit on the screen. I'm not saying you should manually format code though. Personally, I  have set up an editor hotkey that runs `ocamlformat` and I always hit that instead of enter.

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -173,7 +186,17 @@ let pp_fun_def ppf Program.({fdrt; fdname; fdargs; fdbody; _}) =
         functor_suffix pp_sig "operator()" pp_call_str
         ( fdname
         , List.map ~f:(fun (_, name, _) -> name) fdargs @ extra @ ["pstream__"]
-        )
+        ) ;
+      (* Produces the reduce_sum functors that has the pstream argument
+      as the third and not last argument *)
+      let first_two, rest_fdargs = List.split_n fdargs 2 in
+      if String.Set.mem funs_used_in_reduce_sum fdname then
```
nhuurre, line 193: Logically, the comment and the `let` belong inside the the `if` body.

src/frontend/Semantic_check.ml
```diff
@@ -314,6 +315,46 @@ let semantic_check_fn_stan_math ~is_cond_dist ~loc id es =
...
 
+let semantic_check_reduce_sum ~is_cond_dist ~loc id es =
+  let arg_match (x_ad, x_t) y =
+    UnsizedType.check_of_same_type_mod_conv "" x_t y.emeta.type_
+    && UnsizedType.autodifftype_can_convert x_ad y.emeta.ad_level
+  in
+  let args_match a b =
+    List.length a = List.length b && List.for_all2_exn ~f:arg_match a b
+  in
+  let return_generic_error =
```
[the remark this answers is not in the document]
rok-cesnovar: It was used twice before, but one case was removed, so yeah, its unnecessary. Thanks!

src/frontend/Semantic_error.ml
```diff
@@ -401,6 +446,12 @@ let illtyped_ternary_if loc predt lt rt =
 let returning_fn_expected_nonreturning_found loc name =
   TypeError (loc, TypeError.ReturningFnExpectedNonReturningFound name)
 
+let illtyped_reduce_sum loc name arg_tys args =
+  TypeError (loc, TypeError.IllTypedReduceSum (name, arg_tys, args))
+
+let illtyped_reduce_sum_generic loc name arg_tys =
+  TypeError (loc, TypeError.IllTypedReduceSumBasic (name, arg_tys))
```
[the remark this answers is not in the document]
rok-cesnovar: Ah, renamed to generic but wasnt thorough enough. Thanks!

src/middle/Stan_math_signatures.ml
```diff
@@ -128,6 +139,8 @@ let mk_declarative_sig (fnkinds, name, args) =
 let full_lpdf = [Lpdf; Rng; Ccdf; Cdf]
 let full_lpmf = [Lpmf; Rng; Ccdf; Cdf]
 
+let reduce_sum_functions = ["reduce_sum"; "reduce_sum_static"]
```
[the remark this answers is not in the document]
rok-cesnovar: Good one!

src/stan_math_backend/Expression_gen.ml
```diff
@@ -143,6 +143,12 @@ let fn_renames =
 *)
 let map_rect_counter = ref 0
 let functor_suffix = "_functor__"
+let reduce_sum_functor_suffix = "_rsfunctor__"
+
+let functor_suffix_select f =
+  match f with
```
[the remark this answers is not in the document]
rok-cesnovar: Thanks!

src/stan_math_backend/Expression_gen.ml
rok-cesnovar: I typically run `dune build @fmt` which is run with `make format`. I think I forgot to run it here.

rok-cesnovar:
Thanks @nhuurre !

Your comments were are very much on point. I addressed all of them.

Really excited to get this in and see it in action. 
Thanks @wds15 @bbbales2 and @SteveBronder for the Math backend, the majority of work was done there. Thanks @nhuurre and @rybern for all the comments and advices! Very nice team effort!

PR #454, opened 2020-02-04
improve ternary-if typechecking

nhuurre:
Copy Stanc2 error messages for ill-typed ternary operator `? :` expression.

### DIFF
--- a/src/frontend/Semantic_error.ml
+++ b/src/frontend/Semantic_error.ml
@@ -79,17 +79,15 @@ module TypeError = struct
            type %a and rhs has type %a"
           (Pretty_printing.pretty_print_assignmentoperator assignop)
           UnsizedType.pp lt UnsizedType.pp rt
-    | IllTypedTernaryIf (ut1, ut2, ut3) ->
-        if ut1 = UnsizedType.UInt then
-          Fmt.pf ppf
-            "Type mismatch in ternary expression, expression when true is: \
-             %a; expression when false is: %a"
-            UnsizedType.pp ut2 UnsizedType.pp ut3
-        else
-          Fmt.pf ppf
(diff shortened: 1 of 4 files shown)


src/frontend/Semantic_error.ml
```diff
@@ -80,12 +80,16 @@ module TypeError = struct
           (Pretty_printing.pretty_print_assignmentoperator assignop)
           UnsizedType.pp lt UnsizedType.pp rt
     | IllTypedTernaryIf (ut1, ut2, ut3) ->
-        Fmt.pf ppf
-          "Ill-typed arguments supplied to ? : operator. Available \
-           signatures: %s\n\
-           Instead supplied arguments of incompatible type: %a, %a, %a."
-          (Stan_math_signatures.pretty_print_math_sigs "if_else")
-          UnsizedType.pp ut1 UnsizedType.pp ut2 UnsizedType.pp ut3
+        if ut1 = UnsizedType.UInt then
```
seantalts: This should just be part of the pattern match on line 82 I think. You'll end up with two lines,

```ocaml
    | IllTypedTernaryIf (UInt, ut2, ut3) -> ...
    | IllTypedTernaryIf (ut1, ut2, ut3) ->...
```

src/middle/UnsizedType.ml
```diff
@@ -87,6 +87,13 @@ let check_compatible_arguments_mod_conv name args1 args2 =
             && autodifftype_can_convert (fst sign1) (fst sign2) )
           args1 args2)
 
+let rec common_type = function
```
seantalts: Can you add a doc string to this? I'm not sure what it's doing from the name.

Issue #453, opened 2020-02-03
"offset" is not allowed as identifier


wds15:
The variable name "offset" makes stanc3 reject valid Stan programs.

Example:

```stan
data {
  int<lower=0> N;
  int<lower=0,upper=1> y[N];
}
transformed data {
  int offset = 10;
}
parameters {
  real<lower=0,upper=1> theta;
}
model {
  theta ~ beta(1,1);  // uniform prior on interval 0,1
  y ~ bernoulli(theta);
}
```

gives me

```
--- Translating Stan model to C++ code ---
bin/stanc  --o=examples/bernoulli/bernoulli.hpp examples/bernoulli/bernoulli.stan

Syntax error in 'examples/bernoulli/bernoulli.stan', line 6, column 2 to column 5, parsing error:
   -------------------------------------------------
     4:  }
     5:  transformed data {
     6:    int offset = 10;
              ^
     7:  }
     8:  parameters {
   -------------------------------------------------

Expected range constraint or identifier as part of top-level variable declaration.



--- Translating Stan model to C++ code ---
bin/stanc  --o=examples/bernoulli/bernoulli.hpp examples/bernoulli/bernoulli.stan

Syntax error in 'examples/bernoulli/bernoulli.stan', line 6, column 2 to column 5, parsing error:
   -------------------------------------------------
     4:  }
     5:  transformed data {
     6:    int offset = 10;
              ^
     7:  }
     8:  parameters {
   -------------------------------------------------

Expected range constraint or identifier as part of top-level variable declaration.
```

This is a bit unfortunate as "offset" is used by brms, for example...

EDIT: this is the released cmdstan 2.21.1 on Mac (but same problem on Linux)

avehtari:
Release notes should have included new reserved keywords in stanc3.
For offset see https://github.com/stan-dev/stanc3/issues/21

avehtari:
See also discussion at https://discourse.mc-stan.org/t/list-of-stanc3-new-reserved-keywords/12948

avehtari:
@mcol lists `target`, `lower`, `upper`, `offset` and `multiplier` as new reserved keywords, but is not certain whether this is complete list https://discourse.mc-stan.org/t/list-of-stanc3-new-reserved-keywords/12948/5

wds15:
Ok... this is not a bug, but we clearly need better doc. Moreover, the error message is not useful for users. That should be really improved, unless I am overlooking something.

[...]

PR #457, opened 2020-02-04
Separate parser warning for declaring a variables name as a keyword

rybern:
Addressing warning discussion from issue #453 

The upside of merging this is slightly clearer parser errors, the downside is that the parser's a little bigger

Compiling the program
```
transformed data {
  int offset = 10;
}
model { }
```
now results in
```
Syntax error in 'offset.stan', line 2, column 6 to column 12, parsing error:
     1:  transformed data {
     2:    int offset = 10;
                     ^
     3:  }
     4:  model { }

Identifier expected after type in top-level variable declaration.
A reserved keyword was found, which cannot be used as an identifier.
```

The definition of keyword here is currently an arbitrary list of tokens: OFFSET, LOWER, UPPER, MULTIPLIER, TARGET, or PRINT. This list should probably grow to include all the keywords that users are likely to try as names.

The implementation here is a little weird. The difficulty is that you can't write two distinct messages in `parser.messages` unless they correspond to distinct parser states. Previously, there was only parser state for 'unexpected identifier'. I needed to add a new dummy rule to give the parser a distinct state when a keyword was used. But, I also don't ever want that dummy rule to actually be built, so I had it terminate with an UNREACHABLE token which will never be generated.

Note since there's a large diff for parser.messages: the new stuff is at the very end.


wds15:
I can’t review...but this looks cool. Thanks!

mcol:
Is it possible to mention the keyword in the error message?

rybern:
Update messages to @bob-carpenter's message, and added the keyword. New example:
```
     1:  transformed data {
     2:    int offset = 10;
                     ^
     3:  }
     4:  model { }

Expected identifier, but found reserved keyword 'offset'
```


The keyword is possible to include, it just requires writing out a different rule for each case. There is actually a trick to include source tokens (like the keyword used, see [here](https://discuss.ocaml.org/t/referencing-tokens-in-menhirs-messages/4000)) dynamically into parser error messages, but I would have to rearrange how we read in source files.

bob-carpenter:
I don't think we need anything more than a caret pointing into
the right palce and a message about what went wrong.  They can
see directly that it's in a declaration context.

seantalts:
@VMatthijs any chance you'd have some time to take a look at this? Or maybe @bob-carpenter as a backup? I can be 3rd backup if need be, just haven't spent much time in Menhir.

Also @rybern would you mind documenting maybe on the wiki how you went about fixing the error messages? That would be super helpful for folks going forward, I don't think any of us really knew how to do it before, haha.

rybern:
Sure thing. Looking back at this, I think I agree with @bob-carpenter that we don't really need the error message to repeat which keyword was used, so I'm going to revert that unless anyone has objections.

mcol:
I kind of like the redundant keyword in the message, even if we already have the pointer (should it point to beginning of offset rather than at the end?). Also Rust does the same, and it's generally loved for its helpful messages (https://blog.rust-lang.org/2016/08/10/Shape-of-errors-to-come.html). But perhaps this should be judged based on how difficult is coding or maintaining this, and I have no idea about that! In any case, I won't object if you want to revert.

rybern:
Ok, added a Wiki page.

It does add a little bit of extra code, but nothing huge. Sounds like some people prefer it so I'll leave it :)

I do think my list of likely-mistaken keywords to warn about (upper, lower, multiplier, target, print) was pretty arbitrary, maybe someone could some up with a better list.

src/frontend/parser.mly
```diff
@@ -225,6 +245,11 @@ top_var_decl:
                        is_global= true};
        smeta= {loc=Location_span.of_positions_exn $startpos $endpos}}
     }
+  (*
+  | top_var_type OFFSET UNREACHABLE option(dims) SEMICOLON
+    { raise (Failure "This should be unreachable; the UNREACHABLE token should \
+                      never be produced") }
+   *)
```
seantalts: Do you need this commented out code?

test/integration/bad/stanc.expected
```diff
@@ -2455,16 +2455,16 @@ Ill-typed arguments to '~' statement. No distribution 'ormal' was found with the
 
   $ ../../../../install/default/bin/stanc target-reserved.stan
 
-Syntax error in 'target-reserved.stan', line 2, column 2 to column 6, parsing error:
+Syntax error in 'target-reserved.stan', line 2, column 7 to column 13, parsing error:
    -------------------------------------------------
      1:  data {
      2:    real target;
-               ^
+                      ^
```
seantalts: How hard would it be to point this at the beginning of 'target' instead?

seantalts:
[The wiki page](https://github.com/stan-dev/stanc3/wiki/How-to-change-or-add-a-new-custom-parser-error-message) looks awesome, thank you!

test/integration/bad/stanc.expected
rybern: Unfortunately I think pretty hard. The only way to get it to differentiate cases between keywords and anything else is to have the keyword on the stack when the error occurs, which means it points at one later token than the original case. We could change the error printing code to show the whole location span instead of the end of the span, but then I think other cases would underline the previous token (real in this case).

src/frontend/parser.mly
rybern: Whoops, fixed this with commit revise and push --force - turns out that doesnt work. fixed now

VMatthijs:
> Ok, added a Wiki page.
> 
> It does add a little bit of extra code, but nothing huge. Sounds like some people prefer it so I'll leave it :)
> 
> I do think my list of likely-mistaken keywords to warn about (upper, lower, multiplier, target, print) was pretty arbitrary, maybe someone could some up with a better list.

You could look at the lexer for a full list of keywords that should be caught during parsing. (There are some more that are caught during the semantic check, but you don't need to do more work there.)

Some keywords that are perhaps still missing:
functions
data
transformed data
model 
parameters
transformed parameters
generated quantities
and lots of others.

Just look at the tokens that are constructed in the [lexer](https://github.com/stan-dev/stanc3/blob/master/src/frontend/lexer.mll).

bob-carpenter:
I don't think we have to restrict block names because using them as an identifier won't lead to ambiguity anywhere.

PR #461, opened 2020-02-08
Move base_rng__ argument in ud function definitions

### LINKED ISSUE #460: stanc3 generated code cannot compile - rng function
mismatch between argument order in rng function declaration and invocation.

Stan code:
```
functions {
  real test_rng(vector mu) {
    return normal_rng(0, 1);
   }
}
parameters {
  vector[2] mu;
}
model {
}
generated quantities {
  real r = test_rng(mu);
}
```

compiler error:

```
hiroki.hpp:234:11: error: no matching function for call to 'test_rng'
      r = test_rng(mu, base_rng__, pstream__);
          ^~~~~~~~
```

hiroki.hpp function definition:

```
template <typename RNG, typename T0__>
typename boost::math::tools::promote_args<T0__>::type
test_rng(RNG& base_rng__, const Eigen::Matrix<T0__, -1, 1>& mu,
         std::ostream* pstream__) {
```

function invocation - error above.

nhuurre:
Fixes #460 

There was a commented out test, not sure why.


nhuurre:
There's a similar-but-not-identical mother model in this repo too https://github.com/stan-dev/stanc3/blob/master/test/integration/good/code-gen/mother.stan How are these related?

I think the only test needed is to actually call `foo_rng` somewhere in the model, not just define it.

seantalts:
Sounds good. The practical difference today is that the example models
version is used by stanc3 and the performance tests for end to end testing.
The models in the stanc3 repo aren’t the authoritative source for any end
to end tests.

PR #464, opened 2020-02-13
Added exhaustive list of keyword tokens to keyword parser errors

rybern:
Expanded the list of keywords to cover the tokens left out of #457


src/frontend/parser.mly
```diff
@@ -131,12 +131,40 @@ decl_identifier:
      that's distinct from the use of other non-identifiers, so we can assign
      it a different message in the .messages file.
    *)
```
VMatthijs: Nice! Very clever use of the parser generator!

VMatthijs:
@rybern , did you manage to use menhir to update the error messages file to ensure we retain 100% coverage of all error states with an informative error message? Is this what you wrote up in the wiki? I'd be really curious to hear what your experience was with that proces. It's a bit subtle, right? Do you think we need to automate more?

VMatthijs:
I've now read the wiki entry you wrote and I think there are probably still some things missing. I should've documented this better myself. Some remarks:
- to discover the state number of a specific error, you can also just use the flag --debug-parse when parsing a Stan model which triggers the error; it should then list the parsing rules that hit for parsing the file, followed by the number of the error state it ends up in;
- there are a bunch of menhir commands that should be used whenever you change the parser: one for checking whether the .messages file has 100% coverage of all error states and one for automatically generating stubs for the missing states in an .messages file, which can then be completed with informative error messages by the user upon inspection of the stack for that state; these commands were documented in the wiki once upon a time; I think it may now be covered by `dune build @update_messages` followed by `dune promote`;

Could you maybe check whether this works and if so document it in the wiki?

rybern:
Thanks Matthijs -

> * to discover the state number of a specific error, you can also just use the flag --debug-parse when parsing a Stan model which triggers the error; it should then list the parsing rules that hit for parsing the file, followed by the number of the error state it ends up in;
This is definitely nicer, I'll update the wiki.

> * there are a bunch of menhir commands that should be used whenever you change the parser: one for checking whether the .messages file has 100% coverage of all error states and one for automatically generating stubs for the missing states in an .messages file, which can then be completed with informative error messages by the user upon inspection of the stack for that state; these commands were documented in the wiki once upon a time; I think it may now be covered by `dune build @update_messages` followed by `dune promote`;
I have been using update_messages, yeah, I should put that in the wiki as well. It doesn't seem to check for completeness though, and I don't think I've figure out how to do that.

rybern:
Added those two commands to the wiki

VMatthijs:
> It doesn't seem to check for completeness though, and I don't think I've figure out how to do that.

Correct. I usually check completeness by running the following commands from the `src/frontend` folder:

```
menhir --list-errors parser.mly > parser_new.messages ;
menhir parser.mly --compare-errors parser_new.messages --compare-errors parser.messages 
```

This first generates a fresh files `parser_new.messages` which includes dummy error messages for **all** error states. Next, it checks that `parser.messages` has at least at much coverage as `parser_new.messages` (hence 100% coverage). If certain error states, are missed, it will complain about that. For example, if states 68 and 336 had been missed in `parser.messages`, it will say 

```
Read 461 sample input sentences and 461 error messages.
Read 459 sample input sentences and 459 error messages.
File "parser_new.messages", line 6648, characters 0-58:
Error: this sentence leads to an error in state 68.
No sentence that leads to this state exists in "parser.messages".
File "parser_new.messages", line 13297, characters 0-63:
Error: this sentence leads to an error in state 336.
No sentence that leads to this state exists in "parser.messages".
```

The thing to do, is then to copy the entries for states 68 and 336 from `parser_new.messages` to `parser.messages` and to insert informative error messages.

Finally, `parser_new.messages` can be deleted.

I wish we could automate some of this workflow as it's always the same, but menhir doesn't seem to fascilitate that.

Is there any chance you could add this to your wiki? I think it's important information that we shouldn't lose track of.

VMatthijs:
(Then there's also the `--update-errors` command that is useful if the grammar has changed dramatically.
You use it by running 

```
 menhir parser.mly --update-errors parser.messages > parser_updated.messages
```

This will try to repurpose the error messages of `parser.messages` for the updated grammar which may have totally different state numbering. It will put the resulting messages file (which will presumably not be complete) in `parser_updated.messages`. To make it complete, you can again follow the steps above.

Perhaps it's also useful to mention this in the wiki.
)

VMatthijs:
What do you think? Should we create shell scripts to do all of this? @bob-carpenter ?

rybern:
@VMatthijs Finally getting back to this!
Your explanation was super helpful, I've been using those three additional menhir commands to check for completeness whenever I modify the parser. It would be very hard to merge branches with modified parsers without --update-errors.
Menhir shows the messages in this branch to be complete, so I think we can merge.

VMatthijs:
Great! Feel free to merge!

Should we copy the text I wrote about the menhir commands to a wiki so we don't lose track of it?

rybern:
Sure, I can do that now

PR #476, opened 2020-02-26
use underscore in multi word flags

### LINKED ISSUE #468: consistent use of hyphens underscores for compiler option names
old stanc compiler options names had underscores, e.g.:  `allow_undefined`, `include_paths`.

stanc3 compiler options use hyphens, e.g.: `use-opencl`, `warn-uninitialized`

standardize - naming should be uniform - whatever conventions are used for CLI names should be adopted here - makefile options and CmdStan arguments all use underscores - stanc3 interface should too.

rok-cesnovar:
Fixes #468

This changes every multi word flag to use underscore. These looks like it touches a lot of files but the only real changes are in https://github.com/stan-dev/stanc3/commit/3948c70441ba06b18d5c61cc91602af114104d38

Alternatives we could do here are: 
- use hyphens for all which breaks backward compatibility
- use underscores for end user flags only (only need to change user-opencl and warn-uninitialized) tho that feels weird


rybern:
Sorry this never got merged. I would be in favor of using hyphens, because I think that looks better and is much more typical. What do you think?

rok-cesnovar:
I am also in favor of hyphens. Can we specify two flags for an option? Otherwise I will close this and we will do this when we decide to bump major versions.

rok-cesnovar:
Ah ok, you answered that we cant. Will make duplicated options for backward compatibility with a new PR. Thanks!

rybern:
Yeah, duplicated is probably the way to go.

It looks from the [Arg docs](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Arg.html#TYPEdoc) like you could change the usage string to not include the duplicated arguments. That might be a sneaky way of depreciating them and not cluttering the interface while also being backwards compatible.

rok-cesnovar:
Thanks for the link!

PR #491, opened 2020-04-01
[WIP] Sparse Matrix

SteveBronder:
This is a WIP PR for figuring out some issues with sparse matrices. I'll use the following stan program which can be found [here](https://github.com/stan-dev/cmdstan/blob/feature-test/sparse-matrix/examples/sparse/sparse.stan) in a cmdstan branch which should have all the stuff we need one the C++ side

You can get the current implementation of stanc3 with sparse matrices by doing 

```bash
git remote add stevebronder  https://github.com/SteveBronder/stanc3
git fetch stevebronder
git checkout stevebronder/feature/sparse-matrix-static
```

The TL;DR is that this adds the `SizedType` `SSparseMatrix` and `SStaticSparseMatrix` and `UnsizedType` `USparseMatrix` which in C++ should translate to `Eigen::SparseMatrix`. `SSparseMatrix` is passed the non zero rows and columns and `SStaticSparseMatrix` is immutable and is only given the dimensions of the sparse matrix.

I've been talking to @rybern about this and he summed it up nicely below which I'll annotate
- [x] All sparse matrices will have static sparsity structure
- [x] All sparse matrices will either (A) be declared in data or parameters with the nonzero entries specified, or (B) share the sparsity structure of previously defined sparse matrix
- [x] All sparse matrices defined in transformed parameters /data will be immutable
    - For sparse matrices in the data and parameters block this means the user passes in the nonzero rows and columns
   - i.e. `sparse_matrix[nz_rows, nz_cols, N, M] x;`
   - For transformed data and parameters it means that once a sparse matrix is declared in these blocks it can be accessed but not assigned again
   - i.e. good: `sparse_matrix[N, M] xx = x * transpose(x);`
  - bad!: `xx[1, 1] = 5.0`
So the work on Stanc would be
- [x] Adding the new matrix type, with two separate type declaration methods
- [x] Ensuring that the correct type declaration method is allowed in each block
- [x] Disallowing indexed variables of the new type from being on the LHS of an assignment
- [ ] Implementing the codegen for declaration and operations on the new type

EDIT: Thanks to @rybern got pretty much all of the language stuff done! Just need to fix the c++ code gen at this point 

I'll go over each of the 4 working points above with the example stan program below. It transpiles on the current branch but fails to generate the correct C++ and has semantic issues 

```stan
data {
  int N;
  int M;
  int nz_vals;
  int nz_rows[nz_vals];
  int nz_cols[nz_vals];
  sparse_matrix[nz_rows, nz_cols, N, M] x; 
  // Should not be allowed!
  sparse_matrix[N, M] bad_x; // (1)
}

transformed data {
  sparse_matrix[N, M] xx = x * transpose(x);
  // Should be allowed
  real data_cool = xx[1, 1]; // (2)
  // Should not be allowed
  xx[1, 1] = 5.0; // (3)
  xx[1:10, 1:10] = x[1:10, 1:10]; // (3)

}


parameters {
  sparse_matrix[nz_rows, nz_cols, N, M] VV;
}

transformed parameters {
  sparse_matrix[N, M] ZZ = VV * nz_vals;
  // Cool and fine
  real param_cool = ZZ[1, 1]; // (2)
  // Neither are cool or fine
  ZZ[1, 1] = 10.0; // (3)
  ZZ[1:10, 1:10] = VV[1:10, 1:10] * 10.0; // (3)
}
```

1. Semantic Checking that `SStaticSparseMatrices` are not defined in data or parameters block

2. The assignment in (3) such as `ZZ[1, 1] = 10.0;` should throw an error that static sparse matrices are not assignable but I'm having a hard time figuring out how to check this. It may be in part because I was lazy and used `USparseMatrix` as the unsized type for both `SSparseMatrix` and `SStaticSparseMatrix`. I used only one unsized type because I thought that would be easier for generating the stan math signatures.

On the compilation side there's some bad c++ which is also available in the cmdstan branch for the example above [here](https://github.com/stan-dev/cmdstan/blob/feature-test/sparse-matrix/examples/sparse/sparse.hpp)

- The code for `  sparse_matrix[N, M] xx = x * transpose(x);` first generates a for loop over the sparse matrix to fill it with `NA` values

```cpp
      for (size_t sym1__ = 1; sym1__ <= N; ++sym1__) {
        current_statement__ = 13;
        for (size_t sym2__ = 1; sym2__ <= M; ++sym2__) {
          current_statement__ = 13;
          assign(xx,
            cons_list(index_uni(sym1__),
              cons_list(index_uni(sym2__), nil_index_list())),
            std::numeric_limits<double>::quiet_NaN(), "assigning variable xx");
        }}
```

But that sort of defeats the purpose of the sparse matrix. Since we won't know what values of the sparse matrices will be nonzero until after the assignment happens I think the compiler should not generate this for `SStaticSparseMatrix` 

- When we define the compiler tries to [sum up the nz_rows and nz_cols vectors](https://github.com/stan-dev/cmdstan/blob/feature-test/sparse-matrix/examples/sparse/sparse.hpp#L277)
```cpp
      num_params_r__ += nz_rows * nz_cols * N * M;
```

What we want is just one

```cpp
      num_params_r__ += stan::math::size(nz_rows);
```

To make this easier it may be worth changing the signature to 

```stan
  sparse_matrix[nz_idx, N, M] VV;
```

where nz_idx could be something like a vector of arrays (or maybe an array of arrays)? So that this seems more approachable and we don't need to check that the nonzero row and column indices are the same size. Then the above becomes

```cpp
      num_params_r__ += stan::math::size(nz_idx);
```
- The loop in [`constrained_param`](https://github.com/stan-dev/cmdstan/blob/feature-test/sparse-matrix/examples/sparse/sparse.hpp#L533) names is very not good

```cpp
    for (size_t sym1__ = 1; sym1__ <= M; ++sym1__) {
      {
          {
            for (size_t sym3__ = 1; sym3__ <= nz_cols; ++sym3__) {
              {
                for (size_t sym4__ = 1; sym4__ <= nz_rows; ++sym4__) {
                  {
                    param_names__.push_back(std::string() + "VV" + '.' + std::to_string(sym4__) + '.' + std::to_string(sym3__) + '.' + std::to_string(sym2__) + '.' + std::to_string(sym1__));
                  }}
              }}
          }}
      }}
```

What we want it to do is loop over the nonzero entries of the sparse matrix like in the Eigen docs for [Iterating over the nonzero coefficients ](https://eigen.tuxfamily.org/dox/group__TutorialSparse.html)

```cpp
for (int k = 0; k < VV.outerSize(); ++k) {
  for (SparseMatrix<double>::InnerIterator it(VV, k); it; ++it) {
     param_names__.push_back(std::string() + "VV" + '.' + std::to_string(it.row()) + '.' + std::to_string(it.col()));
  }
}
```

[`get_constrained_sizedtypes()`](https://github.com/stan-dev/cmdstan/blob/feature-test/sparse-matrix/examples/sparse/sparse.hpp#L599) where it tries to print out the nonzero rows and column vectors

```cpp
s__ << "[{\"name\":\"VV\",\"type\":{\"name\":\"sparse_matrix\",\"nonzero rows\":" << nz_rows // ...
```

I think the above suggestion of using `nz_idx` and getting the size of it would be better here

```cpp
s__ << "[{\"name\":\"VV\",\"type\":{\"name\":\"sparse_matrix\",\"nonzero indices\":" << stan::math::size(nz_idx) // ...
``` 

I think that's all the issues, they feel pretty resolvable


SteveBronder:
@rybern So I can get this almost compiling now but I'm running into a design issue. The codegen for the following looks good

```stan
data {
  int N;
  int M;
  int nz_vals;
  int nz_rows[nz_vals];
  int nz_cols[nz_vals];
  sparse_matrix[nz_rows, nz_cols, N, M] x;
  vector[M] y;
}

transformed data {
  sparse_matrix[N, M] xx = x * transpose(x);
  // Should be allowed
  real data_cool = xx[1, 1];

}


parameters {
  sparse_matrix[nz_rows, nz_cols, N, M] VV;
}

transformed parameters {
  sparse_matrix[N, M] ZZ = VV * nz_vals;
  real param_cool = ZZ[1, 1];
  vector[M] A;
  real<lower=0> sigma;

}

model {
  A ~ normal(0, 1);
  y ~ normal(xx * A, sigma);
}

generated quantities {
  sparse_matrix[N, M] BB = ZZ * param_cool;
}

```

Except for (un)constrained_param_names when printing out dynamic sparse matrices

```cpp
  void constrained_param_names(std::vector<std::string> &param_names__,
                               bool emit_transformed_parameters__ = true,
                               bool emit_generated_quantities__ = true) const {

    for (size_t sym1__ = 1; sym1__ <= stan::math::size(nz_rows); ++sym1__) {
      {
        param_names__.push_back(std::string() + "VV" + '.' +
                                std::to_string(nz_rows[(sym1__ - 1)]) + '.' +
                                std::to_string(nz_cols[(sym1__ - 1)]));
      }
    }
    if (emit_transformed_parameters__) {
      for (size_t sym1__ = 0; sym1__ < ZZ.outerSize(); ++sym1__)
        for (Eigen::SparseMatrix<double>::InnerIterator sym2__(ZZ, sym1__);
             sym2__; ++sym2__) {
          param_names__.push_back(std::string() + "ZZ" + '.' +
                                  std::to_string(sym2__.row()) + '.' +
                                  std::to_string(sym2__.col()));
        }
      param_names__.push_back(std::string() + "param_cool");
      for (size_t sym1__ = 1; sym1__ <= M; ++sym1__) {
        {
          param_names__.push_back(std::string() + "A" + '.' +
                                  std::to_string(sym1__));
        }
      }
      param_names__.push_back(std::string() + "sigma");
    }

    if (emit_generated_quantities__) {
      for (size_t sym2__ = 0; sym2__ < BB.outerSize(); ++sym2__)
        for (Eigen::SparseMatrix<double>::InnerIterator sym3__(BB, sym2__);
             sym3__; ++sym3__) {
          param_names__.push_back(std::string() + "BB" + '.' +
                                  std::to_string(sym3__.row()) + '.' +
                                  std::to_string(sym3__.col()));
        }
    }

  } // constrained_param_names()
```

I forgot `ZZ` and `BB` were going to be a local variables for log_prob etc. I'm not sure how to handle this since we can't just iterate over the nonzero elements. @bob-carpenter would it be terribly inefficient to print the names of dynamic sparse matrix parameters as if they just had N and M params?

rybern:
So we need to emit names for each nonzero element of ZZ and BB, but are they are local to log_prob, so we can't build their iterator?

If I understand correctly, you could
* Not use sparsity and use M * N, which is a worse complexity class
* Expand the scope of either ZZ and BB themselves or their iterators (which get assigned once when ZZ/BB are assigned, assuming that their sparsity structure never changes?)
* Figure out the sparsity structure of ZZ and BB as a function of nz_rows and nz_cols at compile time
* Not properly report each nonzero element separately for generated quantities/transformed params

bob-carpenter:
Yes, if we're going sparse, we can't have anything that does N * M work as we usually have the number of entries << N * M.

I don't see the problem, though.  Can't you iterate over the non-zero entries given a sparse matrix data structure?  Then we'd still print out a sparse matrix a using a.2.3 for entry a[2, 3].  Why do you need a local data structure?  Isn't it in the parameter or transformed parameter declaration or at least in the variable itself?

SteveBronder:
> So we need to emit names for each nonzero element of ZZ and BB, but are they are local to log_prob, so we can't build their iterator?

Yep! This is just for the static sparse matrices in the transformed blocks where the number of nonzeros rows/cols is not declared explicitly by the user

> If I understand correctly, you could

> - Not use sparsity and use M * N, which is a worse complexity class

Yes though this would be kind of a bummer since if you had say a square matrix like size 1MM with only 1% filled out it would be very wasteful

> - Expand the scope of either ZZ and BB themselves or their iterators (which get assigned once when ZZ/BB are assigned, assuming that their sparsity structure never changes?)

- I thought about this but idt we can because then we make log_prob stateful

> - Figure out the sparsity structure of ZZ and BB as a function of nz_rows and nz_cols at compile time

idt we can figure these out at compile time, but one sort of hacky solution is to have, at the end of the log_prob functions something like 

```cpp
// ZZ_nz(rows)(cols)_ is a class member
std::vector<int> ZZ_nzrows_ = get_nz_rows(ZZ);
std::vector<int> ZZ_nzcols_ = get_nz_cols(ZZ);
```

then in the name generator we just do the same loop we do for the data and parameters sparse matrices.

```cpp
    for (size_t sym1__ = 1; sym1__ <= stan::math::size(ZZ_nz_rows_); ++sym1__) {
      {
        param_names__.push_back(std::string() + "ZZ" + '.' +
                                std::to_string(ZZ_nz_rows_[(sym1__ - 1)]) + '.' +
                                std::to_string(ZZ_nz_cols_[(sym1__ - 1)]));
      }
    }
```

I think this is possible but kind of an ugly solution.

> - Not properly report each nonzero element separately for generated quantities/transformed params

We'd have to talk to Aki et al but yeah if we could just not report them that would be the easiest thing. Though idk if that would work

> Yes, if we're going sparse, we can't have anything that does N * M work as we usually have the number of entries << N * M.

> I don't see the problem, though. Can't you iterate over the non-zero entries given a sparse matrix data structure? 

The problem is that transformed parameters are defined in the log_prob etc. calculations. Since they are deleted after we leave that function we can't pull out their sparsity structure afterwards

Then we'd still print out a sparse matrix a using a.2.3 for entry a[2, 3]. Why do you need a local data structure? Isn't it in the parameter or transformed parameter declaration or at least in the variable itself?

You can see the full stan hpp file [here](https://gist.github.com/SteveBronder/2d932b35258668305784a33fe17196c9#file-sparse_ex-hpp-L281). The problem is the above where `ZZ` is only defined locally, so once we leave that function scope we don't know which entries were nonzero.

```cpp
      Eigen::SparseMatrix<local_scalar_t__> ZZ;
      ZZ = Eigen::SparseMatrix<local_scalar_t__>(N, M);

      current_statement__ = 2;
      assign(ZZ, nil_index_list(), multiply(VV, nz_vals),
             "assigning variable ZZ");
```

rybern:
> idt we can figure these out at compile time, but one sort of hacky solution is to have, at the end of the log_prob functions something like
> ```
> // ZZ_nz(rows)(cols)_ is a class member
> std::vector<int> ZZ_nzrows_ = get_nz_rows(ZZ);
> std::vector<int> ZZ_nzcols_ = get_nz_cols(ZZ);
> ```
Yeah, this is basically what I meant by expanding the scope of the matrix iterators. I agree it's kind of ugly. You could minimize the ugliness by using iterators instead of the two vectors (unless those are better), and only doing the assignment once instead of at the end of every iteration (by checking if it's been changed from some default value) unless the sparsity structure changes over time. I'm not really seeing a better solution if it really is necessary to print them all.

bob-carpenter:
The current model obeys the following invariants:

1.  The size of every parameter, transformed parameter, and generated quantity depends only on data variables and is hence is known as soon as data is loaded and never changes across iterations.  The never changing across iterations is critical to make sure we get rectangular output from sampling and thus all of our downstream analysis like ESS and Rhat will continue to work.
2.  Member variables represent data and are constant once they're loaded.  As a consequence, methods cannot be stateful.  Breaking this contract would wreak havoc on  multithreading.  We desperately need to move to evaluating parallel threads using only a single model instance to cut down on memory pressure from trying to run 8 processes, each with its own copy of the data.
3.  The model log density method `log_prob` is typically evaluated multiple times per iteration. It is assumed this can be evaluated asynchronously and statelessly.  
4.  The model method `write_array` does the output given input of the unconstrained parameters.  It calculates generated quantities and is only evaluated once per iteration using `double` scalars rather than autodiff variables.

> The problem is that transformed parameters are defined in the log_prob etc. calculations. Since they are deleted after we leave that function we can't pull out their sparsity structure afterwards

We don't write from `log_prob`, only from `write_array`, and as noted above, it recreates the transformed parameters and generates the generated quantities.

SteveBronder:
So let's say at the end of log_prob or write array we had something like 

```cpp
// Alias here is just to make the example look nicer
using sparse_iter = Eigen::SparseMatrix<double>::InnerIterator;
// calc_sparsity_ is a class member with type mutable std::atomic<bool> default true
if (ZZ_calc_sparsity_) {
  for (size_t sym1__ = 0; sym1__ < ZZ.outerSize(); ++sym1__) {
    for (sparse_iter sym2__(ZZ, sym1__); sym2__; ++sym2__) {
      // ZZ_iterator is type mutable std::vector<std::array<int, 2>>
      ZZ_iterator_.push_back(std::array<int, 2>{sym2__.row(), sym2__.col()});
    }
  }
  ZZ_calc_sparsity_ = false;
}
```

Then in `constrained_param_names` we would have

```cpp
if (!ZZ_calc_sparsity_) {
      for (int i = 0; i < ZZ_iterator_.size(); i++) {
          param_names__.push_back(std::string() + "ZZ" + '.' +
                                  std::to_string(ZZ_iterator_[i][1]) + '.' +
                                  std::to_string(ZZ_iterator_[i][2]));
      }
}
```

Going through each constraint Bob mentions above I _think_ this satisfies everything

> The current model obeys the following invariants:

> 1. The size of every parameter, transformed parameter, and generated quantity depends only on data variables and is hence is known as soon as data is loaded and never changes across iterations. The never changing across iterations is critical to make sure we get rectangular output from sampling and thus all of our downstream analysis like ESS and Rhat will continue to work.

If we assume the data sparsity pattern will be fixed then this seems satisfied. The weird part of this conversation is that it's all of our other types only care about size in terms of dimensions but for the static sparse matrices we want to allow in transformed parameters we care about size in terms of filled capacity.

> 2. Member variables represent data and are constant once they're loaded. As a consequence, methods cannot be stateful. Breaking this contract would wreak havoc on multithreading. We desperately need to move to evaluating parallel threads using only a single model instance to cut down on memory pressure from trying to run 8 processes, each with its own copy of the data.
The model log density method log_prob is typically evaluated multiple times per iteration. It is assumed this can be evaluated asynchronously and statelessly.

I'm not sure if the solution above break this condition or follows the letter of the law but skirts it's intent. From (1), assuming sparse matrix sparsity patterns are fixed data, then that "data" is not loaded till this function is run once. The atomic bool will mean those class members are only assigned once so while the answer is stateful it feels very minimal / passable

> 3. The model method write_array does the output given input of the unconstrained parameters. It calculates generated quantities and is only evaluated once per iteration using double scalars rather than autodiff variables.

Yes So what we need is for the output from write_array for static sparse matrices to have the same length as the output for sparse matrices from `(un)constrained_param_names` right? Or is there another implication I'm missing here?

>> The problem is that transformed parameters are defined in the log_prob etc. calculations. Since they are deleted after we leave that function we can't pull out their sparsity structure afterwards

> We don't write from log_prob, only from write_array, and as noted above, it recreates the transformed parameters and generates the generated quantities.

Yes I think I was missing point (3)

The other alternative here is to remove the static sparse matrix `sparse_matrix[N, M] ZZ`, but that feels like kind of a bummer since it would make things hard to do for users

SteveBronder:
Another option is to effectively treat static sparse matrices as local parameters and not print them. Though idk what kind of issues that would make in user space

rybern:
> If we assume the data sparsity pattern will be fixed then this seems satisfied. The weird part of this conversation is that it's all of our other types only care about size in terms of dimensions but for the static sparse matrices we want to allow in transformed parameters we care about size in terms of filled capacity.

I've been wondering about this. Is the sparsity of a matrix in e.g. transformed parameters allowed to change, or is it guaranteed to be constant?

SteveBronder:
The data itself can change based on parameters but the sparsity pattern should be fixed over iterations

bob-carpenter:
> Is the sparsity of a matrix in e.g. transformed parameters allowed to change, or is it guaranteed to be constant?

It has to be constant.  None of our posterior analysis makes sense if the set of parameters changes from iteration to iteration.

> The data itself can change based on parameters

The data should be fixed before parameters come into the picture, so I don't see how data can change based on parameters.

> all of our other types only care about size in terms of dimensions

We also care about unconstrained vs. constrained sizes for parameters, which goes beyond dimension.

> then that "data" is not loaded till this function is run once. 

Whichever function (presumably a class method) this is, it could be run during construction in order to satisfy the intent, which is that once data is loaded, we can do things like query shape and size of parameters.

> So what we need is for the output from write_array for static sparse matrices to have the same length as the output for sparse matrices from (un)constrained_param_names

Not just the same length, but the same order as given by the `XXX_names` method.  For parameters, the unconstrained and constrained sizes are the same for sparse matrices.

> The other alternative here is to remove the static sparse matrix sparse_matrix[N, M] ZZ, but that feels like kind of a bummer since it would make things hard to do for users

Is the idea that you're declaring a transformed parameter using `sparse_matrix[N, M]`?  I think this is very problematic and would much prefer `sparse_matrix[N, M, nn, mm]` where the `nn` and `mm` are parallel arrays of indexes with `1 <= nn[n] <= N` and `1 <= mm[m] <= M`.

One thing we'd discussed before is an equivalent of C++'s `decltype`, where you could do something like `sparse_matrix[expression] x;` where the expression is an actual expression, the return value of which is used to define the dimensions and sparsity for `x`.  So that could be something like `sparse_matrix[a * b + d]` and the idea is that you work out dimensions based on dimensions of result of `a * b + d`.  I haven't worked through the details, though, so I don't know how workable it's going to be to consider evaluating `a * b + d` at data-load time, because in the case of something like transformed parameters or generated quantities, these expressions might involve parameters.  It would certainly work if the expression was a data expression (which we already know how to test).

SteveBronder:
>> Is the sparsity of a matrix in e.g. transformed parameters allowed to change, or is it guaranteed to be constant?

> It has to be constant.  None of our posterior analysis makes sense if the set of parameters changes from iteration to iteration.

>> The data itself can change based on parameters

> The data should be fixed before parameters come into the picture, so I don't see how data can change based on parameters.

Word goop on my part, meant the literal values in the parameters not `data` or `transformed data`

>> all of our other types only care about size in terms of dimensions

> We also care about unconstrained vs. constrained sizes for parameters, which goes beyond dimension.

yes but the sizes for those are deducible from the dims right? For printing stuff out for static sparse matrices knowing the overall dimensions are N, M does not tell us how many parameters to print out.

>> then that "data" is not loaded till this function is run once. 

> Whichever function (presumably a class method) this is, it could be run during construction in order to satisfy the intent, which is that once data is loaded, we can do things like query shape and size of parameters.

Oh I like that solution! That would remove the `mutable` and atomic blah blah in my example above

>> So what we need is for the output from write_array for static sparse matrices to have the same length as the output for sparse matrices from (un)constrained_param_names

> Not just the same length, but the same order as given by the `XXX_names` method.  For parameters, the unconstrained and constrained sizes are the same for sparse matrices.

Yes!

>> The other alternative here is to remove the static sparse matrix sparse_matrix[N, M] ZZ, but that feels like kind of a bummer since it would make things hard to do for users

> Is the idea that you're declaring a transformed parameter using `sparse_matrix[N, M]`?  I think this is very problematic and would much prefer `sparse_matrix[N, M, nn, mm]` where the `nn` and `mm` are parallel arrays of indexes with `1 <= nn[n] <= N` and `1 <= mm[m] <= M`.

Yes so what I'm trying to have here is two different initializers for sparse matrices.

1. a standard sparse matrix `sparse_matrix[nonzero_rows, nonzero_cols, N, M]` which can be in any block
2. a "static" sparse matrix `sparse_matrix[N, M]` which can only be declared in transformed and generated quantities blocks. This is for things like `sparse_matrix[N, M] A = B * x;

Ryan and I have (1) working this is about (2)

> One thing we'd discussed before is an equivalent of C++'s `decltype`, where you could do something like `sparse_matrix[expression] x;` where the expression is an actual expression, the return value of which is used to define the dimensions and sparsity for `x`.  So that could be something like `sparse_matrix[a * b + d]` and the idea is that you work out dimensions based on dimensions of result of `a * b + d`.  I haven't worked through the details, though, so I don't know how workable it's going to be to consider evaluating `a * b + d` at data-load time, because in the case of something like transformed parameters or generated quantities, these expressions might involve parameters.  It would certainly work if the expression was a data expression (which we already know how to test).

It feels like for sparse matrices this would end up duplicating work. We also still wouldn't know the nonzero rows and columns from that until the matrix was called once so we would end up back at the same problem. 

Though I think the idea of calling `log_prob` or `write_array` once in the constructor along with the answer above would solve the problem!

bob-carpenter:
> Ryan and I have (1) working

That's awesome.

> yes but the sizes for those are deducible from the dims right? For printing stuff out for static sparse matrices knowing the overall dimensions are N, M does not tell us how many parameters to print out.

Dims and type (e.g., simplex).  And yes, that's right the matrix dimensions M, N aren't enough to determine the output dimensioins of a sparse matrix.

> a "static" sparse matrix sparse_matrix[N, M] which can only be declared in transformed and generated quantities blocks. This is for things like `sparse_matrix[N, M] A = B * x;

How is this going to be controlled so that `A` has exactly the same non-zero entries each iteration when we can write expressions like this:

```
sparse_matrix[M, N] A = theta > 0 ? a : b;
```
where `theta` is a parameter and `a` and `b` have different sparsity patterns?  Under the decltype proposal, it'd be illegal to write

```
sparse_matirx[declsize(theta > 0 ? a : b)] A = theta > 0 ? a : b;
```

because the expression insize `declsize` involves a parameter.

SteveBronder:
>> Ryan and I have (1) working

> That's awesome.

Ty! Yeah once we get the static matrix print out fixed I can make a cmdstan branch that Aki et al can play with. Most of the functions in stan math work with sparse matrices already since @t4c1 has been making the function signatures more generic. There's a couple small things left that rely on the Eigen type traits metaprogramming stuff

>> a "static" sparse matrix sparse_matrix[N, M] which can only be declared in transformed and generated quantities blocks. This is for things like `sparse_matrix[N, M] A = B * x;

> How is this going to be controlled so that `A` has exactly the same non-zero entries each iteration when we can write expressions like this:

> ```
> sparse_matrix[M, N] A = theta > 0 ? a : b;
> ```
> where `theta` is a parameter and `a` and `b` have different sparsity patterns?  Under the decltype proposal, it'd be illegal to write

> ```
> sparse_matirx[declsize(theta > 0 ? a : b)] A = theta > 0 ? a : b;
> ```

> because the expression insize `declsize` involves a parameter.

@rybern how hard would it be to have something in the semantic checker that throws for the above? I wonder if we can look at the assigner and if it's a conditional derived from anything that's not from the data blocks we can throw (that's the condition right Bob?)

rybern:
Sounds tricky. You might need full dependence analysis to do it perfectly, which we aren't set up to do until after semantic analysis.

There is some additional structure that you could use to make more conservative restrictions on the RHS of assignments, since at least one subexpression of the RHS will always be a sparse matrix, and most operations don't change sparsity. Maybe a set of restrictions like:
* No sparse matrix-typed subexpressions can be ternary operators
* No sparse matrix-typed subexpressions can be user-defined function calls or problematic built-in function calls

Those might be sufficient for assignments at the point of declaration, but if you can assign sparse matrices inside of e.g. if statements, it does seem like you'd need the full dependence analysis. There's actually a mostly working check for parameter dependent control flow in pedantic mode already, but again it's not trivial to use it for semantic checking.

Maybe the above rules and also banning sparse matrix assignment within control flow?

bob-carpenter:
> @rybern how hard would it be to have something in the semantic checker that throws for the above? 

The requirmeent is that the non-zero entries of a sparse matrix needs to be declared using data only.  The size and shape of a parameter is data, but the value is not.  The ternary operator was just a simple illustrative example.

SteveBronder:
> The requirmeent is that the non-zero entries of a sparse matrix needs to be declared using data only. The size and shape of a parameter is data, but the value is not. The ternary operator was just a simple illustrative example.

This will be the next thing to tackle after we can grab the nonzero rows and columns for static sparse matrices.

Just to confirm, Bob are you cool with running log_prob once in the constructor? I think that's an pretty good solution. Making sure the sparsity structure is fixed is probably something we will have to do in the semantic checker

bob-carpenter:
Thinking about this more, we'd need to figure out a value for parameters that would actually run.  In some models, even though there's theoretical support over R^N, we fail out in the tails with random inits leading to numerical issues that raise exceptions.  In that case, we wouldn't be able to get off the ground with this strategy.

SteveBronder:
In the worst case could we keep init'ing till we don't throw an exception? Or if the model fails with random inits during construction we could throw and ask the user to set inits

bob-carpenter:
> In the worst case could we keep init'ing till we don't throw an exception?

No.  That could run essentially forever.  It'd hae to time out and throw an exception.  

> Or if the model fails with random inits during construction we could throw and ask the user to set inits

That would change the interface pretty dramatically, requiring an initial unconstrained parameter vector as well as the data in the constructor.

I'm just uneasy with things that increase the complexity of the interface and start extending concerns, like how to initialize data, beyond their natural realm, like requiring data init to be able to find non-throwing parameter values.

SteveBronder:
`decltype()` would also have the same issue right?

I'm not really sure how to move forward with `sparse_matrix[N, M]` then. Should we just cut it? That's going to limit the userspace quite a lot but I'm not really sure right now how else we can get the nonzero elements. The only other answer I can think of is treating the static sparse matrices as local variables and not printing them

SteveBronder:
We could also just print out the full N,M leaving zeros where there's no value

rybern:
> Just to confirm, Bob are you cool with running log_prob once in the constructor?

I think just transformed params + generated quantities would work, and we could further whittle it down with dead code elimination, and even more with a specialized analysis that only keeps code that affects sparsity. But the parameter initialization sounds like it could be a real issue.

Here's one more proposal, which has the pro of solving both issues simply but the con of changing the sparse matrix interface:

Sparse matrix declarations inside parameters, transformed parameters and generated quantities must use either the nz_rows/nz_cols in their size, OR use an existing matrix as their size (symbol only, no expressions). Using existing matrices means that you can point at matrices defined in transformed data.
For example this these declarations from the example program:
```
transformed parameters {
  sparse_matrix[N, M] ZZ = VV * nz_vals;
}
generated quantities {
  sparse_matrix[N, M] BB = ZZ * param_cool;
}
```
become:
```
transformed parameters {
  sparse_matrix[VV] ZZ = VV * nz_vals;
}
generated quantities {
  sparse_matrix[ZZ] BB = ZZ * param_cool;
}
```
This makes it so you always know the sparsity after transformed data is run, therefore the sparsity can't depend on parameters, but you also still have sparsity structure that's an arbitrary function of the data since you can build matrices unrestricted inside transformed data. 
You'd just have to add some runtime errors for breaking the declared sparsity, just like you have out of bounds errors.
You'd have three declaration types:
* N, M, nz_rows, nz_cols
Allowed in any block
* N, M
Allowed in transformed data, model
* S : sparse matrix
Allowed in any block (except data?)

bob-carpenter:
If `decltype`were resricted to pure data expressions, it wouldn't need to be evaluated ahead of time because it would be guaranteed to always produce the same output.  The problem is when we have dependencies on parameters, because those can change iteration to iteration.  It would fail at runtime if the expression threw an exception or something, but that's OK.

We also have the same problem for inits that depend on RNGs.  But I don't think we allow those in size declarations.  At least we shouldn't!  Ack, we do.  At least it crashes during compilation.  Creating new issue:  https://github.com/stan-dev/stanc3/issues/526

P.S.  I'm not suggesting we literally include the identifier `decltype` in our own syntax---it's just a placeholder until we come up with better syntax if we do use this idea.

SteveBronder:
Ryan's is sort of like decltype right? Where the sparse matrix sparsity expression is just defined in `transformed data` and given an explicit name instead of within the `decltype()`. Ryan's idea seems easier to implement. It also gets around the init issue (or at least pushes it on the user very explicitly). I think it would also allow us to store the permutation matrix for things like Cholesky decomposition

bob-carpenter:
> Ryan's is sort of like decltype right?

Yes, @rybern's proposal is identical to my original proposal as far as I can tell.  Use a pure data-only expression to define the sparsity pattern for a sparse matrix.  It should be usable in the data block, too, as long as the argument is data.

The case @rybern called `(N, M)` can be used in general as a local variable.  That'd make them legal in transformed data, transformed parameters, model block, and generated quantities block.

There's a final case for function arguments which doesn't even require the `(N, M)` bit.

rybern:
Yes, sorry @bob-carpenter, I hadn't fully grokked the typedecl discussion when I wrote that out. Definitely the same key idea.

I do think that a substantive difference is allowing expressions vs. restricting to symbols, though. Restricting to symbols would naturally enforce the data restriction without any extra work. I think full expressions would require static analysis and/or introduce a source of runtime errors, with the benefit of more flexibility.

> The case @rybern called (N, M) can be used in general as a local variable. That'd make them legal in transformed data, transformed parameters, model block, and generated quantities block.

Ah, true, I was just thinking of globals.

bob-carpenter:
> Restricting to symbols would naturally enforce the data restriction without any extra work. I think full expressions would require static analysis and/or introduce a source of runtime errors, with the benefit of more flexibility.

It'd be OK to restrict to data variables, but it shouldn't be any harder to generalize to arbitrary expressions.  We already have data-only-expression checks because we need them for the ODE solver, etc.

bob-carpenter:
I should've included details.  First, it's MPI that really needs data only.  ODE solvers just need to know things are constants, so there's a slight difference in transformed data and generated quantities.  

An expression is data only if it only uses variables from the data or transformed data blocks.

All you need to make sure an expression is data only is to make sure it doesn't contain any transformed parameter, model, local

rybern:
> An expression is data only if it only uses variables from the data or transformed data blocks.

This sounds good to me, I'm on the same page now. I was imagining trying to prove that local variables weren't touched by parameters, but restricting to data variables does the trick.

I wonder if it would break anything to always move the sparsity expression into a new transformed data variable to avoid recomputation.

wds15:
Counting variables from - for example - for loops would not qualify as data?

rybern:
@wds15 At least for this purpose, we wouldn't include local variables like loop indices because e.g. the loop bounds might be some function of a parameter. But, since the declaration will be at the top level rather than in a loop, I don't think we could run into loop indices.

wds15:
Interesting. Thanks for the explanation.

bob-carpenter:
> Counting variables from - for example - for loops would not qualify as data?

As @rybern says, the only way a loop variable can be in scope for a declaration is with a local variable.  We should be able to use loop variables in size declarations for local variables.

The subtle distinction here is between:

* *data* :  anything that's not an autodiff variable;  this is the condition for `integrate_ode` data arguments as it's about what needs derivatives
* *constant data*: only depends on transformed data or data variables;  this is the condition for `map_rect` data arguments so that they don't change

rybern:
Yeah, that's the distinction I was tripping up on as well.

@bob-carpenter @SteveBronder Sounds like we're agreed on the constant data-only expression design? I'll help out with the implementation, just want to get confirmation before I start.

bob-carpenter:
> Sounds like we're agreed on the constant data-only expression design?

That would be my preference.  Also, I think we must have the tests for expressions in the code somewhere because we need them for `map_rect`.  (That, or `map_rect` is broken.)

rybern:
Okay, sounds good.

@SteveBronder - we were talking about starting off with a minimal working version that only supports the nz_rows/nz_cols declarations. I guess it makes the most sense for me to hold off on adding expressions until you're satisfied with that as a foundation?

SteveBronder:
Yes finishing that up today!

Issue #458, opened 2020-02-04
missing and inefficient array inits


bob-carpenter:
*This issue is transcribed from [a Discourse post by Niko Huure](https://discourse.mc-stan.org/t/ragged-arrays-bug-in-the-language/12566/8).*

### Description

stanc3 is
1. failing to initialize arrays, and
2. inefficiently initializing vectors in a loop

### Example

```stan
transformed parameters {
  real x[1000];
  vector[100] y;
  x = rep_array(10,1000);
  y = rep_vector(-1,100);
}
```

Stanc2 generates (without comments or statement numbers)
```cpp
validate_non_negative_index("x", "1000", 1000);
std::vector<local_scalar_t__> x(1000, local_scalar_t__(0));
stan::math::initialize(x, DUMMY_VAR__);
stan::math::fill(x, DUMMY_VAR__);
validate_non_negative_index("y", "100", 100);
Eigen::Matrix<local_scalar_t__, Eigen::Dynamic, 1> y(100);
stan::math::initialize(y, DUMMY_VAR__);
stan::math::fill(y, DUMMY_VAR__);
stan::math::assign(x, rep_array(10, 1000));
stan::math::assign(y, rep_vector(-(1), 100));
```

`initialize` and `fill` do the same thing and are redundant in the old code.  There's also an implementation of `assign` in both `stan::math` and `stan::model`, but the one in `model` has a bug that validate sizes before assignment.  As of now @bgoodri is exploiting that bug to create ragged arrays, so we probably don't want to fix it until we have proper ragged arrays.

Stanc3 forgets to initialize arrays.
```cpp
validate_non_negative_index("x", "1000", 1000);
std::vector<local_scalar_t__> x;
x = std::vector<local_scalar_t__>(1000, 0);
validate_non_negative_index("y", "100", 100);
Eigen::Matrix<local_scalar_t__, -1, 1> y;
y = Eigen::Matrix<local_scalar_t__, -1, 1>(100);
for (size_t sym1__ = 1; sym1__ <= 100; ++sym1__) {
  assign(y, cons_list(index_uni(sym1__), nil_index_list()),
    std::numeric_limits<double>::quiet_NaN(), "assigning variable y");
}
assign(x, nil_index_list(), rep_array(10, 1000), "assigning variable x");
assign(y, nil_index_list(), rep_vector(-1, 100), "assigning variable y");
```
[...]

PR #493, opened 2020-04-04
Array initialization fix

nhuurre:
Issue #458 

Copy array initialization behavior from Stanc2:
1. Integer arrays initialized to smallest representable integer instead of zero.
2. `var` arrays initial values copy a prebuilt `DUMMY_VAR__` without creating new `vari`s


test/integration/good/compiler-optimizations/cpp.expected
```diff
@@ -1041,13 +1051,8 @@ class optimizations_model : public model_base_crtp<optimizations_model> {
...
         
-        {
-          {
-            sym112__ = (theta * 34);
-          }
-        }
-        current_statement__ = 85;
-        lp_accum__.add(sym112__);
+        current_statement__ = 83;
+        lp_accum__.add(DUMMY_VAR__);
```
nhuurre, line 1055: This looks like an optimizer bug. Unoptimized code is
```cpp
        local_scalar_t__ paramvar;
        
        current_statement__ = 131;
        paramvar = DUMMY_VAR__;
        current_statement__ = 131;
        paramvar = 3;
        current_statement__ = 133;
        if (logical_gt(42, 1)) {
          current_statement__ = 132;
          paramvar = (theta * 34);
        } 
        current_statement__ = 134;
        lp_accum__.add(paramvar);
```
And if there's an extra `dead_code_elimination` right before the first `copy_propagation` then the optimizer generates
```cpp
        double paramvar;
        
        {
          {
            sym112__ = (theta * 34);
          }
        }
        current_statement__ = 85;
        lp_accum__.add(sym112__);
```

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -497,9 +499,9 @@ and validate_dims_stmt stmt =
 
 let make_fill vident st loc =
   let rhs =
-    Expr.(
-      Helpers.internal_funapp FnNaN []
-      @@ Typed.Meta.create ~type_:UReal ~loc ~adlevel:DataOnly ())
+    Expr.
+      { Fixed.pattern= Var "DUMMY_VAR__"
```
seantalts, line 503: I think the optimizer might break when using a Var that hasn't been defined anywhere - I think those might need special consideration. How do you feel about trying to add the DUMMY_VAR__ declaration to the MIR before we pretty print it? Then we can remove all of the DUMMY_VAR__ string printing in Stan_math_code_gen and maybe have the optimizer work correctly without modifying it. May need to turn on dead code elimination by default but that would be a nice little win too, to avoid initializing the DUMMY_VAR__ even when it's never used.

An alternative is to modify the optimizer to add another special Var that it shouldn't worry about and see if that fixes it. Perhaps it has a bug around this behavior even after that's fixed.

nhuurre: Great! Assigning `FnNaN()` to `DUMMY_VAR__` at the start of the block does fix it! (Declaration didn't matter)

However, when the optimizer sees `DUMMY_VAR__` fully it can and will undo all the work here. The point was after all to create a single NaN-valued `vari` and reuse it everywhere but the optimizer will happily deduce that `DUMMY_VAR__` is `DataOnly`.

I'm starting to think it's better to just not have any initialization in the MIR. Arrays without Eigen matrices are initialized by the backend without any hint in the MIR so why not Eigen types too.

bob-carpenter: Just a couple notes in case this wasn't obvious.

If `stan::math::var` is declared but not defined, the nullary constructor sets its member `vari*` pointer to be undefined.  So if you try to extract a value or adjoint it'll cause problems.  But if you call `var(double x)` instead, then you get memory allocation for the `vari` and its value is set to `x`.  Therefore if initializing to a value, like 0 or even NaN, it's better to create a single `var` (equivalently `vari*`) and copy it everywhere, because there's no extra allocation on the autodiff stack.

Sorry if that was obvious.  With `std::vector`, it's efficient to initialize a size `n` standard vector to `x` using `std::vector<T> y(n, T(x));`

src/stan_math_backend/Statement_gen.ml
```diff
@@ -6,21 +6,32 @@ open Expression_gen
 let pp_call_str ppf (name, args) = pp_call ppf (name, string, args)
 let pp_block ppf (pp_body, body) = pf ppf "{@;<1 2>@[<v>%a@]@,}" pp_body body
 
+let rec contains_eigen = function
+  | UnsizedType.UArray t -> contains_eigen t
+  | UMatrix | URowVector | UVector -> true
+  | _ -> false
```
SteveBronder, line 12: I would say I'm, "good at this" at least enough to do a review but I think it might be nicer here to just do the 

```suggestion
  | UInt |UReal | UMathLibraryFunction | UFun _ -> false
```

To be explicit about the types you want to check against here

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -495,48 +497,6 @@ and validate_dims_stmt stmt =
...
-    when (not (Set.mem no_fill_required decl_id))
-         && type_needs_fill decl_id (SizedType.to_unsized st) ->
-      (* I *think* we only need to initialize eigen types and scalars because we already construct
-       std::vectors with 0s.
-    *)
-      Stmt.Fixed.{pattern= SList [decl; make_fill decl_id st meta]; meta}
-  | {pattern; meta} ->
-      Stmt.Fixed.
-        {pattern= Pattern.map Fn.id (add_fill no_fill_required) pattern; meta}
-
```
SteveBronder, line 539: :+1: :+1: :+1: :+1:

SteveBronder:
Is this waiting on another review or the feature freeze?

rok-cesnovar:
I dont know the details, but isnt this a bugfix? I think its waiting for a review.

nhuurre:
Sorry, I kind of forgot this.
Right now it generates
```cpp
X = Eigen::Matrix<local_scalar_type__, -1, -1>(n, k);
stan::math::fill(X, DUMMY_VAR__);
```
but I was going to change it to
```cpp
X = Eigen::Matrix<local_scalar_type__, -1, -1>::Constant(n, k, DUMMY_VAR__);
```
which I think avoids one null initialization.

SteveBronder:
Looking at a godbolt it seems like the fill option is actually better since the `=` forces a dense assignment 

https://godbolt.org/z/B8Vp5s

SteveBronder:
Check out the assembly code for `main:` in both. The one with

```cpp
X = Eigen::Matrix<local_scalar_type__, -1, -1>::Constant(n, k, DUMMY_VAR__);
```

Has to call `void Eigen::internal::call_dense_assignment_loop`

nhuurre:
What if it's inside an array? like
```cpp
X = std::vector<M>(m, Eigen::Matrix<local_scalar_type__, -1, -1>(n, k));
stan::math::fill(X, DUMMY_VAR__);
```

bob-carpenter:
> Check out the assembly code for main: in both. The one with ... Has to call void Eigen::internal::call_dense_assignment_loop

Whatever gets called, the big hit's going to come if we wind up generating a lot of fresh `vari` in memory.  Each new `vari` has to call memory allocation, which calls constructors, advances pointers, etc.  So as long as we can avoid that, go wild on the other optimizations.

SteveBronder:
This looks a lil crazy but here's the godbolt with a graph of callees and assembly. The left graph is for C++ source 2 (calling Constant) in the top left and the graph on the right is for C++ source 1 (using fill). Just by how complicated the call stack is for the constant version I think the fill version is better here

https://godbolt.org/z/hg56hz

bob-carpenter:
As long as it doesn't duplicate entries on the autodiff stack!

nhuurre:
There's no new `vari`s created, it just makes copies of `DUMMY_VAR__` (a pointer to `vari`).

Reading assembly code is above my pay grade but note that (unlike the Godbolt code) the generated declaration is
```cpp
std::vector<Eigen::Matrix<double, -1, -1>> X;
```
So I think it's a zero-size vector before the assignment which may or may not make a difference.
I'm also not sure if `double` results are representative of `stan::math::var`. You can allocate a double array without initializing it but my understanding is that even "uninitialized" `var`s have their nullary constructor called.

Anyway, I'm going to defer to C++ experts on this so if you don't think `Matrix::Constant` helps I'll just resolve the merge conflict and this is good.

bob-carpenter:
For the record,

```
struct foo {
  foo() { std::cout << "called ctor." << std::endl; }
};

int main() {
  std::cout << "Eigen::Matrix ctor" << std::endl;
  Eigen::Matrix<foo, -1, -1> a(2, 3);

  std::cout << "Eigen::Matrix Constant" << std::endl;
  Eigen::Matrix<foo, -1, -1> b
      = Eigen::Matrix<foo, -1, -1>::Constant(2, 3, foo());

  std::cout << "std::vector, plain" << std::endl;
  std::vector<foo> c(3);

  std::cout << "std::vector, initializer" << std::endl;
  std::vector<foo> d(3, foo());
}
```

produces

```
Eigen::Matrix ctor
called ctor.
called ctor.
called ctor.
called ctor.
called ctor.
called ctor.
Eigen::Matrix Constant
called ctor.
called ctor.
called ctor.
called ctor.
called ctor.
called ctor.
called ctor.
std::vector, plain
called ctor.
called ctor.
called ctor.
std::vector, initializer
called ctor.
```

Only the initializer version for `std::vector` avoids calling the nullary constructor multiple times.  The `Constant` form calls it 7 times to fill a 3 x 2 matrix, so presumably it does a basic init then a fill.  

But as I just realized in the last post (again---I keep forgetting), the `var()` constructo doesn't put things on the stack---it just leaves a dangling pointer.  Now if you get to something like `var(NaN)`, then that's best done with a fill of `var(NaN)` rather than copying in `NaN` or there will be multiple autodiff copies.

bob-carpenter:
> I'm also not sure if double results are representative of stan::math::var. You can allocate a double array without initializing it but my understanding is that even "uninitialized" vars have their nullary constructor called.

Exactly---that's the critical part.  It depends on the level.  I believe if it's just a `var`, it'll get initialized with `var()`, which just leaves a nullptr as the value of `vi_`.  So I'm probalby raising a non-issue here.

rok-cesnovar:
Is this still waiting for a resolution or a review?

nhuurre:
Ready for review.

PR #501, opened 2020-04-09
Add docs for the Stan math C++ codegen printers

SteveBronder:
## Summary

I was reading through the pretty printers for the stan math c++ codegen and figured I'd add docs as I read through it. I'm rather novice so I'd appreciate someone looking through my docs to make sure they make sense

You can make the docs with `dune build @doc-private`

I'd still like to try to find a method to make a full site that includes both the private and public libraries


rok-cesnovar:
Hell yeah! I will have a look. Been playing trial and error for some time now with the printers! Thanks

SteveBronder:
Thanks! Sorry I got caught up in the static matrix stuff

PR #505, opened 2020-04-10
Little C++ optimizations

SteveBronder:
I looked over the C++ and added a few optimizations and cleanups. Notably

## Style 

1. Instead of using `typedef` this uses `using`. This is a style change, I personally find typedefs to be weird to weird since it's like the only thing in C++ where the declaration is `keyword type name` instead of `keyword name = type`. 

3. `vector__.resize(0)` is replaced with `vector__.clear()`. `resize(0)` is a pretty platform/compiler dependent performance enhancement and I think it's weird to read at first if you don't know you don't just call `.clear()`

## Performance

1. methods defined for the model class now have the `inline` attribute. This makes the functions better candidate for inlining.

2. The vector for `un/constrained_param_names` now uses  emplace_back instead of push_back. `emplace_back` and `push_back` differ in that, for `emplace_back`, initialization of the element is done directly in the vector. `push_back` initializes the element then places it in the vector. This is only useful when have something that's largish (more than about 3 words is what I usually ballpark). I think the strings of variable names can have decent odds of being larger than 3 words.

3. idt this function is called often, but thanks to @rybern `get_dims` now uses `emplace_back` and constructs the temporary vector inplace. i.e. previously `get_dims` would be defined as something like

```cpp
  void get_dims(std::vector<std::vector<size_t>>& dimss__) const {
    dimss__.resize(0);
    std::vector<size_t> dims__;
    dimss__.push_back(dims__);
    dims__.resize(0);
    dimss__.push_back(dims__);
    dims__.resize(0);
    dims__.push_back(J);
    dimss__.push_back(dims__);
    dims__.resize(0);
    dims__.push_back(J);
    dimss__.push_back(dims__);
    dims__.resize(0);
  }
```
Every push_back on `dims__` is adding an element to the vector, then that vector is copied into `dimss__ ` and `dims__.resize(0)` clears the vector to be reused again. So instead of doing all those creations and destructions we can just construct the vector that's given to `dims__` inplace ala

```cpp
inline void get_dims(std::vector<std::vector<size_t>>& dimss__) const final {
    dimss__.clear();
    dimss__.emplace_back(std::vector<size_t>{});
    dimss__.emplace_back(std::vector<size_t>{});
    dimss__.emplace_back(std::vector<size_t>{J});
    dimss__.emplace_back(std::vector<size_t>{J});
}
```
 
4. I don't see any tests where we use `resize_to_match__` etc. but I wrote them to pass their `dst` values back by reference instead of making a copy.


seantalts:
I like the idea behind this - no reason to keep generated code looking like stanc2 generated code anymore. I'm curious about `inline` and resize_to_match__ (which maybe we should just delete until it's actually needed?). For `inline` I wouldn't expect the compiler to actually be able to inline these - they're basically part of the ABI for the model class, right? And for resize_to_match__, the objective seems to be kinda unfortunate - to both modify some input as well as return something. I think maybe this was to have the function act in a simplistic way that we could model with the MIR. Are you sure that we need to do anything more complicated than what is there + relying on RVO? I mean probably best to just kill it since we'd want to see benchmarks for end to end stuff anyway, and we aren't using it... Anyway thanks for taking a stab at making the generated code easier to read :)

SteveBronder:
> I like the idea behind this - no reason to keep generated code looking like stanc2 generated code anymore. I'm curious about inline and resize_to_match__ (which maybe we should just delete until it's actually needed?). 

If `resize_to_match__` is not needed then I think it would be a good idea to delete them. They're also just sort of floating around in the global namespace.

> For inline I wouldn't expect the compiler to actually be able to inline these - they're basically part of the ABI for the model class, right? 

It's a weird Q and A for whether the compiler can be smart enough to inline a virtual function. Looking at [this](https://stackoverflow.com/a/733791/2269255) stackoverflow post it seems like they _can_ be

> "The only time an inline virtual call can be inlined is when the compiler knows the "exact class" of the object which is the target of the virtual function call. This can happen only when the compiler has an actual object rather than a pointer or reference to an object. I.e., either with a local object, a global/static object, or a fully contained object inside a composite."

So do we always use the model as a pointer? If so then the answer is nah idt inline would be useful. `final` could actually be nice to encourage devirtualization but I would have to add an optional `attribute` function to `pp_method` and idt it's worth the effort lol.

> And for resize_to_match__, the objective seems to be kinda unfortunate - to both modify some input as well as return something. I think maybe this was to have the function act in a simplistic way that we could model with the MIR. Are you sure that we need to do anything more complicated than what is there + relying on RVO? 

RVO actually doesn't work with input arguments! (I just learned that the other day from a cppcon talk). Essentially the only time a move/forward is useful is whenever the output is one of the input arguments. 

> I mean probably best to just kill it since we'd want to see benchmarks for end to end stuff anyway, and we aren't using it... Anyway thanks for taking a stab at making the generated code easier to read :)

Kill which thing? The changes to `resize_to_match__` or the entirety of `resize_to_match__`?

SteveBronder:
Also would it be hard for the benchmarks to go off for the stanc3 repo like the other repos?

bob-carpenter:
> RVO actually doesn't work with input arguments!

What's an input argument?  Do you have a simple example where RVO won't work?

Just one clarification:

> methods defined for the model class now have the inline attribute. This makes the functions better candidate for inlining.

Class methods are automatically declared as `inline` in C++.  Semantically, the `inline` declaration is more about making it possible to have duplicate definitions in different translation units.   That's why we have it on big functions that'll never actually get inlined.  Can you measure a performance difference between the ones declared `inline` and not?

Also, those functions are going to get called from another translation unit now that we separate compilation.  

Making things `final` should be OK.  I'm curious as to whether that makes a performance difference.  It was huge in Java.

SteveBronder:
> What's an input argument? Do you have a simple example where RVO won't work?

By an input argument I meant one that is used as both incoming and passed as the outcoming, so in/out may have been more appropriate here

Here's a godbolt with an example of how the in/out parameter forces a copy unless you use forward to pass the input value on to the output

https://godbolt.org/z/Qdy-pb

> Class methods are automatically declared as inline in C++. Semantically, the inline declaration is more about making it possible to have duplicate definitions in different translation units. That's why we have it on big functions that'll never actually get inlined. Can you measure a performance difference between the ones declared inline and not?

Ah! Forgot / did not know about that. I'll change those back to non-inline

> Making things final should be OK. I'm curious as to whether that makes a performance difference. It was huge in Java.

It might but I think relative to everything else it will be small. But I figure why leave a little optimization on the table that's just adding a keyword.

bob-carpenter:
Thanks for the reminder.  It makes sense that it'd copy without move semantics---can't have functions just stealing memory on their own!

rok-cesnovar:
@SteveBronder Mind just merging master in? If you are busy I can do that also. I will review this after the conflicts are fixed.

rok-cesnovar:
@SteveBronder do you think you will have time for the final touches here before the release? It would be great to have these in for the next release. If you won't have time, would you mind if I push to this branch?

SteveBronder:
Hey! Sorry I'll look at this today

SteveBronder:
@rok-cesnovar can you give this a lookover?

rok-cesnovar:
Will do!

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -711,37 +730,25 @@ using namespace stan::math; |}
...
-  dst.resize(src.size());
-  return dst;
-}
-std::vector<double> to_doubles__(std::initializer_list<double> x) {
+inline std::vector<double> to_doubles__(std::initializer_list<double> x) {
   return x;
 }
 
-std::vector<stan::math::var> to_vars__(std::initializer_list<stan::math::var> x) {
+inline std::vector<stan::math::var> to_vars__(std::initializer_list<stan::math::var> x) {
```
rok-cesnovar, line 751: Would you mind just removing `to_vars__`, `to_doubles__` and all `resize_to_match__`. Sorry for asking you to remove this after doing all this work on them. 
I triple checked and could not find these used in statement gen, stan math code-gen anywhere.

validate_positive_index, validate_unit_vector_index are used but should be in math/stan. But that is a separate issue/PR.

SteveBronder: Yes good call! I didn't add them originally so if we don't need them I'm fine with getting rid of them

PR #507, opened 2020-04-11
Error on model name that starts with a number or non-underscore symbol

### LINKED ISSUE #506: Error if filename or  model name start with a number or symbol
Stanc3 should stop if the filename or supplied model name starts with a number or symbol other than _.

Currently it passes through but then things break on the C++ level.

stanc2 gives this error:
```
make STANC2=true examples/eight_schools/8schools

--- Translating Stan model to C++ code ---
bin/stanc2  --warn-pedantic --o=examples/eight_schools/8schools.hpp examples/eight_schools/8schools.stan

model_name must not start with a number or symbol other than _
Could not remove output file=examples/eight_schools/8schools.hpp
make: *** No rule to make target 'examples/eight_schools/8schools.hpp', needed by 'examples/eight_schools/8schools'.  Stop.
```

rok-cesnovar:
Fixes #506 by outputing the following error message:

```
Model name must not start with a number or symbol other than underscore.
```

Entire cmdstan is then:
```
make examples/eight_schools/8schools

--- Translating Stan model to C++ code ---
bin/stanc  --o=examples/eight_schools/8schools.hpp examples/eight_schools/8schools.stan
Model name must not start with a number or symbol other than underscore.
make: *** No rule to make target 'examples/eight_schools/8schools.hpp', needed by 'examples/eight_schools/8schools'.  Stop.
```

### DIFF
--- a/src/stanc/stanc.ml
+++ b/src/stanc/stanc.ml
@@ -171,7 +171,7 @@ let model_file_err () =
   exit 127
 
 let model_file_start_char_err () =
-  eprintf "%s" "Model name must not start with a number or symbol other than underscore.";
+  eprintf "%s" "Model name must not start with a number or symbol other than underscore.\n";
   exit 127
 
 let add_file filename =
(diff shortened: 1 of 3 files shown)


rybern:
Looks good

andrewgelman:
Just a quick, comment.  The error should say "File name . . .", not "Model name . . ."

Until we go to blockless, Stan has a "model" block, so we should avoid confusion by not calling a file a model.  Also, in R it is particularly confusing because we give the complied Stan program a name, and it would be natural to call that the "model name".  I think that "file name" is unambiguous and correct here.

Andrew

rok-cesnovar:
The thing is its actually the model name that is problem here. The default model name is the filename without the file extension. But you can override it using STANCFLAGS in makefiles in cmdstan (cmdstanr will also expose model_name argument in the near future to simplify this). But as no one usually does that its more or less always the filename.

It would actually be pretty easy to get rid of this restriction altogether. So I will try that.

PR #508, opened 2020-04-11
Bugfix/issue 504 spurious integer warning

### LINKED ISSUE #504: spurious integer division warnings
We should only have warnings about integer division when dividing two integers.  As is, if I do something like

```
  real c = 1 / 2.0;
```

I get spurious warning messages:

```
Info: Found int division at '/Users/carp/temp2/foo.stan', line 4, column 11 to column 12:
  1 / 2.0
Positive values rounded down, negative values rounded up or down in platform-dependent way.
```

Also, the rounding in C++ is always done by dropping off the fractional part, so the rounding is toward zero.  (The platform-dependent thing was probably also a bug in the original compiler's message.)

rok-cesnovar:
Fixes #504 
Removed int/real from int warning.

The new error message will be:
```
Info: Found int division at 'operators_int.stan', line 10, column 25 to column 30:
  d_int / d_int
Values will be rounded towards zero.
```

### DIFF
--- a/src/frontend/Semantic_check.ml
+++ b/src/frontend/Semantic_check.ml
@@ -599,7 +599,7 @@ and semantic_check_expression cf ({emeta; expr} : Ast.untyped_expression) :
               "@[<hov>Info: Found int division at %s:@   @[<hov 2>%a\n@]%s@.@]"
               (Location_span.to_string x.emeta.loc)
               Pretty_printing.pp_expression {expr; emeta}
-              "Values will be rounded to zero." ;
+              "Values will be rounded towards zero." ;
             (x, y)
         | _ -> (x, y)
       in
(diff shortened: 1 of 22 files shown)


rok-cesnovar:
@bob-carpenter is "Values will be rounded to zero." fine or do we want any more info?

src/frontend/Semantic_check.ml
```diff
@@ -594,13 +594,12 @@ and semantic_check_expression cf ({emeta; expr} : Ast.untyped_expression) :
...
-        | UInt, UReal, Divide | UInt, UInt, Divide ->
+        | UInt, UInt, Divide ->
             Fmt.pr
-              "@[<hov>Info: Found int division at %s:@   @[<hov 2>%a@]@,%s@,@]"
+              "@[<hov>Info: Found int division at %s:@   @[<hov 2>%a\n@]%s@.@]"
               (Location_span.to_string x.emeta.loc)
               Pretty_printing.pp_expression {expr; emeta}
-              "Positive values rounded down, negative values rounded up or \
-               down in platform-dependent way." ;
+              "Values will be rounded to zero." ;
```
rybern, line 602: Rounded to 0 makes me think the result will be 0, maybe "rounded toward zero"?

rok-cesnovar: Oh lol, yes :)

rok-cesnovar: Fixed.

PR #509, opened 2020-04-12
Bugfix/issue 506 move the _model suffix to model_ prefix

rok-cesnovar:
This PR reverts https://github.com/stan-dev/stanc3/pull/507 and fixes the same issue with a different solution. The way it was setup in #507 would break rstan backwards compatibilty (rstan supports model names starting with numbers).

For all uses of the models name in naming namespaces and classes in the generated C++ it uses a `model_` prefix. This removes issues if the models name (derived from the filename or as an argument) starts with a number of symbol.

The ˙model_name()` C++ function returns the name without the prefix.


test/integration/model-name/stanc.expected
```diff
@@ -1,16 +1,4 @@
...
-     1:  data {
-     2:    // declare matching model name should fail
-     3:    int foo;  
-               ^
-     4:  }
-     5:  parameters {
-   -------------------------------------------------
-
-Identifier 'foo' clashes with model name.
-
```
nhuurre, line 14: Where did this error go? It's still supposed to fail, right?

rok-cesnovar: Yeah, this is weird. Will check.

rok-cesnovar:
> Why do C++ namespaces and classes need to be named after the model name, anyway

Not sure actually. @bob-carpenter or @seantalts definitely know more than me on that.

Rstan, cmdstanr and cmdstanpy and I also think pystan compile the model in the temporary directories anyways so its only a thing in cmdstan. Well rstan will do it in javascript now so even less of a problem.


I changed it back to name_model, added your mangling suggestion and if the first character in the name is not an alpha character then add "_" to the `prog_name`. What do you think?

nhuurre:
Sounds good. Either prefix or suffix is fine but prefix has the advantage that you don't need to special-case the first character.

src/stanc/stanc.ml
```diff
@@ -203,7 +198,12 @@ let use_file filename =
     if !print_model_cpp then print_endline cpp )
 
 let remove_dotstan s = String.drop_suffix s 5
-let model_name_check_regex = Str.regexp "^[a-zA-Z_].*$"
+
+let mangle =
+  String.concat_map ~f:(fun c ->
+      Char.(
+        if is_alphanum c || c = '_' then to_string c
+        else "x" ^ Int.to_string (to_int c)) )
```
nhuurre, line 206: IIRC previously Stan converted `'-'` to `"_"`. It probably doesn't matter but we could keep it consistent.

rok-cesnovar: Explicitly changing - to _. Good call.

Issue #511, opened 2020-04-13
make lower/upper and multiplier/offset constraints order independent


bob-carpenter:
As is, `<lower = a, upper = b>` is an OK constraint, but `<upper = b, lower = a>` is not.  

Also, `<offset = a, multiplier = b>` is OK, but not the opposite order.

Both of these should be order independent.

PR #512, opened 2020-04-13
Added reverse-ordered offset/multiplier and lower/upper to the parser

rybern:
Also added complete error messages.

Addresses #511

### DIFF
--- a/test/integration/bad/new/stanc.expected
+++ b/test/integration/bad/new/stanc.expected
@@ -2042,13 +2042,13 @@ Expected "offset = " expression (not containing binary logical operators).
 
   $ ../../../../../install/default/bin/stanc location-scale-bad9.stan
 
-Syntax error in 'location-scale-bad9.stan', line 1, column 27 to column 28, parsing error:
+Syntax error in 'location-scale-bad9.stan', line 1, column 28 to column 29, parsing error:
    -------------------------------------------------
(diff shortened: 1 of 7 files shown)


test/integration/bad/new/stanc.expected
```diff
@@ -2042,13 +2042,13 @@ Expected "offset = " expression (not containing binary logical operators).
...
-Syntax error in 'location-scale-bad9.stan', line 1, column 27 to column 28, parsing error:
+Syntax error in 'location-scale-bad9.stan', line 1, column 28 to column 29, parsing error:
    -------------------------------------------------
      1:  data { vector<multiplier = y,> x;}
-                                     ^
+                                      ^
    -------------------------------------------------
 
-Expected ">" after "multiplier =" expression.
+Expected '>' or lower expression after multiplier expression.
```
nhuurre, line 2051: I think this should be `'>' or offset expression`.

rybern: Fixed. Nice catch!

bob-carpenter:
Thanks.  The code looks so elegant.

I didn't see a test for the different orders of lower and upper that would match the one for different orders of offset and multiplier.

rybern:
@bob-carpenter good point, added a small test

PR #521, opened 2020-04-17
Check for out-of-bounds array access

### LINKED ISSUE #489: Eigen out of bounds access error with simple model
Here is the model:
```
transformed data {
  vector[1] y = to_vector({ 1.0 });
}

parameters {
  real mu;
}

model {
  y[0] ~ normal(mu, 1.0);
}
```

If I build and run it:
```
./model sample
```

I get an Eigen out of bounds error and the program stops:
```
model: stan/lib/stan_math/lib/eigen_3.3.3/Eigen/src/Core/DenseCoeffsBase.h:162: Eigen::DenseCoeffsBase<type-parameter-0-0, 0>::CoeffReturnType Eigen::DenseCoeffsBase<Eigen::Matrix<double, -1, 1, 0, -1, 1>, 0>::operator[](Eigen::Index) const [Derived = Eigen::Matrix<double, -1, 1, 0, -1, 1>, Level = 0]: Assertion `index >= 0 && index < size()' failed.
Aborted (core dumped)
```

There should be some sort of Stan out of bounds access error I'm pretty sure. This is from cmdstan in develop. I assume this is a compiler thing, but maybe it is math somewhere. Can move the issue if that is the case.

nhuurre:
Fixes #489 by forcing the use of `stan::model::rvalue()` instead of direct access. As far as I know that doesn't incur any performance penalty beyond the range check.

The range check is still skipped for compiler generated indexing as those are likely to be correct.

### DIFF
--- a/src/stan_math_backend/Expression_gen.ml
+++ b/src/stan_math_backend/Expression_gen.ml
@@ -56,6 +56,11 @@ let minus_one e =
 
 let is_single_index = function Index.Single _ -> true | _ -> false
 
+let dont_need_range_check = function
+  | Index.Single Expr.Fixed.({pattern= Var id; _}) ->
+      not (Utils.is_user_ident id)
+  | _ -> false
+
 let promote_adtype =
   List.fold
     ~f:(fun accum expr ->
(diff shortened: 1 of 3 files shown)


rok-cesnovar:
@SteveBronder you have been working with rvalue at the stan level. Want to comment on performance? The code change is simple and looks good to me.

rok-cesnovar:
@bob-carpenter @SteveBronder @t4c1 Just want to double-check with you guys on this one. Does forcing the use of `stan::model::rvalue()` instead of direct access incur any performance penalty?

t4c1:
I think it should not.

rok-cesnovar:
@nhuurre please resolve the conflict when you have the time. Then its good to go.

Issue #438, opened 2020-01-21
Request For Help: get optimization working!


seantalts:
@VMatthijs and @rybern implemented a bunch of textbook dataflow analyses and optimization passes, but when hooked up they don't generate correct code. A really awesome project would be to work on getting more and more optimizations in to an `-O1` optimization level mode. 

There may also be opportunities for turning on certain optimizations even in unoptimized mode; the current example is the autodiff level optimization that correctly figures out that some new local variables in e.g. the model block may not need to be `var` but can instead just be `double`, if those variables only depend on data and transformed data.

Thinking about the broader architecture, once we have an optimization level flag in the compiler, we could write a CmdStan harness that builds both the optimized and unoptimized versions of the model under different names and links them together, calling the optimized one each `log_prob` by default but if any exceptions or signals are caught, falling back to the unoptimized `log_prob`.

rybern:
I think I missed an older issue - do we have a collection of cases where optimizations don't generate correct code?

seantalts:
There's no list anywhere, which I guess is the first problem - would be good to get tests set up. But try compiling the mother model with `--O`, for me it hangs.

Issue #193, opened 2019-06-22
Start integrating and testing optimization end to end


seantalts:
We'd like to get to a place where we can have optimizations turned on by default, generate both an optimized and unoptimized `log_prob` and by default run the optimized and on exception or sigsegv fallback to the unoptimized version on an eval-by-eval basis.

VMatthijs:
Or weren't we talking about initially running one evaluation of both and comparing them and only using the optimized code if it agrees and is faster?

seantalts:
That's interesting, I forget. I think figuring out if it agrees and is faster could be tricky in general but is probably fine to just run it once and see. That's a cool idea.

seantalts:
Hm, one problem with the idea of choosing at runtime whether the optimized version is the same is that that adds state to the model class. Though maybe we could just choose some random initial values and do the comparison during construction of the model class?

PR #522, opened 2020-04-20
Optimization debugging

rybern:
I'm working on fixing optimizations. I'm using (modified) scripts and corpus from https://github.com/stan-dev/performance-tests-cmdstan to check for correctness and speed. This extends the work I did in #328, which focused only on getting the results to compile, and had a much smaller and easier corpus.

I'm trying to leave reasonably clear git commit messages (you can expand them in github to see more than the title).

While this is WIP, there are some temporary pieces:
 * Debug code (mostly printing)
 * Partial evaluation of arithmetic expressions into `fma` is disabled, since this introduces small numerical differences and makes it harder to test
 * The numerical CLI argument is currently just a testing tool, but it could be reworked into a permanent interface


rybern:
@VMatthijs Per our discussion yesterday -

I separated out non-duplicatable and non-removable from side-effecting expressions. The only difference between them is currently rng - my intuition is that it's a bad idea to ever optimize any of the internal functions ([diff here](https://github.com/stan-dev/stanc3/pull/522/files#diff-b49246fd9939f572c3f408f101d79740R612)).

I also ran the tests against all the models in stat_comp_benchmarks, and they all passed. (I did have to disable the partial optimization that inserts `mfa` because it gives small numerical differences, I did that for the other corpus too).

SteveBronder:
Might be good to run these against all of the models in the `example_models` folder

https://github.com/stan-dev/performance-tests-cmdstan

SteveBronder:
Also benchmarks would be cool!

rybern:
@SteveBronder Thanks Steve - I should have posted the status here.

I've tested against all stan files in example_models* and stat_comp_benchmarks, and they all compile and produce correct code now.

I haven't done much speed testing but it seems like there's a small average improvement with a very large variance. Now that optimizations are running, we can do some work tuning them, e.g. playing with repeated application and ordering. I think we should merge these fixes and leave speedups for a future PR because optimizations are currently broken in master.

*: Not actually all stan files, only the stan files which passed the test when optimizations were turned off. That's 359/507 files.

SteveBronder:
Cool cool! For clarification, 

> *: Not actually all stan files, only the stan files which passed the test when optimizations were turned off. That's 359/507 files.

Do these models pass for the stanc2 compiler?

rybern:
I'll check

rybern:
Here are performance changes on the benchmark models for one run of 5000 samples. I wouldn't trust these much since they're just one run and I'm not 100% sure what these scripts are using for timing.

```
Name, Old Result, New Result, Ratio, Performance change( 1 - new / old )
stat_comp_benchmarks/benchmarks/gp_regr/gen_gp_data.stan, 0.03, 0.03, 1.0, -0.44% slower
stat_comp_benchmarks/benchmarks/gp_regr/gp_regr.stan, 0.76, 0.64, 1.18, 15.42% faster
stat_comp_benchmarks/benchmarks/arma/arma.stan, 1.61, 1.21, 1.33, 24.84% faster
stat_comp_benchmarks/benchmarks/eight_schools/eight_schools.stan, 0.29, 0.3, 0.97, -3.03% slower
stat_comp_benchmarks/benchmarks/pkpd/one_comp_mm_elim_abs.stan, 61.71, 62.13, 0.99, -0.68% slower
stat_comp_benchmarks/benchmarks/pkpd/sim_one_comp_mm_elim_abs.stan, 0.28, 0.28, 1.01, 1.0% faster
stat_comp_benchmarks/benchmarks/irt_2pl/irt_2pl.stan, 33.87, 34.09, 0.99, -0.65% slower
stat_comp_benchmarks/benchmarks/low_dim_corr_gauss/low_dim_corr_gauss.stan, 0.01, 0.01, 1.0, 0.07% faster
stat_comp_benchmarks/benchmarks/low_dim_gauss_mix/low_dim_gauss_mix.stan, 16.25, 16.36, 0.99, -0.63% slower
stat_comp_benchmarks/benchmarks/sir/sir.stan, 256.93, 241.52, 1.06, 6.0% faster
stat_comp_benchmarks/benchmarks/low_dim_gauss_mix_collapse/low_dim_gauss_mix_collapse.stan, 50.32, 50.14, 1.0, 0.35% faster
stat_comp_benchmarks/benchmarks/arK/arK.stan, 9.37, 7.95, 1.18, 15.12% faster
stat_comp_benchmarks/benchmarks/garch/garch.stan, 2.95, 2.64, 1.12, 10.55% faster
stat_comp_benchmarks/benchmarks/gp_pois_regr/gp_pois_regr.stan, 15.02, 11.77, 1.28, 21.59% faster
performance.compilation, 141.27, 142.59, 0.99, -0.94% slower

```

rybern:
@SteveBronder @rok-cesnovar @wds15 (Sorry to ping everyone!)
I think it's reasonable to merge this given how broken optimizations currently are in master

rok-cesnovar:
I can take a look tomorrow.

I am not familiar with this stuff, so might have some questions. But given that this is an optional feature hidden behind a flag we could also just merge it. And I can ask my questions later :)

bob-carpenter:
> I'm working on fixing optimizations. 

This is too vague for an issue description.  What's the end goal?

> introduces small numerical differences and makes it harder to test

Tests for floating point should (almost) never be exact.  Any reordering of operations is going to cause small floating point differences.  

> I think it's reasonable to merge this given how broken optimizations currently are in master

I don't follow.  Something being broken means it's *not* reasonable to merge it.  Or are you only talking about merging a debugging framework?  This is why it helps to have clearer issue descriptions---I can't tell if this PR is to fix optimization or to just introduce some kind of debugging framework.

>  But given that this is an optional feature hidden behind a flag we could also just merge it.

Optional features need to be treated with the same care in documentation and testing as non-optional features.  It's much much harder to change things after it's been merged into the `develop`.

Now if you want to code something as "experimental" and paste warnings all over it every time it runs and warn developers very clearly away from depending on it in the code, that's a different story.

rok-cesnovar:
This is fixing the behaviour of compiling with the `optimize` flag which is already in master (stanc3 doesn't use develop) and has some issues.

rybern:
@bob-carpenter Optimizations are currently in master in a broken state, meaning they often produce non-compiling or incorrect code. This PR is to debug them, using a testing framework that's not included in the PR (since it relies on CmdStan).

> Tests for floating point should (almost) never be exact. Any reordering of operations is going to cause small floating point differences.

I agree. The tests I ran used a tolerance of 1e-8. That passed for all optimizations/test cases unless `fma` was enabled, I was just saying that it broke the tolerances.

@rok-cesnovar Thanks! Happy to answer questions.

rybern:
> If replacing a regular multiply-add with fma causes tolerance failures below 1e-8, I'd look for a bug somewhere---the tolerance shouldn't be that bad.

The tolerance was on the difference between parameter estimates after running Stan with ~1000 samples, rather than the result of a single computation, and even then it usually only barely broke tolerances.
I also checked the generated code to make sure that `fma` was the only difference between failing/passing tests, and that they appeared to be applied correctly.
So, it's possible there's an issue with `fma`, but I think my results are consistent with a small numerical difference.

> I think it's a very bad idea to have broken code merged into our repos.

Sean's philosophy early on was to err towards keeping things in master, especially if they're behind flags. This has been in master for quite a while. I think optimizations 'work' as a result of this PR (to the extent that I can test them with the cases example_models and stat_comp_benchmarks).

rybern:
@SteveBronder re: the non-running example model issue, it looks from the [issues](https://github.com/stan-dev/example-models/issues) in that repo like they are actually broken example

bob-carpenter:
> The tolerance was on the difference between parameter estimates after running Stan with ~1000 samples

Aha.  That won't work.  If you're running a sampler and the sampler works, what you need to do is check that the two estimates are within their stated MCMC standard errors of each other.    

Did you fix the seed?

> Sean's philosophy early on was to err towards keeping things in master, especially if they're behind flags.

What was the motivation for that philosophy?  I was pushing for getting the core parser working and tested before working on optimization, but that wasn't compatible with what people wanted to work on.  It's a delicate balance in open source when the project can't make decisions and assign work to people.  And now we've taken a step backward in that there's not even a decisionmaking framework in place now that there's no more TWG.

bob-carpenter:
Could someone please create an issue that clearly explains what this PR is meant to solve.

rybern:
> Aha. That won't work. If you're running a sampler and the sampler works, what you need to do is check that the two estimates are within their stated MCMC standard errors of each other.

> Did you fix the seed?

I did fix the seed. If I've fixed the seed, shouldn't almost identical results from two runs be a good indication that they are effectively the same?

> What was the motivation for that philosophy?

I think the idea was to prevent large pieces of code from diverging, thus wasting work. I don't think we're following that philosophy policy any more, I think we're being pretty restrictive about what we add in now.

> Could someone please create an issue that clearly explains what this PR is meant to solve.

Here are some old ones: #438, #193, and all of #274 through #284

I could run my test setup against master to show how often it fails, and put that in a new issue, would that be helpful?

bob-carpenter:
> If I've fixed the seed, shouldn't almost identical results from two runs be a good indication that they are effectively the same?

Yes.  But the reverse isn't true.  Non-identical results can be effectively the same.  And that's what we really care about.  I'm not, by the way, pushing for making this PR more complex!  I'd start simple with things that are easy to test.

> Here are some old ones: #438, #193, and all of #274 through #284

Is this PR going to address all of those?

rybern:
> Yes. But the reverse isn't true. Non-identical results can be effectively the same.

Right, I think it could give some false failures but not false passes, which I think has worked out (with the exception of `fma`).

> Is this PR going to address all of those?

Yeah, I think it does address all of those. The only borderline piece is whether there are enough unit tests for individual optimizations as suggested in #274 - #284 (although I did add a number of pathological cases as I encountered them). I would probably close #438 and #193 and open a new issue to tune the optimization order/replication for speed.

rok-cesnovar:
Sorry I dropped the ball here. I took a dive, but it would basically require you to explain everything from the start so if @VMatthijs would have time in the not so distant future to look at this, that would be more efficient. Else I can start bugging with the basic questions :)

rybern:
No worries @rok-cesnovar. It's not really anyone's responsibility and we don't really have a solution for reviewing specialized sections. If you do want to dive in, I'd be happy to answer your questions as code comment discussions or maybe a video call.

rybern:
I wrote issue #543 to clarify why this PR is needed.

Running that same test suite on this branch, the results out of 359 are:
Failed to compile: 0
Gave wrong answer: 0
Gave right answer: 359

I've also added a debugging flag `preserve_stability` to make this test reproducible by turning off all of the features that might produce small numerical differences when the random seed is the same (like `fma` discussed above). Maybe that flag could be folded into a user-facing interface, like optimization levels 1-3, but I'll leave it for another PR.

rok-cesnovar:
Ok, I took a deep dive in optimizations and re-read mostly all the optimization PRs/issues. I will not pretend I understand all the details of every optimization implementation. I do however have a basic understanding of how we got here and what is going on.

I also went in and checked the testing corpus and verified that this PR is a strict improvement of what is currently on master. 

I do have to note these optimizations badly need more docs explaining for each optimization case what they do. But that is an issue with the state of optimizations as a whole, not this PR. The docs will make it easier for someone to pick this up without trying to backtrack what went on. It will also help with debugging, support, etc.

I would suggest the following plan for these optimizations:
- merge this PR to master as it is a huge improvement
- add optimization levels (@rybern has already noted that is the plan, see https://github.com/stan-dev/stanc3/issues/544#issuecomment-633285347) and make all current optimizations whichever optimization level is "undocumented/experimental". So if we have levels from 0 to 3, make that 4.
- add optimizations to their appropriate level in separate PRs. For example a PR that will move dead code elimination to optimization level 1. Preferably each optimization would be added in its own PR or maybe group the easier ones together. That PR will only be merged once the optimization is properly documented in stan-dev/docs if that is where we want them or anywhere else. Preferably with an easy to understand example. If want users to use this we need docs on what is being done.

Does this sound good?

@rybern Please merge this with recent master.

rybern:
@rok-cesnovar Thanks for doing the deep dive!

Sounds like a reasonable plan. Certainly the optimization levels will need docs, and I can write simple docs for each optimization as well.

Merged master.

rok-cesnovar:
@bob-carpenter would you be fine with the proposed plan above? It's not ideal but neither is the current state.

bob-carpenter:
Thanks @rok-cesnovar!  Sounds good to me.  I like the idea of adding things in a controlled way rather than trying to debug everything at once.

PR #524, opened 2020-04-22
Feature/standalone functions

### LINKED ISSUE #500: stanc3 lacks compile_functions.hpp
The stanc3 compiler is lacking the ability to call the `compile_functions` from

https://github.com/stan-dev/stan/blob/develop/src/stan/lang/compile_functions.hpp
https://github.com/stan-dev/stan/blob/develop/src/stan/lang/generator/generate_standalone_functions.hpp

which is used by `rstan::expose_stan_functions`. There are some weird things about how this works currently, but once `stanc3` gets to the point where it can do all the things `stanc2` does, I can drop the C++ parser.

rok-cesnovar:
Fixes #500 

In order to get this standalone functions C++ use the `--standalone-functions` argument.

Changes compared to what stanc2 produced:
- return type is auto (can easily be deduces from the function in the namespace)
- the header included is the regular stan/model/model_header.hpp we include when compiling models, not stan/model/standalone_functions_header.hpp
- the namespace used is the same as if it were an actual model (`basic_model_namespace` vs `basic_functions`). Is `_functions` suffix required? See example 1.
- calls to user defined _lpdf, _lpmg and  _lp  are only supplied the `<false>` template parameter for propto, the rest is not required. See example 2.

Example 1:

stanc2:
```cpp
double
my_log1p_exp(const double& x, std::ostream* pstream__ = nullptr){
  return 
basic_functions::my_log1p_exp<double>(x, pstream__);
}
```

stanc3:
```cpp
auto my_log1p_exp(const double& x, std::ostream* pstream__ = nullptr)  
{
return basic_model_namespace::my_log1p_exp(x, pstream__);
}
```
As you can see the scalars are still passed by reference. This was the easiest to do after all, but can be changed if required.

Example 2:

stanc2:
```cpp
// [[stan::function]]
void
test_lp(const double& a, double& lp__, stan::math::accumulator<double>& lp_accum__, std::ostream* pstream__ = nullptr){
  
special_functions_functions::test_lp<double, double, stan::math::accumulator<double> >(a, lp__, lp_accum__, pstream__);
}

// [[stan::function]]
double
test_rng(const double& a, boost::ecuyer1988& base_rng__, std::ostream* pstream__ = nullptr){
  return 
special_functions_functions::test_rng<double, boost::ecuyer1988>(a, base_rng__, pstream__);
}

// [[stan::function]]
double
test_lpdf(const double& a,
              const double& b, std::ostream* pstream__ = nullptr){
  return 
special_functions_functions::test_lpdf<false, double, double>(a, b, pstream__);
}

```

stanc3 (ignore the different namespace - its in a different test model now):
```cpp
auto test_lp(const double& a, double& lp__,
             stan::math::accumulator<double>& lp_accum__,
             std::ostream* pstream__ = nullptr)  
{
return basic_model_namespace::test_lp<false>(a, lp__, lp_accum__, pstream__);
}

auto test_rng(const double& a, boost::ecuyer1988& base_rng__,
              std::ostream* pstream__ = nullptr)  
{
return basic_model_namespace::test_rng(a, base_rng__, pstream__);
}

auto test_lpdf(const double& a, const double& b,
               std::ostream* pstream__ = nullptr)  
{
return basic_model_namespace::test_lpdf<false>(a, b, pstream__);
}
```


rok-cesnovar:
Here is the JS for stanc for this branch: https://jenkins.mc-stan.org/job/stanc3-test-binaries/69/artifact/bin/stanc.js

rok-cesnovar:
Currently I test with the following functions (most taken from stan-dev/stan):

```cpp
real my_log1p_exp(real x) {
    return log1p_exp(x);
}

real array_fun(real[] a)
{
  return sum(a);
}

real int_array_fun(int[] a)
{
  return sum(a);
}

vector my_vector_mul_by_5(vector x)
{
  vector[num_elements(x)] result = x * 5.0;
  return result;
}

int int_only_multiplication(int a, int b) {
  return a*b;
}

real test_lgamma(real x) {
      return lgamma(x);
  }

// test special functions
void test_lp(real a) {
    a ~ normal(0, 1);
}

real test_rng(real a) {
  return normal_rng(a, 1);
}

real test_lpdf(real a, real b) {
  return normal_lpdf(a | b, 1);
}
vector integrand(vector x) {
  return exp(-square(x));
}

real[] integrand_ode(real r, real[] f, real[] theta, real[] x_r, int[] x_i) {
  real df_dx[1];
  real x = logit(r);
  df_dx[1] = exp(-square(x)) * 1/(r * (1-r));
  return(df_dx);
}

real ode_integrate() {
  int x_i[0];
  // ok:
  //return(integrate_ode_rk45(integrand_ode, rep_array(0.0, 1),
  //1E-5, rep_array(1.0-1E-5, 1), rep_array(0.0, 0), rep_array(0.0,
  //0), x_i)[1,1]);
  // not ok
  return(integrate_ode_bdf(integrand_ode, rep_array(0.0, 1), 1E-5, rep_array(1.0-1E-5, 1), rep_array(0.0, 0), rep_array(0.0, 0), x_i)[1,1]);
}
```

@bgoodri  do you have any more in the rstan repo? Or know of any that caused issues?

bgoodri:
To answer the questions about the stanc2 vs. stanc3 behavior

(1) I think `auto` returns are fine
(2) I can't remember why `standalone_functions_header.hpp` was created to be different than `model_header.hpp` but it is entirely possible that we don't need the distinction any more
(3) I think that is fine too, and the model's namespace is the md5sum of the `.stan` in rstan in order to avoid name clashes
(4) `propto = FALSE` is better than `propto = TRUE` and probably it is best to just have one

The most extensive test of the standalone functions is actually in the rstanarm unit tests. I'll have to get a branch of rstan to utilize this before I can run that, but it looks good so far.

bob-carpenter:
It might be that the standalone functions don't need autodiff.  Though we could be exporting differentiable functions if we could figure out an interface.

wds15:
But the exported functions may need nested ad (stiff ode solver). It’s not a hard requirement, but certainly good to have.

rok-cesnovar:
Does the current rstan functionality support that? I think everything was always double as far as I looked. 

But i agree that exposing AD stuff would be nice to have.

bob-carpenter:
Good point about the ODE solver.  So we don't have any truly double-only steps.  I'll have to stop saying that about generated quantities!

bgoodri:
The current version can only take double / int arguments but it may use
autodiff internally. I have some code to pass a var back and forth between
R and C++ but it is not merged yet.

bgoodri:
Where it used to say 
```
// [[stan::function]]
```
can it now say
```
// [[Rcpp::export]]
```
or another comment? I have to add that in order for it to be exposed, and it is easier if there is already a comment to mark the locations.

rok-cesnovar:
Done. Updated stanc.js below (just remove the .txt added due to Github).

[stancjs.js.txt](https://github.com/stan-dev/stanc3/files/4531316/stancjs.js.txt)

rok-cesnovar:
The `stanc()` JS function now supports stancflags with its third argument. The following are supported
"--allow_undefined" (and "--allow-undefined")
"--version"
"--standalone-functions"

Examples:

```js
var undef_test = stanc.stanc("testing", model, "--allow-undefined");
var stanfuncs_test = stanc.stanc("stanfuncs", functions, "--standalone-functions");
```

You can also specify multiple flags, separated with spaces.

Will add the following if the above work fine for you:
"--warn-uninitialized" and "--warn-pedantic" (need to figure out how to put warnings in the "warnings" node of the returned object, not stderr)
"--auto-format" and "--print-canonical" (these return stan model back, not cpp)

I am not sure on what to do with "--include_paths". How does this work with JS without acces to the filesystem?
Did you already hash this out? I was not following the stancjs discussion before. Are you preprocessing includes?

The following are not going to be ported to stancjs:
"--name" (not needed, first argument of stanc() call)
"--o" (cant output to file)
"--print-cpp" (stanc() call could print directly, but that seems weird)

Also no reason to include debug/development flags:
"--dump-stan-math-signatures" (maybe this one?)
"--debug-lex"
"--debug-parse"
"--debug-ast"
"--debug-decorated-ast"
"--debug-generate-data"
"--debug-mir"
"--debug-mir-pretty"
"--debug-optimized-mir"
"--debug-optimized-mir-pretty"
"--debug-transformed-mir"
"--debug-transformed-mir-pretty"

Here is the built .js file to test:

[stancjs.bc.zip](https://github.com/stan-dev/stanc3/files/4707468/stancjs.bc.zip)

And a test file I used to test this with `node`. Just run `node test.js`.
[test.zip](https://github.com/stan-dev/stanc3/files/4707469/test.zip)

nhuurre:
IMO the third argument as options object is more natural than space-separated string.
```js
var undef_test = stanc.stanc("testing", model, {"allow-undefined": true});
var stanfuncs_test = stanc.stanc("stanfuncs", functions, {"standalone-functions": true});
```

rok-cesnovar:
Yes, for stancjs that is definitely more native.

I think my reasoning was too much in the sense of cmdstan that just passes the STANCFLAGS makefile flag to the stanc3. In that case, using strings is simpler because it requires no preprocessing. In rstan that will be the first and probably main consumer of this, allow_undefined and such are arguments, so yeah, no reason to not make this more JS-native.

rok-cesnovar:
@bgoodri I am guessing you haven't had the time to test this with the rstan changes and all. I should have time in the second part of August to finish/polish this, so if you can at least give me some pointers on how I can test it, that would be great. Not urgent, I will probably get to this after the 15th or so.

standalone functions are also the last thing keeping us from being able to run all tests in stan-dev/stan with stanc3 and that would then unblock other CI improvements.

rok-cesnovar:
With this PR now, the standalone functions tests (`./runTests.py test/integration/compile_standalone_functions_test`) in the Stan repo pass with stanc3 as well, so I think this is ready for review and merge. 

This will finally enable the removal of stanc2 from the stan repo and cmdstan that has been on hold due to this.

seantalts:
Does this need a reviewer? I could take a look this weekend.

rok-cesnovar:
Yes, no one has volunteered so far. That would be great!

test/integration/good/code-gen/standalone_functions/cpp.expected
```diff
@@ -0,0 +1,578 @@
...
+stan::promote_args_t<T0__,
+T1__>
+operator()(const T0__& a, const T1__& b, std::ostream* pstream__)  const 
+{
+return test_lpdf<propto__>(a, b, pstream__);
+}
+};
+}
+
+// [[Rcpp::export]]
```
seantalts, line 321: Is the need for this just to allow Rcpp to not need to use our `basic_model_namespace`? Maybe this is old and @bgoodri  knows? 

If so maybe we can just elide the namespace part for standalone functions and add the Rcpp::export comments to the original definitions...?

seantalts (changes requested):
Thanks for doing this! Wondering about if the standalone file could be simpler now that the compiler is a little easier to hack on.

test/integration/good/code-gen/standalone_functions/cpp.expected
rok-cesnovar: Do you mean in the sense why don't we just put the stuff in line 52 in `test/integration/good/code-gen/standalone_functions/cpp.expected` outside the namespace?

seantalts: Yeah, and add the `Rcpp::export` comment.

rok-cesnovar: That was added due to this comment by Ben: https://github.com/stan-dev/stanc3/pull/524#issuecomment-619211360

rok-cesnovar: I was aiming at being close to the old version, but yeah. If Rcpp can handle those templates and all that would make it simpler.

seantalts:
Ah yeah. @bgoodri is there a way we can test Rcpp or RStan with this
output?

test/integration/good/code-gen/standalone_functions/cpp.expected
bgoodri: The exported functions need to be in the global namespace and not templated, which is why we have been exporting untemplated functions that call the corresponding versions within the namespace.

bgoodri:
> Ah yeah. @bgoodri is there a way we can test Rcpp or RStan with this output?

To test it for real, we would have to call `rstan::expose_stan_functions` with a branch that uses the javascript stanc3 rather than the C++ one. In the meantime, you can probably test that things at least compile by passing the filename of the generated C++ to `Rcpp::sourceCpp` after adding
```
// [[Rcpp::depends(StanHeaders)]]
// [[Rcpp::depends(rstan)]]
// [[Rcpp::depends(RcppEigen)]]
// [[Rcpp::depends(BH)]]
#include <stan/math/prim/mat/fun/Eigen.hpp>
#include <boost/integer/integer_log2.hpp>
#include <exporter.h>
#include <RcppEigen.h>
```

test/integration/good/code-gen/standalone_functions/cpp.expected
rok-cesnovar: Thanks. In that case we need these non-templated functions and there is no need to remove the namespace from the templated ones.

seantalts:
Thanks! I will try this. I though `auto` was equivalent to templating in some ways, but I forget the details and might be wrong.

@bgoodri do you know if it's that we just need to explicitly instantiate the templated functions?

bgoodri:
If explicit instantiation is sufficient, it isn't documented as being sufficient by Rcpp. I do know (although it isn't documented) that if all the template parameters have defaults, that calling `foo<>(...)` is sufficient.

seantalts:
Rcpp::sourceCpp doesn't seem to like `nullptr` or `enable_if_t` or a bunch of other C++1y stuff. Do I need to set something in Makevars?

rok-cesnovar:
Add 

CXX14 = g++ -std=c++1y

to makevars

bgoodri:
But it still doesn't like nullptr because it doesn't have a native R
representation. I have to change nullptrs to 0.

rok-cesnovar:
Also in the Math backend or just the top R-facing level?

bgoodri:
In general, any default arguments that the R function derives from the C++
function must be valid R syntax. The C++ wrapper functions have to have
signatures that Rcpp can handle but the function bodies can do pretty much
anything.

rok-cesnovar:
@seantalts do we take care of this at the stanc.js level by default or do we add a --rcpp flag that will enable cleaning any of these rcpp-specific stuff?

@bgoodri is there anything else you do at the rstan level that we could do for you in the compiler?

bgoodri:
Originally the plan was for PyStan to have equivalent functionality so the generated C++ code didn't have anything that was specific to R. But I think PyStan has given up on this idea. Or you could just put things inside an `#indef USING_R`. The `expose_stan_functions_hacks` function at

https://github.com/stan-dev/rstan/blob/develop/rstan/rstan/R/expose_stan_functions.R

adds the aforementioned

```
// [[Rcpp::depends(StanHeaders)]]
// [[Rcpp::depends(rstan)]]
// [[Rcpp::depends(RcppEigen)]]
// [[Rcpp::depends(BH)]]
#include <stan/math/prim/mat/fun/Eigen.hpp>
#include <boost/integer/integer_log2.hpp>
#include <exporter.h>
#include <RcppEigen.h>
```

rewrites the functions that pass in an `accumulator`, and changes all the `nullptr`s to 0. It also adds any headers that define undefined functions, but I guess that has to be done from R.

seantalts:
Thanks! I still can't get it to work from RStudio... have either of you managed to test it? 

The furthest I've gotten involves adding `// [[Rcpp::plugins(cpp14)]]` to the top of the file (and I have the CXX14 set as well). But I'm getting SEXP errors (what is this, OCaml?):
```
> Rcpp::sourceCpp("~/github/stanc3/test/integration/good/code-gen/standalone_functions/basic.cpp")
basic.cpp: In function ‘SEXPREC* sourceCpp_1_test_lp(SEXP, SEXP, SEXP, SEXP)’:
basic.cpp:485:73: error: invalid use of void expression
  485 |     rcpp_result_gen = Rcpp::wrap(test_lp(a, lp__, lp_accum__, pstream__));
      |                                                                         ^
In file included from /home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/Rcpp/include/Rcpp/as.h:25,
                 from /home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/Rcpp/include/RcppCommon.h:159,
                 from /home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/rstan/include/exporter.h:4,
                 from basic.cpp:9:
/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/Rcpp/include/Rcpp/internal/Exporter.h: In instantiation of ‘Rcpp::traits::Exporter<T>::Exporter(SEXP) [with T = stan::math::accumulator<double>; SEXP = SEXPREC*]’:
/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/Rcpp/include/Rcpp/as.h:87:41:   required from ‘T Rcpp::internal::as(SEXP, Rcpp::traits::r_type_generic_tag) [with T = stan::math::accumulator<double>; SEXP = SEXPREC*]’
/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/Rcpp/include/Rcpp/as.h:152:31:   required from ‘T Rcpp::as(SEXP) [with T = stan::math::accumulator<double>; SEXP = SEXPREC*]’
/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/Rcpp/include/Rcpp/InputParameter.h:46:54:   required from ‘Rcpp::ReferenceInputParameter<T>::ReferenceInputParameter(SEXP) [with T = stan::math::accumulator<double>; SEXP = SEXPREC*]’
basic.cpp:483:102:   required from here
/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/Rcpp/include/Rcpp/internal/Exporter.h:31:31: error: no matching function for call to ‘stan::math::accumulator<double>::accumulator(SEXPREC*&)’
   31 |       Exporter( SEXP x ) : t(x){}
      |                               ^
In file included from /home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/StanHeaders/include/stan/math/prim/mat.hpp:53,
                 from /home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/StanHeaders/include/stan/math/rev/mat.hpp:12,
                 from /home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/StanHeaders/include/stan/math.hpp:4,
                 from /home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/StanHeaders/include/src/stan/model/model_header.hpp:4,
                 from basic.cpp:13:
/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/StanHeaders/include/stan/math/prim/mat/fun/accumulator.hpp:32:3: note: candidate: ‘stan::math::accumulator<T>::accumulator() [with T = double]’
   32 |   accumulator() : buf_() {}
      |   ^~~~~~~~~~~
/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/StanHeaders/include/stan/math/prim/mat/fun/accumulator.hpp:32:3: note:   candidate expects 0 arguments, 1 provided
/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/StanHeaders/include/stan/math/prim/mat/fun/accumulator.hpp:24:7: note: candidate: ‘stan::math::accumulator<double>::accumulator(const stan::math::accumulator<double>&)’
   24 | class accumulator {
      |       ^~~~~~~~~~~
/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/StanHeaders/include/stan/math/prim/mat/fun/accumulator.hpp:24:7: note:   no known conversion for argument 1 from ‘SEXP’ {aka ‘SEXPREC*’} to ‘const stan::math::accumulator<double>&’
make: *** [/usr/lib/R/etc/Makeconf:177: basic.o] Error 1
g++ -std=c++1y -w -std=gnu++14 -I"/usr/share/R/include" -DNDEBUG   -I"/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/Rcpp/include/"  -I"/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/RcppEigen/include/"  -I"/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/RcppEigen/include/unsupported"  -I"/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/BH/include" -I"/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/StanHeaders/include/src/"  -I"/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/StanHeaders/include/"  -I"/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/RcppParallel/include/"  -I"/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/rstan/include" -DEIGEN_NO_DEBUG  -DBOOST_DISABLE_ASSERTS  -DBOOST_PENDING_INTEGER_LOG2_HPP  -DSTAN_THREADS  -DBOOST_NO_AUTO_PTR  -include '/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/StanHeaders/include/stan/math/prim/mat/fun/Eigen.hpp'  -D_REENTRANT -DRCPP_PARALLEL_USE_TBB=1   -I"/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/Rcpp/include" -I"/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/StanHeaders/include" -I"/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/rstan/include" -I"/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/RcppEigen/include" -I"/home/xitrium/R/x86_64-pc-linux-gnu-library/3.6/BH/include" -I"/home/xitrium/github/stanc3/test/integration/good/code-gen/standalone_functions"   -fpic  -g -O2 -fdebug-prefix-map=/build/r-base-jbaK_j/r-base-3.6.3=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g  -c basic.cpp -o basic.o
Error in Rcpp::sourceCpp("~/github/stanc3/test/integration/good/code-gen/standalone_functions/basic.cpp") : 
  Error 1 occurred building shared library.
```

rok-cesnovar:
I am guessing this is related to what @bgoodri mentioned

> rewrites the functions that pass in an accumulator, and changes all the nullptrs to 0. It also adds any headers that define undefined functions, but I guess that has to be done from R.

@bgoodri can you link to where you rewrite this or provide a bit more detail on what exactly you have to do to w.r.t to accumulators. Thanks!

seantalts:
Ah, is that it? I think his functionality for doing that is in this: 
https://github.com/stan-dev/rstan/blob/develop/rstan/rstan/R/expose_stan_functions.R I just need to decipher the R and try again, haha.

rok-cesnovar:
Try replacing

`stan::math::accumulator<double>& lp_accum__, std::ostream* pstream__ = nullptr){
`
with 

`std::ostream* pstream__ = nullptr){\nstan::math::accumulator<double> lp_accum__;`

This is something we should be doing in stanc3 I guess.

seantalts:
Is that... the same?

seantalts:
I tried that and I still get 
```cpp
basic.cpp: In function ‘SEXPREC* sourceCpp_1_test_lp(SEXP, SEXP, SEXP)’:
basic.cpp:484:61: error: invalid use of void expression
  484 |     rcpp_result_gen = Rcpp::wrap(test_lp(a, lp__, pstream__));
      |                                                             ^
make: *** [/usr/lib/R/etc/Makeconf:177: basic.o] Error 1
```

Because `test_lp` returns void.

In general I think it's a good idea to try to get this in basically as-is for the next release if possible, but I also just want to make sure it works. Though I guess bgoodri could always adjust the output further, hah.

rok-cesnovar:
Thanks!

So it is mostly working apart from functions that have a void return type? Will check what current rstan does with void.

seantalts:
> So it is mostly working apart from functions that have a void return type? Will check what current rstan does with void.

Believe so? That's the only error message I get now when I try to use `Rcpp::sourceCpp`.

rok-cesnovar:
A few changes. 

Added a `--rcpp` flag that also does all the stuff required to compile with Rcpp. 

Now both test files with functions compile. As do the rstanarm functions. The integrate one does not link (needs cvodes) but that is not a stanc3 thing, would have to supply cvodes, but not the point of this.

Had to add

```
CXX14 = g++ -std=c++1y
CXX14FLAGS = -I"/home/rok/R/x86_64-pc-linux-gnu-library/4.0/StanHeaders/include/src" -Wno-ignored-attributes -Wno-deprecated-declarations
```
and run 
```
Sys.setenv(USE_CXX14 = 1)
```
@bgoodri is this expected?

The StanHeaders/include/src one was needed because it didnt find the model header hpp file otherwise. Is requiring USE_CXX14 ok?

Code for testing:

```R
path_to_stanc3 <- "/home/rok/Desktop/stanc3"
stanc3_bin <- file.path(path_to_stanc3, "_build", "default", "src", "stanc", "stanc.exe")

test_models <- c(
  file.path(path_to_stanc3, "test", "integration", "good", "code-gen", "standalone_functions", "basic.stan")
  #file.path(path_to_stanc3, "test", "integration", "good", "code-gen", "standalone_functions", "integrate.stan")
)

for (model in test_models) {
  cpp_file <- tempfile(fileext = ".cpp")
  processx::run(stanc3_bin, args = c(
    model,
    paste0("--o=", cpp_file),
    "--standalone-functions",
    "--rcpp"
  ),
  echo_cmd = TRUE)
  Rcpp::sourceCpp(file = cpp_file)
}
```

bgoodri:
The `expose_stan_functions()` function does a `Sys.setenv(USE_CXX14 = 1)`
internally so that is fine for testing.

rok-cesnovar:
Cool, thanks!

In that case, this is almost ready then. Rcpp stuff compiles fine now, there is only one more thing, that Ben will know off the top of his head probably.

Example:
```stan
functions {
real my_log1p_exp(real x) {
  return log1p_exp(x);
}  
}
```
If I run it with rstan::expose_stan_functions() I can do `my_log1p_exp(1)`. If I run it for my test with the stanc3 generated code
I have to do
```
o <- rstan::get_stream()
my_log1p_exp(1, o)
```

otherwise I get `Error in my_log1p_exp(1) : Expecting an external pointer: [type=integer].`

rok-cesnovar:
stanc.js to test:
[stancjs.bc.zip](https://github.com/stan-dev/stanc3/files/5323356/stancjs.bc.zip)

Example to test:

```js
let example = 
`
functions {
    real my_log1p_exp(real x) {
        return log1p_exp(x);
    }
}
`
let example_result = stanc.stanc("example", example, ["standalone-functions", "rcpp"]);
```

bgoodri:
Yeah, all functions have the pstream__ argument, even if they do not print
anything.

rok-cesnovar:
Ah okay, that is something you set in `expose_stan_functions` also (https://github.com/stan-dev/rstan/blob/5a97a4674779d743b42c22a97b694697273180c8/rstan/rstan/R/expose_stan_functions.R#L118). 

In that case, this is ready for a final review.

seantalts:
I'm reading over it now and making some changes. @bgoodri would you mind if we kept it backend-agnostic and you had to keep your existing `expose_stan_functions.R` script? This part of the code is very hair and it's getting exponentially hairier the more we add different special cases that appear throughout the code like this.

bgoodri:
I don't really mind as long as the functions are pre-marked with comment.
Finding all the starts of all the functions to insert the //
[[Rcpp::export]] is otherwise difficult in general.

seantalts (approved):

seantalts:
@rok-cesnovar thank you very much for doing this! I made some changes to try to simplify things as much as we can while still being reasonably good Stan ecosystem citizens, haha. I think I want to go through the code gen at some point and go ahead and create a very simple Lower Intermediate Representation that hopefully really cleans up that code... Thanks for dealing with it all :)

seantalts:
(also please use a little extra scrutiny looking over my changes; I have to run and have been slightly less careful in trying to get this out).

rok-cesnovar:
Thanks @seantalts. This looked good, only had to cleanup JS tests and had one minor cleanup. Other than that this looks good to me. I still struggle a bit with the formatters so I tend to write code that is not as pretty. Thanks for making that much simpler!

seantalts:
Woops, thank you!!

bgoodri:
@rok-cesnovar How do I call `stanc` now with the extra argument?
```
ctx <- V8::v8()
ctx$source("https://github.com/stan-dev/stanc3/releases/download/nightly/stanc.js")
stopifnot(ctx$validate("stanc"))
ctx$call("stanc", "", "", "--version")
```
fails for me.

rok-cesnovar:
`let version = stanc.stanc("version-test", "", ["version"]);`

So the last one is an array of flags (and its without -- since this isn't really command line flags).

See https://github.com/stan-dev/stanc3/blob/master/test/stancjs/version.js
and other tests files https://github.com/stan-dev/stanc3/tree/master/test/stancjs

bgoodri:
Do you mean like

str(ctx$call("stanc", "", "", "[version]"))

? That does not error but it does not return the version number.

rok-cesnovar:
I have to familirize myself with the V8 R package but this works

```
ctx$eval("stanc('name','',['version'])")
```
but that returns an object so its

```
ctx$eval("stanc('name','',['version']).result")
```

bgoodri:
Ah. So, to get the version, I need to call ctx$eval(...) but for actual
parsing I need to call ctx$call(...) .

rok-cesnovar:
I think you can also use the call, I just never worked with the V8 R package so it was a start. It has to be doable. 
Let me get back to you in a few minutes.

rok-cesnovar:
```js
ctx <- V8::v8(typed_arrays = FALSE) // you need to set typed_arrays = FALSE so c("version") works

ctx$source("https://github.com/stan-dev/stanc3/releases/download/nightly/stanc.js")
stopifnot(ctx$validate("stanc"))

model_code <- 
'
parameters {
  real y;
}
model {
  y ~ std_normal();
}
'
ctx$assign("model_code", model_code)
ctx$assign("flags", c("version"))
ctx$assign("stanc_return", V8::JS("stanc('basic', model_code, flags)"))
ver <- ctx$get(V8::JS("stanc_return.result"))
cat(ver)

ctx$assign("model_code", model_code)
ctx$assign("flags", c("warn-pedantic")) // or empty array
ctx$assign("stanc_return", V8::JS("stanc('basic', model_code, flags)"))
model_code <- ctx$get(V8::JS("stanc_return.result"))
cat(model_code)
```

outputs the following: 

```
> cat(ver)
stanc 62a45cd9

> cat(model_code)

// Code generated by stanc 62a45cd9
#include <stan/model/model_header.hpp>
namespace basic_namespace {


...
```

Does that work for you?

rok-cesnovar:
I updated the last comment but also found the solution for `$call `if that is what you prefer that. It does look simpler.

```js10
ctx$call("stanc", "", "", as.array("version"))
model_code <- 
'
parameters {
  real y;
}
model {
  y ~ std_normal();
}
'
ctx$call("stanc", "model", model_code  ,as.array(""))

model_code <- 
'
functions {
  real foo(real a);
}
parameters {
  real y;
}
model {
  y ~ normal(foo(0), 1);
}
'
ctx$call("stanc", "model", model_code  , c("warn-pedantic", "allow-undefined"))
```
So you need the `as.array()` if its only 1 flag...

Issue #165, opened 2019-06-10
Emit warnings for all deprecated language constructs


VMatthijs:
```src/stan/lang/grammars/semantic_actions_def.cpp:  msgs << "Info: Function name '" << old_name << "' is deprecated"
src/stan/lang/grammars/semantic_actions_def.cpp:  msgs << "Info: Deprecated function '" << f.name_ << "';"
src/stan/lang/grammars/semantic_actions_def.cpp:  error_msgs << "Info: assignment operator <- deprecated"
src/stan/lang/grammars/semantic_actions_def.cpp:    error_msgs << "Info: the lkj_cov_log() sampling distribution"
src/stan/lang/grammars/semantic_actions_def.cpp:    error_msgs << "Info:" << std::endl
src/stan/lang/grammars/semantic_actions_def.cpp:  error_msgs << "Info: increment_log_prob(...);"
src/stan/lang/grammars/semantic_actions_def.cpp:  error_msgs << "Info: the integrate_ode() function is deprecated"
src/stan/lang/grammars/semantic_actions_def.cpp:    error_msgs << "Info: get_lp() function deprecated." << std::endl
src/stan/lang/grammars/semantic_actions_def.cpp:    error_msgs << "Info: Function abs(real) is deprecated"
src/stan/lang/grammars/semantic_actions_def.cpp:    error_msgs << "Info: the lkj_cov_log() function"
src/stan/lang/grammars/semantic_actions_def.cpp:    error_msgs << "Info: the if_else() function"
src/stan/lang/grammars/semantic_actions_def.cpp:    error_msgs << "Info: integer division"
src/stan/lang/grammars/semantic_actions_def.cpp:               << "Info:" << std::endl
src/stan/lang/grammars/semantic_actions_def.cpp:  error_msgs << "Info: Comments beginning with #"
src/test/unit/lang/parser/assignment_test.cpp:               "Info: assignment operator <- deprecated"
src/test/unit/lang/parser/expression_grammar_test.cpp:               "Info: Function abs(real) is deprecated in the Stan language.");
src/test/unit/lang/parser/if_else_deprecation_test.cpp:               "Info: the if_else() function is deprecated.");
src/test/unit/lang/parser/lkj_cov_deprecation_test.cpp:               "Info: the lkj_cov_log() sampling distribution is deprecated.");
src/test/unit/lang/parser/lkj_cov_deprecation_test.cpp:               "Info: the lkj_cov_log() function is deprecated.");
[...]

PR #525, opened 2020-04-24
Variadic ODE

### LINKED ISSUE #481: Add new ODE interface
Could we get a prototype for the new ODE stuff similar to what we have for reduce_sum (https://github.com/stan-dev/stanc3/pull/451)? No need for super error checking and whatnot at this point. Just wanna start playing with ODEs in the new interface.

@rok-cesnovar, @rybern , whoever feels like putting it together.

The mission here is similar. Instead of the ODE signature:

```
real[ , ] integrate_ode_bdf(function ode_rhs, real[] initial_state,
                            real initial_time, real[] times,
                            real[] theta, real[] x_r, int[] x_i,
                            data real rel_tol, data real abs_tol, data int max_num_steps)
```

with ode right hand side function:
```
real[] my_ode_func(real, real[], real[], data real[], data int[]):real[]
```

we want (these names are all subject to change I just wanted something short for testing):

```
real[ , ] ode_bdf(function ode_rhs, real[] initial_state,
                  real initial_time, real[] times,
                  data real rel_tol, data real abs_tol, data int max_num_steps,
                  T1 arg1, T2 arg2, ...)
```

where the ode right hand side function looks like:

```
real[] my_ode_func(real time, real[] state, T1 arg1, T2 arg2, ...)
```

The real_tol/abs_tol/max_num_steps args are no longer optional with the new form.

The C++ signature of the new ode integrate function is:
```
template <typename F, typename T_initial, typename T_t0, typename T_ts, typename... T_Args>
std::vector<std::vector<typename stan::return_type<T_initial, T_t0, T_ts, T_Args...>::type>>
ode_bdf(const F& f, const std::vector<T_initial>& y0, const T_t0& t0,
        const std::vector<T_ts>& ts,
        double relative_tolerance,
        double absolute_tolerance,
        long int max_num_steps,
        std::ostream* msgs,
        const T_Args&... args)
```

An example C++ functor that will work with the new integrator signature is:
```
struct CosArg1 {
  template <typename T0, typename T1, typename... T_Args>
  inline std::vector<typename stan::return_type<T1, T_Args...>::type>
  operator()(const T0& t, const std::vector<T1>& y,
             const T_Args&... a, std::ostream* msgs) const {
    return ...;
  }
};
```

Here is an old Stan model (taken from Bob's Lotka-Volterra case study -- https://mc-stan.org/users/documentation/case-studies/lotka-volterra-predator-prey.html):
```
functions {
  real[] dz_dt(real t,       // time
               real[] z,     // system state {prey, predator}
               real[] theta, // parameters
               real[] x_r,   // unused data
               int[] x_i) {
    real u = z[1];
    real v = z[2];
   
    real alpha = theta[1];
    real beta = theta[2];
    real gamma = theta[3];
    real delta = theta[4];
   
    real du_dt = (alpha - beta * v) * u;
    real dv_dt = (-gamma + delta * u) * v;
    return { du_dt, dv_dt };
  }
}
data {
  int<lower = 0> N;           // number of measurement times
  real ts[N];                 // measurement times > 0
  real y_init[2];             // initial measured populations
  real<lower = 0> y[N, 2];    // measured populations
}
parameters {
  real<lower = 0> theta[4];   // { alpha, beta, gamma, delta }
  real<lower = 0> z_init[2];  // initial population
  real<lower = 0> sigma[2];   // measurement errors
}
transformed parameters {
  real z[N, 2]
  = integrate_ode_rk45(dz_dt, z_init, 0, ts, theta,
                       rep_array(0.0, 0), rep_array(0, 0),
                       1e-5, 1e-3, 5e2);
}
model {
  theta[{1, 3}] ~ normal(1, 0.5);
  theta[{2, 4}] ~ normal(0.05, 0.05);
  sigma ~ lognormal(-1, 1);
  z_init ~ lognormal(log(10), 1);
  for (k in 1:2) {
    y_init[k] ~ lognormal(log(z_init[k]), sigma[k]);
    y[ , k] ~ lognormal(log(z[, k]), sigma[k]);
  }
}
```

Here is code that should compile with the new interface:
```
functions {
  real[] dz_dt(real t,       // time
               real[] z,     // system state {prey, predator}
               real alpha,
               real beta,
               real gamma,
               real delta) {
    real u = z[1];
    real v = z[2];
   
    real du_dt = (alpha - beta * v) * u;
    real dv_dt = (-gamma + delta * u) * v;
   
    return { du_dt, dv_dt };
  }
}
data {
  int<lower = 0> N;          // number of measurement times
  real ts[N];                // measurement times > 0
  real y_init[2];            // initial measured populations
  real<lower = 0> y[N, 2];   // measured populations
}
parameters {
  real<lower = 0> alpha;
  real<lower = 0> beta;
  real<lower = 0> gamma;
  real<lower = 0> delta;
  real<lower = 0> z_init[2];  // initial population
  real<lower = 0> sigma[2];   // measurement errors
}
transformed parameters {
  real z[N, 2]
  = ode_bdf(dz_dt, z_init, 0, ts,
            1e-5, 1e-3, 5e2,
            alpha, beta, gamma, delta);
}
model {
  alpha ~ normal(1, 0.5);
  gamma ~ normal(1, 0.5);
  beta ~ normal(0.05, 0.05);
  delta ~ normal(0.05, 0.05);
  sigma ~ lognormal(-1, 1);
  z_init ~ lognormal(10, 1);
  for (k in 1:2) {
    y_init[k] ~ lognormal(log(z_init[k]), sigma[k]);
    y[ , k] ~ lognormal(log(z[, k]), sigma[k]);
  }
}
```

Here is data for it (this pull has the branch that should run it, but I wouldn't be surprised if it fails -- it's not the most tested: https://github.com/stan-dev/math/pull/1641):
```
N <- 20
ts <-
c(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20)
y_init <-
c(30, 4)
y <-
structure(c(47.2, 70.2, 77.4, 36.3, 20.6, 18.1, 21.4, 22, 25.4, 27.1, 40.3, 57, 76.6, 52.3, 19.5, 11.2, 7.6, 14.6, 16.2, 24.7, 6.1, 9.8,
35.2, 59.4, 41.7, 19, 13, 8.3, 9.1, 7.4, 8, 12.3, 19.5, 45.7, 51.1, 29.7, 15.8, 9.7, 10.1, 8.6),
.Dim = c(20, 2))
```

rok-cesnovar:
Fixes #481


rok-cesnovar:
@bbbales2 once https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3-test-binaries/detail/stanc3-test-binaries/87/pipeline finishes

put this in make/local

STANC3_TEST_BIN_URL = https://jenkins.mc-stan.org/job/stanc3-test-binaries/87/artifact

rm bin/stanc and make build.

The old interface doesnt work right now but the variadic does.

Also, it treats 5e2 as real so this is valid

 vector[2] z[N]
  = ode_bdf_tol(dz_dt, z_init, 0, ts,
		1e-5, 1e-3, 500,
		alpha, beta, gamma, delta);

this isnt

 vector[2] z[N]
  = ode_bdf_tol(dz_dt, z_init, 0, ts,
		1e-5, 1e-3, 5e2,
		alpha, beta, gamma, delta);

rok-cesnovar:
Lord Jenkins was not pleased the first time. The stanc3 binaries are now ready with the magic number 88. 

So use:
STANC3_TEST_BIN_URL = https://jenkins.mc-stan.org/job/stanc3-test-binaries/88/artifact

rok-cesnovar:
@bbbales2 @wds15 the _tol signatures should probably not allow the scientific notation of max steps. This will not be valid:
```
vector[2] z[N]
  = ode_bdf_tol(dz_dt, z_init, 0, ts,
		1e-5, 1e-3, 5e3,
		alpha, beta, gamma, delta);
```
The alternative is to handle this at the C++ level somehow also. See https://github.com/stan-dev/stanc3/issues/596

wds15:
fine for me to not allow this (I guess to make it integer).

bbbales2:
Yeah it's totally fine.

wds15:
@rok-cesnovar does this PR display deprecation messages when the parser finds any of the old integrate* functions in a Stan file? I think we should do that as the older function will be slightly slower than the new ones due to the need to use adapters.

rok-cesnovar:
They do not, because we do not currently have the deprecation warnings for functions working (outside of the parser level ones (# for comment, <- for assignments, etc.).

There are 3 issues for that open (https://github.com/stan-dev/stanc3/issues/600, https://github.com/stan-dev/stanc3/issues/165 and https://github.com/stan-dev/stanc3/issues/261) and this seems like something simple to implement (to my naive mind) so maybe we can add this in the meantime while waiting you guys to hash out the math PR.

rok-cesnovar:
@bbbales2 

Just double checking. In your test models you have

zd = ode_rk45_tol(f_1_arg, vd, rd, rad, 1e-6, 1e-6, 100, rd);
zd = ode_rk45_tol(f_0_arg, vd, rd, rad, 1e-6, 1e-6, 100);

but those should not be allowed right? I was under the impression that two variables follow the max_num_steps?
Or am I missing something?

bbbales2:
> but those should not be allowed right? I was under the impression that two variables follow the max_num_steps?
Or am I missing something?

Those are all the pass through arguments. It makes sense to pass through zero or one argument.

rok-cesnovar:
Oh ok. Thanks. Will fix.

rok-cesnovar:
This is ready for review. Tests will only pass once Math PR is in but the review does not have to wait.

bbbales2:
@rok-cesnovar just asked a question over [here](https://github.com/stan-dev/stan/pull/2937#pullrequestreview-446578416) that's relevant

src/frontend/Semantic_check.ml
```diff
@@ -347,6 +348,92 @@ let semantic_check_reduce_sum ~is_cond_dist ~loc id es =
       |> Semantic_error.illtyped_reduce_sum_generic loc id.name
       |> Validate.error
 
+let semantic_check_variadic_ode_tol ~is_cond_dist ~loc id es =
+  let arg_match (x_ad, x_t) y =
```
rybern, line 352: There's some code replication between this function, semantic_check_variadic_ode, and semantic_check_reduce_sum. I think we can at minimum pull out arg_match and args_match

src/frontend/Semantic_check.ml
```diff
@@ -347,6 +348,92 @@ let semantic_check_reduce_sum ~is_cond_dist ~loc id es =
...
+                :: (_, UnsizedType.UVector) (* initial_state *)
+                    :: fun_args
+              , ReturnType UnsizedType.UVector ); _ }; _ }
+    :: initial_state
+       :: initial_time
+          :: times
+             :: rel_tol
+                :: abs_tol
+                   :: { emeta= {type_= UInt; ad_level= UnsizedType.DataOnly; _}; _
+                      } (* max_num_steps *)
```
rybern, line 374: Ideally, we would encapsulate the information about the signature of the ODEs inside stan_math_signatures. I know it's not obvious how to do that until someone takes the time to make the signatures flexible enough to encode these variadic functions, but maybe some of this signature-specific information can be pulled out?

src/frontend/Semantic_check.ml
```diff
@@ -347,6 +348,92 @@ let semantic_check_reduce_sum ~is_cond_dist ~loc id es =
...
+                      
+                      :: args
+    when arg_match (AutoDiffable, UReal) initial_time
+         && arg_match (AutoDiffable, UVector) initial_state
+         && arg_match (AutoDiffable, UArray UReal) times
+         && arg_match (DataOnly, UReal) rel_tol
+         && arg_match (DataOnly, UReal) abs_tol ->
```
rybern, line 381: It seems like these two lines and error types are the only difference between this function and semantic_check_variadic_ode? Can we call that one here, or have them share a helper function? That'd save ~50 lines of replicated code

src/frontend/Semantic_error.ml
```diff
@@ -152,6 +162,96 @@ module TypeError = struct
...
+        Fmt.pf ppf
+          "Ill-typed arguments supplied to function '%s'. Expected \
+           arguments:@[<h>%a@]\n\
+           @[<h>Instead supplied arguments of incompatible type: %a@]"
+          name
+          Fmt.(list UnsizedType.pp ~sep:comma)
+          generate_ode_sig
+          Fmt.(list UnsizedType.pp ~sep:comma)
+          arg_tys
+    | IllTypedVariadicODEGenericTol (name, arg_tys) ->
```
rybern, line 226: Again I'm wondering if there's a less redundant way to write this, since it only differs from IllTypedVariadicODEGeneric by the list of types - but it might not be worth it here, up to you

src/frontend/Semantic_error.ml
```diff
@@ -152,6 +162,96 @@ module TypeError = struct
...
+            Pretty_printing.pp_unsizedtype UReal
+            (* rel_tol *)
+            Pretty_printing.pp_unsizedtype UReal
+            (* abs_tol *)
+            Pretty_printing.pp_unsizedtype UInt
+          (* max_num_steps *)
+        in
+        Fmt.pf ppf
+          "Ill-typed arguments supplied to function '%s'. Available arguments:\n\
+           %sWhere T1 and T2 are of any type.@[<h>Instead supplied arguments \
```
rybern, line 250: do we still have this minimum of two types, T1 and T2?

src/middle/Stan_math_signatures.ml
```diff
@@ -140,8 +140,26 @@ let mk_declarative_sig (fnkinds, name, args) =
 let full_lpdf = [Lpdf; Rng; Ccdf; Cdf]
 let full_lpmf = [Lpmf; Rng; Ccdf; Cdf]
 let reduce_sum_functions = ["reduce_sum"; "reduce_sum_static"]
+
+let variadic_ode_functions =
```
rybern, line 144: tiny nitpick - these might be slightly faster as a set, since the only operation seems to be checking membership

rok-cesnovar: Thanks. Will move to set!

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -775,6 +797,21 @@ let pp_register_map_rect_functors ppf p =
     (list ~sep:cut pp_register_functor)
     (List.sort ~compare (Hashtbl.to_alist map_rect_calls))
 
+let fun_used_in_ode_bdf p =
```
rybern, line 800: this function seems the same as `fun_used_in_reduce_sum` except for the function matching predicate. I'd suggest making a local helper function and pulling out that predicate as an argument

src/frontend/Semantic_error.ml
rok-cesnovar: No. Missed this one. Thanks.

src/frontend/Semantic_error.ml
rok-cesnovar: Will try to reduce the redundancy. Good call

src/frontend/Semantic_check.ml
rok-cesnovar: Will try to add something to signatures. Seems reasonable yeah.

rybern:
Overall looks quite good to me! Mostly just some small suggestions.

It would be best not to have to handle this as a special case in semantic check, just like reduce_sum, but someone would need to make the stan math type signatures flexible enough to represent these variatic functions.

wds15:
@rok-cesnovar How can I grab the latest stanc3 pre-build binary? What URL do I need? I wanted to rerun the SIR examples, just to make sure that using the new interface gives the right performance. Thanks.

rok-cesnovar:
The latest binary is still https://jenkins.mc-stan.org/job/stanc3-test-binaries/97/

But that one only works for the new interface and not for the old one (we reverted to the old functor). I am doing that update together with addressing the review comments. I am in the middle of fixing that but probably wont finish until the afternoon. 

So for now you can use the new interface with this binary and the old binary with cmdstan develop. Not ideal, I know.

wds15:
Ah... ok.. so I can at least use the 97 to process a variadic ODE Stan model... will do that.

rok-cesnovar:
Yes, for variadic ODE 97 is fine. The only things that changed since 97 for the variadic interface are cosmetic changes to the OCaml code, the generated code is still the same.

rok-cesnovar:
And the code-gen will stay the same, so these tests that you will perform will be valid for the final version too. All the requested changes are cosmetic or touch the old interface.

src/frontend/Semantic_check.ml
rok-cesnovar: Pulled arg match and args_match out of the semantic checks and collapsed the separate _tol and normal check.

rok-cesnovar:
I addressed all review comments, moved all signature types related info to stan math signatures and reduced the code duplication. This passes tests now https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3/detail/PR-525/43/pipeline but Github is having some issues ATM so its not showing up.

wds15:
@rok-cesnovar This could be a bit late... but we were just talked over it at the Stan Math meeting (with @bbbales2 @SteveBronder and @syclik ) ... the issue with variadic ODE is that it slows down a lot the old integrate_ode* stuff. So what about if the stanc3 compiler translates the calls of the old intrgrate_ode* directly to the new variadic ODE interface? This could be done by emitting respective C++ code or by mapping things within the Stan language (which is hopefully the easier way with all the AST magic you have).

If not doable for now... then ideally in a little bit. This would allow all existing code to keep running at full speed without any user change to their models.

rok-cesnovar:
Sorry I missed the meeting. I think this is doable and a great idea. Let me give it a try.

rok-cesnovar:
So we will have to change the input y0 to a vector and the result from an array of vectors to a 2d array. @wds15 @bbbales2 I am guessing this will hurt performance but no to a 20% extent?

wds15:
It won’t affect performance at all if you just do that conversion once. The issue with the current adaptier is that it’s done every time when the ode rhs is called, but you should be able to code it up such that it’s done a single time such that there is no penalty.

rok-cesnovar:
Oh, so currently the conversion is done for each step in the solve?

wds15:
Yes

bbbales2:
> should be able to code it up such that it’s done a single time such that there is no penalty.

@wds15 I think we always have to cast the output. We'd have to modify the internals of the function to make it return a vector.

wds15:
But the parser can just create the old and a converted ode rhs code in order to avoid the conversion upon each call.

I hope it’s that easy...it does sound easy to me, but I don’t deal with the parser...

wds15:
Ah...I see...the issue is that the user function must be modified...hmm.

bbbales2:
I don't think it will be easy to convert all the variations in how stuff can be written:

```
real out[1];

out[1] = theta[1] * y[1];

return out;
```

```
real out[1] = { theta[1] * y[1] };
return out;
```

```
return { theta[1] * y[1] };
```

to make them vector-like. I think the difficulty of automatically converting the output to be a vector would be very high. The input should be do-able though, and then if we just convert the output that's 50% of the losses gone.

bbbales2:
Yeah yeah, but I think the input state could still be handled. We'll see. This would have been better if we did it earlier, but deprecation warnings are good enough (I'm very happy to just stay with what we've got if that means a smooth release).

wds15:
Unless the parser can do magic with the ast, mir or anything like that we should probably let this alone. The idea is sweat, but dies on the details.

rok-cesnovar:
No magic yeah. So just a review left to do. Attaching the Linux binary. unfortunately, I cant build on Jenkins right now, some Windows machine issues.

[stanc.zip](https://github.com/stan-dev/stanc3/files/4933847/stanc.zip)

wds15:
Is this still on track for tomorrow? Would be good to get it in 2.24... as we slow down ODEs by 30% otherwise...

rok-cesnovar:
The first review comments were addressed so just a final review needed.

src/frontend/Semantic_error.ml
```diff
@@ -152,6 +156,68 @@ module TypeError = struct
           (String.concat ~sep:"" lines)
           Fmt.(list UnsizedType.pp ~sep:comma)
           arg_tys
+    | IllTypedVariadicODE (name, arg_tys, args) ->
```
rybern, line 159: I think this could still have been two errors like you had it, I only meant to suggest that they could both use a helper function because their logic overlapped so much. But this works too, I'm sorry if I put you through extra effort

src/frontend/Semantic_error.ml
```diff
@@ -152,6 +156,68 @@ module TypeError = struct
...
+        let types x = List.map ~f:(fun (_, t) -> t) x in
```
rybern, line 160: small note: `(fun (_, t) -> t)` is also called `snd`

src/frontend/Semantic_error.ml
```diff
@@ -152,6 +156,68 @@ module TypeError = struct
...
+        let optional_tol_args =
+          if Stan_math_signatures.is_variadic_ode_tol_fn name then
+            List.map
+              ~f:(fun (_, y) -> y)
```
rybern, line 164: could use your `types` function here?

src/frontend/Semantic_error.ml
```diff
@@ -152,6 +156,68 @@ module TypeError = struct
...
+              Stan_math_signatures.variadic_ode_tol_arg_types
+          else []
+        in
+        let args_types = List.map ~f:(fun (_, t) -> t) args in
```
rybern, line 168: again could use `types` function

src/frontend/Semantic_error.ml
```diff
@@ -152,6 +156,68 @@ module TypeError = struct
...
+        let generate_ode_sig =
+          [ UnsizedType.UFun
+              ( Stan_math_signatures.variadic_ode_mandatory_fun_args @ args
+              , ReturnType Stan_math_signatures.variadic_ode_fun_return_type )
+          ]
+          @ types Stan_math_signatures.variadic_ode_mandatory_arg_types
+          @ optional_tol_args @ args_types
+        in
+        let variadic_ode_generic_signature =
```
rybern, line 177: IMO it's not clear why you're building these things, a comment describing your strategy throughout this section might make it more clear

src/frontend/Semantic_error.ml
```diff
@@ -152,6 +156,68 @@ module TypeError = struct
...
+            else []
+          in
+          let nth_fun_arg n =
+            List.nth_exn
+              (types Stan_math_signatures.variadic_ode_mandatory_fun_args)
+              n
+          in
+          let nth_arg n =
+            List.nth_exn
+              (types Stan_math_signatures.variadic_ode_mandatory_arg_types)
```
rybern, line 190: this might recompute the result of `(types ...)` every time it's called - very tiny issue, but FWIW one alternative way of writing it might be something like

```
let arg0 :: arg1 :: arg2 :: _ = types ...
```

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -242,7 +255,16 @@ let pp_fun_def ppf Program.({fdrt; fdname; fdargs; fdbody; _})
         | _ ->
             raise_s
               [%message
-                "Ill-formed reduce_sum call! This is bug in the compiler."] )
+                "Ill-formed reduce_sum call! This is bug in the compiler."]
+      else if String.Set.mem funs_used_in_ode_bdf fdname then
+        (* Produces the reduce_sum functors that has the pstream argument
```
rybern, line 260: outdated comment here i think

src/frontend/Semantic_error.ml
rok-cesnovar: No problem, I think this is more concise and also makes less code to maintain or fix when we add the type for variadic functions.

src/frontend/Semantic_error.ml
rok-cesnovar: Thanks. Fixed.

src/frontend/Semantic_error.ml
rok-cesnovar: Yeah, good call. Done.

src/frontend/Semantic_error.ml
rok-cesnovar: Added this comment: 

This function is used to generate the generic signature for variadic ODEs, i.e. with ... representing the variadic parts of the signature. This should be removed once a type representing variadic arguments is added. The generic signature is printed when there is a semantic error with one of the non-variadic arguments in the signature. If there is a mismatch in the variadic arguments, we print the non-generic expected signature (with explicit types for variadic args).

src/stan_math_backend/Stan_math_code_gen.ml
rok-cesnovar: Indeed. Fixed.

src/frontend/Semantic_error.ml
rok-cesnovar: Thanks. But this way I get a warning for a not-exhaustive pattern matching that I am unable to fix: 
File "src/frontend/Semantic_error.ml", line 192, characters 10-471:
Error (warning 8): this pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
(_::_::[]|_::[]|[])

rok-cesnovar: Ignore the above. Fixed.

rybern: Oh, yeah, that's unfortunate. You could get around it like:

```
let (arg0, arg1, arg2) = match types ... with
   a0 :: a1 :: a2 :: _ -> (a0, a1, a2)
   _ -> error "oops"
```

but again it's a very small issue that probably doesn't deserve our attention :P

rok-cesnovar: Went with:

```ocaml
match
            ( types Stan_math_signatures.variadic_ode_mandatory_arg_types
            , types Stan_math_signatures.variadic_ode_mandatory_fun_args )
          with
          | arg0 :: arg1 :: arg2 :: _, fun_arg0 :: fun_arg1 :: _ ->
              Fmt.strf "(%a, %a, ...) => %a, %a, %a, %a, %a ...\n"
                UnsizedType.pp fun_arg0 UnsizedType.pp fun_arg1 UnsizedType.pp
                Stan_math_signatures.variadic_ode_fun_return_type
                UnsizedType.pp arg0 UnsizedType.pp arg1 UnsizedType.pp arg2
                Fmt.(list UnsizedType.pp ~sep:comma)
                optional_tol_args
          | _ ->
              raise_s
                [%message
                  "This should not happen. Variadic ODE functions have exactly three \
                   mandatory arguments and the function supplied to the variadic ODE \
                   function has exactly two mandatory arguments."]
```

rok-cesnovar:
Thanks @rybern ! 

I addressed all your comments. If you are good with the pattern matching thing on args, I will merge once tests pass.

src/frontend/Semantic_error.ml
rybern: Looks good!

rybern:
Looks good to me

PR #528, opened 2020-05-02
[WIP] Fixed Size Sparse Matrices

SteveBronder:
## Summary

Adds a `sparse_matrix[array<int>, array<int>, int, int]` type that can be declared in `data` and `parameters` blocks.

## Description

This works on the following program

```stan
data {
  int N_dt;
  int M_dt;
  int nz_vals_dt;
  int nz_rows_dt[nz_vals_dt];
  int nz_cols_dt[nz_vals_dt];
  sparse_matrix[nz_rows_dt, nz_cols_dt, N_dt, M_dt] X_dt;
  int N_par;
  int M_par;
  int nz_vals_par;
  int nz_rows_par[nz_vals_par];
  int nz_cols_par[nz_vals_par];
}

transformed data {
  // Should be allowed
  real data_cool = X_dt[1, 1];
  // Fails correctly
  // YY_tdt[1, 1] = 10;
}


parameters {
  sparse_matrix[nz_rows_par, nz_cols_par, N_par, M_par] XX_par;
}

```

A couple points:

1. Related to @rybern's comment on [#1827](https://github.com/stan-dev/stanc3/pull/491#issuecomment-620134464) I wrote this restricting the `sparse_matrix[array, array, int, int]` declaration to only be legal in `data` and `parameters` because I'm worried for a `transformed parameter` of the decleration we could have an instance of something like

```stan
transformed parameters {
  sparse_matrix[nz_rows, nz_cols, N_par, M_par] XX_par = X * transpose(X);
}
```

where the expression on the right hand side would not have the correct nonzero rows and columns. I think we'd have to valid at runtime that the RHS expression has the correct nz rows and cols, which I'm not sure we want to have to do? Or should this be supported in each block? 

2. I think this would look nicer but can't figure out the proper place / how to handle a `sparse_matrix[array<int>[2, nz_vals], int, int]` like

```stan
 int N;
 int M;
 int nonzero_N;
 int[2, nonzero_N] nonzero_vals;
 sparse_matrix[nonzero_vals, N, M] X;
```

For the `SizedType` I get that it would be 

```ocaml
  | SSparseMatrix of 'a t * 'a * 'a
```

Though we want a very specific type here (`int[2, nz_vals]`). For checks on that type would the semantic checker be in charge of verifying the correct type is there? I also found this made a lot of the functions a bit difficult. I can take another stab at doing this style tmrw.

3. This PR is waiting on a few stan and stan math PRs before it can be merged

4. I have commented out code [here](https://github.com/stan-dev/stanc3/pull/528/files#diff-5b95435ef954991e37fcffd291911d96R778) to perform a check that the nonzero row and column arrays both have a dim deriving from the same object. But I'm having some trouble with the monad there to sort out how to check against the dims of both types


bob-carpenter:
I think we do want to validate at runtime that the RHS and LHS have the same sizes.  We'd have to do that for assignment anyway in order not to segfault.  So I don't see any reason not to allow a transformed parameter or generated quantity.

bob-carpenter:
Good question about semantic checking of size 2.  I don't think it'd be terrible to leave that until runtime.  That's where we have to check that sizes aren't negative, for example, since they can come in as integer data.

SteveBronder:
> I think we do want to validate at runtime that the RHS and LHS have the same sizes. We'd have to do that for assignment anyway in order not to segfault. So I don't see any reason not to allow a transformed parameter or generated quantity.

What would that look like in the C++ code? So for

```stan
transformed parameters {
  sparse_matrix[nz_rows, nz_cols, N_par, M_par] XX_par = X * transpose(X);
}
```

would we have to first assign the expression to a temporary and then iterate over the temporary and `XX_par` to make sure the sizes match up?

SteveBronder:
Thinking about the C++ code for the above would it have to look like

```cpp
Eigen::Matrix<double> XX_par_tmp = X * transpose(X);
check_sparse_sizes(nz_rows, nz_cols, N_par, M_par, XX_par_tmp); 
XX_par = XX_par_tmp;
```

Or is there another way to check that at runtime? It feels v expensive but I'm not sure how to get around it

bob-carpenter:
`nz_rows` and `nz_cols` will be fixed at compile time.  So we can store those once and for all.  Then we do the assign and check it has the right `nz_rows` and `nz_cols` afterward.  

Right ow, I think we store the sizes in the objects themselves.  To continue that pattenr, we don't quite have to copy the whole thing, just keep `nz_rows` and `nz_cols` around long enough (which may mean copying them).

Issue #441, opened 2020-01-23
restrict map_rect to legal data-only args


bob-carpenter:
The following program should not compile because `xi[i]` isn't an entirely data-only expression.  The value `xi` is a data variable, but the index `i` is local.  That allows illegal programs like this to get compiled: 

```
functions {
  vector f(vector beta, vector theta, data real[] xr, data int[] xi ) {
    return [1.0]';
  }
}
transformed data {
  vector[2] beta;
  vector[2] theta[2];
  real xr[2, 2];
  int xi[2, 2, 2];
}
parameters {
  real y;
}
transformed parameters {
  real s = 0;
  for (i in 1:2) {
    // BAD!  DATA CHANGING BETWEEN CALLS
    s += sum(map_rect(f, beta, theta, xr, xi[i]));
  }
}
```

These will produce different results with MPI turned on and turned off, which we don't ever want to happen.

rok-cesnovar:
This is for MPI or any map_rect case? Also tagging @wds15

wds15:
Only mpi needs these additional restrictions.

wds15:
I just want to point out that our user doc does correctly state the additional restrictions which we need for map_rect when used with MPI:

https://mc-stan.org/docs/2_21/stan-users-guide/map-function.html

The thing is that for MPI we call "data" something which never changes throughout program execution - that is different as compared to calling "data" something which we do not autodiff.

The crux with the example above is that when MPI is not used (no threading on or we use threading), then `map_rect` will use the data as given in each loop iteration, since no data caching is happening. Only if MPI is used - so something happening after the parser gets involved - and only then data caching with the assumption of static data (as being defined in data or transformed data) is being used.

The GPU caching facilities should be affected by these issues as well.

FIY, we decided during the development of `map_rect` that using hashes of the data would be too costly in order to decide if one would need to update the data.

The hardest thing with this parser "bug" is to not restrict map_rect too much since the additional requirements are really only applicable to MPI which is probably anyway not getting used so much any longer given that the Intel TBB is far easier to deploy, run & there are no more speed penalties for using threading compared to MPI (these went away with the TLS refactor and the use of the TBB).

So before putting additional constraints on the usability, we should consider to build into the `map_rect` MPI functionalities which make it possible to detect if data caching holds or not. The MPI user group is likely a lot smaller than the TBB user group.

rok-cesnovar:
> The GPU caching facilities should be affected by these issues as well.

[...]

Issue #540, opened 2020-05-20
allow data-only functions in generated quantities


nhuurre:
Generated quantities block does not use autodiff so data-only restrictions don't matter.
Local variables are DataOnly but parameters visible from previous blocks are still considered autodiffable. This is a bug; everything should be DataOnly.

This model shows the inconsistency:
```stan
functions {
  real add_one(data real x) {
    return x+1;
  }
}
parameters {
  real p;
}
model {
  p ~ std_normal();
}
generated quantities {
  real r = p;
  real s = add_one(r); // this works
  real z = add_one(p); // this fails; should work
}
```

bob-carpenter:
I agree that everything should be data only.  This is probably a carryover from the original implementation that used the origin of the variables rather than the block that they're in.

PR #542, opened 2020-05-22
Change parameters to DataOnly inside generated quantities block

nhuurre:
issue #540 

There's also a problem with `map_rect` because it has a different interpretation what `DataOnly` means.
`DataOnly` just means there's no autodiff but `map_rect` expect the data to be persistent between iterations. We should probably add `StaticData` qualifier for that case.


rok-cesnovar:
> `map_rect` expect the data to be persistent between iterations

If using MPI yes.

> We should probably add `StaticData` qualifier for that case.

That or refactor the MPI interface yes. I was thinking we could just add the requirement that non-literal indexing can not be used with map_rect inputs if using MPI. So x[0], x[1], y[0,0] would work but x[i] would not. Which would fix https://github.com/stan-dev/stanc3/issues/441
Would that work or would that be too confusing?

And once we move the copying to the model level we could also remove that restriction.

bob-carpenter:
Yes, we defintely need to split `const data` from `data` or whatever we want to call it.  The former is for map_rect and the latter for ODE solvers, etc.

bob-carpenter:
> I was thinking we could just add the requirement that non-literal indexing can not be used with map_rect inputs if using MPI. So x[0], x[1], y[0,0] would work but x[i] would not. Which would fix #441 Would that work? 

I think we'll still have problems with this:

```
cond ? x[0] : x[1]
```

nhuurre:
Restricting the indexing of `map_rect` arguments is probably the best solution for the MPI static data issue. In any case, that is not a blocker for this PR. I mentioned it here only because I had to change a few of test files in `test/integration/bad/map_rect/`. Those tests check that certain uses of `map_rect` are invalid and they can no longer catch non-static data in generated quantities.

Issue #545, opened 2020-05-25
Add optimization levels


rok-cesnovar:
- [ ] Add optimization levels (--optimize=0 to --optimize=3 for example) and put all current optimizations to the highest level. See https://github.com/stan-dev/stanc3/issues/544#issuecomment-633285347 for example

- [ ] move optimizations to their levels in separate PRs and add docs to stan-dev/docs. See https://github.com/stan-dev/stanc3/pull/522#issuecomment-633453999 for more.

- [ ] decide what should be the default level

A good candidate for the first one would be https://github.com/stan-dev/stanc3/issues/544

PR #549, opened 2020-05-26
[WIP] Optimization level interface

rybern:
This PR adds a command-line interface for applying optimizations by level, like GCC's -O1, -O2, etc. See #545 and #522 for context. See [here](https://github.com/stan-dev/docs/issues/193) for proposed docs.

The CLI usage currently looks like this:
```
Usage: %%NAME%% [option] ... <model_file.stan>
...
  -O0                             Do not apply optimizations to the Stan code.
  -O1                             Apply level 1 compiler optimizations (only basic optimizations).
  -O2                             Apply level 2 compiler optimizations (all optimizations which are not likely to have small numerical differences from the original program).
  -O3                             Apply level 3 compiler optimizations (all optimizations).
  --O                             Apply level 3 compiler optimizations (all optimizations).
...
```

We should discuss some things:
#### 1. What should be the organizing principal of each optimization level?
Currently this PR implements the following scheme:
-   **O0**
    No optimizations are applied.
-   **O1**
    Only optimizations which are simple to understand, do not dramatically change the program, and are unlikely noticeably slow down compile times are applied.
-   **O2**
    All optimizations which are unlikely introduce numerically differences relative to the unoptimized program are applied.
-   **O3**
    All optimizations are applied.

I don't know if the idea for **2** is necessary. It was a helpful concept for debugging, but here it might just be confusing and rarely useful. An alternative would be to separate the numerical stability idea into a separate CLI flag, then have level 2 be something like:

-   **O2**
    All optimizations which do not significantly change the structure of the program.

So level 2 would include partial evaluation and all of its helpers, but would not include e.g. lazy code motion or function inlining. Again, I'm not sure if this is a helpful distinction. On the one hand, avoiding restructuring optimizations may help people to read the output code, but on the other hand, partial evaluation could be considered the more aggressive optimization.

#### 2. Should each optimization be added one-by-one to the interface in separate PRs?

In #545, @rok-cesnovar suggested that optimizations be added one-by-one to the optimization levels in separate PRs. Here are some concerns I have (keeping in mind that I 100% see where Rok's caution is coming from):
- Many optimization don't make sense without others. E.g., expression propagation doesn't make sense without lazy code motion; copy propagation and expression propagation don't make much make sense without partial evaluation, etc. So the optimizations are not really independent units.
- Every combination of optimizations that's merged into master needs to be fully tested, since they can have complex interactions. My testing effort did not (could not) test every combination. Those tests each take many hours to run, and the one-by-one proposal would require many rounds of this. It might actually be valuable to test random subsets of optimizations to try to turn up more bugs, but I do not have the time to do that right now.
- I'm not sure what additional discussion the individual optimization PRs would enable that we couldn't have in this PR.

Tagging @rok-cesnovar @bob-carpenter @VMatthijs @SteveBronder, looking forward to anyone's input.


SteveBronder:
I think this is around the right ballpark. Though it's hard because gcc essentially sets levels at space-time tradeoff for `O2/O3` (use more space -> make things faster like loop unrolling, propogate constants) then an `Ofast` for optimizations that can cause numerical instability

But this is pretty different since optimizations like swapping `log(1+x)` for `log1p(x)` can actually make things more numerically stable, it's just different from what they literally wrote. Is there an actual math reason someone would want `log(1 + x)` over `log1p(x)`? imo if we can make things more numerically stable for users then that's something that should always be on.

I wouldn't mind phrasing this like gcc/clang where 

- O1 Simple optimizations that can only help with speed / numerics

- O2 take more space and allows larger compilation times

- O3 perform multiple sweeps, always trading off space for performance

- Ofast Optimizations that could lead to numerically unstable results (if we have any optims like that).

 Looking at the [gcc optims](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html) we could do something like this

- O/O1
    - With -O, the compiler tries to reduce code size and execution time, without performing any optimizations that take a great deal of compilation time. 

```
inline-functions-called-once (This doesn't exist but could be good to add)
Constant propagation*
Copy propagation*
Dead code elimination
Auto-differentiation level optimization
```

- O2
    -  stanc3 performs nearly all supported optimizations that do not involve a space-speed tradeoff. As compared to -O, this option increases both compilation time and the performance of the generated code. 

- O2 turns on all optimization flags specified by -O1. It also turns on the following optimization flags: 

```
Function inlining*
Static loop unrolling
One step loop unrolling
Expression propagation*
Partial evaluation*
Lazy code motion*
```
* I'd remove the stars here and just give them different names if they are different optimizations than what's at O3

- O3
   -O3 turns on all optimizations specified by -O2 and also turns on the following more expensive optimization flags: 

```
(give the duplicate names here different names if they are different / more aggressive)
Function inlining
Static loop unrolling
One step loop unrolling
Constant propagation
Expression propagation
Copy propagation
Dead code elimination
Partial evaluation
Lazy code motion
Auto-differentiation level optimization
```

idk, grouping things is hard!

rok-cesnovar:
We don't really need to go one by one. If there are dependencies between them and one does not make sense without the other, that definitely does not make sense to separate it.

I think the good docs you started with will help a lot. There was some discussion already on the docs issue https://github.com/stan-dev/docs/issues/193 but I think it makes more sense to have the implementation discussion here and the docs tweaks there. Otherwise it will be impossible to follow the discussion. Also ideas for additional optimizations probably deserve their own issue.

One question that should be definitely be addressed is the limit for static loop unrollling. I am not sure what the exact limit should be, but there definitely should be one.

Otherwise, once you think the doc is ready lets make a binary and put out a call for testing.

rok-cesnovar:
Should we set the default level to 1? Or should we hold off for one release cycle and promote the use of optimizations?

rybern:
I would be a little bit more comfortable with promoting optimizations for stress testing before they are turned on by default, even just for dead code elimination. But then there's the question of whether they should be tuned for performance per #555 before they're promoted as well, so people don't get turned off of them if they aren't effective enough.

SteveBronder:
Okay so I got this updated with master. I built the mother model with the new settings and turned off some things. Like I think `block_fixing` and `optimize_ad_levels` gives the wrong code. It was setting transformed parameters to data whenever they were initialized with data or literal types. But that should just be parameters initialized with data right? Or is that false and actually transformed parameters initialized with just data can be data? If so that is very cool and we should have that on always.

I think I'd like to pull this back a bit, for now I think we should have just `-O0`, and `-O1` with things we know 100% work and are good. I also think `O3` takes too long, for the mother model I let it run for 5 minutes and then just stopped it. I think if we expose `O0` and `O1` we should put everything else into something like `Od` where `d` stands for dangerous since they are untested and we are not sure they will produce the correct thing or even be faster. I'm thinking something like 

```ocaml

let level_optimizations (lvl : optimization_level) : optimization_settings =
  match lvl with
  | O0 -> no_optimizations
  | O1 ->
    { function_inlining= false
    ; static_loop_unrolling= false
    ; one_step_loop_unrolling= false
    ; list_collapsing= true
    ; block_fixing= false
    ; constant_propagation= true
    ; expression_propagation= false
    ; copy_propagation= true
    ; dead_code_elimination= true
    ; partial_evaluation= true
    ; lazy_code_motion= false
    ; optimize_ad_levels= false
    ; preserve_stability= false }
  | Od -> all_optimizations
```

though tbh I'd probably feel safer just removing Od or at least leaving it undocumented since it's really for internal testing use

rybern:
Thanks for working on this Steve!

> Like I think block_fixing and optimize_ad_levels gives the wrong code. It was setting transformed parameters to data whenever they were initialized with data or literal types.

 * I'm surprised that block_fixing would have an impact on that
 * I'm actually not seeing the problem with setting transformed parameters to non-AD if they don't depend on parameters, but I wouldn't be surprised if I were wrong
 * I did test that these optimizations gave correct results, and I think I included the mother model, but that was a while ago and I don't think we're testing that in our CI

I'm fine with O0/O1/Od, but I'd suggest documenting Od. It also would be easy to let users turn individual optimizations on/off from the CLI.

SteveBronder:
> I'm surprised that block_fixing would have an impact on that

What does block fixing do? Lol I looked over the docs but I didn't see it listed anywhere. The only comments / docs in the code say
```
(* TODO: implement SlicStan style optimizer for choosing best program block for each statement. *)
(* TODO: add optimization pass to move declarations down as much as possible and introduce as
   tight as possible local scopes *)
(* TODO: add tests *)
(* TODO: add pass to get rid of redundant declarations? *)
```

> I'm actually not seeing the problem with setting transformed parameters to non-AD if they don't depend on parameters, but I wouldn't be surprised if I were wrong

Yeah actually now that I think about it I think your right. Because transformed params that are not based on parameters, that's data. I think that makes sense?

> I did test that these optimizations gave correct results, and I think I included the mother model, but that was a while ago and I don't think we're testing that in our CI

Yeah tbh I could be a bit conservative here. Let me read through a few more Stan hpp files maybe I just need to get more comfortable with a few of these

rybern:
> What does block fixing do?

It's just cleaning up some uses of nested `SList` and `Block` MIR nodes that can get messy from other optimizations

SteveBronder:
Oh these are all well documented in Optimize.mli!

Okay so I settled on this

```
let level_optimizations (lvl : optimization_level) : optimization_settings =
  match lvl with
  | O0 -> no_optimizations
  | O1 ->
      { function_inlining= false
      ; static_loop_unrolling= false
      ; one_step_loop_unrolling= false
      ; list_collapsing= true
      ; block_fixing= true
      ; constant_propagation= true
      ; expression_propagation= false
      ; copy_propagation= true
      ; dead_code_elimination= true
      ; partial_evaluation= true
      ; lazy_code_motion= false
      ; optimize_ad_levels= true
      ; preserve_stability= false }
  | Od -> all_optimizations
```

I chose those semi-arbitrarily based on whether the c++ looked a little odd and how long it took to compile the mother.stan model. I kinda wish we could break these up between data and parameters etc because the loop unrolling etc would be good for working with doubles but idt we want that when looping over vars

SteveBronder:
Okay here is my reasoning behind these 

```
       function_inlining= false
```

takes too long for the mother model

```
      ; static_loop_unrolling= false
      ; one_step_loop_unrolling= false
```

For data this optimization is always good, but for parameters idt this optimization is useful. since we can't differentiate when to use it for each I say no-go for now

```
      ; list_collapsing= true
      ; block_fixing= true
```

Internal things that seem good. maybe we could even just always have these things on even for O0? Is it even possible for this to produce something bad? It seems like it's always good

```
      ; constant_propagation= true
```

Letting the c compiler know compile time constant values is always good

```
      ; expression_propagation= false
```

Bad without lazy code motion, no lazy code motion so off 

```
      ; copy_propagation= true
```

Letting the compiler know two things are the same thing, seems good!

```
      ; dead_code_elimination= true
```

Removing dead code, good! We should probably even turn this on for `O0`

```
      ; partial_evaluation= true
```

```
      ; lazy_code_motion= false
```

This generates a lottt of extra code in the C++ and I haven't had time to evaluate whether these optimizations are meaningful. I think we need something like a `Dead Object Elimination` pass after lcm so that any lcm objects made during this pass that are not used get eliminated.

```
      ; optimize_ad_levels= true
```

Things that are data, as data. Good! tbh I think another candidate that we could even turn on in `O0`. The fact that we make all types autodiffable in the model block when they don't need to be almost feels like an implementation detail rather than something the user would ever rely on.

```
      ; preserve_stability= false }
```

This one I'm fine with at O1 as it actually improves numerical accuracy. I'd say at O0, but at O0 idt we should change the actual functions called by the user.

rybern:
Sounds good, a few comments:

>       ; static_loop_unrolling= false
>       ; one_step_loop_unrolling= false

The idea is to boost partial evaluation and constant propagation (and LCM), but it does generate lots of extra code (especially static unrolling). Not sure why it differs for parameters?

> Removing dead code, good! We should probably even turn this on for O0

It would be nice for O0, especially to get rid of those assignments to fill declared variables, but it's also possibly a little risky. Sometimes it's hard to tell if a statement has a side effect or not, so there's a chance of bugs.

> Things that are data, as data. Good! tbh I think another candidate that we could even turn on in O0

Ditto - agreed it'd be good in O0 but there's a (very small) chance of bugs

SteveBronder:
> The idea is to boost partial evaluation and constant propagation (and LCM), but it does generate lots of extra code (especially static unrolling). Not sure why it differs for parameters?

For data (doubles) loop unrolling tells the compiler that it can definitely use SIMD instructions in a particular place. But with parameters we don't have SIMD instructions. Each step in the loop we have to do the forward pass, store data for the reverse pass, and then setup a callback for the reverse pass. So we don't really get the benefit like we do for doubles. We actually want to do the opposite and roll those loops up to vectorized code so that we can allocate memory in big chunks and setup one reverse pass callback for the whole vector. 

> It would be nice for O0, especially to get rid of those assignments to fill declared variables, but it's also possibly a little risky. Sometimes it's hard to tell if a statement has a side effect or not, so there's a chance of bugs.

> Ditto - agreed it'd be good in O0 but there's a (very small) chance of bugs

Okay let's set those off to O1 then until we are more sure

SteveBronder:
@rybern if tests pass are you cool with this impl? Should we have someone else check it out?

rybern:
Oof sorry @SteveBronder I didn't see this. Yep I think this would be fine

VMatthijs:
My real question when reviewing this is: how have you convinced yourselves that these optimizations are correct? I remember writing a lot of this code. I tried as hard as I could to get things correct. I tried to test optimizations with unit tests, but I never really had good end to end tests. What's your confidence in the correctness of this code? If it turns out to have bugs after all, how do we plan to find out?

SteveBronder:
> My real question when reviewing this is: how have you convinced yourselves that these optimizations are correct? I remember writing a lot of this code. I tried as hard as I could to get things correct. I tried to test optimizations with unit tests, but I never really had good end to end tests. What's your confidence in the correctness of this code? If it turns out to have bugs after all, how do we plan to find out?

It's a good Q, @rybern mentioned previously that when he cleaned up the optimization code he tested with them all on and got results back that were within an reasonable MCSE of the non-optimized version (at least I believe I remember him saying that Ryan correct me if I'm wrong). The reason I set these to O1 and not on by default is that I can only read so much C++ to make sure these things are happening properly. 

For most of the ones I chose for O1 they would lead to a catastrophic failure that would be easy to see is wrong (like trying to assign var types to doubles). There could still be a few tricky ones, for instance partial evaluation. 

It's a really hard Q on how to test these, especially because of how setting them off in certain orders can effect things. Maybe we need a Jenkins step that compiles with O1 and then no optims, then runs either the examples or benchmark models and verifies the results are within some MCSE?

VMatthijs:
> There could still be a few tricky ones, for instance partial evaluation.

Agreed! I wonder if we should write some tests for each of those partial evaluations. It'd be a lot of work... They should all be fine, seeing that I basically used the compound functions that are listed in the manual. Then again, it's so easy to make a mistake when implementing these things...

rybern:
> > How have you convinced yourselves that these optimizations are correct?

> It's a good Q, @rybern mentioned previously that when he cleaned up the optimization code he tested with them all on and got results back that were within an reasonable MCSE of the non-optimized version (at least I believe I remember him saying that Ryan correct me if I'm wrong).

That's about right, thanks @SteveBronder. I wrote a bit about what I did in #522 and #555. I used a corpus of example Stan programs and checked that the unoptimized and optimized versions produced (almost) exactly the same log densities at random points (I didn't actually use MCSE). When I finished #522, the whole corpus passed with all of the optimizations active. It'd probably be a good idea to run those tests regularly.

I'm not sure I'll really be confident until we have some kind of randomized testing or verification system. If I had to bet I'd say there are still correctness bugs, but I'm guessing they're rare and mostly in the complicated ones. I think it's reasonably safe to activate the simpler ones.

> There was also the idea a while back to always compile both an optimized and non-optimized version of the program, to compare them for one functions/gradient evaluation, and then to stick with the optimized one if they agree (and otherwise tell the user to report a bug with their model).

Yeah, I like this idea. It'll have to be in CmdStan rather than stanc3 though.

SteveBronder:
> Agreed; partial evaluation seems like one of the trickiest ones! I wonder if we should write some tests for each of those partial evaluations. It'd be a lot of work... They should all be fine, seeing that I basically used the compound functions that are listed in the manual. Then again, it's so easy to make a mistake when implementing these things...

I feel like that would be v difficult and I'm not sure how much value we would get out of it. Since we are saying "x is y" if we actually had a stan program and compared the two I think it just double checks our logic no? imo I think we can just read them over and be conservative in knowing "x is exactly y"

Does anyone have any disagreements with merging this as is? Personally I think the ones at O1 are v good and it's a good idea to give them to users.

On a side note, I looked over the code for `optimize_ad_levels` and imo I'd almost call it a performance bug that we don't have that on by default. The code just checks if an object is ever assigned to from an autodiff type, so unless there's a bug in the compiler giving a DataOnly type when it should be an autodiff type idt it can produce errors?

Another kind of interesting thing would be to have something like the partial evaluator on during pedantic mode. So if we see an expression that could be optimized we could say, "Hey you wrote X, but you could also write Y". Same could be said about users doing `log(1 + x)` instead of `log1p(x)`.

SteveBronder:
Just bumping this, the single function inlining would be v nice for the new matrix type!

SteveBronder (approved):
Oh I just realized since I'm not the author I'm the one who approves. @rybern @VMatthijs I'd like to merge this as I have other little optims I'd like to add and the handful of the ones here at O1 will also help with the new matrix type. Is this fine to merge?

rybern:
Sorry @SteveBronder thought I replied here. I think this is fine to merge

WardBrian:
I know I’m joining this late, but I’m not sure  if the idea of optimization levels is the right fit for this stage in the process. I think it would be better to enable or disable individual optimizations through the command line interface, rather than group them, until we have a much better idea of which optimizations are good/beneficial

rybern:
I'd be happy to have that as well, that's why the interface to Optimize.ml is the way it is, but I also thinks it'd be reasonable to have recommended collections so as not to require users to sort through all of the optimizations

rybern:
I'd be okay with that option, although I don't see the downside of including levels. Is it that they'd be too convenient a foot-shooting gun, or that the individual activation would let us debug more easily?

WardBrian:
To me it's both. Individual control lets us test and analyze much easier, _and_ I think a lot of users will immediately see that `--O3` is an option and assume it will work/be the best/ etc, regardless of any warning to the contrary

rybern:
Gotcha. I agree on the point that individual controls will help us debug and help users narrow down any issues they run into. But, I think we can find a way to warn users sufficiently about the experimental nature of any debugging levels. For example, we could include it in the CLI description, we could always emit a warning to stderr, we could even call it `--O3-experimental` or something. We could also put in a message to recommend that users play with the individual controls if they run into an issue with a level.

WardBrian:
I think `--O3-experimental` is probably the only thing which would actually discourage people. Tons of users are using advi, for example, despite it doing all the stdout/doc warnings about being experimental

SteveBronder:
@serban-nicusor-toptal @rok-cesnovar looking at Jenkins it seems be failing at the Verify stage. It looks like it's missing a line to checkout a particular branch? Looking at the Jenkins I actually can't tell where that's being called from. Is it in the `StanUtils` library?

https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3/detail/PR-549/50/pipeline

SteveBronder:
> I think --O3-experimental is probably the only thing which would actually discourage people. Tons of users are using advi, for example, despite it doing all the stdout/doc warnings about being experimental

I changed the name to just `-Oexperimental` and made it clearer in the docs that this is not as well tested. Previously I had it as `-Od` where the `d` stood for dangerous.

WardBrian:
Thanks @SteveBronder 

Speaking of doc, the docs PR should be updated. The content can stay the same, but I think it should live in the new stanc3 chapter (https://mc-stan.org/docs/2_28/stan-users-guide/using-the-stan-compiler.html)

serban-nicusor-toptal:
@SteveBronder I'm on it

SteveBronder:
@serban-nicusor-toptal much appreciated!

SteveBronder:
@WardBrian just put up https://github.com/stan-dev/docs/pull/435! I'm not sure if I like the wording of `-O0` since in #1029 is going to turn back on the NaN thing. How do you feel about rewording it to something like 

```
- -O0 (Default) Apply only basic optimizations.
- -O1 Apply level 1 compiler optimizations.
```

Looking at gcc and clangs `-help` file they don't even talk about `-O` settings! Though I like [gcc docs](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html) where they list out the different optims at each level. Though I think we should do that once we update to 4.12 then fixup the args so it's easier to parse and add something like the `-f{no}-{optim_name}`

WardBrian:
I agree we will need to expand the documentation to include a description of each. Ryan's suggestions [here](https://github.com/rybern/optimization-docs/blob/master/optimization-docs.md) still look good to me, though I think some optimizations have been added since?

I think the 4.12 PR should be ready today, I'm just doing one last local clean install of everything to make sure it all builds as it should and then we can update our docker CI and merge it

SteveBronder:
@rybern @WardBrian I updated the docs PR to reflect this PR, so I think once #1079 is in we can merge this one?

https://github.com/stan-dev/docs/pull/262

rok-cesnovar:
I can review the docs PR over the next few days.

WardBrian:
I’m not going to be active until the new year most likely, will review things when I get back if still needed

rok-cesnovar:
I just went in and reviewed it, looks good to me, so I think we are good.

SteveBronder:
This actually has #1079 merged into it and @rybern already approved #1079 so if we want we can just merge this PR

PR #552, opened 2020-05-27
array-valued transform arguments

### LINKED ISSUE #547: vector valued transform arguments
Cases come up where a different lower and/or upper bound is needed for every element of a vector of array.  It'd be nice to be able to write:

```
real L[K];
real U[K];
vector<lower = L, upper = U>[K] alpha;
```

It would also be possible to allow the containers to be vectors or row vectors.  The bounds should allow mixing of scalars and vectors in the same way as probability functions---the scalar argument is broadcast to the size of the container argument.  If there are multiple container arguments, their size must match and must be consistent with the size of the container being declared.

nhuurre:
Fix #547 
Non-scalar values are now allowed in `lower`, `upper`, `offset` and `multiplier` transforms if the value has the same type and size as the declared variable.

Note that container bounds check error messages have always been kind of bogus.
```stan
parameters {
  real p[5,5];
}
transformed parameters {
  real<lower=0> t[5,5] = p;
}
```
```
Rejecting initial value:
  Error evaluating the log probability at the initial value.
Exception: example_model_namespace::log_prob: t[sym1__, sym2__] is -0.574295,
  but must be greater than or equal to 0 (in 'example.stan', line 5, column 2 to column 27)
```
The opaque indexing is being fixed in math (https://github.com/stan-dev/math/issues/1635) but it doesn't support array-valued bounds.


bob-carpenter:
I don't quite feel prepared to review the OCaml, but this is huge!  We've been wanting to do this for ages.  

We've been improving indexing into error messages every place we can, so that'll also be great if that can be fixed.

rok-cesnovar:
With develop cmdstan you can set this in make/local to test this:

STANC3_TEST_BIN_URL = https://jenkins.mc-stan.org/job/stanc3-test-binaries/106/artifact

rok-cesnovar:
@nhuurre can you merge master in? I can review this today.

wds15:
whow 10k lines of code!

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -723,6 +723,61 @@ std::vector<stan::math::var> to_vars__(std::initializer_list<stan::math::var> x)
   return x;
 }
 
+template <typename T>
+void FnCheckShape__(const char* var_name, size_t depth,
+                    const T& obj) {
+}
+
+template <typename T>
+void FnCheckShape__(const char* var_name, size_t depth,
```
rok-cesnovar, line 732: Could we move this functions to the stan-dev/stan repo? Not sure of the best location but I guess under /model/check.hpp and then add that header to the model_header? They would probably have to be named more in the style of math / stan C++ conventions we use. So something like `check_shape()`.

The other functions that dont depend on model specific should probably also go there but that is not this PRs fault and most of them can actually just be removed.

src/frontend/Semantic_error.ml
```diff
@@ -55,6 +56,12 @@ module TypeError = struct
     | IntOrRealExpected (name, ut) ->
         Fmt.pf ppf "%s must be of type int or real. Instead found type %a."
           name UnsizedType.pp ut
+    | TypeExpected (name, (UInt | UReal), ut) ->
+        Fmt.pf ppf "%s must be of type int or real. Instead found type %a."
```
rok-cesnovar, line 60: must be a scalar to make it shorter and consistent with the others?

src/stan_math_backend/Stan_math_code_gen.ml
nhuurre: Yeah, I was definitely being lazy. I think this actually belongs in the math library. FnCheckShape__ is a generalization of [check_consistent_size](https://github.com/stan-dev/math/blob/develop/stan/math/prim/err/check_consistent_size.hpp). Alternatively it could be replaced by [check_matching_dims](https://github.com/stan-dev/math/blob/develop/stan/math/prim/err/check_matching_dims.hpp) if that recursed into arbitrary arrays.

src/frontend/Semantic_error.ml
nhuurre: Agreed , "scalar" is better.

src/stan_math_backend/Stan_math_code_gen.ml
rok-cesnovar: I am fine with either of the proposed solutions.

src/frontend/Semantic_check.ml
```diff
@@ -755,6 +747,16 @@ and semantic_check_expression_of_int_or_real_type cf e name =
       Semantic_error.int_or_real_expected ue.emeta.loc name ue.emeta.type_
       |> error)
 
+let semantic_check_expression_of_scalar_or_type cf t e name =
```
rok-cesnovar, line 750: Minor thing, not sure the or is necessary. I guess a replace all error? :)

nhuurre: The "or" _is_ necessary because it's not "scalar type", it's "_either_ scalar _or_ the given type".

I don't like the `scalar_or_type` name but I'm not sure what is a good alternative.

rok-cesnovar: Ah, makes sense.

rok-cesnovar:
Thanks!

nhuurre:
Hey, @rok-cesnovar ,just curious, why do you squash the history when merging these?
When I try to delete the branch locally git thinks it hasn't been merged.

rok-cesnovar:
No reason, its the default option I think.

nhuurre:
I don't think that's a good default. It doesn't matter for small fixes like this PR but optimization debugging was a [sequence](https://github.com/stan-dev/stanc3/pull/522/commits/6f312535c657687c75f802636b8d750fdfd1e817) of small [changes](https://github.com/stan-dev/stanc3/pull/522/commits/0ccd903e339d5a643a0d03ecc406e63e50b9d9d6) with associated [explanations](https://github.com/stan-dev/stanc3/pull/522/commits/5fca414cb9f38ab11f470351f067446efc1feb02) and now all of that is a [single huge blob](https://github.com/stan-dev/stanc3/commit/c60d3c6a00c60e9fcb4ba87e888616de484cb65c). Good luck figuring out which parts of the code those bullet points refer to.

bob-carpenter:
I don't think it is the default.  At least it wasn't when I set things up.  We had a long go around on this ages ago and decided to keep all the commit and merge history bubbles.  If you are going to squash history, whatever you do, don't rewrite it on the origin (GitHub)---only collapse commits locally before pushing to origin.

rok-cesnovar:
stanc3 has these three options on:

Allow merge commits 
Add all commits from the head branch to the base branch with a merge commit.

Allow squash merging 
Combine all commits from the head branch into a single commit in the base branch.

Allow rebase merging 
Add all commits from the head branch onto the base branch individually.

Should we just allow the first option then?

nhuurre:
Yes, I think the first is the preferred action.

rok-cesnovar:
Done.

Issue #93, opened 2019-04-03
Store comments on statements in AST for pretty printer


VMatthijs:
During lexing and parsing attach comments to the metadata of the closest statement, so we do not need to throw them away during pretty printing.

adamhaber:
I'll try... will probably need some help on how to get started, still new to all this. :-)

VMatthijs:
Fantastic! The challenge here is on how to work within the restrictions that ocamllex and menhir impose. Happy to discuss and help out if you get stuck! :-)

adamhaber:
Great. I'd probably need some more information to get started - I don't have any experience with writing compilers, and still learning most of the terminology. Is this change relevant to a specific file, or several files? Which parts of them are to be modified?

seantalts:
Hey Adam, not sure if anyone has the time, but if you're interested in doing some self-learning here there's a ton of material online. This issue takes place in the parser (frontend/parser.mly, which is a special file read by Menhir). A compiler is split into a few phases, typically lexing, parsing, and semantic checking are the first three. Lexing produces tokens and parsing produces an AST, which is the big complex tree that represents an entire Stan program defined in frontend/Ast.ml. 
I'd look at these resources:
https://dev.realworldocaml.org/parsing-with-ocamllex-and-menhir.html
https://eu.udacity.com/course/programming-languages--cs262 -- taught by one of my favorite professors from undergrad.

adamhaber:
Hi! I'm trying to go back to this and I have a couple of questions - trying to understand if I'm in the right direction... 
1. Is it correct that in `parser.mly`, `smeta` saves the location of the current statement, and should also save comments, if they exists?
2. Can these comments be captured by pattern matching against (atmoic/nested) statements?
3. Will this require adding additional tokens (`COMMENT_START`, `<string> COMMENT` etc)?

seantalts:
I think this will require additional tokens, `smeta` saves the location and should be updated to be a record with both a location and an optional comment, and I'm not sure what question 2 means. Hopefully @VMatthijs can pop in and help you get set up; he marked this as `good-first-issue` but I'm not sure the best way to do it in the parser without upsetting the custom error messages.

PR #560, opened 2020-06-02
Allow array sizes after type in declarations

rybern:
Declarations of the form:
```
type[n,..] identifier;
type[n,..] identifier = value;
```
are now parsed, and have the same semantics as if the `[n,..]` were after the identifier. For example:
```
int[2] a;
vector[3][3,2] b = ..;
matrix[3,3][5,4] c = ..;
```
is now equivalent to:
```
int a[2];
vector[3] b[3,2] = ..;
matrix[3,3] c[5,4] = ..;
```

Let me know if you come up with more test cases.

Paging @bob-carpenter


rok-cesnovar:
@nhuurre would you be comfortable reviewing this?

nhuurre:
Sure, I'll give it a try.

src/frontend/parser.mly
```diff
@@ -246,10 +258,22 @@ sized_basic_type:
...
 
 top_var_decl_no_assign:
-  | tvt=top_var_type id=decl_identifier d=option(dims) SEMICOLON
+  | tvt=top_var_type id=decl_identifier SEMICOLON
+    {
+      grammar_logger "top_var_decl_no_assign" ;
+      {stmt=
+         VarDecl {decl_type= Sized (fst tvt);
+                  transformation=  snd tvt;
+                  identifier= id;
```
nhuurre, line 267: I was going to complain about the formatting (two spaces before `snd`, semicolons at the end of lines) but `ocamlformat` doesn't work on `.mly` files and the formatting is messed up a bit all over the file so whatever, this is good enough.

test/integration/good/pretty.expected
```diff
@@ -1157,221 +1157,355 @@ data {
...
   real b0;
   real b1[4];
   real b2[5, 6];
+  real b3[4];
+  real b4[5, 6];
   vector[7] c0;
   vector[7] c1[8];
   vector[7] c2[8, 9];
+  vector[7] c3[8];
+  vector[7] c4[8, 9];
```
nhuurre, line 1171: So the pretty printer reverts code to the old syntax. That seems odd; isn't the new syntax meant to be a better replacement for the old? If so, the printer should output should reflect that.

Or maybe the AST should remember which syntax was used and pretty printer preserves it. I'd like to keep syntax information in the AST. The canonicalizer can change it to the "preferred" form.

src/frontend/parser.mly
rybern: Yep, fair. Also this is fixed in #561

test/integration/good/pretty.expected
rybern: Agreed. My plan was to make that change in a future PR, but I could easily add it if that's better

nhuurre: The pretty printer can wait. I think the AST change would belong to this PR.

rybern:
Thanks @nhuurre!

> The change I'm requesting is that a user who types
> 
> ```stan
> real[2] y[2];
> real[2][2] z;
> ```
> 
> needs a meaningful error that explains what they did wrong.

Good catch, I'll see what I can do. It's sometimes quite difficult to differentiate erroneous parser states from each other. I would personally prefer if this syntax were allowed but that's a bigger change.

> I also have a question about the syntax. The declaration has the matrix dimensions before array dimensions. This is backwards compared to accessing the elements.
> 
> ```stan
> matrix[K,L][N,M] mat;
> print(mat[n,m,k,l]);
> ```
> 
> I find that a bit counter-intuitive. Some new users have mixed up the order even with the current syntax where the dimensions are clearly separated.

I agree that `mat[n, m, k, l]` is counter-intuitive in this case, but I think that `mat[n, m][k, l]` is correct, and I think it's reasonable (and possibly necessary) for the former syntax to match the latter. (I say the latter is correct because for any value `a` of type `T[]`, I expect `a[n]` to be of type `T`, so `mat[n, m]` should be of type matrix.)

> Has anyone considered unified size declaration like
> 
> ```stan
> matrix[N,M|K,L] mat;
> ```
> 
> where `|` separates the matrix dimensions from the array-like dimensions and the dimensions are in the same order as they will be when indexing the object.

I personally don't find that syntax easier, but I'd understand if it were easier for the new users you mentioned.

test/integration/good/pretty.expected
rybern: Is the AST currently sensitive to style anywhere? It currently throws out e.g. whitespace and comments. IMO, preserving code style is a pretty significant change in the goal of the AST, and this would be the only instance AFAIK. Do you have an important use case in mind?

rybern:
On the parser message, I can only add one message for all unexpected cases. I could change the message for e.g. `real [2] y[2];` to something like "Expected assignment or semicolon after identifier in declaration (a second array index is not allowed).", but that message would be more confusing when triggered by e.g. `real [2] z` with a missing semicolon.

test/integration/good/pretty.expected
nhuurre: I don't know but [here](https://github.com/stan-dev/stanc3/issues/65#issuecomment-468362126) Matthis says the current AST is "full fidelity" and people discuss why that is a good thing. Throwing out comments is a [mistake](https://github.com/stan-dev/stanc3/issues/93). I consider it a critical bug that precludes the use of `--auto-format`.

nhuurre:
>I can only add one message for all unexpected cases. I could change the message for e.g. real [2] y[2]; to something like "Expected assignment or semicolon after identifier in declaration (a second array index is not allowed)."

Can't you just make the parser expect `real[2] y[2]` but then panic when it sees it?

Regarding the syntax options, I didn't see a stanc3 issue for this. Where's the design discussion? Is it part of the ragged arrays design-doc? Somewhere on Discourse?

test/integration/good/pretty.expected
rybern: I see what you're saying; you're considering it a bug that e.g. comments are thrown out, and this would make that bug worse by not including this element of style in the AST.

But I was misleading, this isn't a feature of style like whitespace or comments; we're planning on depreciating the old syntax, so really we want to always print out the new syntax. When we do that we won't need this information in the AST, it would become obsolete once the pretty printer changes are made, right?

rybern:
> Can't you just make the parser expect real[2] y[2] but then panic when it sees it?

Yep, that's essentially the trick I mentioned. I try to avoid it because it mangles the parse tree a bit. I can add it here if you think that's going to be a sufficiently common point of confusion.

> Regarding the syntax options, I didn't see a stanc3 issue for this. Where's the design discussion? Is it part of the ragged arrays design-doc? Somewhere on Discourse?

It's a necessary step preceding tuples; @bob-carpenter may know about a design discussion

bob-carpenter:
I definitely agree on meaningful errors for doubling up array syntax.

The order of stacked operators is an evergreen debate topic in algebraic circles because no ordering is optimal for everyting.  

I like `matrix[M, N][K1, ..., KJ]` because 

1.  it preserves the simple declaration of array syntax as `T[K1, ..., KJ]`, where here `T = matrix[M, N]`, and

2.  it preserves straightforward array indexing semantics, so that if `x` is an array of type `T[K1, ..., KJ]`, then `x[k1]` is of type `T[K2, ..., KJ]` if `k1 in 1:K1`.  

As @nhuurre points out, this produces a reversed index ordering of `K1, ..., KJ, M, N` for `matrix[M, N][K1, ..., KJ]`, which will definitely confuse some users.  The particular cases for concern are:

* `matrix[M, N][I, J]`,
* `vector[M][I]`,

and their constrained equivalent, because those won't raise compiler errors if the indexes are reversed.

I tend to lean toward the solution that's easiest to specify and hence document.  That usually leans toward generality.  But that's how we wound up with the general array issue in the first place where we have `real[]` because I just wanted to say that whenever we have a type `T`, we can have an array type `T[]` without having to say things like "except for `T = real`".

test/integration/good/pretty.expected
nhuurre: The problem with omitting comments is that `--auto-format` doesn't improve the readbility of a complex model--without comments it's even worse. I don't care about other syntax changes.

Note that  `GetLP` and `GetTarget` are exactly the same except that one is deprecated. The current AST considers them separate anyway. The pretty printer prints the AST faithfully; deprecated constructs are cleaned up by the canonicalizer. This division of labor makes sense to me.

nhuurre:
> Yep, that's essentially the trick I mentioned. I try to avoid it because it mangles the parse tree a bit. I can add it here if you think that's going to be a sufficiently common point of confusion.

You said you'd prefer to allow `real[N][K] x[M];`. I'd like design where the parser accepts it and semantic check rejects it. Of course that requires keeping at least some formatting info in the AST. 
The canonicalizer could even fix such "unambiguous but nonconforming" syntax.

test/integration/good/pretty.expected
rybern: Good point about the precedent with `GetLP` and `GetTarget`, especially because it was necessary to parse them differently in order to emit a depreciation warning. I agree it's necessary to put them in the AST for that reason at minimum. I'll add that!

rybern:
I don't know about extending the AST to include something that we don't support just to reject it, I think that's higher cost than explicitly rejecting it with the parser trick.

I'll make the changes you suggested @nhuurre - better error message and parse each syntax separately.

bob-carpenter:
This is off topic, but are omments not in the AST?  If so, should I create an issue to add them? That way, we can use the AST as the basis of a proper pretty printer for Stan programs.  I'd like to autoformat, say, all the programs in our example repos.

nhuurre:
> This is off topic, but are omments not in the AST? If so, should I create an issue to add them?

There's already an issue for it, #93 , but it's buried on page 5 (!!) of our issue tracker far behind many issues fixed or forgotten long ago. I looked into it last week but despite being marked as "good first issue" it's seems rather difficult.

rok-cesnovar:
Agreed, we need to sweep this issue tracker a bit.

nhuurre:
> The order of stacked operators is an evergreen debate topic in algebraic circles because no ordering is optimal for everyting.

I opened a poll on discourse to get more feedback on this
https://discourse.mc-stan.org/t/new-array-declaration-syntax/16011

rybern:
This is superseded by #669, which implements the `array[]` syntax

rybern:
Closing in favor of #669

PR #561, opened 2020-06-02
Parse multiple-identifier declarations and declare-defines; refactor declaration parsing

rybern:
This PR builds on #560 to extend our declaration syntax.

I'll keep this marked as WIP until I add more tests, but I thought I would open it for comment.

These are examples of multiple-identifier declarations that can now be parsed:
```
real x, y, z;
int[5] i, j;
```
The above is parsed equivalently to:
```
real x;
real y;
real z;
int i[5];
int j[5];
```
The following are **not** parsed (but could be pretty easily):
```
int x = 5, y = 6;
int i[2], j[3];
```
EDIT: `int x = 5, y = 6;` is now parsed as well.

In addition to the new syntax, I heavily refactored the declaration parsing code to be simpler and to avoid code replication.

Paging @bob-carpenter for comment


bob-carpenter:
Awesome.  I think it'd be ideal to allow declare/define everywhere.

```
int x = 5, y = 6;
```

but I wouldn't worry about this 

```
int i[2], j[3];
```

as we'll be deprecating the `int i[2]` form.  I realize there's no equivalent with prefix form

```
int[?] i, j;
```

but I think that's OK.  There won't be varying size vector or matrix declarations either.

rybern:
`int x = 5, y = 6;` is parsed now. I'll just add some tests.

rybern:
Added some tests, I think this is ready for review/merging

rybern:
Not sure who I should tag for this and #560 - @bob-carpenter to finalize the design, @seantalts 
 or @VMatthijs for the parser in case they're around?

bob-carpenter:
There are two things going on here:

1.  allow `T[N] x;` and deprecate `T x[N];` for array declarations.  

2.  allow `T x1 = y1, x2 = y2, ..., xN = yN;`  where each of the defines is optional.

We need (1) for delcaring tuple types like `(int[N], matrix[2, 3])`.  The most contentious part of this is that arrays of vectors now get declared as `vector[2][3, 4]` for a 3 x 4 array of 2-vectors.  

Change (2) is just a convenience to shorten vertical space in programs.

I'm very much in favor of both changes.

rok-cesnovar:
(1) is done in https://github.com/stan-dev/stanc3/pull/560 right?
and this PR addresses (2)?

rybern:
@rok-cesnovar That's right

seantalts:
Sorry for not responding earlier - is this ready for review? I can take a look :)

rok-cesnovar:
Sorry to barge in (not my PR) but yes, this is ready I think. 

#560 is on hold as the general feedback is that most would prefer 
```
array[N] real y;
array[2,3] vector[4];
```
over what is proposed there.

src/frontend/parser.mly
```diff
@@ -5,6 +5,16 @@ open Core_kernel
...
+(*
+
+     -Only top-level variable declarations are allowed in data and parameter blocks.
+                                                                   -
+                                                                   +(Parse error state 723)
+-Expected top-level variable declaration, statement or "}".
+-
++(Parse error state 775)
+
+ *)
```
seantalts, line 17: Was this comment meant to stay in?

src/frontend/parser.mly
```diff
@@ -33,6 +43,8 @@ let reducearray (sbt, l) =
 %token PRINT REJECT
 %token TRUNCATE
 %token EOF
+
+(* Unreachable tokens are useful for differentiating states *)
```
seantalts, line 47: Interesting, is there some online resource we could link to explaining why that is? I realize you probably weren't the person to add this but if you know of one and can add it that would be swell.

src/frontend/parser.mly
```diff
@@ -218,61 +230,108 @@ unsized_dims:
   | LBRACK cs=list(COMMA) RBRACK
     { grammar_logger "unsized_dims" ; List.length(cs) }
 
-(* declarations *)
+(* Never accept this rule, but return the same type as expression *)
+no_assign:
+  | UNREACHABLE
+    { (* This code will never be reached *)
```
seantalts, line 236: Is there any time when this could be executed? I didn't see anything in the lexer that generates this token. If there's nothing that could cause it to be executed, might be nicer to just raise here instead.

seantalts:
I'm taking a look now - lookin' good! This PR includes #560, right? If so, I think we should either
1. Get #560 code out of this PR and then review + merge this one first while we wait for a decision on #560, or 
2. Make a decision on #560, review, and merge that before reviewing this. 

Does that sound right? I'll submit the few comments I already wrote now.

src/frontend/parser.mly
rybern: Nope!

src/frontend/parser.mly
rybern: I was actually the person to add this in #457, and I don't know of an online resource. I explained it a bit in the PR but I could definitely add a paragraph comment here.

src/frontend/parser.mly
rybern: Yeah good point, this should raise an error.

rybern:
Thanks Sean! I was thinking we'd do option 2 and wait for #560, since there isn't a big rush on this feature and it's a bit work intensive especially to update the error messages.

rok-cesnovar:
Would we be able to do the currently most popular option `array[N] vector[M]` and still support the current syntax?

seantalts:
sounds good to me, please feel free to ping me on that one and/or this one when it's time :)

rybern:
This is superseded by #670, because that one uses the new `array` syntax. @seantalts if you're still up for reviewing, I think I'm in a good place

seantalts:
@rybern I'm down and should have some time soon! Does that mean we should close this one (#561) and I should review #670?

rybern:
@seantalts Awesome. Yeah I'll close #561, I was just leaving it open in case we decided on the old array syntax. Yeah, #670 should be first.

rybern:
Sorry, I meant that #669 should be first, since #670 depends on that. Closing these.

PR #564, opened 2020-06-03
use _lpdf/_lpmf rather than _log

### LINKED ISSUE #563: The sampling statement (~) should use *_lpdf, not _log
```stan
y ~ normal(0,1);
```
uses `normal_log` in the generated C++. 

Same for `y ~ beta`, etc. I didn't go and check, but I sense its the same for all.

This does not change anything for the user as normal_log just call normal_lpdf, but it would make some issues easier to handle (see discussion in https://github.com/stan-dev/stanc3/pull/562). It should be cleaned up anyways for when we finally drop support for `*_log`.

rok-cesnovar:
Fixes #563 

Uses _lpdf or _lpmf instead of _log in tilde statements. See issue for examples.

The suffices were listed as 
```ocaml
["_log"; "_lpmf"; "_lpdf"]
```
and stanc3 used the first one that existed. So for `y ~ normal` it basically looked for `normal_log`, `normal_lpmf` and `normal_lpdf`, then filtered the ones that actually exists and took the first one.

Changed the suffices list to 
```ocaml
["_lpmf"; "_lpdf"; "_log"]
```
and that is it. The rest changes are just for tests.

### DIFF
--- a/src/frontend/Ast_to_Mir.ml
+++ b/src/frontend/Ast_to_Mir.ml
@@ -463,7 +463,7 @@ let dist_name_suffix udf_names name =
 
 let%expect_test "dist name suffix" =
   dist_name_suffix [] "normal" |> print_endline ;
-  [%expect {| _log |}]
+  [%expect {| _lpdf |}]
 
 let rec trans_stmt ud_dists (declc : decl_context) (ts : Ast.typed_statement) =
   let stmt_typed = ts.stmt and smeta = ts.smeta.loc in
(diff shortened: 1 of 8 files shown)


rok-cesnovar:
@nhuurre or @rybern anyone of you has some time to review this?

nhuurre:
What about this?
https://github.com/stan-dev/stanc3/blob/546dce69f11fa6cf569a33563a76643a12310924/src/middle/Stan_math_signatures.ml#L253
That line and `is_distribution_name` and `is_propto_distribution` are just duplicates of their `Utils`versions.

rok-cesnovar:
I am cleaning that up in https://github.com/stan-dev/stanc3/pull/562 but I can move it here. Yep there were quite a few duplicates.

rok-cesnovar:
I kind of thought that it made sense to use the Utils one because that makes this not dependant on the backend which I guess would be better.

rok-cesnovar:
I cherry-picked the cleanup to this PR. It makes more sense to do it here, if we decide to not gor with th `_lu*` PR. Should be good now.

Issue #565, opened 2020-06-03
Signatures for vectorised binary functions


andrjohns:
I'm trying to add the signatures (or figure out the general approach) for vectorised binary functions (the framework in math is just about finished going through, so it's almost time for the signatures). But I'm running into a problem that I'm not familiar enough with OCaml to figure out.

Vectorised binary functions work in a very similar way to the vectorised unary functions, except that both arguments can now be arrays/vectors/matrices of the same dimension, or one argument can be an array/vector/matrix and the other a scalar.

My current approach is to add a type for just ```int``` and ```real```:
```
let rec expand_arg = function
  ...
  | DVPrimitive -> [UnsizedType.UReal; UInt]
  ...
```

Then to use the existing structures for vectorised unary functions:
```
let math_sigs =
  ...
  
  ; ([UnaryVectorized], "beta", [DDeepVectorized; DVPrimitive]) 
  ; ([UnaryVectorized], "beta", [DVPrimitive; DDeepVectorized]) ]
```
Which gives the signatures for the various combinations of int/real and container.

However, when I try to add the signatures for two containers of the same type (e.g. ```beta(matrix[,], matrix[,])```):
```
  ; ([UnaryVectorized], "beta", [DDeepVectorized; DDeepVectorized]) 
```
The combinatorics pull out possible combinations:
```
...
(matrix[,,,,,,], int[,,,,,,]) => matrix[,,,,,,]
(matrix[,,,,,,], real[,,,,,,]) => matrix[,,,,,,]
(matrix[,,,,,,], vector[,,,,,,]) => matrix[,,,,,,]
(matrix[,,,,,,], row_vector[,,,,,,]) => matrix[,,,,,,]
(matrix[,,,,,,], matrix[,,,,,,]) => matrix[,,,,,,]
```

Is there a way to declare a return type (i.e. BinaryVectorised, or the like) that doesn't extract all binary combinations of ```DDeepVectorized```?

andrjohns:
Alternatively, is there a wholly different approach that's more appropriate here?

rok-cesnovar:
You can do this:

```ocaml
for i = 0 to 7 do
    for j = 0 to bare_types_size - 1 do
      add_unqualified
        ("something", ReturnType (bare_array_type (bare_types j, i)), [(bare_array_type (bare_types j, i)); (bare_array_type (bare_types j, i))]);
      add_unqualified
[...]

PR #569, opened 2020-06-04
Allow variable names that shadow math function names

### LINKED ISSUE #567: Allow variable names that match math function names
Currently, Stan variables are allowed to use names that match nullary functions from the Math library. So for example

```stan
data {
  real pi;
}
```
is fine, but the following is not:

```stan
data {
  real logit;
}
```
Also, see discussion in #565

rok-cesnovar:
Fixes #567 

The following model now compiles (I labeled the newly accepted stuff):
```stan
parameters {
  real e;
  real pi;
  real log2;
  real log10;
  real sqrt2;
  real not_a_number;
  real positive_infinity;
  real negative_infinity;
  real machine_precision;
  real inv_logit;   // NOW SUPPORTED
  real logit;   // NOW SUPPORTED
  real bernoulli_logit_glm_lpmf;   // NOW SUPPORTED
  real reduce_sum;   // NOW SUPPORTED
}
transformed parameters {
  real mu;
  mu = e() + pi() + log2() + log10() + sqrt2() + not_a_number()
    + positive_infinity() + negative_infinity() + machine_precision();
  mu += logit + bernoulli_logit_glm_lpmf + reduce_sum;   // NOW SUPPORTED
}
```

@bob-carpenter just checking if this is what you meant? I am pretty sure you did. User-defined functions remain restricted.


rok-cesnovar:
Any thoughts on this? Do we want to do this at all? 

@bob-carpenter @nhuurre @rybern @seantalts @VMatthijs

seantalts:
I haven't been in any of the discussions, so let me know if I'm missing anything. 

Is this allowed in Stan 2? If so seems like we should and I wouldn't expect much controversy. Need a reviewer?

rok-cesnovar:
Sorry, should have provided a bit more context. The relevant discussion starte from this https://github.com/stan-dev/stanc3/issues/565#issuecomment-638055261

Stanc2 allows shadowing for nullary functions and in identifiers of function arguments (which I found when adding stanc3 to /stan tests, I havent seen this documented). Stanc3 also allows shadowing for nullary functions.

The main problem right now would be beta(a,b) which would break a bunch of models if included. Even if we ignore beta this will likely come up again.

Allowing only nullary functions to shadow seems arbitrary. And Bob mentions that there is no reason to not allow this.

seantalts:
> The main problem right now would be beta(a,b) which would break a bunch of models if included. Even if we ignore beta this will likely come up again.

Why does beta break models if included? What's the current behavior? Do we currently sort of keep the namespaces separate depending on the type? e.g. if you define `real beta;` but then later call `beta(a, b);` the compiler just knows you mean the Math function and outputs `stan::math::beta` or something?

seantalts:
What happens with `beta` now? It kinda seems like we just don't check, so shadowing is allowed in a sense and when we actually emit code `beta()` and `beta` don't actually clash, because the former ends up being `beta_lpdf` or whatever. So we kinda have this separate namespace for distributions vs. other variables. I feel like, while it's not a hugely popular choice, certain languages do have separate namespaces for different kinds of variables. Like in OCaml you can have both a type and a variable with the same name and it just knows which one you mean every time.

rok-cesnovar:
If we would have a beta() function (not a distribution) any “beta” identifier would error that shadowing is not allowed. And beta is a very common identifier. 

We might not need the beta function but I just think this problem will eventually come up again.

seantalts:
I think we want this - let me try to merge master in and review!

PR #570, opened 2020-06-05
[WIP] Closures prototype

nhuurre:
Add closures to the Stan language. The math library implementation is https://github.com/stan-dev/math/pull/1924

See also the desing-doc https://github.com/stan-dev/design-docs/pull/6 and the forum discussion https://discourse.mc-stan.org/t/closures-discussion/14932/

A closure definition begins `functions` keyword and otherwise looks just like a function definition. The keyword isn't ideal but something is needed to avoid ambiguity with ordinary variable declarations.


rybern:
Hi Niko. You mentioned on Discourse that the parser isn't behaving. I just went through the gauntlet with the parser for declarations, so I thought I'd have a look.

The issue is a reduce/reduce conflict between `return_type` inside `closure_def` and `sized_basic_type` inside `var_decl`. You could `%inline return_type`, but then you run into the more fundamental issue of a reduce/reduce conflict between `basic_type` and `sized_basic_type`. Essentially, when the parser sees `int`..., it doesn't know whether to reduce the `int` to an unsized type (leading up to a declaration) or a sized type (leading up to a closure). It becomes unambiguous when the parentheses are read, but it needs to be unambiguous at the point of decision between sized and unsized.

I came up with a trick to solve a similar issue in #561 that I think I can apply here. But, since it'll take some work, I'd rather do it when you're stable here and when #561 is merged (since it included some refactoring).

nhuurre:
Closing in favor of #742

PR #573, opened 2020-06-11
integer division operator

### LINKED ISSUE #533: promote int to real, double-based divide function
#### to_real()

Add a special function:

```
real to_real(int);
```
- [ ] code
- [ ] tests
- [ ] doc in manual

Suggested implementation:

```
double to_real(int x) { return x; }
```

Would this clearer with a static cast?
#### div()

Add a special function

```
real div(real,real);
```
- [ ] code
- [ ] tests
- [ ] doc in manual

Suggested implementation:

```
double div(double x, double y) { return x / y; }
```
#### Motivation

The motivation is that I'm implementing some mark-recapture models that involve integer populations, but I need a fractional bound on a real parameter that's currently cumbersome to write:

```
data {
  int<lower=0> M;                       // marked
  int<lower=0> C;                       // captured
  int<lower=0,upper=min(M,C)> R;        // recaptured
}
transformed data { // HACK!
  real theta_max;     
  theta_max <- M;         
  theta_max <- theta_max / (C - R + M);
}
parameters {
  real<lower=0,upper=theta_max> theta;  // proportion marked
}
...
```

I would prefer to get rid of the transformed data division hack and just write:

```
parameters {
  real<lower=0,upper=div(M, C - R + M)> theta;  // proportion marked
}
```

Without the constraint, you get inconsistent values of theta that are above the upper bound.

nhuurre:
Fix #533 
Adds new operator `%/%` for integer division that rounds toward zero. The warning the ordinary division operator emits when applied to integers is now more informative.

The `IntDivide` operator is not implemented in the Stan Math Library  as such a function would be redundant in C++. Instead the compiler special-cases the operator when type-checking and turns it into an ordinary division in the backend.

### DIFF
--- a/src/analysis_and_optimization/Partial_evaluator.ml
+++ b/src/analysis_and_optimization/Partial_evaluator.ml
@@ -41,6 +41,7 @@ let apply_operator_int (op : string) i1 i2 =
         | "Minus__" -> i1 - i2
         | "Times__" -> i1 * i2
         | "Divide__" -> i1 / i2
+        | "IntDivide__" -> i1 / i2
         | "Modulo__" -> i1 % i2
         | "Equals__" -> Bool.to_int (i1 = i2)
         | "NEquals__" -> Bool.to_int (i1 <> i2)
@@ -509,6 +510,7 @@ let rec eval_expr (e : Expr.Typed.t) =
             | "pow", [x; {pattern= FunApp (StanLib, "Divide__", [y; z]); _}]
               when is_int 1 y && is_int 2 z ->
                 FunApp (StanLib, "sqrt", [x])
+                (* This is wrong; if both are type UInt the exponent is rounds down to zero. *)
             | "square", [{pattern= FunApp (StanLib, "sd", [x]); _}] ->
                 FunApp (StanLib, "variance", [x])
             | "sqrt", [x] when is_int 2 x -> FunApp (StanLib, "sqrt2", [])
@@ -558,7 +560,7 @@ let rec eval_expr (e : Expr.Typed.t) =
                 FunApp (StanLib, "expm1", l')
             | "Plus__", [{pattern= FunApp (StanLib, "Times__", [x; y]); _}; z]
              |"Plus__", [z; {pattern= FunApp (StanLib, "Times__", [x; y]); _}]
-               when not preserve_stability ->
+              when not preserve_stability ->
                 FunApp (StanLib, "fma", [x; y; z])
             | "Minus__", [x; {pattern= FunApp (StanLib, "gamma_p", l); _}]
               when is_int 1 x ->
@@ -593,7 +595,7 @@ let rec eval_expr (e : Expr.Typed.t) =
                 FunApp (StanLib, "matrix_exp_multiply", [a; b])
             | "Times__", [x; {pattern= FunApp (StanLib, "log", [y]); _}]
              |"Times__", [{pattern= FunApp (StanLib, "log", [y]); _}; x]
-               when not preserve_stability ->
+              when not preserve_stability ->
                 FunApp (StanLib, "lmultiply", [x; y])
             | ( "Times__"
               , [ {pattern= FunApp (StanLib, "diag_matrix", [v]); _}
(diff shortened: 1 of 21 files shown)


src/analysis_and_optimization/Partial_evaluator.ml
```diff
@@ -509,6 +510,7 @@ let rec eval_expr (e : Expr.Typed.t) =
             | "pow", [x; {pattern= FunApp (StanLib, "Divide__", [y; z]); _}]
               when is_int 1 y && is_int 2 z ->
                 FunApp (StanLib, "sqrt", [x])
+                (* This is wrong; if both are type UInt the exponent is rounds down to zero. *)
```
rok-cesnovar, line 513: So if someone does `pow(x, 1/2)` that would technically have to return 1, but we fix that and replace with sqrt(x). I feel what we do is fine, its just not consistent.

src/frontend/parser.mly
```diff
@@ -464,6 +464,8 @@ common_expression:
     {  grammar_logger "infix_times" ; Operator.Times }
   | DIVIDE
     {  grammar_logger "infix_divide" ; Operator.Divide }
+  | IDIVIDE
+    {  grammar_logger "infix_divide" ; Operator.IntDivide }
```
rok-cesnovar, line 468:
```suggestion
    {  grammar_logger "infix_int_divide" ; Operator.IntDivide }
```

src/analysis_and_optimization/Partial_evaluator.ml
nhuurre: The result of `pow(x, 1/2)` depends on optimization level. At levels O2 or higher the partial evaluator changes it to `sqrt(x)`. At lower optimization level the partial evaluator does not run and it is up to the backend to decide what to do with `1/2`. C++ rounds it to zero, while Python (TFP) does proper division and gets `0.5`.
I agree this "inconsistency" is fine. That's why I didn't change the code, just added a comment.

rok-cesnovar: But `pow(x, 1/2)` will still warn right?

nhuurre: Yes, the warning is always there.

bob-carpenter:
This is just defining `%/%` as a synonym for `/` that doesn't raise a warning, right?  Is there going to be another division for int / int that returns a real value, or are we just going to recommend the hack of multiplying by 1.0?

I was thinking that the warning for int / int is going to go into pedantic mode.  At least I hope it goes into pedantic mode, since it currently issues false positives.

rok-cesnovar:
> This is just defining `%/%` as a synonym for `/` that doesn't raise a warning, right?

Yes.

> Is there going to be another division for int / int that returns a real value

We can add another division or wait for the major version bump and define that int / int returns a real value. 
Would you prefer the former? Any ideas on the operator if so. I might be more in favor of the latter, but no strong opinion.

> At least I hope it goes into pedantic mode, since it currently issues false positives.

What false positives does it issue? There was this issue: https://github.com/stan-dev/stanc3/issues/504 but that was fixed.
Did you notice others?

nhuurre:
The new operator %/% is indeed synonymous with /. The *only* functionality it adds is the ability to remove "false positive" int division warnings. The warning should stay always on.

I'm not aware of any languages that have a special "real division" operator for integers. Python and R just use the regular / for that. So I'm in favor of just recommending the hack until the version bump.

bob-carpenter:
Not so much false positives as it always gives a warning when there is int divided by int, right?  Sometimes that's intentional, so it's a kind of false positive warning that users who meant int division shouldn't have to see.  

I also don't know of languages with special real division operators; C++ just has you cast.  I see Python 2 used / for integer division just like Stan and C, but then in Python 3, they switched / to be real division and // to be "floor" division as they call it (which rounds down, not toward zero).  MATLAB also just does real division and then provides `idivide` for integer division rounding toward zero.  

Looks like we're heading down the same route as Python.  Just wish we could find a better symbol than `%/%`.  I'd actually prefer `idivide(i, j)` or just overloading elementwise division to scalars, `./`.

PR #588, opened 2020-06-23
Add a check that the Ocaml code is formatted

### LINKED ISSUE #587: Enforce formatting on PRs
How would everyone feel on enforcing formatting on PRs?

I am certainly guilty of not running `make format` on my PRs. We could add this to the CI in two different ways:

a) fail if running make format produces any changes (the PR needs to be formatted)
b) push make format changes automatically by the CI

Option a) is probably a bit easier to implement and less involved.

rok-cesnovar:
Fixes #587

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -70,8 +70,9 @@ pipeline {
                     then
                         echo "The source code was not formatted. Please run 'make format | dune promote' and push the changes."
                         echo "Also consider installing a pre-commit git hook for formatting with the above command."
+                        exit 1;
                     fi
-                    """)
+                """)
             }
             post { always { runShell("rm -rf ./*") }}
         }        
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -73,6 +73,8 @@ pipeline {
                         exit 1;
                     fi
                 """)
+                sh 'echo $error'
+
             }
             post { always { runShell("rm -rf ./*") }}
         }        
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -69,7 +69,7 @@ pipeline {
                     (
                         set +x &&
                         echo "The source code was not formatted. Please run 'make format | dune promote' and push the changes." &&
-                        echo "Also consider installing a pre-commit git hook for formatting with the above command." &&
+                        echo "Please consider installing a pre-commit git hook for formatting with the above command." &&
                         exit 1;
                     )
                 """
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -65,11 +65,13 @@ pipeline {
                 sh 'printenv'
                 runShell("""
                     eval \$(opam env)
-                    make format || 
-                    (echo "The source code was not formatted." &&
-                     echo "Please run 'make format | dune promote' and push the changes." &&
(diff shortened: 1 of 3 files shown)


rok-cesnovar:
This is ready.

After a deliberate formatting error:
https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3/detail/PR-588/4/pipeline/

After fixing:
https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3/detail/PR-588/5/pipeline

Jenkinsfile
```diff
@@ -53,6 +53,23 @@ pipeline {
...
+                    filename 'docker/debian/Dockerfile'
+                    //Forces image to ignore entrypoint
+                    args "-u root --entrypoint=\'\'"
+                }
+            }
+            steps {
+                sh 'printenv'
+                runShell("""
+                    eval \$(opam env)
+                    make format
```
seantalts, line 68: Consider printing a message about:
1. That this is a formatting error that they can resolve with "make format"
2. They should install the git hook

rok-cesnovar: I did that initially, but make format stops this runShell. I can work around that though.

nhuurre:
Why does the formatting run before the tests? My thinking is that it is "just another test" and maybe even one that doesn't matter if there are other failures.

Jenkinsfile
seantalts: Can you use `make format || echo "<message>"`?

seantalts:
> Why does the formatting run before the tests? My thinking is that it is "just another test" and maybe even one that doesn't matter if there are other failures.

My theory of testing with resource constraints is that you run the cheap tests first and fail early to conserve compute resources - the person is going to have to resubmit and run all tests anyway. Though if we were comfortable with the amount of resources we have then I could see preferring to run all tests and report all failures at once.

nhuurre:
Makes sense. I was writing from the point of view of developer convenience but I have no idea how resource-constrained the testing is.

rok-cesnovar:
> I have no idea how resource-constrained the testing is.

stanc3 on its own isn't too demanding (if you don't run "compile all good integration models"), but it shares Jenkins workers with other stan repos (mainly math/stan/cmdstan) and especially the former two are very resource consuming.

Jenkinsfile
rok-cesnovar: Took me way too many tries but this message is now printed.

The source code was not formatted. Run 'make format | dune promote' and push the changes.
Please consider installing a pre-commit git hook for formatting with the above command.

https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3/detail/PR-588/14/pipeline

Copy-edit on the message welcome as always.

seantalts:
LGTM. I made one minor change in the message - I don't think we want to pipe the results of make format into dune promote so I changed it to `;`. I also don't have a real opinion on the resource issue, could see it going either way.

rok-cesnovar:
I think we should keep this order. The 'make format; dune promote' is easy to run. If this test demanded more manual labor it would make sense to see if all tests pass before wasting an hour on formatting. But this takes a few seconds.

Thanks!

Issue #407, opened 2019-12-24
TFP backend - LKJ parameter shapes


adamhaber:
The following program:
```stan
data {
  int K;
}
parameters {
  cholesky_factor_corr[K] L;
}
model {
   L ~ lkj_corr_cholesky(2.0);
}
```
Produces the following parameter_shapes method:
```python
def parameter_shapes(self, nchains__):
    K = self.K
    return [(nchains__, K * K - tf__.cast(1, tf__.float64) / tf__.cast(2, tf__.float64))]
```
I assume this should be K * (K-1) / 2?

nhuurre:
Looks like the code gen adds casts too often. We could change this
https://github.com/stan-dev/stanc3/blob/c63095d306f45b97364d985950b8cfd62509a7c9/src/tfp_backend/Code_gen.ml#L14-L17
to cast only `Lit (Real, s)` and not `Lit (Int, s)`. I don't know what side effects that has. The problem is that one can write an integer literal where the code expects a real number. I assume we need a cast in that case. Maybe we need to first implement stan-dev/stanc3#373.

adamhaber:
You're right - it seems like stan-dev/stanc3#411 fixed the parenthesis issue, and now there's just the extra casts.

adamhaber:
I guess we can close this one as well? I don't think the `tf__.cast` is actually problematic, but I might be missing something. Perhaps we're OK with the auto-casting but we prefer to cast to int instead?

brianwa84:
Presumably parameter_shapes is used to initialize a chain as the shape argument to something like `tf.random.uniform(shape=param_shp, minval=-2, maxval=2.)` In this case float tensors won't work. You could consider using `_default_event_space_bijector()` on `CholeskyLKJ` (and others) to pull back to an unconstrained shape.

Something like:
```python
nchains = 7
@tf.function
def unconstrained_shape():
  d = tfp.distributions.CholeskyLKJ(5, tf.cast(.5, tf.float64))
  return d._default_event_space_bijector().inverse(d.sample(nchains))
f.get_concrete_function().structured_outputs
```
(Using `tf.function`/`get_concrete_function` avoids actually materializing a possibly big sample, and just traces the execution.)

adamhaber:
> Presumably parameter_shapes is used to initialize a chain as the shape argument to something like tf.random.uniform(shape=param_shp, minval=-2, maxval=2.) In this case float tensors won't work. 

[...]

PR #592, opened 2020-06-25
cast param shapes to tf.int32

adamhaber:
Following [this](https://github.com/stan-dev/stanc3/issues/407#issuecomment-647777378), parameter shapes are cast to tf.int32 so they would be valid shape arguments for chain initialisation.

### DIFF
--- a/src/tfp_backend/Code_gen.ml
+++ b/src/tfp_backend/Code_gen.ml
@@ -193,8 +193,11 @@ let get_params p =
 
 let pp_shapes ppf p =
   let pp_shape ppf (_, {Program.out_unconstrained_st; _}) =
-    pf ppf "(nchains__, @[<hov>%a@])" (list ~sep:comma pp_expr)
-      (SizedType.get_dims out_unconstrained_st)
+    let shapes_lst = SizedType.get_dims out_unconstrained_st in
+    if List.length shapes_lst > 0 then
+      pf ppf "(nchains__, tf__.cast(@[<hov>%a@], tf__.int32))"
+        (list ~sep:comma pp_expr) shapes_lst
+    else pf ppf "(nchains__, )"
   in
   let ppbody ppf =
     pf ppf "%a@ " pp_extract_data p ;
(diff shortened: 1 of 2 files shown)


src/tfp_backend/Code_gen.ml
```diff
@@ -193,8 +193,11 @@ let get_params p =
 
 let pp_shapes ppf p =
   let pp_shape ppf (_, {Program.out_unconstrained_st; _}) =
-    pf ppf "(nchains__, @[<hov>%a@])" (list ~sep:comma pp_expr)
-      (SizedType.get_dims out_unconstrained_st)
+    let shapes_lst = SizedType.get_dims out_unconstrained_st in
+    if List.length shapes_lst > 0 then
+      pf ppf "(nchains__, tf__.cast(@[<hov>%a@], tf__.int32))"
+        (list ~sep:comma pp_expr) shapes_lst
+    else pf ppf "(nchains__, )"
```
seantalts, line 200: Is it possible for there to be more than one thing in `shapes_lst`? If so that would give `tf__.cast(J, K, L, tf__.int32)` which seems to not work for tf.cast.

src/tfp_backend/Code_gen.ml
```diff
@@ -193,8 +193,11 @@ let get_params p =
 
 let pp_shapes ppf p =
   let pp_shape ppf (_, {Program.out_unconstrained_st; _}) =
-    pf ppf "(nchains__, @[<hov>%a@])" (list ~sep:comma pp_expr)
-      (SizedType.get_dims out_unconstrained_st)
+    let shapes_lst = SizedType.get_dims out_unconstrained_st in
+    if List.length shapes_lst > 0 then
```
seantalts, line 197: I think another way to write a version that casts each shape individually is to keep the original but wrap pp_expr with code to wrap it in a tf.cast call, 

```ocaml
let cast_expr ppf e = pf ppf "tf__.cast(%a, tf__.int32)" pp_expr e in
pf ppf "(nchains__, @[<hov>%a@])" (list ~sep:comma cast_expr)
      (SizedType.get_dims out_unconstrained_st)
```

(not sure if it's worth putting a format box around the cast argument, didn't seem like it).

adamhaber: Thanks! That's certainly a better way to do this.

PR #595, opened 2020-06-26
Check for invalid literals

nhuurre:
Fix parser to deal with some bad literals.
* Line breaks in string literals mess up line numbering. Now disallowed. (stanc2 did allow a line break but ignored everything after it.)
* Integers larger than signed 32-bit would fail C++ compilation. Now disallowed.
* In contrast to integers, too-big reals simply overflow to infinity. Although stanc2 did not allow these they are currently allowed.

### DIFF
--- a/src/frontend/Semantic_check.ml
+++ b/src/frontend/Semantic_check.ml
@@ -678,10 +678,13 @@ and semantic_check_expression cf ({emeta; expr} : Ast.untyped_expression) :
   | Variable id ->
       semantic_check_variable cf emeta.loc id
       |> Validate.apply_const (semantic_check_identifier id)
-  | IntNumeral s ->
-      mk_typed_expression ~expr:(IntNumeral s) ~ad_level:DataOnly ~type_:UInt
-        ~loc:emeta.loc
-      |> Validate.ok
+  | IntNumeral s -> (
+    match int_of_string_opt s with
+    | Some i when i < 2_147_483_648 ->
+        mk_typed_expression ~expr:(IntNumeral s) ~ad_level:DataOnly ~type_:UInt
+          ~loc:emeta.loc
+        |> Validate.ok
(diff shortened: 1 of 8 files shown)


rok-cesnovar:
Nice! I will take a look soon. 

As we are already at literals, one thing that bothered me some time ago, 1e5 and such are handled as reals, right? 
So 1e100000000 would overflow to infinity?

nhuurre:
Yes.
```stan
transformed data {
  print("max: ",  1e100000000);
}
```
```
warning: magnitude of floating-point constant too large for type 'double'; maximum is
      1.7976931348623157E+308 [-Wliteral-range]
        stan_print(pstream__, 1e100000000);
                              ^
1 warning generated.
```
```
max: inf
```
Clang issues a warning but still compiles. Python (TFP) backend doesn't even care.

rok-cesnovar:
Ok thanks.

I got confused by this actually. which is a separate issue 

```
real[] integrate_ode_bdf(function ode, real[] initial_state, real initial_time, real[] times, real[] theta, data real[] x_r, data int[] x_i, data real rel_tol, data real abs_tol, dta int max_num_steps)`
```
but the stanc3 signature is

```
real[,] integrate_ode_bdf((real, real[], real[], data real[], data int[]) => real[],
                            real[], real, real[], real[], real[], int[],
                            real, real, real)
```
So 
```
y_hat = integrate_ode_bdf(sho, y0_d, t0, ts, theta_d, x, x_int, 1e-10, 1e-10, 10000.5);
```
compiles but fails at C++ because integrate_ode_bdf accepts an integer as the last arg. Same with stanc2.
This was probably a way to allow 1e8 as valid input.

rok-cesnovar:
I apologize, it does not fail with clang, it just prints a C++ warning, which is odd for the user.

Issue #577, opened 2020-06-16
stanc3 online demo for pedantic mode, canonicalizer, auto format, etc


rok-cesnovar:
Given that we build a javascript version of `stanc3` with each release, it would be very simple to make a static web page hosted on github that would enable demos of pedantic mode, canonicalizer (--print-canonical) and other options like that. The "compiling/transpiling/formatting/...." steps are done on the users end.

If we make Github Pages for this repository, that could be hosted under `https://mc-stan.org/stanc3/demo`

It probably does not make sense to add printing of C++.

Any thoughts?

wds15:
Cool !!!

bob-carpenter:
Sounds cool.  

I think printing the C++ would also be useful.

Does GitHub let you host servers or would you be plumbing in other servers?

rok-cesnovar:
This would use github pages like we do for hosting the website. Given that its only html and a javascript file its a static page that github can handle.

rok-cesnovar:
There is now a prototype available: https://rok-cesnovar.github.io/stanc3js-demo/index.html

It lacks any CSS styling to make it prettier but it can:
- print C++
- do a pedantic check
- list supported signatures 
- run auto formatter and canonicalizer
- show how stanc3 optimization changes the C++

WardBrian:
I think especially with #894 merged, it would be really cool to put this somewhere more official. It reminds me a lot of stuff like https://try.ocamlpro.com/

rok-cesnovar:
Yeah, we could probably place it somewhere like mc-stan.org/stanc3-try or -demo or something like that. Still need to pretty-up the interface, havent touched this in a while.

WardBrian:
I was impressed that it had kept up to date the whole time, but (of course!) it's just pulling the nightly stanjs from Github

rok-cesnovar:
I finally got back to fixing this. The link again https://rok-cesnovar.github.io/stanc3js-demo/

Fixes: 
- supports the new style of signatures so that the function search now works, as does linking to a specific function: https://rok-cesnovar.github.io/stanc3js-demo/signatures.html?function=fabs
- fixed pedantic check (previously it looked for stderr, it now looks for out.warnings)
- a bit nicer interface but not really :)

TODO: 
- add Stan syntax highlighting (this should be solved whenever we solve it for @spinkney's helpful Stan functions repo, I think with highlight.js?)
- link to Stan docs from the signatures subpage
- support modular canonicalizer (after #1058 )

[...]

PR #598, opened 2020-06-28
Add stanc info function to the model

rok-cesnovar:
Adds the following function to the model:

```cpp
std::vector<std::string> model_compile_info() const {
  std::vector<std::string> stanc_info;
  stanc_info.push_back("stanc_version = stanc3 66c38187 (Unix)");
  stanc_info.push_back("stancflags = --warn-pedantic");
  return stanc_info;
}
```

This will be used to add the compile info to the CSV and the standard output:

```
# init = 2 (Default)
# random
#   seed = 21139381 (Default)
# output
#   file = output.csv (Default)
#   diagnostic_file =  (Default)
#   refresh = 100 (Default)
# stanc_version = stanc3 66c38187 (Unix)
# stancflags = --warn-pedantic
```

### DIFF
--- a/src/stan_math_backend/Stan_math_code_gen.ml
+++ b/src/stan_math_backend/Stan_math_code_gen.ml
@@ -24,14 +24,13 @@ open Statement_gen
 
 let stanc_args_to_print =
   let sans_model_and_hpp_paths x =
-    not (String.is_suffix ~suffix:".stan" x || String.is_prefix ~prefix:"--o" x)
+    not String.(is_suffix ~suffix:".stan" x || is_prefix ~prefix:"--o" x)
   in
   (* Ignore the "--o" arg, the stan file and the binary name (bin/stanc). *)
-  let stanc_flags_sans_model_and_hpp_paths =
-    List.filter ~f:sans_model_and_hpp_paths
-      (List.tl_exn (Array.to_list Sys.argv))
-  in
-  String.concat ~sep:" " stanc_flags_sans_model_and_hpp_paths
+  Array.to_list Sys.argv
+  |> List.tl_exn
+  |> List.filter ~f:sans_model_and_hpp_paths
+  |> String.concat ~sep:" "
 
 let pp_unused = fmt "(void) %s;  // suppress unused var warning@ "
 
@@ -678,12 +677,12 @@ let pp_model ppf ({Program.prog_name; _} as p) =
 
   std::vector<std::string> model_compile_info() const {
     std::vector<std::string> stanc_info;
-    stanc_info.push_back("stanc_version = %s");
+    stanc_info.push_back("stanc_version = %%NAME%%3 %%VERSION%%");
     stanc_info.push_back("stancflags = %s");
     return stanc_info;
   }
   |}
-    "%%NAME%%3 %%VERSION%%" stanc_args_to_print ;
+    stanc_args_to_print ;
   pf ppf "@ %a@]@]@ };" pp_model_public p
 
 (** The C++ aliases needed for the model class*)
--- a/src/stan_math_backend/Stan_math_code_gen.ml
+++ b/src/stan_math_backend/Stan_math_code_gen.ml
@@ -27,8 +27,7 @@ let stanc_args_to_print =
     not String.(is_suffix ~suffix:".stan" x || is_prefix ~prefix:"--o" x)
   in
   (* Ignore the "--o" arg, the stan file and the binary name (bin/stanc). *)
-  Array.to_list Sys.argv
-  |> List.tl_exn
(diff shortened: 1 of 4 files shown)


src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -22,6 +22,17 @@ open Fmt
 open Expression_gen
 open Statement_gen
 
+let stanc_args_to_print =
+  let sans_model_and_hpp_paths x =
+    not (String.is_suffix ~suffix:".stan" x || String.is_prefix ~prefix:"--o" x)
```
seantalts, line 27: super minor, but fun OCaml trick:
```ocaml
    not String.(is_suffix ~suffix:".stan" x || is_prefix ~prefix:"--o" x)
```

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -662,6 +673,17 @@ let pp_model ppf ({Program.prog_name; _} as p) =
...
+    {|
+
+  std::vector<std::string> model_compile_info() const {
+    std::vector<std::string> stanc_info;
+    stanc_info.push_back("stanc_version = %s");
+    stanc_info.push_back("stancflags = %s");
+    return stanc_info;
+  }
+  |}
+    "%%NAME%%3 %%VERSION%%" stanc_args_to_print ;
```
seantalts, line 686: What happens if you put these two in the string literal itself? e.g. 

`    stanc_info.push_back("stanc_version = %%NAME%%3 %%VERSION%%");`

I would have thought that would work but maybe it makes it harder to read or something?

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -22,6 +22,17 @@ open Fmt
...
+let stanc_args_to_print =
+  let sans_model_and_hpp_paths x =
+    not (String.is_suffix ~suffix:".stan" x || String.is_prefix ~prefix:"--o" x)
+  in
+  (* Ignore the "--o" arg, the stan file and the binary name (bin/stanc). *)
+  let stanc_flags_sans_model_and_hpp_paths =
+    List.filter ~f:sans_model_and_hpp_paths
+      (List.tl_exn (Array.to_list Sys.argv))
+  in
+  String.concat ~sep:" " stanc_flags_sans_model_and_hpp_paths
```
seantalts, line 34: Another way to write this that you might already know about (but just in case)

```ocaml
Array.to_list Sys.argv
|> List.tl_exn
|> List.filter ~f:sans_model_and_hpp_paths
|> String.concat ~sep:" "
```

src/stan_math_backend/Stan_math_code_gen.ml
rok-cesnovar: Nice, thanks! Still discovering these little tricks as I go.

src/stan_math_backend/Stan_math_code_gen.ml
rok-cesnovar: Doh, yes. Thanks!

src/stan_math_backend/Stan_math_code_gen.ml
seantalts: (also TBC I think reasonable people can have different opinions on whether the above is more or less clear! Just pointing out an option)

rok-cesnovar: Honestly, I am still not too friendly with the pipe syntax so I usually don't use it until I am at the stage of "ok now it works let's try to make it nicer". I do prefer this.

src/stan_math_backend/Stan_math_code_gen.ml
rok-cesnovar: Do you know if this %%VERSION%% names work with stancjs? I tried it once but failed. Or is this only for building the stanc binary?

seantalts: Ah! If it's not working, I probably have a bug in the stancjs deployment code in Jenkins. The way the %%VERSION%% stuff works is very simple, you just run `dune promote` before building and it does the appropriate string substitutions.

rok-cesnovar: Ah ok, works locally now! Thanks!! And there is a `dune subst` in the Jenkinsfile so it should also work for the stuff built on Jenkins. Want to add args to stancjs for https://github.com/stan-dev/stanc3/issues/577 and rstan stuff in general and started with --version and bumped into that :)

rok-cesnovar:
Thanks @seantalts for the review and Ocaml sugar ideas!

PR #599, opened 2020-06-28
Feature/468 Use hyphens in all arguments

### LINKED ISSUE #468: consistent use of hyphens underscores for compiler option names

rok-cesnovar:
Fixes #468 by deprecating `--allow_undefined` and `--include_paths` in favor of `--allow-undefined` and `--include-paths`
to consistently use hyphens in all stanc arguments.

### DIFF
--- a/src/stanc/stanc.ml
+++ b/src/stanc/stanc.ml
@@ -111,19 +111,39 @@ let options =
     ; ( "--print-cpp"
       , Arg.Set print_model_cpp
       , " If set, output the generated C++ Stan model class to stdout." )
-    ; ( "--allow_undefined"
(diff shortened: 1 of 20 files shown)


bob-carpenter:
Thanks!  Outside of identifiers in programs where they are mandated, I don't like underscores!

PR #602, opened 2020-07-02
fix python power op string

adamhaber:
Change power op string from Stan's ^ to Python's **


adamhaber:
Failed on a docker build error - @rok-cesnovar any chance you could try to re-run the checks?

rok-cesnovar:
Done, passing now.

PR #603, opened 2020-07-03
Store comments in AST

nhuurre:
Here's a shot at issue #93

Comments are stored in statement metadata. Technically Stan allows comments _anywhere_ but I don't want to clutter the AST with comment holders everywhere so the lexer simply discards comments that aren't in "statement-position".


rok-cesnovar:
Nice!

Can you give an example of which comments would be discarded?

nhuurre:
This mess
```stan
// comments outside blocks are fine
functions {
  void // this function doesn't return anything
        foo/*whitespace?*/(real a, // can't document arguments
                           vector s ) { // first line in the function body
    real x[3] = { // the array loses the comments
      1, // one
      2, // two
      3  // three
    };
  
    if // why is there a comment here?
      (x[1 /* index */] < 0)
      // this branch won't happen
      print("never"); // ok, this one's fine
    else {
      // do nothing
    }
  }
}
// end of file works too
```
becomes
```stan
// comments outside blocks are fine
functions {
  void foo(real a, vector s) {
    // first line in the function body
    real x[3] = {1, 2, 3};
    if (x[1] < 0) 
      print("never");// ok, this one's fine
    else {
      // do nothing
    }
  }
}
// end of file works too
```

rok-cesnovar:
Thanks! That does seem fine to me. Not sure if I have the knowledge here to review.

@rybern would you have time to take a look?

rybern:
Cool! I'll have some time sporadically,

I'd like to get up to speed on the intended use cases and design.

The idea is that we need comments for auto-formatting and canonicalizing, and those are the only use cases, right? I'm not familiar with how we're going to do canonicalization, does that need to be an AST-to-AST transformation?

If we were only talking about auto-formatting, I would probably advocate for doing token-to-token transformation instead of parsing to the AST. This describes the approaches of parsing vs. not parsing: https://beza1e1.tuxen.de/articles/formatting_code.html

If we do have a compelling use case for comments in the AST, metadata seems like a reasonable way. But is it okay to silently drop non-statement comments like those in your example? Maybe we could at least log the comments that the tokenizer encountered when save_comments is off, and print them to warnings when auto-formatting/canonicalizing?

nhuurre:
Yes, `auto-format` and `--print-canonical` are the only use cases.

A typical canonicalizing transform is changing `normal_log(x, mu, sigma)` to `normal_lpdf(x| mu, sigma)`. Note how the first comma changes to a bar. That requires some context-awareness. Also, if #569 is merged then `normal_log` will be a valid name for a local variable.

The most difficult case is the division operator. Following C++, Stan's division operator `/` gives an integer result if both operands are integers. The implicit rounding is potentially confusing to users coming from Python or R so we recently added a new integer division operator `%/%` to make the rounding explicit. The substitution `/ -> %/%` is possible only after type-checking both operands.

Dropping things silently is a bad idea. Would it be enough to just print "Warning: some comments ignored" or should it print the location of every missing comment?

rybern:
Gotcha, does sound like AST.

> Dropping things silently is a bad idea. Would it be enough to just print "Warning: some comments ignored" or should it print the location of every missing comment?

Maybe a list of line numbers? Or at least the location of the first instance, so they get some direction for how to correct it.

rybern:
It looks like you have three ways you store comments in the AST:
1. Metadata on a statement stores a comment after that statement
2. Metadata on a Blank statement stores a comment that starts at the beginning of a block, before any other statement
3. Top level comments are stored in the program

Do I have that right?

In your example above, the comment "// ok, this one's fine" stays on the same line as its host statement. How does the AST remember that it's on the same line? If it doesn't or if that's not a necessary features, could we eliminate the metadata approach entirely and use only methods 2 and 3 above (maybe renaming Blank to Comment)? The advantage would be simplicity, the disadvantage would be that it couldn't be generalized to e.g. expression metadata.

I'll drop some inline comments as well but I won't have time to finish now..

src/frontend/Ast.ml
```diff
@@ -8,6 +8,9 @@ type identifier =
   {name: string; id_loc: Location_span.t sexp_opaque [@compare.ignore]}
 [@@deriving sexp, hash, compare]
 
+type comment = LineComment of string | MultiComment of string list
```
rybern, line 11: Did you consider making comments just list of strings and deciding the format upon output? That would also save us from needing lists of comments later on, but I understand it would lose some of the original style.

src/frontend/Ast.ml
```diff
@@ -137,6 +140,7 @@ type ('e, 's, 'l, 'f) statement =
   | Print of 'e printable list
   | Reject of 'e printable list
   | Skip
+  | Blank
```
rybern, line 143: I think this would be a helpful place to leave a comment explaining the meaning and usage of Blank

src/frontend/Ast.ml
```diff
@@ -202,19 +213,27 @@ type typed_statement =
...
+  ; comments3: comment list
   ; parametersblock: 's list option
+  ; comments4: comment list
   ; transformedparametersblock: 's list option
+  ; comments5: comment list
   ; modelblock: 's list option
-  ; generatedquantitiesblock: 's list option }
+  ; comments6: comment list
+  ; generatedquantitiesblock: 's list option
+  ; comments7: comment list }
```
rybern, line 236: I think the naming here could be more clear - what do you think about something like
```
header_comment
functionblock
functionblock_comment
datablock
datablock_comment
  ...
```

src/frontend/Pretty_printing.ml
```diff
@@ -385,21 +395,45 @@ let pp_opt_block ppf block_name opt_block =
...
+     ; transformedparametersblock= btp
+     ; comments5
+     ; modelblock= bm
+     ; comments6
+     ; generatedquantitiesblock= bgq
+     ; comments7 } :
+      untyped_program) =
   Format.pp_open_vbox ppf 0 ;
+  Fmt.list pp_comment ppf comments0 ;
+  if not (List.is_empty comments0) then Fmt.cut ppf () ;
```
rybern, line 416: Maybe we can pull out these two lines into pp_comments for clarity/clutter reduction?

src/frontend/Ast.ml
nhuurre: The naming definitely could be more clear. Those names are fine.

nhuurre:
> It looks like you have three ways you store comments in the AST:
> Do I have that right?

Yes, that's right.

> How does the AST remember that it's on the same line?

It doesn't remember. The first comment after a statement goes on the same line.

> could we eliminate the metadata approach entirely and use only methods 2 and 3 above (maybe renaming Blank to Comment)?

I considered placing all comments inside Blank/Comment statements but let's look at that example again
```stan
if (x[1] < 0)
  print("never"); // ok, this one's fine
else {
  // do nothing
}
```
There's no braces around the if-branch. It's not a block, just a single statement. The IfElse AST node has room for only one statement; often it's a Block (which can hold any number of statements) but in this case it's a Print with no room for additional Comment statements. Although I suppose always wrapping the branch in a block statement would be acceptable.

rybern:
> It doesn't remember. The first comment after a statement goes on the same line.

Will something like the following still work well enough?
```
statement1();

/* description of
   statement2 */
statement2();
```
IIUC turns into 
```
statement1(); /* description of
   statement2 */
statement2();
```

> There's no braces around the if-branch. It's not a block, just a single statement. The IfElse AST node has room for only one statement; often it's a Block (which can hold any number of statements) but in this case it's a Print with no room for additional Comment statements. Although I suppose always wrapping the branch in a block statement would be acceptable.

Ah I see, it makes a new context for parsing multiple statements. I do think a block statement would be best there (although it's not obvious to me how best to get something like `if(..) /*c1*/ print(); /*c2*/` to produce 3 statements). In my opinion, if we aren't worried about capturing more comment locations in the future, we should go with just the Blank/Comment statements + top-level comments to keep things as simple as possible.

rok-cesnovar:
Gentle bump on this. Would be nice to have this. Anything I could help with?

nhuurre:
Sorry, I abandoned this due to lack of a clear spec.
It's not always obvious what is the "nearest" statement or expression that the comment should be associated with.
This PR implements whatever I felt was easiest but I'm not really satisfied with it.

You can test this PR out and if it's adequate then I'll fix the merge conflicts and clean up the things Ryan pointed out above.

Some questions to consider:
1. Even before getting to comments, what to do about whitespace formatting:
```stan
parameters {
  real x;
  real y;

  real z;
  real w;
}
```
The empty line is probably meant to group related parameters together so it would make sense to preserve it.
2. Do we need comments outside of blocks?
```stan
/* Comment at the start of the file
   information relating to the entire model */
data {
  real x;
}
parameters {
  real y;
}
/* is this comment needed here?
   why isn't it inside the model block? */
model {
  x ~ normal(y,1);
}
```
3. Comments in places where a statement is expected are pretty simple but there's that `if (c) /* abc */ stmt();`
4. Comments inside expressions. I think most these are just weird clutter that is _technically_ allowed but doesn't help.
But unfortunately some do seem worth keeping and it's a mess.
Take this function signature for example
```stan
real[] dz_dt(real t,       // time
             real[] z,     // system state {prey, predator}
             real[] theta, // parameters
             real[] x_r,   // unused data
             int[] x_i) {
```
Logically the comment is associated with the expression _before_ the comma. Note that the simple way to pretty print that would produce
```stan
real[] dz_dt(real t /* time */,
             real[] z /* system state {prey, predator} */,
             real[] theta /* parameters */,
             real[] x_r /* unused data */,
             int[] x_i) {
```
Is that acceptable?
I'm not sure if printing the comments with the original clever alignment is even possible with the `Fmt` module we're using.

rok-cesnovar:
Thanks! Will try it out in the next few days.

> The empty line is probably meant to group related parameters together so it would make sense to preserve it.

If possible yes, this is also what clang-format does for C++, 1 line is fine, multiple empty lines get trimmed to a single line.

> Do we need comments outside of blocks?

I would say yes. For example starting comments (where the model was taken from, if it was generated using some Stan generating process, etc) are pretty common.

> if (c) /* abc */ stmt();

I would not be mind if these are ingored.

> 4. Is that acceptable?

To me this is nicer:

```stan
real[] dz_dt(real t /* time */,
             real[] z /* system state {prey, predator} */,
             real[] theta /* parameters */,
             real[] x_r /* unused data */,
             int[] x_i) {
```

seantalts:
What do you guys think about adding Newline and Comment variants to the
statement type? Would that solve the problem? The comments don’t need to be
represented as metadata on the AST. If we wanted to preserve them into the
MIR, that would probably be as metadata and we’d have to answer those
questions, but we don’t need to do that (yet) right?

WardBrian:
Closing due to #894 merge

Issue #418, opened 2020-01-04
Remove Stan fit files from tests


seantalts:
Copying the goals from discussion on #409,

Hmm. I think we could formulate these as cram tests (aka gold tests, or expect tests in OCaml). Then we can have a way to automatically generate the correct values from Stan and an automated test of the TFP backend against those values. How does that sound?

I don't want to be prescriptive here - I think some goals in maybe descending order of preference from my perspective would be:

1.    Easy to understand what is being tested and what is going wrong when a test failure occurs
1.    Easy to create new tests or update the tested values when something changes (though these are testing end-to-end stuff that isn't supposed to change except for bug fixes and major version number bumps)
1.    Has a small footprint
 1.   Separates code requiring the Stan toolchain from code requiring the TFP toolchain (not a huge priority but could be nice to have)

So it's possible that the best solutions to 1 and 2 have a larger footprint and that's okay, we can just organize around it.

bob-carpenter:
We've already polluted our math lib code with the Apache 2 license for the Intel TBB, so dependencies to TFP should be OK as long as they're stable.  

It'd be nice if we don't have to update anyhting other than the TFP generator.

adamhaber:
Resurfacing this. The current TFP integration tests are not ideal (see the discussions on stan-dev/stanc3#409 and stan-dev/stanc3#388 ), and I'd like to improve them. Specifically, removing the Stan fit files and automating the log_prob comparison would be great. Would be happy to hear your thoughts on this.

A question in this regard - given a .stan file that has K parameters, can we compute the lp__ of N predefined vectors in R^K (unconstrained) during testing?

PR #606, opened 2020-07-04
Add a test for tfp syntax errors

adamhaber:
Somewhat related to #602 , the idea is to make sure that all the generated TFP code is "valid python". An easy way to check this is to include a test which runs all the .py files using exec(). I'm sure there are better ways to do this, but maybe it's a good place to start?

### DIFF
--- a/test/integration/tfp/tests.py
+++ b/test/integration/tfp/tests.py
@@ -18,6 +18,7 @@ import numpy as np
 import tensorflow as tf
 import tensorflow_probability as tfp
 from os.path import dirname, join
+from glob import glob
 tfb = tfp.bijectors
 
 def significant_digit(x):
@@ -108,6 +109,13 @@ class TestModels(unittest.TestCase):
         self.assertTrue(np.all(theta_lb>=0))
         self.assertTrue(np.all(theta_ub<=1))
 
+    def test_synatx(self):
+        tfp_files = glob(join(dirname(__file__), "tfp_models/*.py"))
+        for file in tfp_files:
+            with open(file, "r") as handle:
+                current_file = handle.read()
+            current_file = current_file[current_file.find('import'):]
+            exec(current_file, {}, {})
 
 if __name__ == '__main__':
     unittest.main()


nhuurre:
First, if the idea is to test whether the generated code is syntactically valid wouldn't `compile()` be enough?

Secondly, `tests.py` already imports and executs all models except `nested_expr`, `python_kwrds` and `transformed_data`. If every model had a unit test (as they should) then a separate syntax test is redundant. Would it be possible to automatically generate these `_compare_lp()` tests?

adamhaber:
> First, if the idea is to test whether the generated code is syntactically valid wouldn't compile() be enough?

Thanks, that's a better idea. I'll implement it if we choose to go through with this PR.

> Secondly, tests.py already imports and executs all models except nested_expr, python_kwrds and transformed_data. If every model had a unit test (as they should) then a separate syntax test is redundant. Would it be possible to automatically generate these _compare_lp() tests?

The test_[disc/cont]\_* functions aren't really unit tests, in the sense that they don't test parameter_[shapes/bijectors/names]. More generally, this is related to #418; it's probably a good idea to rewrite most of the tfp integration tests, but I'm not sure what would be a good way to do so. We probably want to do something like: for Stan program P_1 and its generated TFP counterpart P_2, for N different (unconstrained) parameter vectors X_i, assert that P_1.lp(X_i)==P_2.lp(X_i).  Is there a straightforward way to compute P_1.lp(X_i)?

seantalts:
> for Stan program P_1 and its generated TFP counterpart P_2, for N different (unconstrained) parameter vectors X_i, assert that P_1.lp(X_i)==P_2.lp(X_i). Is there a straightforward way to compute P_1.lp(X_i)?

I worry that this test could end up being too strict - it's a bit difficult to compare numerical library implementations across languages, math libraries, compilers, etc. I think in theory it's not impossible, so might be worth a shot.

That said, I'd be fine with checking in some gold tests for P_1 in your example. One slightly annoying way to do that might be to just run any inference algorithm with X_i as the initial values for 1 step and then check the lp__ in the results.

adamhaber:
> I worry that this test could end up being too strict - it's a bit difficult to compare numerical library implementations across languages, math libraries, compilers, etc. I think in theory it's not impossible, so might be worth a shot.

This is roughly what's happening now with all the Stan fit files, and it works reasonably well; the places in which it failed were ones in which TFP and Stan actually implement things differently (like the bijector for CholeskyLK). However, it's a clunky way to set up tests, and adding/changing is quite annoying.

> That said, I'd be fine with checking in some gold tests for P_1 in your example. One slightly annoying way to do that might be to just run any inference algorithm with X_i as the initial values for 1 step and then check the lp__ in the results.

... and then compare to log_prob in P_2? I'm not sure I understand the difference from what we have right now with the stan fit files. Is it possible to get the lp__ of random initial values without running any inference at all?

seantalts:
Woops, yeah - I haven't had much time for Stan the past ~5 months or so until recently; I needed to go reload all of this back into my working memory :P Sorry about that.

I don't feel strongly about #418 at this point (despite having created the issue). I think the property that one can develop on the TFP backend without needing a working CmdStan or PyStan installation is a nice one and worth having some of these files checked in, assuming we can autogenerate the correct files(!).

Maybe we can take a step back and think about the pain points. Is this the main concern:
> However, it's a clunky way to set up tests, and adding/changing is quite annoying.

If so, maybe just adding some tooling that essentially re-creates all of these sample files given some CmdStan install location would solve the problem? Jenkins could even be programmed to re-generate on pull requests and check in the results so that someone could submit a PR with a new test model and have Jenkins fill in the .txt sample files.

> Is it possible to get the lp__ of random initial values without running any inference at all?

I thought running just 1 non-warmup iteration of HMC would give this to you? Might be off by 1. 

>  the places in which it failed were ones in which TFP and Stan actually implement things differently (like the bijector for CholeskyLK)

I think this is interesting; I guess it's true that we want to keep track of all of these differences and probably generate code around them if possible in an ideal world. On the other hand I wouldn't mind if we just said "Hey, we're generating the correct TFP code for this model and it's up to TFP to make sure those results are correct." That might mean pushing our tests much more towards gold tests that check that our TFP samples haven't changed with a given commit. 

Sorry, this is a bit of a brainstorming post. I'm very open minded towards what you all think is best here and would be happy to even close #418 as a bit of a misfeature.

WardBrian:
If this is still relevant it should be re-opened against https://github.com/stan-dev/stan2tfp/. Thanks!

Issue #609, opened 2020-07-05
Broken indexing in the transform_inits function


rok-cesnovar:
Another generate_quantities related issue.

```stan
parameters {
  matrix[1,2] p_ar_mat[3];
}
model {
    for (i in 1:3)
      for (k in 1:1)
        for (l in 1:2)
          p_ar_mat[i, k, l] ~ normal(0,1);
}
generated quantities {
  matrix[1,2] gq_ar_mat[3];
      for (i in 1:3)
        for (k in 1:1)
          for (l in 1:2)
            gq_ar_mat[i, k, l] = p_ar_mat[i, k,l];
}
```
When sampling, this model produces:
```
p_ar_mat.1.1.1,p_ar_mat.2.1.1,p_ar_mat.3.1.1,p_ar_mat.1.1.2,p_ar_mat.2.1.2,p_ar_mat.3.1.2,gq_ar_mat.1.1.1,gq_ar_mat.2.1.1,gq_ar_mat.3.1.1,gq_ar_mat.1.1.2,gq_ar_mat.2.1.2,gq_ar_mat.3.1.2
0.968647,0.666071,-1.08735,-1.59251,-1.8494,-0.563961,0.968647,0.666071,-1.08735,-1.59251,-1.8494,-0.563961
```
which is expected.

But if we then use these fitted params with standalone generated quantities we get:
```
gq_ar_mat.1.1.1,gq_ar_mat.2.1.1,gq_ar_mat.3.1.1,gq_ar_mat.1.1.2,gq_ar_mat.2.1.2,gq_ar_mat.3.1.2
0.968647,-1.08735,-1.8494,0.666071,-1.59251,-0.563961
```
which is wrong obviously. I narrowed it down to the `transform_inits` function. If I replace that function with the stanc2 implementation of the same function, everything works.

Given that generated quantities in cmdstan currently dont work if any parameter is non-scalar (stanc3 or stanc2), no one was able to use this and so no one noticed. Working on https://github.com/stan-dev/stan/pull/2937, I used the service directly (bypassing cmdstan) and found this bug.

rok-cesnovar:
stanc3 transform_inits
```cpp
  inline void transform_inits(const stan::io::var_context& context__,
                              std::vector<int>& params_i__,
                              std::vector<double>& vars__,
                              std::ostream* pstream__) const
    final {
    using local_scalar_t__ = double;
    vars__.clear();
    vars__.reserve(num_params_r__);
    
    try {
[...]

PR #610, opened 2020-07-05
fix indexing in transform_init

nhuurre:
I think this should fix issue #609

Pretty sure indexing in `unconstrained_param_names()` is also broken but [a comment](https://github.com/stan-dev/stanc3/blob/5de57e01f7e070b745696307aacfe229d4774958/src/stan_math_backend/Stan_math_code_gen.ml#L505-L507) says it's probably always been broken in other ways too.

### DIFF
--- a/src/middle/SizedType.ml
+++ b/src/middle/SizedType.ml
@@ -50,6 +50,7 @@ let rec associate ?init:(assocs = Label.Int_label.Map.empty) = function
       associate ~init:(Expr.Labelled.associate ~init:assocs e) st
 
 let is_scalar = function SInt | SReal -> true | _ -> false
+let rec inner_type = function SArray (t, _) -> inner_type t | t -> t
 
 let rec dims_of st =
   match st with
--- a/src/middle/Stmt.ml
+++ b/src/middle/Stmt.ml
@@ -338,9 +338,11 @@ module Helpers = struct
 
   (** Exactly like for_scalar, but iterating through array dimensions in the
   inverted order.*)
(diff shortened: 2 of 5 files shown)


rok-cesnovar:
Model

```stan
parameters {
  matrix[1,2] p_ar_mat[3];
}
model {
    for (i in 1:3)
      for (k in 1:1)
        for (l in 1:2)
          p_ar_mat[i, k, l] ~ normal(0,1);
}
generated quantities {
  matrix[1,2] gq_ar_mat[3];
      for (i in 1:3)
        for (k in 1:1)
          for (l in 1:2)
            gq_ar_mat[i, k, l] = p_ar_mat[i, k, l];
}
```

fails with sampling with this fix.

nhuurre:
> fails with sampling with this fix.
Does sampling even use `transform_inits`? Oh, right `gen_write` is used for other stuff too. Oops.

rok-cesnovar:
> Does sampling even use `transform_inits`?

No, sampling does not use it. Otherwise, we would see this one earlier.

nhuurre:
Ok, how about now?

rok-cesnovar:
Sampling is fine now, but the generate_quantities fails with:

`Exception: matrix[uni,uni] indexing, row: accessing element out of range. index 2 out of range; expecting index to be between 1 and 1 (in 'examples/bugs/tgq.stan', line 2, column 2 to column 26)`

What I run is:

`./examples/bugs/tgq sample output file=fit.csv`
and then just
`./examples/bugs/tgq sample output file=fit.csv`

Issue #205, opened 2019-06-26
Make Debug_data_generator generate better values for constrained matrix types


VMatthijs:
Something fixed that is not the identity would already be a step forward. Indeed, we expect the identity to behave in a special way w.r.t. all kinds of constrain and unconstrain operations.

adamhaber:
This refers to changing https://github.com/stan-dev/stanc3/blob/3aff87223c6a11dbd5f35092473d48bf759cff01/src/frontend/Debug_data_generation.ml#L120 to produce more interesting matrices, right? 

Two followup questions:
1. Having some sort of vector-to-matrix reshaping (and back) and dot-product/matrix multiplication functionality is probably needed; could this be imported from somewhere, or should it be written from scratch?
2. ATM, this seems to always produce the same mock data; is this the desired behaviour? If not - maybe we want to give the user the option to pass a seed?

nhuurre:
Yes, that's the code.
1. I don't think the OCaml libraries Stanc3 uses have much matrix functionality. Hopefully writing it from scratch isn't too burdensome.
2. Looks like we're using [Core_kernel.Random](https://ocaml.janestreet.com/ocaml-core/latest/doc/base/Base/Random/index.html). It could take a seed but that's probably not necessary.

PR #613, opened 2020-07-07
Issue 205 - Added support for additional matrix types

adamhaber:
Following #205 , this adds covariance matrices that are not the identity.

This is still WIP - I wanted to get an idea if this seems in the right direction; if so I'll proceed with the other kinds of matrices.

Also - would it make sense to put all the linear algebra helper functions elsewhere?


src/frontend/Debug_data_generation.ml
```diff
@@ -2,6 +2,41 @@ open Core_kernel
 open Middle
 open Ast
 
+let rec transpose = function
+  | [] :: _ -> []
+  | rows ->
+      let hd = List.map ~f:List.hd_exn rows in
+      let tl = List.map ~f:List.tl_exn rows in
```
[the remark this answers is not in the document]
adamhaber: Not sure I understand this one. hd :: tl would give me the first row (since rows is a list of lists) and the rest, but I want the head of each individual row and the rest for the transpose. Am I missing something?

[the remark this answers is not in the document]
nhuurre: Oh, right, there's a List.map. Sorry, wasn't paying attention.

src/frontend/Debug_data_generation.ml
```diff
@@ -2,6 +2,41 @@ open Core_kernel
...
+let matprod x y =
+  let y_T = transpose y in
+  if List.length x <> List.length y_T then
+    failwith "Matrix multiplication dim. mismatch"
+  else List.map ~f:(fun row -> List.map ~f:(dotproduct row) y_T) x
+
+let rec vect_to_mat l m =
+  let rec drop n h =
+    match n with 0 -> h | _ -> drop (n - 1) (List.tl_exn h)
+  in
```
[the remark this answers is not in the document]
adamhaber: Thanks! I was looking for this at the docs, but I think I went over Base.List instead of Core_kernel.List

src/frontend/Debug_data_generation.ml
```diff
@@ -2,6 +2,41 @@ open Core_kernel
...
+
+let dotproduct xs ys =
+  if List.length xs <> List.length ys then
+    failwith "Must be called on vectors of equal size"
+  else
+    let rec rec_dotprod xs ys =
+      match (xs, ys) with
+      | [], [] -> 0.
+      | x :: xs, y :: ys -> (x *. y) +. rec_dotprod xs ys
+      | _ -> failwith "This should never happen"
```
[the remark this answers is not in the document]
adamhaber: Thanks, changed.

src/frontend/Debug_data_generation.ml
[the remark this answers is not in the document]
nhuurre: I don't even look at the docs, I just start typing and let Visual Studio Code autocomplete find it.

adamhaber:
I think this should be ready now. One possible change that I'm considering - once we have gen_[corr/cov]_cholesky, we don't really need gen_[corr/cov] since we can simply generate the cholesky factor and multiply... what do you think?

src/frontend/Debug_data_generation.ml
```diff
@@ -104,26 +129,84 @@ let gen_vector m n t =
...
+  in
+  List.mapi ~f:fill_row m
+
+let pad_mat mm m n =
+  let padding_mat =
+    List.init (m - n) ~f:(fun _ -> List.init n ~f:(fun _ -> 0.))
+  in
+  wrap_real_mat (mm @ padding_mat)
+
+let gen_cov_cholesly m n =
```
nhuurre, line 166: `cholesky` (with a `k`)

src/frontend/Debug_data_generation.ml
```diff
@@ -104,26 +129,84 @@ let gen_vector m n t =
...
+  let diag_mat = gen_diag_mat (List.init ~f:(fun _ -> Random.float 2.) n) in
+  let filled_mat = fill_lower_triangular diag_mat in
+  let row_normalizer l =
+    let row_norm =
+      Float.sqrt (List.fold ~init:0. ~f:(fun accum x -> accum +. (x *. x)) l)
+    in
+    List.map ~f:(fun x -> x /. row_norm) l
+  in
+  let normed_mat = List.map ~f:row_normalizer filled_mat in
+  if m <= n then wrap_real_mat normed_mat else pad_mat normed_mat m n
```
nhuurre, line 181: A correlation matrix is always full rank so the cholesky factor must be square. Stan language does not even allow declaring two dimensions for `cholesky_factor_corr`.

src/frontend/Debug_data_generation.ml
```diff
@@ -104,26 +129,84 @@ let gen_vector m n t =
...
+let fill_lower_triangular m =
+  let fill_row i l =
+    let _, tl = List.split_n l i in
+    List.init ~f:(fun _ -> Random.float 2.) i @ tl
+  in
+  List.mapi ~f:fill_row m
+
+let pad_mat mm m n =
+  let padding_mat =
+    List.init (m - n) ~f:(fun _ -> List.init n ~f:(fun _ -> 0.))
```
nhuurre, line 162: Does this need to be all zeros?

src/frontend/Debug_data_generation.ml
adamhaber: Oops, thanks :-)

src/frontend/Debug_data_generation.ml
adamhaber: No, I did this to be consistent with the previous behaviour (zeros padding the identity matrix), but it's probably better to put random numbers instead. Changed.

src/frontend/Debug_data_generation.ml
adamhaber: Thanks, changed.

PR #615, opened 2020-07-07
Add multinomial_logit distribution

nhuurre:
The new distribution added to math: https://github.com/stan-dev/math/pull/1927

The signatures are
* `real multinomial_logit_log(int[], vector)`
* `real multinomial_logit_lpmf(int[], vector)`
* `int[] multinomial_logit_rng(vector, int)`


rok-cesnovar:
@nhuurre will you be able to work on the docs before the release ? In that case we could merge this.

wds15:
If @nhuurre does not get to write the doc, I can write it before the release. Is it fine for you if we do it these days or does it have to be in place before the rc tag?

rok-cesnovar:
Yes, that is fine by me. I can merge develop in and then merge this.

PR #624, opened 2020-07-10
Bugfix/611 false positive too large integer in stancjs

### LINKED ISSUE #611: false positive about numeric literals
If you do in rstan

> stan_model(model_code = 'parameters {real y;} model {y ~ normal(0,1);}')

this parses under the C++ stanc but not the nightly javascript stanc3. The error message is

> Semantic error in 'string', line 1, column 39 to column 40:
Integer literal cannot be larger than 2_147_483_647.

rok-cesnovar:
Fixes #611

### DIFF
--- a/src/frontend/Semantic_check.ml
+++ b/src/frontend/Semantic_check.ml
@@ -679,8 +679,8 @@ and semantic_check_expression cf ({emeta; expr} : Ast.untyped_expression) :
       semantic_check_variable cf emeta.loc id
       |> Validate.apply_const (semantic_check_identifier id)
   | IntNumeral s -> (
-    match int_of_string_opt s with
-    | Some i when i < 2_147_483_648 ->
+    match float_of_string_opt s with
+    | Some i when i < 2_147_483_648.0 ->
         mk_typed_expression ~expr:(IntNumeral s) ~ad_level:DataOnly ~type_:UInt
           ~loc:emeta.loc
         |> Validate.ok
--- a/test/integration/bad/numeric-literal/int-bad4.stan
+++ b/test/integration/bad/numeric-literal/int-bad4.stan
@@ -1,0 +1,3 @@
+transformed data {
+    int n = 2_147_483_648;
+}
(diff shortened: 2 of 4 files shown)


src/frontend/Semantic_check.ml
```diff
@@ -679,8 +679,8 @@ and semantic_check_expression cf ({emeta; expr} : Ast.untyped_expression) :
       semantic_check_variable cf emeta.loc id
       |> Validate.apply_const (semantic_check_identifier id)
   | IntNumeral s -> (
-    match int_of_string_opt s with
-    | Some i when i < 2_147_483_648 ->
+    match float_of_string_opt s with
```
rok-cesnovar, line 682: `int_to_string` seems to not parse an integer literal from a string model. I have not figured why, but parsing and comparing it as a float works for both the binary and stancjs version.

nhuurre: so it just produced `None`? IIRC Javascript doesn't have a native integer type, maybe that has something to do with it.

rok-cesnovar: Sorry, I meant the parsing does not work correctly.
It behaves as if 0 is not < max int. Example:
```ocaml
match int_of_string_opt s with
    | Some i when i < 2_147_483_648 ->
        mk_typed_expression ~expr:(IntNumeral s) ~ad_level:DataOnly ~type_:UInt
          ~loc:emeta.loc
        |> Validate.ok
    | Some _i when -> (* This gets used. If you print _i, you get 0. *)
        mk_typed_expression ~expr:(IntNumeral s) ~ad_level:DataOnly ~type_:UInt
          ~loc:emeta.loc
        |> Validate.ok
    | _ -> Semantic_error.bad_int_literal emeta.loc |> Validate.error )
```

rok-cesnovar: But yeah, JS integer support is probably the issue here. Though we also use int_of_string for other line and column numbers in errors and those seem to work fine.

nhuurre: Huh, weird.

PR #629, opened 2020-07-11
Array valued bounds in debug data generation

### LINKED ISSUE #612: Array-valued bounds in Debug_data_generation
Pull request #552 generalized data bounds to arrays but Debug_data_generation was not updated.
For example this model complies correctly but `--debug-generate-data` crashes.
```stan
data {
  vector[10] x;
  vector<lower=x>[10] y;
}
```

Debug_data_generation is also not able to handle indexes
```stan
data {
  int K[2];
  vector[K[1]] x[K[2]];
}
```
but that's a known problem with the partial evaluator #202 
Still, the error message could be better. Maybe this
https://github.com/stan-dev/stanc3/blob/458a7933728debdb8b3ce14e5af6960faef95513/src/frontend/Debug_data_generation.ml#L12
Could be something like
```ocaml
  | _ -> failwith ("Cannot evaluate expression at" ^ Location_span.to_string e.meta.loc)
```

adamhaber:
Fixes #612


src/frontend/Debug_data_generation.ml
```diff
@@ -89,8 +89,65 @@ let gen_int m t = wrap_int (gen_num_int m t)
...
 
 let gen_row_vector m n t =
-  { expr= RowVectorExpr (repeat_th n (fun _ -> gen_real m t))
-  ; emeta= {loc= Location_span.empty; ad_level= DataOnly; type_= UMatrix} }
+  let extract_var e =
+    match e with {expr= Variable x; _} -> Map.find_exn m x.name | _ -> e
+  in
+  let gen_bounded t e =
+    match e with
+    | {expr= RowVectorExpr unpacked_e; _} | {expr= ArrayExpr unpacked_e; _} ->
```
nhuurre, line 97: It could be a column vector too. They look like this
```ocaml
| {expr=PostfixOp ({expr=RowVectorExpr unpacked_e;_},Transpose);_}
```

src/frontend/Debug_data_generation.ml
```diff
@@ -89,8 +89,65 @@ let gen_int m t = wrap_int (gen_num_int m t)
...
+        wrap_row_vector (List.map ~f:(fun x -> gen_real m (t x)) unpacked_e)
+    | _ ->
+        failwith
+          ( "Bad bounded (upper OR lower) expr: "
+          ^ Sexp.to_string [%sexp (e : (typed_expr_meta, fun_kind) expr_with)]
+          )
```
nhuurre, line 103: I recently discovered a better alternative to `failwith (Sexp.to_string _)`
```ocaml
raise_s [%message "Bad bounded (upper OR lower) expr: " (e : (typed_expr_meta, fun_kind) expr_with)]
```
It's both shorter and formats the s-expression instead of cramming it all on a single line.

PR #634, opened 2020-07-13
Fixed a path issue that emerged from switching to ec2 instances in Jenkinsfile

serban-nicusor-toptal:


rok-cesnovar:
One question here. Is it actually required to use Windows WSL here given that this is the same as running Linux?

Wouldnt this work?

```
stage("Build & test a Windows binary") {
    agent {
        dockerfile {
            filename 'docker/static/Dockerfile'
            //Forces image to ignore entrypoint
            args "-u 1000 --entrypoint=\'\'"
        }
    }
    steps {
        runShell("""
            eval \$(opam env)
            dune subst
            dune build -x windows
        """)
        echo runShell("""
            eval \$(opam env)
            time dune runtest --verbose
        """)
        sh "mkdir -p bin && mv `find _build -name stanc.exe` bin/windows-stanc"
        sh "mv `find _build -name stan2tfp.exe` bin/windows-stan2tfp"
        stash name:'windows-exe', includes:'bin/*'
    }
    post {always { runShell("rm -rf ./*")}}
}
```

rok-cesnovar:
Nothing against WSL, it's more that we have more (and faster) Linux agents and we don't need to fight this fight with paths and whatnot.

serban-nicusor-toptal:
Totally agree, I thought this needs to build on windows because of `dune build -x windows`.
If we get the same results by using a docker image it would indeed simplify a lot.

rok-cesnovar:
Building in WSL is basically the same as building on a Linux machine AFAIK. If you can try building with the above approach I can verify on my Windows machine.

serban-nicusor-toptal:
Looks like failing at the windows toolchain step https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3/detail/PR-634/7/pipeline/

rok-cesnovar:
Yeah, docker would need the cross-compile windows stuff: https://github.com/stan-dev/stanc3/blob/master/scripts/README.MD#installing-build-dependencies--inside-the-ubuntul-1004-wsl-

Forgot about that.

rok-cesnovar:
This latest one seems to work right?

serban-nicusor-toptal:
Yup, works fine now, will merge it into master.
I will install the pre-requisites on the linux ec2 and switch it to use docker from them tomorrow since it would speed up things a little.

rok-cesnovar:
Great, just uncomment and I will approve.

Jenkinsfile
```diff
@@ -179,7 +179,7 @@ pipeline {
             }
         }
         stage("Build and test static release binaries") {
-            when { anyOf { buildingTag(); branch 'master' } }
+            //when { anyOf { buildingTag(); branch 'master' } }
```
rok-cesnovar, line 182: This one and its ready.

PR #646, opened 2020-07-18
Move building and testing windows to ec2 instances

serban-nicusor-toptal:


rok-cesnovar:
@serban-nicusor-toptal does this otherwise work on Linux but not in Docker? Or are you unable to install this on the Linux machines directly?

serban-nicusor-toptal:
Hey @rok-cesnovar sorry, went for a walk.
I've updated the ec2 instances template to contain all the required dependencies. They build just fine, already tested!
I chose to make this work by Docker so we can be ready for [this](https://discourse.mc-stan.org/t/docker-stan/14966/25) thread when we'll get to it.
I am not sure why this doesn't work, I'll debug locally a bit more since it's easy with Docker but ... it has the same versions as what's on the ec2 instance.
What I think it happens, it's trying to compile a package for 32 bit on 64 bit and it messes up on an asm push instruction.

rok-cesnovar:
> What I think it happens, it's trying to compile a package for 32 bit on 64 bit and it messes up on an asm push instruction.

That does sound right yeah. Don't have any ideas why this works on WSL though or at least on the WSL on the new Windows machine.

serban-nicusor-toptal:
So installing `ocamlformat.0.8` before `dune 1.11.3` would force dune to install as 2.6 then when we pin dune to 1.11.3 would downgrade, somewhere in the middle it would break a package that later wants to rebuild a 32 bit binary... Removed it and the image builds fine now.
What's strange is that it says opam is not initialized even if it was in the image, looks like the other images hmmh not sure what's wrong on this one.

serban-nicusor-toptal:
Everything seems to be fine in the second to last [build](https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3/detail/PR-646/26/pipeline).
This is ready to merge.

rok-cesnovar:
Sorry to double review, but I now remembered that we should add the same change to the test binaries jenkins file. But that can be a separate PR if you want.

serban-nicusor-toptal:
No worries, I'll add it to this one.
I wonder why is [this](https://github.com/stan-dev/stanc3/pull/646/files#diff-24fba5cdd9447f8adc146959954ef135) showing up in the PR.

rok-cesnovar:
That was a merge to master. I think uts all good. Will merge when it passes.

PR #653, opened 2020-07-22
Improved support for Nix development

rybern:
- Lowered the version of Nixpkgs so that ocamlformat matches our current standard of 0.8
- Added a shell.nix that manages a bunch of the useful dev environment tools, like dune, ocamlformat and merlin
- Added support for utop
- Updated the readme to match


rybern:
Thanks @rok-cesnovar! Good idea to leave it until after the release.

PR #656, opened 2020-07-24
Revert 521

rok-cesnovar:
This PR reverts #521 

There is nothing wrong with the stanc3 implementation itself, but rvalue seems to does have a performance penalty the way it is implemented right now. I will dig a bit more in the rvalue implementation if we can maybe not revert the stanc3 PR. If we can't find a solution, it will have to be reverted for the release.

### DIFF
--- a/src/stan_math_backend/Expression_gen.ml
+++ b/src/stan_math_backend/Expression_gen.ml
@@ -56,11 +56,6 @@ let minus_one e =
 
 let is_single_index = function Index.Single _ -> true | _ -> false
 
-let dont_need_range_check = function
-  | Index.Single Expr.Fixed.({pattern= Var id; _}) ->
-      not (Utils.is_user_ident id)
-  | _ -> false
-
 let promote_adtype =
   List.fold
     ~f:(fun accum expr ->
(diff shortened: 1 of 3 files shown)


wds15:
Is there a stanc3 magic url I can use from this PR to test this out?

rok-cesnovar:
Building https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3-test-binaries/detail/stanc3-test-binaries/130/pipeline

nhuurre:
I was afraid of this. I assume the problem is that non-checking index access can return a reference/pointer but `rvalue()` must return a copy? Is [stan::math::get_base1](https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/get_base1.hpp) a lightweight alternative to `rvalue`? Stanc2 had that but I wasn't sure why.

rok-cesnovar:
Could be yes. I was debating this with @t4c1 and the first thing that popped up was the checks, but commenting out checks made no difference.

t4c1:
> I assume the problem is that non-checking index access can return a reference/pointer but rvalue() must return a copy?

I kinda doubt it. Usually it is faster to return scalars by value than by reference. It might make a difference for `var`. Anyway compilers are quite good at optimizing returns.

On the other hand I have no idea what it could be, so something I don't understand must be going on.

It might be worth trying `get_base1`. I have no idea if it will help, but at least its a bit simpler without those awkward index structs.

rok-cesnovar:
Should we just merge this and go without this check feature for the release? And if we come up with something until Monday we merge that?

rok-cesnovar:
@wds15  if you wish to confirm this one on your system:

STANC3_TEST_BIN_URL = https://jenkins.mc-stan.org/job/stanc3-test-binaries/130/artifact

rok-cesnovar:
The simplest model where I can reproduce this issue is the non-vectorized Bernoulli example model.

```stan
data { 
  int<lower=0> N; 
  int<lower=0,upper=1> y[N];
} 
parameters {
  real<lower=0,upper=1> theta;
} 
model {
  theta ~ beta(1,1);  // uniform prior on interval 0,1
  for(i in 1:N) {
    y[i] ~ bernoulli(theta);
  }  
}

```

rok-cesnovar:
Sebastian confirmed this fixes the perf. regression. Anyone wants to approve?

PR #660, opened 2020-07-31
Add a complex type to the stan language

SteveBronder:
## Summary

This is a first go at writing a `complex` type and it's containers for the stan language. The code here isn't complete but can parse a lot of stuff. The stan compiler can parse and make the right C++ for things like the example here

```stan
data {
  int N;
  int M;
  vector[N] y;
  matrix[N, M] x;
}

parameters {
  complex_vector[M] beta;
  complex<lower=0> sigma;
}

model {
  target += normal_lpdf(y | x * beta, sigma);
}
```

I mostly wrote this so I could get an idea of what stuff we need to write in the `stan` repo to connect everything (you can see the output cpp of the above model [here](https://gist.github.com/SteveBronder/536e20753d6e6ff7ab8125a8f6a92eb1)).

Though I'll also say I h8 the way I wrote this and am very open to better patterns. This branch treats `complex`, `complex_vector`, `complex_rowvector`, and `complex_matrix` as additional types. This adds a ton of what I'd absolutely call boilerplate to the compiler. It's also a bad pattern to set precedence on, if we add `sparse`, `static`, and other attributes in the future we'll end up with some pretty ugly stuff. Also I haven't figured it out yet but something in `stan2tfp` is seg-faulting

Another way to do this would be to add an additional param (idk what the name for this is) to the unsized and sized types that indicate whether it's just a real value or a complex value or any other attribute we want to add in the future. idk how to do that or what the pattern would look like but with some guidance (or bardic inspiration) I could probably figure something out.

Doing ^ we could keep the name of each as `complex`, `complex_vector` etc. or we could have attributes on types like `complex real`, `complex vector` yadayada.

This also makes error messages nearly gcc levels of bloated, where a mismatch of types produces errors like

```
(real, real[], complex_vector) => real
+|(real, real[], complex_row_vector) => real
+|(real, real[], real[]) => real
+|(real, real[], complex[]) => real
+|(real, complex[], real) => real
+|(real, complex[], complex) => real
+|(real, complex[], vector) => real
+|(real, complex[], row_vector) => real
+|(real, complex[], complex_vector) => real
+|(real, complex[], complex_row_vector) => real
+|(real, complex[], real[]) => real
+|(real, complex[], complex[]) => real
```

But for 500+ lines because of all the combinations. I'd assume that anything that can take a real in `stan` should also be able to take a `complex` type (same for all the containers)


SteveBronder:
Also the +/- at the top is silly, there's like 500 lines of code change here

bob-carpenter:
>  I'd assume that anything that can take a real in stan should also be able to take a complex type 

Other way around.  Anything that can take a `complex` should be able to handle a `real`.

> Also the +/- at the top is silly, there's like 500 lines of code change here

Silly because it's a lot or a little?

> Another way to do this would be to add an additional param (idk what the name for this is) to the unsized and sized types that indicate whether it's just a real value or a complex value or any other attribute we want to add in the future. idk how to do that or what the pattern would look like but with some guidance (or bardic inspiration) I could probably figure something out.

I assume we're talking about the difference between a new full type that just coincidentally looks like another type, such as `complex_matrix[M, N]` and a type qualifier `complex matrix[M, N]`.  Keep in mind, there are constraints---we don't wanbt `complex int`, `complex simplex`, etc.  Same for `sparse`, because that only applies to matrix types.  It doesn't make sense to have `sparse real`, etc.  So you can either choose to fully elaborate the types or enforce the syntactic co-occurrence restrictions some other way.

So if we have `complex real<lower = 0, upper = 1>`, what would that mean?  Are both the real and imaginary parts constrained?  Or do we rule that out, too?  

I'm all for minimizing work, but unless there's a plan where the parser/intermediate layer/generator code is made easier, I don't see the point.

How would having type qualifiers make our error messages shorter?  Are you imagining we have something like `foo((complex) real, real)` to indicate two signatures, `foo(complex real, real)` and `foo(real, real)`. 

Also, to my eye, having a qualifier like `complex real` just looks wrong.  It's not a complex real number, it's a complex number that happens to have a real and imaginary component.   I would strongly prefer the type `complex` for a complex scalar.

bob-carpenter:
P.S.  Super cool that you're doing this.  However we want to do it, I really want to get this in.  Otherwise I'm going to have to start writing things like FFT functions that take parallel arrays of reals to represent complex inputs and outputs.

SteveBronder:
> Other way around. Anything that can take a complex should be able to handle a real.

Ah, this changes my assumptions

>> Also the +/- at the top is silly, there's like 500 lines of code change here

> Silly because it's a lot or a little?

it says `+263,585` lines added at the top but code-wise I changed about +/- 500 lines

>> Another way to do this would be to add an additional param (idk what the name for this is) to the unsized and sized types that indicate whether it's just a real value or a complex value or any other attribute we want to add in the future. idk how to do that or what the pattern would look like but with some guidance (or bardic inspiration) I could probably figure something out.

> I assume we're talking about the difference between a new full type that just coincidentally looks like another type, such as complex_matrix[M, N] and a type qualifier complex matrix[M, N].

Yeah exactly. About 70% of the code changes here are just matching against both a type and it's complex form like `UMatrix` and `UComplexMatrix`.

> Keep in mind, there are constraints---we don't wanbt complex int, complex simplex, etc. Same for sparse, because that only applies to matrix types. It doesn't make sense to have sparse real, etc. So you can either choose to fully elaborate the types or enforce the syntactic co-occurrence restrictions some other way.

I'd leave this to someone with more ocaml knowledge than me but either way we can have `complex_matrix`/`complex_corr_matrix` or `complex matrix`/`complex corr_matrix`. The question of fully elaborated types or something fancier is a question of how much boilerplate we want in the codebase

> So if we have `complex real<lower = 0, upper = 1>`, what would that mean? Are both the real and imaginary parts constrained? Or do we rule that out, too?

My programmer brain is translating this to "`std::complex<double>` that has the constrain functions called on it" too quickly for me to understand how a statistician would see this. For me, reading that right to left it's just a bounded floating point complex type. If we did `complex` as an attribute it would be weird to also have a shorthand `complex` that meant `complex real` since then `complex` is both an attribute and a type.

idk about that one way or another. Maybe this would be a good place to use the new Stan dev voting thing Jonah posted on discouse.

> I'm all for minimizing work, but unless there's a plan where the parser/intermediate layer/generator code is made easier, I don't see the point.

Yeah that's the other gotcha here, tbf I'm not clever enough to write this in a prettier way. @rybern @nhuurre would one of yinz have any time to help me figure something nicer out?

> How would having type qualifiers make our error messages shorter? Are you imagining we have something like `foo((complex) real, real)` to indicate two signatures, `foo(complex real, real)` and `foo(real, real)`.

This was based on thinking that all functions that take reals would also take complex types (where it's the other way around).

The actual issue is that I don't like how ginormous the list of type signatures becomes when a user hit's an error. Maybe the actual answer is to have more "meaningful" error message somehow. Like if we assume they got one input correct we only print out the signatures that contain either of the input types. Otherwise I just worry that adding more types is going to make our error messages a big mess to read through

> Also, to my eye, having a qualifier like complex real just looks wrong. It's not a complex real number, it's a complex number that happens to have a real and imaginary component. I would strongly prefer the type complex for a complex scalar.

tbh I've never worked with complex numbers so if `complex real` looks weird to folks then sure we could just have `complex`/`complex_matrix` etc.

> P.S. Super cool that you're doing this. However we want to do it, I really want to get this in. Otherwise I'm going to have to start writing things like FFT functions that take parallel arrays of reals to represent complex inputs and outputs.

Yeah np! I meant to do this earlier but totally forgot. If I have time this weekend I can slam out a PR in stan and we can try compiling some models with this

SteveBronder:
@bob-carpenter should the line [here](https://gist.github.com/SteveBronder/536e20753d6e6ff7ab8125a8f6a92eb1#file-complex_test_code-hpp-L201) for complex be

```cpp
    stan::math::accumulator<std::complex<T__>> lp_accum__;
```
or is it fine as
```cpp
    stan::math::accumulator<std::complex<T__>> lp_accum__;
```

rok-cesnovar:
> it says `+263,585` lines added at the top but code-wise I changed about +/- 500 lines

Yeah, the signatures file has +262,646 changes, because all functions that take in real were added another signature with complex, so that is expected. We should think of a way of compressing this list. If foo() can take in any type we should print it as T foo(T), I think.

> Maybe the actual answer is to have more "meaningful" error message somehow. 

I think this is the answer. Maybe print complex signatures only if any of the types used in the call are complex or if the function is only allowed with complex types but not real (if will there be any such)? There might be better solutions

bob-carpenter:
>  If foo() can take in any type we should print it as T foo(T), I think.

Not quite.  We have a lot of functions that look like `real foo(real | int)`, for example, and functions that look like `real foo(complex)`.

bob-carpenter:
> You can replace some of the UInt | UReal | UComplex with something like is_primitive_type,

This shouldn't be `is_primitive`.  Only the int and real are primitive---complex is made up of two reals (i.e., it's a compound, not a primitive).

The single word that refers to all of them is "scalar", but we've used "scalar" to just mean real in the past, so this is all going to be very confusing until we get make our metaprograms consistent.  (Steve and I did some of that for the complex PRs, but it's still not complete because we didn't touch some things that weren't getting used.)

> Maybe the actual answer is to have more "meaningful" error message somehow.

I agree with @rok-cesnovar that this would be ideal.  I'm just not sure how to achieve it.  At the extreme level of customization, we could tailor these manually for each function.  I don't see how to summarize automatically.  

The simplest expedient is to just truncate the list and send the user to the manual.

>or is it fine as
> `stan::math::accumulator<T__> lp_accum__;`

It's not just fine, it hjas to be real.  Densities are real, not complex!

> The question of fully elaborated types or something fancier is a question of how much boilerplate we want in the codebase

Is there a proposal for eliminating the boilerplate?  I really see a problem given our basic types:

```
int, real, vector, row_vector, matrix
```

We need a way to have complex versions of everything but `int`.  But while the basic types can get lower and upper bounds, the complex versions cannot.

Some of the more general types might have complex versions, but I'm not sure we really want to worry about positive-definite Hermitian complex matrices and their Cholesky factors (I don't even know if the latter exist, but the general rule seems to be that moving to complex makes more things exist, not fewer).  It won't make sense for other types like correlation matrices, simplex, unit_vector, ordered, etc.

So my question remains how much more complicated does the code have to get to allow `complex` as a type qualifier compared to just adding four new types.

> If we did complex as an attribute it would be weird to also have a shorthand complex that meant complex real since then complex is both an attribute and a type.

You're right that `complex real` doesn't make sense---that's why I'm bringing it up.  So it'd be `complex vector` but just `complex` instead of `complex real`?   Then it seems like it's going to have to act both as a basic type and as a type qualifier.

Maybe I should just wait for you to make a concrete proposal rather than trying to guess what you intend.

>  it's just a bounded floating point complex type

What does that mean?  A complex number has a real component and an imaginary component. Is the bound on radius in complex space (`sqrt(real^2 + imag^2)`) or what?  If that's the case, the minimum value is 0.  

> Maybe this would be a good place to use the new Stan dev voting thing Jonah posted on discouse.

We can't vote with only one proposal! My first instinct is to just use `complex`, `complex_vector`, `complex_row_vector`, and `complex_matrix` as new basic types which do not admit any constraints.  For a vote, there would need to be an alternative proposal!

> The actual issue is that I don't like how ginormous the list of type signatures becomes when a user hit's an error. 

I completely agree this is an issue.  It's already too large now for the basic arithmetic because of all the vector arithmetic options we've added.  I think just truncating may be a decent first approximation.

> tbh I've never worked with complex numbers so if complex real looks weird

It's weird because a complex number consists of a real number and an imaginary number (represented as a real multiple of `i = sqrt(-1)`).  So a complex number is two real numbers computationally.  That means things like ordering aren't so straightforward.

SteveBronder:
If people are cool with the boilerplate'd version then I'm fine with it as well.

Bob I'm A-OK with `complex`, `complex_vector`, etc. I can remove the stuff for bounds I wrote in here if they don't make sense for complex numbers
>> or is it fine as
```
    stan::math::accumulator<T__> lp_accum__;
```
> It's not just fine, it has to be real. Densities are real, not complex!

So if `T__` is var or double idk how this will work because the distributions will return a complex value. Like the return in normal's signature is

```
template <bool propto, typename T_y, typename T_loc, typename T_scale>
inline return_type_t<T_y, T_loc, T_scale> normal_lpdf(const T_y& y,
                                                      const T_loc& mu,
                                                      const T_scale& sigma) 
```

Where `return_type_t` is 

```
template <typename T, typename... Ts>
struct return_type<T, Ts...> {
  using type
      = scalar_lub_t<scalar_type_t<T>, typename return_type<Ts...>::type>;
};
```

and `scalar_lub_t` and `scalar_type_t` return a `complex<T>` for a complex value. The accumulator will be summing up vars or doubles so what should it do when it sees a complex value it needs to accumulate?

Some other things prodding around trying to get this to compile

1. Do we need a signature in math to accept multiplying complex and real matrices? I think in some places we need to replace where we have `require_arithmetic_t` what's some sort of type trait name we could use for a function that should take in both real and complex<double> values? 

2. `write_array` tries to write values to a `std::vector<double>` called `vars__`. When we have a `complex_vector` like

```cpp
      beta = Eigen::Matrix<std::complex<double>, -1, 1>(M);
      stan::math::fill(beta, std::numeric_limits<double>::quiet_NaN());
      beta = in__.complex_vector(M);
```

what should happen when we add values to `vars__` i.e.
```
      for (int sym1__ = 1; sym1__ <= M; ++sym1__) {
        vars__.emplace_back(beta[(sym1__ - 1)]);
      }
```
How is `vars__` used later? Should that loop load up both the real and imaginary part?

3. How should complex types be handled in the `var_context`? Another good example in the code Iink above

```cpp
      Eigen::Matrix<std::complex<double>, -1, 1> beta;
      beta = Eigen::Matrix<std::complex<double>, -1, 1>(M);
      stan::math::fill(beta, std::numeric_limits<double>::quiet_NaN());
      
      {
        std::vector<std::complex<local_scalar_t__>> beta_flat__;
        current_statement__ = 1;
        assign(beta_flat__, nil_index_list(), context__.vals_r("beta"), // THIS LINE HERE
          "assigning variable beta_flat__");
        current_statement__ = 1;
        pos__ = 1;
        current_statement__ = 1;
        for (int sym1__ = 1; sym1__ <= M; ++sym1__) {
          current_statement__ = 1;
          assign(beta, cons_list(index_uni(sym1__), nil_index_list()),
            beta_flat__[(pos__ - 1)], "assigning variable beta");
          current_statement__ = 1;
          pos__ = (pos__ + 1);}
      }
```

For the line I marked above, `vals_r` returns back an `std::vector<double>`. For an assignment to `std::vector<std::complex<double>>` what should assign do with a double vector assigned to a complex vector? I'm not really sure how the real and imaginary parts should be handled here

bob-carpenter:
> because the distributions will return a complex value

No they won't!  Complex arguments are not legal for the probability functions we have.  Maybe we should jump on a call and talk about this.  

The key here is that a real value can be promoted to a complex value, but complex values can't be demoted to real generically.  The distributions are all real-only as we have them defined.  So they're never going to be returning complex values.

For efficiency, we can't double the size of every real number and just assume everything's complex underyingly.  It'd also need runtime checks on the complex components being zero in lots of places.

Return type for complex functions varies.  Sometimes they return real values, sometimes they return complex values.  Some functions take multiple real values and return complex values.  

> a signature in math to accept multiplying complex and real matrices?

That's already in place.  All the mixed arithmetic works and is thoroughly tested up to 3rd order derivatives.  Check out the unit tests for the basic functions and operators.

> write_array tries to write values to a std::vector<double> called vars__. When we have a complex_vector like

We only need to worry about the basic types. When we write a `complex` value out, it spits out two real numbers, the real component of the complex number and the imaginary component.  It's literally just this:

```
template <typename T, typename = is_real_v<T>>
complex(const T& real, const T& imag) : real_(real), imag_(imag) { }

template <typename T, typename = is_real_v<T>>
complex(const T& real) : complex(real, 0) { }
```

where `is_real_v<T>` evaluates to `true` if `T` is a real scalar type, which is just `int, double, var, fvar<T>` for us.

The second constructor explains how promotion works.

To save a complex number, we need to dump both the real and imaginary component.  Same for input---it requires two real numbers to make a complex number.

> How should complex types be handled in the var_context? 

Again, we just need to work out how `complex` works, then containers just `vector<complex>` just work elementwise.

Complex numbers need to read two values just like a `real[2]` or `vector[2]` would.  Probably the easiest thing to do is require them to be an array.  We have to think about the JSON and Rdump representation, and both of those will be easier by just treating them like a 1-dimensional container with a fixed size of 2.  That's how they're laid out in memory, after all :-)




```

**Comment by bob-carpenter** [comment]
I thought I already responded, but don't see a response.  The log density is going to be real, not complex.  As are all the return values of densities and other probability functions.

**Comment by SteveBronder** [comment]
>> because the distributions will return a complex value

> No they won't! Complex arguments are not legal for the probability functions we have. Maybe we should jump on a call and talk about this.

I'll hit you up to jump on a call cuz yes ftr I never took a class on complex numbers so I'm not really sure about the example programs I should be writing to check against. I'll probs need your help here to add what stuff we want to support in the mother model and any other stan programs we want to test compile correctly. Also need to add the `imag` and `real` functions to the compiler

>> a signature in math to accept multiplying complex and real matrices?

> That's already in place. All the mixed arithmetic works and is thoroughly tested up to 3rd order derivatives. Check out the unit tests for the basic functions and operators.

I'm seeing an error for this when trying to compile models so I think there might be a bug in our type deduction we need to sort out. I'm also not seeing tests for `multiply` with an `Eigen::Matrix<complex<T>>` in `mix/prim/rev` tests for math

(you can ignore the normal_lpdf part of the error message below since the multiply needs to be wrapped in a `real()` call here)

```
examples/bernoulli/bernoulli.hpp:218:57: error: no matching function for call to ‘multiply(const Eigen::Matrix<double, -1, -1>&, Eigen::Matrix<std::complex<double>, -1, 1>&)’
  218 |         lp_accum__.add(normal_lpdf<propto__>(y, multiply(x, beta), sigma));
      |                                                 ~~~~~~~~^~~~~~~~~
In file included from stan/lib/stan_math/stan/math/prim/fun.hpp:212,
                 from stan/lib/stan_math/stan/math/prim.hpp:10,
                 from stan/lib/stan_math/stan/math/rev.hpp:9,
                 from stan/lib/stan_math/stan/math.hpp:19,
                 from stan/src/stan/model/model_header.hpp:4:
stan/lib/stan_math/stan/math/prim/fun/multiply.hpp:28:13: note: candidate: ‘template<class Mat, class Scal, stan::require_eigen_t<T>* <anonymous>, stan::require_stan_scalar_t<Scal>* <anonymous> > auto stan::math::multiply(const Mat&, Scal)’
   28 | inline auto multiply(const Mat& m, Scal c) {
      |             ^~~~~~~~
stan/lib/stan_math/stan/math/prim/fun/multiply.hpp:28:13: note:   template argument deduction/substitution failed:
stan/lib/stan_math/stan/math/prim/fun/multiply.hpp:44:13: note: candidate: ‘template<class Scal, class Mat, stan::require_stan_scalar_t<T>* <anonymous>, stan::require_eigen_t<S>* <anonymous> > auto stan::math::multiply(Scal, const Mat&)’
   44 | inline auto multiply(Scal c, const Mat& m) {
      |             ^~~~~~~~
stan/lib/stan_math/stan/math/prim/fun/multiply.hpp:44:13: note:   template argument deduction/substitution failed:
stan/lib/stan_math/stan/math/prim/fun/multiply.hpp:67:13: note: candidate: ‘template<class Mat1, class Mat2, stan::require_all_eigen_vt<std::is_arithmetic, T1, T2>* <anonymous>, stan::require_any_not_same_t<double, typename stan::value_type<T, void>::type, typename stan::value_type<T_container, void>::type>* <anonymous>, stan::require_not_eigen_row_and_col_t<Mat1, Mat2>* <anonymous> > auto stan::math::multiply(const Mat1&, const Mat2&)’
   67 | inline auto multiply(const Mat1& m1, const Mat2& m2) {
      |             ^~~~~~~~
stan/lib/stan_math/stan/math/prim/fun/multiply.hpp:67:13: note:   template argument deduction/substitution failed:
stan/lib/stan_math/stan/math/prim/fun/multiply.hpp:93:13: note: candidate: ‘template<class Mat1, class Mat2, stan::require_all_eigen_t<Mat1, Mat2>* <anonymous>, stan::require_all_same_t<double, typename stan::value_type<T, void>::type, typename stan::value_type<T_container, void>::type>* <anonymous>, stan::require_not_eigen_row_and_col_t<Mat1, Mat2>* <anonymous> > decltype ((m1 * m2).eval()) stan::math::multiply(const Mat1&, const Mat2&)’
   93 | inline auto multiply(const Mat1& m1, const Mat2& m2)
      |             ^~~~~~~~
stan/lib/stan_math/stan/math/prim/fun/multiply.hpp:93:13: note:   template argument deduction/substitution failed:
stan/lib/stan_math/stan/math/prim/fun/multiply.hpp:130:13: note: candidate: ‘template<class RowVec, class ColVec, stan::require_all_not_var_t<typename stan::scalar_type<T, void>::type, typename stan::scalar_type<T, void>::type>* <anonymous>, stan::require_eigen_row_and_col_t<RowVec, ColVec>* <anonymous> > auto stan::math::multiply(const RowVec&, const ColVec&)’
  130 | inline auto multiply(const RowVec& rv, const ColVec& v) {
      |             ^~~~~~~~
stan/lib/stan_math/stan/math/prim/fun/multiply.hpp:130:13: note:   template argument deduction/substitution failed:
stan/lib/stan_math/stan/math/prim/fun/multiply.hpp:146:40: note: candidate: ‘template<class Scalar1, class Scalar2, stan::require_all_stan_scalar_t<ScalarA, ScalarB>* <anonymous> > stan::return_type_t<T1, T2> stan::math::multiply(Scalar1, Scalar2)’
  146 | inline return_type_t<Scalar1, Scalar2> multiply(Scalar1 m, Scalar2 c) {
      |                                        ^~~~~~~~
stan/lib/stan_math/stan/math/prim/fun/multiply.hpp:146:40: note:   template argument deduction/substitution failed:
In file included from stan/lib/stan_math/stan/math/rev/fun/matrix_exp_multiply.hpp:6,
                 from stan/lib/stan_math/stan/math/rev/fun.hpp:100,
                 from stan/lib/stan_math/stan/math/rev.hpp:11,
                 from stan/lib/stan_math/stan/math.hpp:19,
                 from stan/src/stan/model/model_header.hpp:4:
stan/lib/stan_math/stan/math/rev/fun/multiply.hpp:560:13: note: candidate: ‘template<class Mat1, class Mat2, stan::require_all_eigen_t<Mat1, Mat2>* <anonymous>, stan::require_any_eigen_vt<stan::is_var, Vec1, Vec2>* <anonymous>, stan::require_not_eigen_row_and_col_t<Mat1, Mat2>* <anonymous> > auto stan::math::multiply(const Mat1&, const Mat2&)’
  560 | inline auto multiply(const Mat1& m1, const Mat2& m2) {
      |             ^~~~~~~~
stan/lib/stan_math/stan/math/rev/fun/multiply.hpp:560:13: note:   template argument deduction/substitution failed:
stan/lib/stan_math/stan/math/rev/fun/multiply.hpp:597:12: note: candidate: ‘template<class RowVec, class ColVec, stan::require_any_var_t<typename stan::value_type<T, void>::type, typename stan::value_type<T_container, void>::type>* <anonymous>, stan::require_eigen_row_and_col_t<RowVec, ColVec>* <anonymous> > stan::math::var stan::math::multiply(const RowVec&, const ColVec&)’
  597 | inline var multiply(const RowVec& m1, const ColVec& m2) {
      |            ^~~~~~~~
stan/lib/stan_math/stan/math/rev/fun/multiply.hpp:597:12: note:   template argument deduction/substitution failed:
```

I think where we have checks for arithmetic in prim in some places we need to have checks just that it's not an autodiff type. We can sort this out in the call

> To save a complex number, we need to dump both the real and imaginary component. Same for input---it requires two real numbers to make a complex number.

Cool cool that's what I was doing in the reader. I'll probably have to write something in the compiler to handle collecting those values for `write_array`

>>    How should complex types be handled in the var_context?

> Again, we just need to work out how complex works, then containers just vector<complex> just work elementwise.

> Complex numbers need to read two values just like a real[2] or vector[2] would. Probably the easiest thing to do is require them to be an array. We have to think about the JSON and Rdump representation, and both of those will be easier by just treating them like a 1-dimensional container with a fixed size of 2. That's how they're laid out in memory, after all :-)

Cool yes if we have them passed in as if they are arrays then I think everything should just sort itself out. So a single complex type would come in the same form as `real x[1, 2]` and complex arrays as `real x[N, 2]` then I think we are good

**Comment by rok-cesnovar** [comment]
Bumping this PR a bit as its a new release period and trying to revive a few things for the next period. 

This seems like a somewhat easy win if we can define a clear language spec and I would be interested in helping out if needed. Do not want to hijack anyone elses project though. I think going with `complex`, `complex_vector`, `complex_row_vector` and `complex_matrix` is the clearer way to implement this but we need more eyes on this.

It seems that there is a bit of a misunderstanding of which function would support complex inputs. We need to make a list of functions that currently support complex inputs in Math.

@bob-carpenter when you have a minute can you write an example model of what you imagine would be a valid stan model (assume the types are complex, complex_vector and complex_matrix).

This model makes no sense, fft also doesnt exist, but lets imagine it did. Was something like this the idea?
```stan
parameters {
      vector[N] y;
      ... 
}
model{
    complex_vector[N] y_fft = fft(y);
    ...
    ...
    abs(y_fft) ~ normal(x, 1);

}
```

bob-carpenter:
This would be great.  I can help write the spec, but it's basically just as @rok-cesnovar outlined in terms of new types

* `complex`, `complex_vector`, `complex_row_vector`, `complex_matrix`

We want to support all the of the arithmetic and matrix arithmetic operations, including mixed real/complex operations. 

The entire C++ spec for complex is implemented, so all those functions should be included.

It'd be good to add the useful functions like FFT and asymmetric eigendecomposition.

Assignment needs to be extended so that real and integer values can be assigned to complex.  The assignment hierarchy is now `int` -> `real` -> `complex` in terms of assignability.

In terms of models, I was thinking about [phase retrieval](https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=8587152) specifically.  You observe magnitudes of FFTs with noise, then try to work out the underlying signal that would've produced that observation---it's a classic "inverse problem" in that sense.

```
data {
  vector[N] y;  // measurements of magnitude of spectrum (not complex)
parameters {
  complex_vector[N] alpha;  // underlying signal to reconstruct (complex)
  real<lower = 0> sigma;  // observation noise
model {
  complex_vector[N] fft_alpha = fft(alpha);  // fft maps complex vector to complex vector
  vector[N] mu = abs(fft_alpha);  // complex abs() (aka "modulus") produces real value elementwise
  y ~ normal(mu, sigma);  // noisy measurement of y
  alpha ~  some prior
  sigma ~ some prior
}
```

bob-carpenter:
The basic matrix operations for Eigen need to extended using their extension mechanism to implement the mixed double/complex matrix operations.  The FFT is used as a test case, so I know that works from Eigen's implementation, but it's not wrapped up as a function.  That's the most important thing to get in.  Users can build a lot of their own functionality with the C++ lib.

rok-cesnovar:
Thanks @bob-carpenter ! That makes it a lot clearer. 

The  bigger challenges where decisions will need to be made are: 
- how to create complex type from to primitve types, for example `complex a = to_complex(r,i)` or something else
  
  Here I feel `to_complex` follows the `to_matrix`, `to_vector`, etc already in Stan and would be fine.

- how to get real/imag parts of a complex number (`a.real/a.imag or `get_real(a)`/`get_imag(a)` or something else)

  The `.imag`,`.real` would be something new in the language syntax, though tuples will probably have something similar.
  I do not really like the get_real(a)/get_imag(a), just posting an alternative . I would prefer real(a)/imag(a) but real is already a type.

- how to read in complex types from data files 

  I guess an array of length 2 is the obvious solution here. We need to support this for both JSON and Rdump.
  This is something we could do last as we could test the rest by working with the below and defer to this once we know the rest of it works.

```stan
data {
   vector[N] a; 
   vector[N] b;
  }
  transformed data {
    complex_vector c = to_complex(a,b);
 } 
```

- CSV ouput
 The obvious solution here is to output a.real and a.imag for each complex item.

Will give it a few more days and we can then collaborate on making a language spec. I think if we define everything and decide this has a very good shot for 2.27.

bob-carpenter:
> (@rok-cesnovar) The .imag,.real would be something new in the language syntax, though tuples will probably have something similar. I do not really like the get_real(a)/get_imag(a), just posting an alternative that would be deviate the least from current language and would thus be easiest to implement. I would prefer real(a)/imag(a) but real is already a type.

Good point.  I checked again and C++ uses `real(z)`, `imag(z)`, as well as `z.real()` and `z.imag()`.  The member accessors `z.im` and `z.re` are not public---I was just used to seeing them in the internal code.  I'm OK with however this is done, but we probably need vectorized versions, so something like

```
vector real(complex_vector x);
```

or whatever we want to call it (given that as @rok-cesnovar points out, `real` is a reserved word).

> (@rok-cesnovar) how to read in complex types from data files ... I guess an array of length 2 is the obvious solution here. 

That was what I was imagining

Also, we're going to want a constant for `i` because it shows up in arithmetic.  Don't know what to call that as we can't exactly reserve `i` or `I`.

rok-cesnovar:
> but we probably need vectorized versions, so something like

Yes, I agree.

> Don't know what to call that as we can't exactly reserve i or I.

The current use of Stan is as nullary functions, so: `e()`, `machine_precision()`. So I guess `i()`?

bob-carpenter:
> So I guess i()?

D'oh.  Of course.  We can have those without reserving `i`.

SteveBronder:
How about `im()`?

bob-carpenter:
It should be `i()`, not `im()`---that's the universal mathematical notation.

SteveBronder:
Gonna close this and open a new one @rok-cesnovar can you file an issue in stan math about the piece you said we are missing for complex there? Then I think once we have that sorted I can write a new one of these

rok-cesnovar:
Sure thing!

PR #669, opened 2020-08-09
New array type syntax

rybern:
This PR implements the array syntax discussed at https://discourse.mc-stan.org/t/new-array-declaration-syntax/16011. The old syntax is still allowed.

An important thing to note about the new syntax is that **`array` is a new keyword**. I don't know how big of a problem that is. I miiiight be able to avoid it if it's a big deal, but for now there's an informative error message: `Expected identifier, but found reserved keyword 'array'`

The following program compiles:

```
functions {
  array[,,,] matrix f(array[,,,] matrix m, array[] int d) {
    return m;
  }
}
transformed data {
  array[1, 2, 3, 4] matrix[1, 2] m;
  array[4] int d;
  m = f(m, d);
}
```
I also updated the pretty printer to default to the new syntax, so running `--auto-format` on the example above returns the exact same program.

I looked into tagging the AST with the style that the array type was written in like @nhuurre suggested, but it would be a big organizational change for unsized types (it would be easy for sized types). I think that when we decide to put out depreciation warnings, we can either make that big change or we can just emit warnings directly from the parser (which isn't hard).

This supersedes #560 and #561, because those implemented the other array format.


src/frontend/parser.mly
```diff
@@ -218,61 +232,125 @@ unsized_dims:
...
+ *
+ * The value returned is a function from a bool (is_global, which controls
+ * whether the declarations should be global variables) to a list of statements
+ * (which will always be declarations).
+ *
+ * The rules match declarations with/without assignments, with/without array
+ * dimensions, single/multiple identifiers, and dimensions before/after the
+ * identifier.
+ *)
+decl(type_rule, rhs):
```
seantalts, line 265: 🤯 This is super slick! I don't know if this would have occurred to me. Nice DRYing up of the code :)

What do you call these rules that take arguments? Just wondering for my own edification.

src/frontend/parser.mly
```diff
@@ -218,61 +232,125 @@ unsized_dims:
...
+  (* When dims are after identifier, do not allow multiple identifiers *)
+  | ty=type_rule id=decl_identifier dims=dims rhs_opt=optional_assignment(rhs)
```
seantalts, line 267: So this is matching the new syntax and the next production is matching the old syntax, right? I would love a comment that says that and just gives a concrete example - would help me reading this code again later. Could also mark the old syntax with some "TO BE DEPRECATED" string we can then search from when Stan 3.0 does come out and break things.

src/frontend/parser.mly
```diff
@@ -218,61 +232,125 @@ unsized_dims:
...
+        }
+    })
+    }
+  (* Array dimensions option must be inlined, else it will conflict with first
+     rule. *)
+  | dims_opt=ioption(arr_dims) ty=type_rule
+      id_rhs=id_and_optional_assignment(rhs) SEMICOLON
+    { (fun ~is_global ->
+      (* map over each variable in v (often only one), assigning each the same
+         type. *)
```
seantalts, line 289: what `v` and `map` is this referring to?

src/frontend/parser.mly
```diff
@@ -218,61 +232,125 @@ unsized_dims:
...
+top_var_decl_no_assign:
+  | d_fn=decl(top_var_type, no_assign)
+    { grammar_logger "top_var_decl_no_assign" ;
+      d_fn ~is_global:true
     }
 
 sized_basic_type:
   | INT
-    { grammar_logger "INT_var_type" ; SizedType.SInt }
+    { grammar_logger "INT_var_type" ; (SizedType.SInt, Identity) }
```
seantalts, line 345: What are these `, Identity` used for? Having trouble tracking it down.

src/frontend/parser.mly
rybern: Not sure what they're officially called, I think Menhir refers to them as formal parameters

src/frontend/parser.mly
rybern: Whoops, this comment is actually for the multiple declarations version. I actually started with the multiple declarations version and removed the multiple declarations capability, so that I wouldn't have to redo all the refactoring I did there.

src/frontend/parser.mly
rybern: They're transformations, they're set to non-identity in e.g. top_var_type

rybern:
Thanks Sean!

> Can't wait for tuples (though I'm slightly sad we didn't want named records)

Don't see why records couldn't be next! They'd sure be easy after tuples.

bob-carpenter:
> though I'm slightly sad we didn't want named records

We do want named records, aka structs.  I just figured it'd be a lot harder to design in a first pass and tuples are the limiting case where the names are 1, 2, 3, ...

seantalts:
I'm trying to go through and resolve the conflicts here... what's your normal strategy on that for parser.messages, since all of the autogenerated stuff changes every time a parser change happens?

seantalts:
I'm trying now to use `dune build @update_messages` which seems reasonable, though it adds back in all of the autogenerated stuff I had deleted, haha. I see now why that's tricksy.

rybern:
Yeah, unfortunately that is what I typically do

seantalts:
I'm writing a Python script to automate some of this, haha.

seantalts:
is something wrong with github on this one? It's saying there's a lot of changes in this PR that I thought were in master already...

rybern:
Oh, not sure, is it from these new commits?

seantalts:
I think it might be, let me try to change the base and change it back - apparently that fixes it sometimes.

seantalts:
I think we might have to open a new PR... I can't figure out how to fix it but the command line shows it having the appropriate diff

rybern:
Yeah, when I run `git diff master --name-only | wc -l` I get 180 files changed, but GitHub says 260, is that what you're seeing?

seantalts:
yep

rybern:
Okay, closing to try to fix the diff issue..

PR #670, opened 2020-08-10
New syntax for multiple identifiers/definitions in one declaration statement

rybern:
This PR introduces new syntax for declaring and defining more than one variable with the same declaration statement, given that they have the same type.

This builds on #669 and supersedes #561.

The following program is parsed:
```
data {
  real x, y, z;
  array[5] int i, j;
}
transformed data {
  real a = 5, b = 6;
  real c = 5, d, e = 7;
}
```
Running it through `--auto-format` produces this equivalent program:
```
data {
  real x;
  real y;
  real z;
  array[5] int i;
  array[5] int j;
}
transformed data {
  real a = 5;
  real b = 6;
  real c = 5;
  real d;
  real e = 7;
}
```


rybern:
Could also do this for `~` statements. In the golf case study there's:
`[sigma_angle, sigma_distance, sigma_y] ~ normal(0, 1);`
which could be:
`sigma_angle, sigma_distance, sigma_y ~ normal(0, 1);`

src/frontend/dune
```diff
@@ -22,26 +22,25 @@
...
-  (run menhir %{dep:parser.mly} --update-errors %{dep:parser.messages}))
-)
+  (run menhir %{dep:parser.mly} --update-errors %{dep:parser.messages})))
 
-(alias
- (name update_messages)
-  (action (diff parser.messages parser_updated.messages)))
+(rule
+ (with-stdout-to parser_updated_trimmed.messages
+  (run python3 %{dep:strip_redundant_parser_state.py} %{dep:parser_updated.messages})))
```
rybern, line 29: @seantalts what do you think about calling python3 explicitly here, instead of relying on the shebang? The issue being that my python installation isn't where the shebang points

rybern: BTW, I used the script to update the messages for this branch and it worked great 👍

seantalts:
Ah nice!

I had issues with dune - it didn’t like the entire dune file if any of the
commands weren’t found on parsing it. We could switch back; then it might
make sense to put that back under runtest and enabled_if the relevant
stanzas. Or you could make symlinks to where “normal people” keep their
snakes 😂 Or I possibly fucked up the location. All great options.

rybern:
Ah gotcha. It's probably the most annoying part of my hipster distro. I'll revert it for now and find a workaround

I think this branch is ready for review if you get the chance, although it'll be a lot easier to see the diff once the array syntax PR is merged

rybern:
Okay, diffs are fixed now (had to work around the github bug [here](https://github.com/isaacs/github/issues/750)) so this is ready for review.

rok-cesnovar:
@rybern can you merge current master in here and we can discuss how to proceed here?

rybern:
@rok-cesnovar All merged up

seantalts:
I can take a look at this this weekend. Thank you!

src/frontend/parser.mly
```diff
@@ -276,8 +276,19 @@ decl(type_rule, rhs):
...
+
+  (* (\* Array dimensions option must be inlined, else it will conflict with first
+   *    rule. *\)
+   * | dims_opt=ioption(arr_dims) ty=type_rule
+   *     vs=separated_nonempty_list(COMMA, id_and_optional_assignment(rhs)) SEMICOLON
+   *   { (fun ~is_global ->
+   *     (\* map over each variable in v (often only one), assigning each the same
+   *        type. *\)
+   *     let dims = Option.value dims_opt ~default:[] in
+   *     List.map vs ~f:(fun (id, rhs_opt) -> *)
```
seantalts, line 290: Does this comment and code still need to be here?

seantalts (approved):
I think one errant comment, otherwise looks good!

I am so happy we have Menhir and scripts to properly diff and update parser.messages. Is it easier now? Too bad we can't teach github (or Menhir) to ignore e.g. changes in state machine state id numbers, lol.

rybern:
Thanks Sean!

Messages are definitely a bit easier with the script, yeah.

I was actually just messing around with messages again after I ran into some issues updating the tuples branch, and I discovered that menhir's gotten some nice updates: https://gitlab.inria.fr/fpottier/menhir/blob/master/CHANGES.md
We could definitely make use of the --merge function there.
I was also thinking that we could scrub the comments before committing and regenerate them when needed; might make it more readable with smaller diffs.

seantalts:
The `--merge-errors` one, right? Are you thinking for when two people are working on the parser at the same time?

rybern:
@seantalts 
Yeah `--merge-errors` is the main one I looked at. They have a little demo project with some message management utils that look nice: https://gitlab.inria.fr/fpottier/menhir/-/blob/master/demos/calc-syntax-errors/Makefile.messages.maintenance

PR #687, opened 2020-09-04
Add vectorised signatures - falling_factorial, fdim, fmax, fmin, fmod

andrjohns:
Continuing the addition of vectorised signatures, this PR adds signatures and tests for:
 - ```falling_factorial```
 - ```fdim```
 - ```fmax```
 - ```fmin```
 - ```fmod```


rok-cesnovar (changes requested):
The stanc3 side of this PR is fine. There is however an issue that some of the `fmin`/`fmax` signatures fail in the expressions tests. I think all of them fail for the same reason. The expressions tests matches output values to see if expression results match non-expression ones. And in some cases for `fmax`/`fmin` the outputs are `nan`, so I imagine this is the `nan != nan` issue. You can run the expressions tests with these new signatures by running

```
./runTests.py test/expressions --only-function new.sigs
```

So before this gets merged we need to either add those signatures to the exception list in Math or fix this issue. Or leave fmin/fmax out of the PR and add it later.

andrjohns:
That's concerning! Can I just clarify how to run the expression tests?

I'm guessing this is run in the math repo:
```
./runTests.py test/expressions --only-function new.sigs
```
What's the ```new.sigs``` part of it? Is that pointing to file with a list of signatures to test?

rok-cesnovar:
> That's concerning!

Its not that concerning, the expression testing framework needs to be extended. Though this also uncovered what we think is a Math bug. The first problem here is that the expression testing framework uses the same values for all inputs and in fwd mode it returns nan for `fmin(x,x)` which discussing with @t4c1 is a bug. And since nan != nan, those tests failed. Which in this case is great as it uncovered this bug.

> What's the `new.sigs` part of it?

I am sorry, I was not clear here. That is a file that I create locally and put in all the signatures that are added with a PR and run the expressions tests.

t4c1:
> which discussing with @t4c1 is a bug.

The fact that it returns `NaN` is not a bug. It is, however, inconsistent with what rev mode calculates. `d fmin(x,y) / d x` is undefined if `x == y`, so fwd returns `NaN`. But in the limits if we are approaching the point `x == y` from either side we get either  0 or 1, one of which is pragmetically returned by rev implemetation.

> What's the new.sigs part of it?

You can also directly specify signatures on the command line, such as `"matrix fmin(real, matrix)" "matrix fmin(real, matrix)"`, or create a file and put in one signature per line.

t4c1:
A pushed changes that needed to be done to expression testing framework to enable specifying different values for different argumets here: https://github.com/bstatcomp/math/tree/expression_test_independent_arg_values

andrjohns:
Ah, good to hear this isn't a binary vectorisation bug. So it looks like the Math problem is that the ```rev``` and ```fwd``` implementations of these functions don't handle edge cases in the same way. I can open an issue and fix that.

 > enable specifying different values for different argumets here

How would the command-line call look for that?

t4c1:
I already opened the issue here: https://github.com/stan-dev/math/issues/2060

t4c1:
You will need to specify the values to test with in the table: https://github.com/bstatcomp/math/blob/5925700c275156d4c2c3bd76fd668d72096a550e/test/expressions/generateExpressionTests.py#L130

andrjohns:
@rok-cesnovar Now that the fix has been merged in math (thanks!), this is passing expression tests

rok-cesnovar:
Great, thanks!

rok-cesnovar (approved):

Issue #649, opened 2020-07-21
Remove lower, upper, offset, multiplier from reserved keywords


rok-cesnovar:
This popped up in https://github.com/stan-dev/stanc3/issues/620 and is a backward compatibility issue with stanc2. 

See discussion in the linked issue.

Issue #671, opened 2020-08-10
CI should check that parser messages are complete


rybern:
When I went to update the parser messages in #669, I found that the parser was already missing some messages. We should include that in our testing.

The relevant commands are:
``` bash
cd src/frontend

# Produce a parser message file with updating error state numbers
menhir parser.mly --update-errors parser.messages > parser_updated.messages
# Check for differences between the updated messages and the current messages
# This should produce no output
diff parser_updated.messages parser.messages

# Produce a complete parser message file for the current parser
menhir --list-errors parser.mly > parser_new.messages
# Check the current parser message file against the complete one, issues to stderr
# This should produce two identical lines to stderr that look like: 
#    Read 534 sample input sentences and 534 error messages.
menhir parser.mly --compare-errors parser_new.messages --compare-errors parser.messages
```

Issue #648, opened 2020-07-21
The .* operator should work with scalars


rok-cesnovar:
This is a backward compatibility issue that popped up now that rstan checks compatibility with stanc3. The following model works with stanc2:

```
data {
    real p;
    int i;
}
transformed data {
    real p1 = p .* p;
    real p2 = i .* p;
    real p3 = p .* i;
    int i1 = i .* i;
}
```

should both work.

rok-cesnovar:
There are a few ways of fixing this, none of which are ideal. In my opinion, ideally, `.*` would not be allowed with scalars, but if we don't proclaim this as a bug of stanc2, we should support this.

The problem we have is that the way stanc2 handled this was a bit weird. While `real p1 = p .* p;` was allowed, `real rr = elt_multiply(r, r);` was not. And the latter failed with this message:
```
SYNTAX ERROR, MESSAGE(S) FROM PARSER:
No matches for: 

  real .* real

Available argument signatures for operator.*:

  vector .* vector
  row_vector .* row_vector
  matrix .* matrix
```
which is obviously false as `real .* real` with the operator is allowed. So maybe this was not intentional and is a bug?
If not we have a few options:

a) handle things the way stanc2 did (with a better message)

b) handle things the way stanc2 did (with a better message) and deprecate .* for scalars

c) support `elt_multiply(scalar, scalar)` which will automatically also support `scalar .* scalar`. This does need a Math update (adding c++ functions for elt_multiply).

@bob-carpenter do you have any thoughts on this? Is the behavior of stanc2 intentional?

bob-carpenter:
> Is the behavior of stanc2 intentional?

Yes.  It was intended to support `vector .* real` and `real .* vector` too and same for `row_vector` and `matrix`.  That makes `.*` work a lot like `*` in R.

> @bob-carpenter do you have any thoughts on this?

I'm OK with (a) or (b).  As far as I can tell, (c) is an implementation detail for (a).

seantalts:
(c) seems like a good route to me. Would that mean we should move this issue to stan-dev/math?

rok-cesnovar:
Yeah lets go with (c). Will make a Math issue tomorrow. Lets leave this one here as it will also then need a stanc3 touch up.

PR #697, opened 2020-09-18
Unreserve constraint keywords "upper" "lower "multiplier" "offset"

seantalts:
Fixes https://github.com/stan-dev/stanc3/issues/649

Also took a machete to the error messages. I think the result is a lot easier to modify and keep up to date...

## Release notes

"offset, "multiplier," "lower," and "upper" are no longer reserved keywords and can be used in the names of identifiers. 

## Copyright and Licensing


src/frontend/parser.mly
```diff
@@ -110,18 +110,21 @@ generated_quantities_block:
   | GENERATEDQUANTITIESBLOCK LBRACE tvds=list(top_vardecl_or_statement) RBRACE
     { grammar_logger "generated_quantities_block" ; tvds }
 
+build_id(ID):
+  | ID { fun id -> (
+           grammar_logger ("identifier " ^ id);
+           {name=id; id_loc=Location_span.of_positions_exn $startpos $endpos})}
```
rybern, line 116: I think this is equivalent to defining a normal ocaml function and calling that. You can define ocaml functions at the top of the file (see `reducearray`). Then you can call that function where you're currently calling `f`, and also in the `id=IDENTIFIER` rule

seantalts: I tried and it didn't have access to `$startpos` and `$endpos`, tho I guess the rest of this would possibly work if we passed those in at each use-site? Would that be better?

rybern:
It looks like there are two states missing from the error messages, if I run: 

> menhir --list-errors parser.mly > parser_new.messages ;
> menhir parser.mly --compare-errors parser_new.messages --compare-errors parser.messages 

the result is:

> File "parser_new.messages", line 12739, characters 0-55:
> Error: this sentence leads to an error in state 131.
> No sentence that leads to this state exists in "parser.messages".
> File "parser_new.messages", line 5447, characters 0-87:
> Error: this sentence leads to an error in state 475.
> No sentence that leads to this state exists in "parser.messages".

src/frontend/parser.mly
rybern: Yeah, you'd have to pass in $startpos and $endpos. I do think that would be a little bit better

seantalts:
Woops, good catch, thank you! I'll fix that and also see if I can add a test to automatically check that...

rok-cesnovar:
I think  https://github.com/stan-dev/stanc3/issues/671 has all the code we need to test it?

seantalts:
Nice! Didn’t see that, tho the commands are the same from the awesome wiki
Ryan wrote about updating messages. This PR adds Dune tests to the front
end to run these commands and check for completeness.

seantalts:
Woops, I forgot to go and change to the OCaml function - will do that tonight.

seantalts:
When I try to do that, 
```
let build_id name start_pos end_pos =
  grammar_logger name;
  {name; id_loc=Location_span.of_positions_exn start_pos end_pos}
```
```
  | id=IDENTIFIER { build_id ("identifier " ^ id) $start_pos $end_pos }
```
I get
```
File "src/frontend/parser.mly", line 119, characters 61-62:
Error: Unbound value $
```

rybern:
@seantalts it's just a bad error message, the problem is the underscores in $start_pos and $end_pos, should just be $startpost and $endpos

seantalts:
🤦 🤦 🤦

bgoodri:
With this PR are there any keywords that are reserved now that weren't reserved in 2.19.x?

rok-cesnovar:
I think this only leaves target.

rybern (approved):
LGTM!

seantalts:
🥳 thanks!

seantalts:
@rybern @rok-cesnovar @bgoodri do folks want to unreserve target as well? I haven't been to any meetings in a while, not sure if that's been talked about.

rok-cesnovar:
I havent seen that one reported ever. So I dobt think its needed. Offset/multiplier were probabyl the most problematic.

In terms of stanc2 compliance and readiness for rstan I think we are down to:

https://github.com/stan-dev/stanc3/issues/648 And
https://github.com/stan-dev/stanc3/pull/524

rybern:
I don't think we should unreserve target, I think that would be confusing

PR #698, opened 2020-09-19
stancjs arguments and testing

### LINKED ISSUE #557: arguments to javascript version
The javascript version of stanc3 does not seem to have the ability to pass optional arguments such as `--allow_undefined`. This causes otherwise valid Stan programs that provide C++ definitions of user-declared Stanfunctions to not parse. This is also relevant to #500 where we need to pass an argument in order to generate standalone functions.

More generally, it seems like the javascript version should accept all the same arguments as the OCaml version.

### LINKED ISSUE #625: stancjs tests
We should add tests for stancjs. 

At minimum tests that would compile all /good models. 

I think the easiest way is to use nodejs.

Example:

```
var stanc = require('path/to/stancjs.bc.js');

var model = `
parameters {
    real y;
}
model {
    y ~ normal(0, 1)
}
`

var test_model_res = stanc.stanc("testing", model);
if (test_model_res.errors) {
    console.error("ERROR:")
    console.error(stantest_model_resfuncs_test.errors)
}
if (test_model_res.warnings) {
    console.warn("WARNINGS:")
    console.warn(test_model_res.warnings)
}
```
Run with `node filename.js` and check if there is anything on stderr.

rok-cesnovar:
Fixes #557  and fixes #625 

stancjs now accepts: 
```
--warn-uninitialized
--allow-undefined (and --allow_undefined)
--warn-pedantic
--version
--O
--use-opencl
--auto-format
--print-canonical
```

The rest of the flags are not applicable in the case of stancjs:
```
--debug-lex
--debug-parse
--debug-ast
--debug-decorated-ast
--debug-generate-data
--debug-mir
--debug-mir-pretty
--debug-optimized-mir
--debug-optimized-mir-pretty
--debug-transformed-mir
--debug-transformed-mir-pretty
--dump-stan-math-signatures
--o
--print-cpp
--include-paths (and --include_paths)
--name
```

This also adds unit tests for stancjs. In order to run them, you need to have `node`.
The tests can be run with:
`dune build @runjstest --profile release`

## Release notes

Stancjs: added argument handling and testing.

## Copyright and Licensing


src/stancjs/stancjs.ml
```diff
@@ -5,8 +5,28 @@ open Analysis_and_optimization
...
+    ^ "' may not have been initialized.\n"
+  in
+  let filtered_uninit_vars =
+    Set.filter ~f:(fun (span, _) -> span <> Location_span.empty) uninit_vars
+  in
+  Set.Poly.to_list
+    (Set.Poly.map filtered_uninit_vars ~f:(fun v_info -> show_var_info v_info))
+
+let stan2cpp model_name model_string flags_string =
+  let flags = String.split_on_chars ~on:[' '] flags_string in
```
rok-cesnovar, line 24: For now, the arguments are a string with space-delimited flags. It would be more Javascriptish if it were an array of string. Are "--" necessary in that case?

src/stancjs/stancjs.ml
```diff
@@ -22,29 +42,42 @@ let stan2cpp model_name model_string =
...
+           let tx_mir = Transform_Mir.trans_prog mir in
+           let opt_mir =
+             if is_flag_set "--O" then Optimize.optimization_suite tx_mir
+             else tx_mir
+           in
+           let cpp = Fmt.strf "%a" Stan_math_code_gen.pp_prog opt_mir in
+           if is_flag_set "--warn-uninitialized" then
+             Pedantic_analysis.print_warn_uninitialized mir ;
+           if is_flag_set "--warn-pedantic" then
+             Pedantic_analysis.print_warn_pedantic mir ;
```
rok-cesnovar, line 62: For now these warnings are all printed and not returned. I want to also refactor that, but that is a separate issue/PR. We could return this as part of warnings, but we still have some other warnings that would get printed (like integer division). So a separate PR to unify that would be better.

test/stancjs/run_js_on_args.ml
```diff
@@ -0,0 +1,16 @@
...
+  let chns = [stdout; stderr] in
+  let out = List.map ~f:In_channel.input_lines chns in
+  ignore (Unix.close_process_full (stdout, stdin, stderr)) ;
+  String.concat ~sep:"\n" (List.concat out)
+
+let () =
+  let files = Array.(sub Sys.argv ~pos:1 ~len:(length Sys.argv - 1)) in
+  Array.stable_sort ~compare:String.compare files ;
+  Array.iter files ~f:(fun arg ->
+      let cmd = "node " ^ arg in
```
rok-cesnovar, line 15: This requires node to run. But it does not make node a dependency if you develop for stanc/stan2ftp.

rok-cesnovar (commented):
A couple of comments. @seantalts you probably know the stancjs the most. Mind taking a look for some feedback. No rush of course.

src/stancjs/dune
```diff
@@ -1,7 +1,8 @@
 (executable
   (name stancjs)
   (libraries js_of_ocaml frontend stan_math_backend)
-  (preprocess (pps js_of_ocaml-ppx)))
+  (preprocess (pps js_of_ocaml-ppx))
+  (modes js))
```
seantalts, line 5: what does this modes js do? this breaks on my machine locally, but maybe I need to upgrade dune?

seantalts: I upgraded dune and it works now!

src/stancjs/stancjs.ml
seantalts: I don't think they're necessary. I will push an attempt at the array version but I can't get the tests to run locally - any tips?

seantalts:
Okay, got the tests running locally, fixed the array syntax, and added these new JS tests to Jenkins (thank you!!!). For now I added node to our standard ubuntu dockerfile out of laziness, but maybe we should have a separate Docker image or file for just the JS tests? I guess the benefit would be that we can avoid accidentally depending on node or the js_of_ocaml libs in some non-JS code. What do you think?

rok-cesnovar:
Thanks for setting up the Docker stuff. I would also rather keep this separate. I was actually thinking about adding a Github Actions job, looking at it, it should be simple.

Going to add the two missing flags today then there is only one more thing here. The fact that stanc("name", model) would fail now without args. And given that rstan already uses that, that would cause issues.

I played around with the default args Ocaml stuff but failed. Maybe adding a stanc() and stanc3() or stanc_args() I dont know.

seantalts:
> Going to add the two missing flags today then there is only one more thing here. The fact that stanc("name", model) would fail now without args. And given that rstan already uses that, that would cause issues.

I fixed this with the array changes, I think (check out the js test file changes as well). Now the 3rd argument is a Js.(js_string t js_array t opt). lol.

What's the pitch on github actions? Is it like free compute? Does it use Docker?

src/stancjs/stancjs.ml
```diff
@@ -22,29 +45,44 @@ let stan2cpp model_name model_string =
...
+let stan2cpp_wrapped name code (flags : Js.string_array Js.t Js.opt) =
+  stan2cpp (Js.to_string name) (Js.to_string code)
+    Js.(
+      Opt.map flags (fun a ->
+          str_array a |> Js.to_array |> Array.map ~f:to_string )
+      |> Opt.to_option)
+  |> wrap_result
 
-let () = Js.export "stanc" (wrap2 stan2cpp)
+let _ = Js.export "stanc" stan2cpp_wrapped
```
seantalts, line 88: Woops, I think this should stay `let () = ` no? The style guide doesn't like `_` unless necessary and then with type annotations (e.g. `let (_ : unit) =` (but `let () = ` is more concise).

src/stancjs/stancjs.ml
seantalts: (I got the tests to run!)

But not sure if we need to include the "--". I can see arguments either way. Happy with whatever you decide here.

rok-cesnovar:
> I fixed this with the array changes, I think (check out the js test file changes as well). Now the 3rd argument is a Js.(js_string t js_array t opt). lol.

Niiiiice!

> What's the pitch on github actions? Is it like free compute? Does it use Docker?

Yes, its free compute for open source, native to Github, and for these simple and short jobs its really great. An example of a workflow: https://github.com/marketplace/actions/set-up-ocaml#example-workflow

I found it a bit easier to handle compared to Docker, but that is probably because I am not super well versed in docker. 
Seeing as this now works no need to explore other stuff.

Thanks for the help!

rok-cesnovar:
Everything should be ready now for a review and merging. Will open an issue to rework the warnings once this is in. @seantalts 
 thanks again for the help!

seantalts:
Thank you! 

PS I am very happy to move stuff onto github actions as applicable - free is good! We have a huge AWS bill now that we've run out of credits :(

seantalts (approved):
Super cool, thank you so much! Very happy to have the js stuff under some tests now. I really didn't see the js version going as it has... haha.

PR #708, opened 2020-09-30
New array type syntax (Fixed)

rybern:
Please see #669 for a description and discussion history for this PR. I'm reopening it to fix a GitHub issue with a bad diff.


seantalts:
I think this test error is (finally) legit

rybern:
Yep, I clobbered the messages, should be fixed

rybern:
@seantalts 
I thought a bit more about avoiding the array keyword. I don't think the parser massaging I was thinking of will work, because I forgot that indices are often not just lists of integers like array sizes should be, so that's a whole other ambiguity. I don't think it's worth that bigger mess.

So I think the options are: 1) reserve the keyword, or 2) parse the array[..] grammar as though it were an indexed expression, and then give a custom error message if it does not exactly match the pattern of "array" with integer sizes. Kind of an uncomfortable hack for a few reasons, but not hard to write.

seantalts:
Hmm... How bad is the hack? It would require Semantic_check.ml to be aware of this, right?

rybern:
Not necessarily, depends on where and how we end up handling the error. To handle it in Semantic_check, we'd probably have to encode array dimensions in the AST as indexed expressions, which would be further spreading the very bad abstraction. Instead I think we'd keep the the AST the same, but throw an error during parsing. The parser should call existing error printing code in Error.ml, just like it would do if the parser itself were actually finding the error.

seantalts:
Ah that makes sense. I think I'm okay with that hack, tbh - seems pretty localized. And it does seem non-trivial to not reserve the word "array."

rybern:
We decided to go with #714 instead.

PR #714, opened 2020-10-07
New array type syntax - no reserved keyword

rybern:
Alternate version of #708 

@seantalts have a look at the diff here.

It really is a hack, but as we said it's fairy contained.

I'm not 100% confident in my test coverage here.


seantalts:
I think this approach is good with me for now - doesn't seem like a huge amount of technical debt (30 lines?) and its very localized in not just one file but basically just 1 semantic action, more or less. Is this good with you? If so happy to review and we can move on to tuples!

seantalts:
Took a crack at merging in master and updating the messages. The last commit was a new message that I wasn't quite sure what to do with yet, but I have to run for now. Feel free to revert any/all of my commits here

seantalts:
I was messing around with the automation here and I think it's a lot cleaner now (I hope) - two python scripts, add_new_messages.py (changed since you last saw it to be part of the dune run and thus a little cleaner) and strip_redundant_parser_state.py, which I've added to @update_messages. I also got rid of that parser_comparison.expected file; it wasn't working correctly but we also didn't need it :)

In the last commit I went crazy and decided maybe we don't need a separate @update_messages and we can just run those actions as part of runtest. What do you think about that? It seems like with the new scripts, the messages file will be kept as we like it automatically now, I think. Though I'm sure there are edge cases I'm not yet thinking of.

seantalts (approved):
Looks good to me! Thank you so much, excited for this and tuples! 

There's a freeze on so not merging yet; also want to make sure you're okay with the stuff I did (mostly around scripting parser message stuff). I also changed the location of the new error messages slightly, I like this a little better but if you prefer the other that's fine too.

seantalts:
@serban-nicusor-toptal Could you take a look at these builds? They're failing with a Docker error in this PR and in master. Thanks!

serban-nicusor-toptal:
Base image of debian doesn't have `python3` as needed in here [src/frontend/dune](https://github.com/rybern/stanc3/blob/array-type-syntax-no-keyword/src/frontend/dune#L29). Master branch looks fine to me as it doesn't have `/src/frontend/dune`.
I will commit in a moment an updated Dockerfile.

serban-nicusor-toptal:
Hey @seantalts python3 error seems to be fixed, it's now running into:
```
File "src/frontend/dune", line 27, characters 0-142:

27 | (rule

28 |  (with-stdout-to parser_updated_trimmed.messages

29 |   (run python3 %{dep:strip_redundant_parser_state.py} %{dep:parser_updated.messages})))

Error: No rule found for src/frontend/strip_redundant_parser_state.py

      menhir src/frontend/parsing_errors.ml
```

Please let me know if I can help with anything else.

seantalts:
Thanks Nic :) 

I'm wondering if we should allow the tests to require Python 3, or just move that functionality back into the separate @update_messages build target... @rybern do you have thoughts on that? I'm a bit torn.

rybern:
@seantalts Sorry for the delay. I'm slowly unburying myself..

I would guess that Python3 is ubiquitous and would be fine to assume for tests, but since since it is a dependency and the different in UX is pretty small, I'm happy to keep them separate like you have it.

Do you think this is a good time to merge this?

seantalts:
Yeah, I’m happy with this. We can iterate on the testing and message update situation in future PRs; it’s already testing more and automating more than the old and I think it’s pretty decent.

For future reference: If you like it being part of the runtest (seems nice in some ways) then maybe the easiest thing to do is use enabled_if to only enable all of the related targets if profile<>static...

seantalts:
I understated my happiness- I’m really excited about this PR! Can’t wait for tuples. If you make a post on discourse or something and tag me, Steve, others about what there is to do and what you could use help with I bet people would volunteer. Thanks for your work!

rok-cesnovar:
Very nice! :tada:

rybern:
Sounds good, I'll merge when CI passes. Thanks a bunch @seantalts!

This frees up the multiple declaration syntax, I'll work on updating that next.

seantalts:
@rybern I forget where we were talking about this, but we might be able to use `#!/usr/bin/env python3` instead of `#!/usr/bin/python3` - the former should search the path for `python3`, assuming nix has that present it should work.

PR #716, opened 2020-10-08
Add the --filename-in-msg CLI argument 

### LINKED ISSUE #715: Add an option to not "overwrite" the filename in error messages
All cmdstan wrappers (including the julia wrapper) compile the models in temp folders. Which means that the compiler errors look like:

```
Syntax error in 'a04866f9e9.stan' line 3, column 23 to column 24, parsing error:
     1:  data {
     2:    real x;
```
The same name is mentioned in exceptions.

Same for rstan, but rstan will use stanc3 in Javascript. But we have a similar issue there where the error is:
```
Syntax error in 'string' line 3, column 23 to column 24, parsing error:
     1:  data {
     2:    real x;
```

Could we have a cmd line arg to overwrite this name with an actual name?

rok-cesnovar:
Fixes #715 

In addition to the compiler messages, this also fixes exception messages

Example
```
parameters {
    real y;
}
model {
    y ~ normal(0,1);
    reject("msg");
}
```
put it in something.stan and add `STANCFLAGS= --filename-in-msg=foo.stan` and the message will be
`
Rejecting initial value:
  Error evaluating the log probability at the initial value.
Exception: msg (in 'foo.stan', line 6, column 4 to column 18)
`

I always found those messages in rstan or cmdstan wrappers with "in a82fbd.stan" confusing.

## Release notes

Added the --filename-in-msg CLI argument to overwrite filename in messages.

## Copyright and Licensing


rok-cesnovar:
This is missing the stanc.js parts. Will add those once we settle on the argument name and if we even want this.

rok-cesnovar:
@jgabry @mitzimorris any thoughts on this? See the issue for more context.

jgabry:
Yeah the nonsense file names caused by using the temp directories has always bothered me too! If we add this then I think the wrappers should always use it automatically.

seantalts:
@rok-cesnovar I saw this before but thought it needed JS flags first?

rok-cesnovar:
Yes, I was waiting for any initial feedback if we even want to do this. If you are ok with this will add the js flag shortly. Thanks!

seantalts:
Seems good to me!

rok-cesnovar:
Should be ready now. Open to suggestion for better argument names. Thanks!

seantalts:
I was a little worried about include situations as well as someday forgetting that that flag affects how locations are printed, so I rewrite it to explicitly rewrite the program with the new filename in the Locations module. And if the file is included from some other file, it follows the chain and replaces only the base filename.

seantalts:
Wait, I actually kinda don't think stanc3 should be handling this at all, though I agree that the error messages for some of the interfaces are annoying. The way that an interface would handle that, though, would be to give whatever filename they would pass to this `--filename-in-msg` flag and just... name the file that.

rok-cesnovar:
> I was a little worried about include situations as well as someday forgetting that that flag affects how locations are printed, so I rewrite it to explicitly rewrite the program with the new filename in the Locations module. And if the file is included from some other file, it follows the chain and replaces only the base filename.

Thanks!

> Wait, I actually kinda don't think stanc3 should be handling this at all, though I agree that the error messages for some of the interfaces are annoying. The way that an interface would handle that, though, would be to give whatever filename they would pass to this --filename-in-msg flag and just... name the file that.

There are two different things here: 

- stanc.js can not do that as it only gets a string. There we could maybe just say that the first argument to `stanc()` is the model filename, not the model name. 

- for the other interfaces you are probably right that those could handle that on their own. The problem is that even if they name the model basename properly it will still be `Error in '/tmp/somethingRandom/model.stan'`, which is confusing. Though maybe the solution is to only print the basename of the model file. But that makes it harder to use this for error messages in text editors/IDEs. Exposing and using the `--o` flag and changing the subsequent build steps to build with the hpp in temp folder would work for cmdstanr/py/.jl , probably also for httpstan which pystan uses, but not sure there. Or regexing stderr messages. It should not be that bad. 

So a stanc.js fix is definitely needed, for other interfaces we can work around this some other way. Anyway, no rush here in any case. The Stanc.js fix can wait for after the freeze is over.

Thanks again!

seantalts:
Ah yeah, that's fair. Let me see if I can fix this so the semantic errors are changed too...

seantalts:
Okay, hopefully this one works. I left the regular stanc with the flag as well, why not.

I saw we released RCs - are we still doing the freeze on develop? If people actually actively test the RCs and they're cheap to make ( ? ) then maybe we don't need the feature freeze?

rok-cesnovar:
> are we still doing the freeze on develop?

Yes, based on the feedback on the Math issue I decided to retract my suggestion of changing the freeze period. It was a bit late for any changes anyways.

Will take a look at the test error. This can wait for after the freeze.

test/integration/cli-args/filename_good.expected
```diff
@@ -0,0 +1,329 @@
...
+using stan::model::index_multi;
+using stan::model::index_omni;
+using stan::model::nil_index_list;
+using namespace stan::math;
+using stan::math::pow; 
+
+static int current_statement__ = 0;
+static const std::vector<string> locations_array__ = {" (found before start of program)",
+                                                      " (in 'filename_good.stan', line 2, column 4 to column 11)",
+                                                      " (in 'filename_good.stan', line 3, column 4 to column 19)"};
```
rok-cesnovar, line 63: This should be foo.stan

test/integration/cli-args/filename_error.expected
```diff
@@ -0,0 +1,13 @@
+  $ ../../../../install/default/bin/stanc --filename-in-msg=foo.stan  filename_error.stan
+
+Semantic error in 'filename_error.stan', line 3, column 17 to column 18:
```
rok-cesnovar, line 3: This should be " Semantic error in 'foo.stan' "

rok-cesnovar (commented):
Two "note-to-self" comments.

test/integration/cli-args/filename_error.expected
seantalts: woops

rok-cesnovar: The stancjs tests pass so I think this is something minor.

seantalts:
Hey @rok-cesnovar, for normal stanc we need to make one other serious change for this to work correctly. Do you want to hop on a call at some point? I feel bad for hijacking this PR; I didn't realize how much work would be needed to make it work "properly" (arguable) and felt bad making you do the extra work for the sake of properness. I just feel from experience that global config should ideally not spread throughout the code but be processed with all the other I/O in the "driver" (stanc.ml, stancjs.ml) if possible. 

I think the next step for this is either
1. Give up on stanc.ml supporting this flag (and it does seem superfluous to me, though maybe someone can comment on why it's not), or
2. Change the program so that all files are only read in once at the beginning and stored for any later printing needs. I think we should have done this ideally in the beginning, but we could correct it now.

> The problem is that even if they name the model basename properly it will still be Error in '/tmp/somethingRandom/model.stan', which is confusing

Do we actually print the full filename? I thought we just printed the "model.stan" part.

rok-cesnovar:
> Do you want to hop on a call at some point?

Sure, just say when.

> 1. Give up on stanc.ml supporting this flag

I don't oppose this. The main reason for doing this PR was that stanc.js outputs the "error in 'string'" which is just bad haha. And since I was already doing this I thought I would also tackle the "compile in temp folder issue", because I kind of thought they would overlap. But now I see that doing this properly messes up a ton of stuff and I am not sure we need to do this. 

I was under the impression it would be easier here, but the regex in stderr seems simpler than this honestly :)

> Do we actually print the full filename?

We print the path supplied to the stanc binary. So either a relative or absolute path, depending what you supply. When working with tmp files it's usually the absolute path.

seantalts:
> But now I see that doing this properly messes up a ton of stuff and I am not sure we need to do this.

I would say that doing it properly causes us to need to clean up some previous technical debt :sweat_smile: Which I'm down to do; I'll take a crack at trying to limit I/O to the drivers and remove it from the printing of error messages.

seantalts:
Oh also, if you would prefer to do this I'm happy to let you handle it with as much or little support from me as you want; I have many other tech debt things to tackle here haha.

rok-cesnovar:
Ah sorry, forgot to reply here. Feel free to tackle this of course. It would be good to get this resolved for the rstan stanc3 release that will hopefully be coming soon. 

But I guess we can also release with the "error in 'string'" messages. Users have been seeing "Error in tmp/3423bacad.stan" and no one seemed to mind much :)

rok-cesnovar:
Closing in favor of #780

PR #717, opened 2020-10-08
Small Nix installation fix

rybern:
Fixing an issue with Nix installation that came up here: https://discourse.mc-stan.org/t/cmdstan-on-arm-error/18485/3

The issue occurred because of a workaround I added to simplify the shell.nix file, but it wasn't necessary so I reverted it.


rok-cesnovar (approved):
Looks good. Thanks!

PR #721, opened 2020-10-17
typo fix in a semantic error message

avehtari:
Typo fix in a semantic error message (+ corresponding test)
Probabilty -> Probability
No functional effect.


rok-cesnovar (approved):
Thanks!

Issue #600, opened 2020-06-29
Deprecation warnings are missing for functions


rok-cesnovar:
This model:

```
model {
  real foo;
  foo = multiply_log(1, 1);
}
```
issue a warning with STANC2:
```
DIAGNOSTIC(S) FROM PARSER:
Info: Function name 'multiply_log' is deprecated and will be removed in a later release; please replace with 'lmultiply'
```

but nothing with stanc3. 

stanc3 does handle some deprecation warnings for the uses of #, <-, increment_log_prob and get_lp.

We use a list of deprecated functions in the canonicalizer: https://github.com/stan-dev/stanc3/blob/master/src/frontend/Canonicalize.ml#L8

That list should be moved out of the canonicalizer mode and warn in regular use and replace in canonicalizer.
Also warn for the _log suffix, _log UDFs, add cov_exp_quad to the deprecated list in favor of gp_cov_exp_quad. There might be others.

Reference manual list of deprecated features: https://github.com/stan-dev/docs/blob/068bda7038671e89fb4574d2cc56a3f6e351d01f/src/reference-manual/deprecations.Rmd

adamhaber:
Hi,

> Reference manual list of deprecated features: 

Some followup questions:
1. "Replace `lkj_cov_log(...)` with an `lkj_corr` distribution on the correlation matrix and independent lognormal distributions on the scales. " - should this be included? 
2. `if_else` replaced with `a ? b : c` - presumably should be handled but not included in the list, since it's not a simple-replace-this-function-with-this-function? 
3. "Use of the `abs` function with real-valued arguments is deprecated; use functions `fabs` instead." - same question.

> That list should be moved out of the canonicalizer mode and warn in regular use and replace in canonicalizer.

Moved the list to ast.ml (not sure it's the best place). Do we want to repeat all the canonicalizer `replace_deprecated_X` with `warn_deprecated_X` in ast (or a better location)?

rok-cesnovar:
Hey,

for 1. I am not the best one to say (maybe @bob-carpenter , @bbbales2 )?

I am not sure why the canonicalizer doesn't handle 2 and 3. @nhuurre I guess 2 is not easy, any specific reason for 3? 
I personally think we should warn for both.

> Do we want to repeat all the canonicalizer `replace_deprecated_X` with `warn_deprecated_X` in ast (or a better location)

I think so yes. The location seems fine to me.

nhuurre:
[...]

PR #723, opened 2020-10-18
Cleanup tests of deprecated functions

rok-cesnovar:
## Release notes

Cleaned up distribution tests of deprecated _log functions.

## Copyright and Licensing


rok-cesnovar:
This is required for https://github.com/stan-dev/stanc3/issues/600 because these distribution test files would generate a few 100k deprecation warnings.

seantalts:
Did you run a script for this? Could you paste the script somewhere? Seems easier than reviewing the 800,000 lines of changes :D

rok-cesnovar:
I ran it with regex in Visual Studio code (first for continuous (lpdf) and discrete(lpmf). I can re-do this with a script if that would make it easier, not a problem. This is easy.

Otherwise, all the changes are inside the `test/integration/good/function-signatures/distributions/`: *.stan files and pretty.expected files.

I think the easiest check would be that there is no error in `test/integration/good/function-signatures/distributions/.*/pretty.expected`.

There is a catch though. There is one error in a pretty.expected file now. And it's legit:
https://github.com/stan-dev/stanc3/blob/master/src/middle/Stan_math_signatures.ml#L1509 and
https://github.com/stan-dev/stanc3/blob/master/src/middle/Stan_math_signatures.ml#L1517

should represent the same signature I suppose (given that in _log calls the _lpmf) version.
I didn't want to fix that here as no one can spot that. I also have to double-check which signatures actually work and are listed in docs.

rok-cesnovar:
Will re-do with a script. I realized this is impossible to review.

rok-cesnovar:
Also not sure why I didn't just use print-canonical... doh

There is no useful or necessary comments in these files.

PR #724, opened 2020-10-18
Cleanup deprecated in distribution tests

rok-cesnovar:
Another attempt, this time much cleaner.

For https://github.com/stan-dev/stanc3/commit/49bd2a43771731841d7dbf96a68aa833bc63356f I ran `--print-canonical` on all models in test/integration/good/function-signatures/distributions. A side effect of that is this deleted all comments in those models, but there were not many of them, none of them relevant.
There are a few in [this file](https://github.com/stan-dev/stanc3/blob/master/test/integration/good/function-signatures/distributions/rngs.stan), but nothing necessary. The rest of the files only had a comment like [this](https://github.com/stan-dev/stanc3/blob/master/test/integration/good/function-signatures/distributions/univariate/continuous/beta/beta_log.stan#L879), which also seems unnecessary.

The code to run that was: 
```python
import os
import sys
import re
import glob
import subprocess

def files_in_folder(folder):
    """Returns a list of files in the folder and all
    its subfolders recursively. The folder can be
    written with wildcards as with the Unix find command.
    """
    files = []
    for f in glob.glob(folder):
        if os.path.isdir(f):
            files.extend(files_in_folder(f + os.sep + "**"))
        else:
            files.append(f)
    return files

for f in files_in_folder("stanc3/test/integration/good/function-signatures/distributions/"):
    if f.endswith(".stan"):
        command = "cmdstan/bin/stanc --print-canonical " + f + " > tmp.stan"
        p1 = subprocess.Popen(command, shell=True)
        p1.wait()
        command = "cat tmp.stan > " + f
        p1 = subprocess.Popen(command, shell=True)
        p1.wait()
```

The two commits that followed are renames of files, done with:
```python
for f in files_in_folder("stanc3/test/integration/good/function-signatures/distributions/"):
    if f.endswith("_ccdf_log.stan"):
        f_new = f.replace("_ccdf_log.stan", "_lccdf.stan")
    elif f.endswith("_cdf_log.stan"):
        f_new = f.replace("_cdf_log.stan", "_lcdf.stan")
    elif f.endswith("_log.stan"):
        if "continuous" in f:
            f_new = f.replace("_log.stan", "_lpdf.stan")
        else:
            f_new = f.replace("_log.stan", "_lpmf.stan")
    else:
        f_new = f
    if f_new != f:
        command = "mv " + f + " " + f_new
        p1 = subprocess.Popen(command, shell=True)
        p1.wait()
```

Then I went and checked where _log is still used and renamed those files back manually. Those either are not deprecated or not handled by the canonicalizer.

## Release notes

Cleaned up deprecated functions in test models for distributions.

## Copyright and Licensing


rok-cesnovar:
@serban-nicusor-toptal this one is also occuring a lot currently: 

```
docker build -t 58bf8a26e98f0a813fe1402d348abf8e675a06c6 -f docker/static/Dockerfile .
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
script returned exit code 1
```

any idea?

serban-nicusor-toptal:
Hey @rok-cesnovar I've found the issue and fixed it, should not happen anymore.

rok-cesnovar:
Thanks, @serban-nicusor-toptal!

@seantalts this is ready.

seantalts:
Thanks! Nic what was the solution?

serban-nicusor-toptal:
Docker on the instances was started up by the init script which proved to be unreliable sometimes, I've configured the ami to now start docker at boot using [systemd](https://docs.docker.com/config/daemon/systemd/) as stated in the Docker docs.

seantalts:
Thanks both of you! @rok-cesnovar how difficult would it be for you to run the script with `git mv` instead of `mv`? I'm seeing some stuff not tracking that and I think if it's easy enough to re-run it might be better to do that. E.g. shows `normal_ccdf_log.stan` as deleted and I had to ctrl-F to check that we'd appropriately renamed it to `normal_lccdf.stan` https://github.com/stan-dev/stanc3/pull/724/files#diff-6d13160a1ca5dad21d10679a7f84e0748447b4c79f58814f64db6c3a106a13ae

But if it takes you more than 10 minutes it's probably not worth it...

rok-cesnovar:
> how difficult would it be for you to run the script with `git mv` instead of `mv`

Done, that was simple. It seems that didnt do much though. If you look at files changed it still looks as if files were deleted and created. However if you look at the commit it looks like a bunch of renames: https://github.com/stan-dev/stanc3/pull/724/commits/cd7d5a53cc59021d75cfcf5d24cd76fa064b0194 which it actually was.

So maybe it would be easier to review the 4 commits? :/ Sorry for the difficult review here.

rok-cesnovar:
@adamhaber you can merge this branch in and that should fix the issues with tests for the deprecation warnings.

seantalts:
You ran the `git mv` after the `mv` and that did something? That's interesting, haha. Anyway this is fine, thank you!! I think the script looks good to me, a few pokes in the results also look good, and the tests are passing so let's go with it.

seantalts (approved):

rok-cesnovar:
> You ran the `git mv` after the `mv` and that did something?

Nah, I reset the branch to the commit pre the mv commit. And then did git mv again.

Thanks for the review. Will merge once the freeze is over.

PR #726, opened 2020-10-20
Add function deprectation warning

adamhaber:
This addresses #600 - deprecated functions warnings are moved out of the canonicalizer mode, warn in regular use and replace in canonicalizer. I've tried to make as little code duplication as possible between `Canonicalizer.ml` and `Deprecation_analysis.ml`, but in some places it seemed clearer to be a little more repetitive. 

## Release notes

Added deprecated functions warnings to regular use, and replace in canonicalizer.

## Copyright and Licensing


rok-cesnovar:
Thanks Adam. Will take a look immediately after the release on Thursday.

adamhaber:
Now with 500K less loc to review :-)

test/integration/canonicalize/canonical.expected
```diff
@@ -53,6 +53,39 @@ generated quantities {
   real y_hat[3, 2] = integrate_ode_rk45(sho, y0, 0.0, ts, theta, x_r, x_i);
 }
 
+
+Warning: deprecated language construct used in 'deprecated.stan', line 55, column 20 to column 69:
+integrate_ode is deprecated, use integrate_ode_rk45 instead
```
seantalts, line 58: Can we make these one line instead for ease of grep downstream? Maybe something like

```Warning: integrate_ode is deprecated and will be removed in the future. Use integrate_ode_rk45 instead``` 
which is close to stanc, or a one line version with line num if you prefer. Don't need to say "deprecated language construct" I think.

File: src/frontend/Deprecation_analysis.mli [hunk 2]
```diff
@@ -0,0 +1,28 @@
...
+
+val userdef_distributions :
+     ( (located_meta, unit) expr_with
+     , located_meta
+     , ((located_meta, unit) expr_with, located_meta) lval_with
+     , unit )
+     statement_with
+     sexp_list
+     sexp_option
+  -> string sexp_list
```
seantalts, line 19: How did you get these types? Is there a way to generate an .mli file? I think merlin will sometimes show "sexp_list" etc but we don't need to write the `sexp_` prefix for these three.

src/frontend/Deprecation_analysis.mli
```diff
@@ -0,0 +1,28 @@
...
+val is_distribution : string -> bool
+val rename_distribution : string -> string
+val rename_function : string -> string
+
+val userdef_distributions :
+     ( (located_meta, unit) expr_with
+     , located_meta
+     , ((located_meta, unit) expr_with, located_meta) lval_with
+     , unit )
+     statement_with
```
seantalts, line 16: Does this have a short-hand type in the AST module?

src/frontend/Deprecation_analysis.ml
```diff
@@ -0,0 +1,168 @@
...
+  || String.is_suffix ~suffix:"_lpmf" name
+  || String.is_suffix ~suffix:"_lcdf" name
+  || String.is_suffix ~suffix:"_lccdf" name
+
+let userdef_distributions stmts =
+  List.filter_map
+    ~f:(function
+      | {stmt= FunDef {funname= {name; _}; _}; _} ->
+          if
+            String.is_suffix ~suffix:"_log_lpdf" name
```
seantalts, line 45: If you feel like doing some refactoring while you're in there (I know you didn't originate this code) there are a couple of ways to avoid typing String so many times: 
* `let open String in` at the top, or
* wrap the expression with `String.()`

src/frontend/Deprecation_analysis.ml
```diff
@@ -0,0 +1,168 @@
...
+  then String.drop_suffix name 4
+  else name
+
+let warn_deprecated (loc_span, message) =
+  Fmt.pf Fmt.stderr
+    "@[<v>@,Warning: deprecated language construct used in %s:@,%s@]@."
+    (Location_span.to_string loc_span)
+    message
+
+let replace_suffix = function
```
seantalts, line 73: This looks like it's also back in the old file.

src/frontend/Deprecation_analysis.ml
```diff
@@ -0,0 +1,168 @@
...
+    | FunApp (UserDefined, {name; id_loc}, l) ->
+        let newname = String.Table.find deprecated_userdefined name in
+        if Option.is_some newname then
+          warn_deprecated
+            ( emeta.loc
+            , "Use of the _log suffix in user defined function " ^ name
+              ^ " is deprecated, use " ^ Option.value_exn newname ^ " instead."
+            ) ;
+        FunApp (UserDefined, {name; id_loc}, List.map ~f:warn_deprecated_expr l)
+    | _ -> map_expression warn_deprecated_expr ident expr
```
seantalts, line 138: Hmm, there may be another way to iterate over the tree for warnings without needing to return the tree. Maybe fold_expression et al with unit accumulator?

seantalts (dismissed):
Hey, thanks for doing this! Some food for thought.

test/integration/canonicalize/canonical.expected
adamhaber: Sure. Do you prefer the location information to be removed, as well? I thought it's quite helpful.

src/frontend/Deprecation_analysis.mli
adamhaber: :-)

I tried `typed_statement` before, and it caused errors in `repair_syntax`. I now realize that this happens before `let typed_ast = ...`, and that `untyped_statement` works, thanks.

src/frontend/Deprecation_analysis.ml
adamhaber: Thanks, changed. I currently use `let open String in` in 4 different functions - maybe it's better to open in the global scope (hope I'm getting the terminology right)?

src/frontend/Deprecation_analysis.ml
```diff
@@ -0,0 +1,168 @@
+open Core_kernel
+open Ast
+open Middle
+
+let deprecated_functions =
+  String.Map.of_alist_exn
+    [ ("multiply_log", "lmultiply")
+    ; ("binomial_coefficient_log", "lchoose")
+    ; ("integrate_ode", "integrate_ode_rk45")
```
rok-cesnovar, line 9: We need to expand/modify this:

integrate_ode -> ode_rk45
integrate_ode_rk45 -> ode_rk45
integrate_ode_bdf -> ode_bdf
integrate_ode_adams -> ode_adams

This was added recently (version 2.24).

src/frontend/Deprecation_analysis.ml
adamhaber: Thanks - I tried to comment it out and got an `unbound value` error. I now see that when I add it to `Deprecation_analysis.mli`, it resolves the problem.

rok-cesnovar (dismissed):
We need to add a few more deprecated functions. Sorry these were not listed in the issue.

src/frontend/Deprecation_analysis.ml
adamhaber: Thanks, updated.

rok-cesnovar: Thanks!

src/frontend/Deprecation_analysis.mli
adamhaber: I've tried to automatically generate the mli file with `ocamlc -i`, but got `Error: Unbound module Core_kernel`, so I added it myself until the type checker was happy and things made sense. I'm sure there's a better way. :-)

test/integration/canonicalize/canonical.expected
```diff
@@ -50,42 +50,31 @@ model {
...
-integrate_ode is deprecated, use integrate_ode_rk45 instead
+Warning: in 'deprecated.stan', line 55, column 20 to column 69: integrate_ode is deprecated and will be removed in the future. Use ode_rk45 instead
 
-Warning: deprecated language construct used in 'deprecated.stan', line 28, column 11 to column 46:
-The function `if_else` is deprecated. Use the conditional operator (x ? y : z) instead.
+Warning: in 'deprecated.stan', line 28, column 11 to column 46: The function `if_else` is deprecated. Use the conditional operator (x ? y : z) instead.
 
-Warning: deprecated language construct used in 'deprecated.stan', line 28, column 24 to column 42:
-multiply_log is deprecated, use lmultiply instead
+Warning: in 'deprecated.stan', line 28, column 24 to column 42: multiply_log is deprecated and will be removed in the future. Use lmultiply instead
```
rok-cesnovar, line 61: Minor: Some of these warnings end with a period and some don't.

adamhaber: Oops! Any preference? What would be more consistent with other places?

rok-cesnovar: Huh, I do not have a preference. I am also not huge friend with punctuation in such cases haha. My gut feeling is period everywhere but not super confident.

src/frontend/Deprecation_analysis.ml
adamhaber: I like the idea but I'm not sure how to implement it. At the moment there's no `fold_expression`, right (the closest I've found is `fold_expr` in `Mir_utils`)? Does this mean changing expression's `@@deriving` statement somehow?

test/integration/canonicalize/canonical.expected
adamhaber: Periods galore.

test/integration/canonicalize/canonical.expected
seantalts: I personally think the column information is superfluous, the line number seems plausibly helpful.

src/frontend/Deprecation_analysis.ml
seantalts: Hmm, according to the doc: https://github.com/ocaml-ppx/ppx_deriving it looks like there's even an `iter` method. Though I will say it's better functional style to accumulate the error messages rather than printing them out procedurally when they are encountered. For the procedural style you could add `iter` to all of the relevant `@@deriving` statements in Ast.ml; for functional style you could add `fold` and accumulate a list of warnings to be printed out later.

src/frontend/Deprecation_analysis.ml
seantalts: Hm. Style-wise OCaml seems to prefer few global opens, but this module does revolve around strings quite a bit, heh. If one were to import String it would sort of be equivalent to saying this module's main type that it deals with is strings though, and I think maybe that's not quite true. The reason it's that strong is that String contains many things besides the functions we're using, like a type `t` and possibly other things that would be confusing and maybe even collide. So I'd say just keep the local opens or specifically import functions using `let is_suffix = String.is_suffix` :/

test/integration/canonicalize/canonical.expected
adamhaber: I agree that the column information is not very important (in general imo); is there a way to print just line numbers? If not, and you think it's too cluttered as it is, I'll remove the Location_span part altogether.

src/frontend/Deprecation_analysis.ml
adamhaber: OK, keeping the local opens than.

src/frontend/Deprecation_analysis.ml
adamhaber: I tried to implement the fold-based approach. I hope it's somewhat similar to what you had in mind... any feedback would be much appreciated. 

One nice side effect that this change has is that now warnings are automatically sorted by line (at least in the examples I went over, not sure if it'll always hold), which I didn't expected.

src/frontend/Canonicalize.ml
```diff
@@ -1,95 +1,20 @@
...
 
 let rec repair_syntax_stmt user_dists {stmt; smeta} =
   match stmt with
   | Tilde {arg; distribution= {name; id_loc}; args; truncation} ->
       { stmt=
           Tilde
-            { arg= repair_syntax_expr arg
+            { arg
             ; distribution= {name= without_suffix user_dists name; id_loc}
-            ; args= List.map ~f:repair_syntax_expr args
```
seantalts, line 85: What happened to these lines btw?

adamhaber: @nhuurre mentioned in the [original thread](https://github.com/stan-dev/stanc3/issues/600#issuecomment-692846128) that `repair_syntax_expr` is obsolete.

src/frontend/Deprecation_analysis.ml
```diff
@@ -0,0 +1,163 @@
...
+      acc @ warn_deprecated_lval [] l @ warn_deprecated_expr [] e
+  | FunDef {body; _} -> warn_deprecated_stmt acc body
+  | _ ->
+      fold_statement warn_deprecated_expr warn_deprecated_stmt
+        warn_deprecated_lval
+        (fun l _ -> l)
+        acc stmt
+
+let emit_warnings (program : typed_program) : unit =
+  String.Table.clear deprecated_userdefined ;
```
rybern, line 161: We should avoid using global mutable state like `deprecetaed_userdefined` unless strictly necessary. Instead accumulate the results of functions into their return values. Global mutable state is a lot less clear and more error prone.

I think this issue predates this PR, though..

nhuurre: Yes, the global `deprecated_userdefined` was my fault back when I had no experience with functional programming.

So right now `replace_suffix` fills in `deprecated_userdefined` and then `warn_deprecated_expr` and `replace_deprecated_expr` look things up in that global hashtable.
A better design would be a `replace_suffix` that builds and returns a `String.Set.t` which is then explicitly passed to `warn_deprecated_expr`/`replace_deprecated_expr`. That requires passing it through a few other functions too but nothing too radical.
The code here, at the beginning of `emit_warnings` would be something like
```ocaml
let deprecated_userdefined =
  program.functionblock |> Option.value ~default:[]
  |> List.filter_map ~f:replace_suffix
  |> String.Set.of_list
in
```

By the way, I'm not a fan of the name `replace_suffix` because the function does not _replace_ anything, rather it _finds_ replaceable distribution names.

src/frontend/Deprecation_analysis.ml
```diff
@@ -0,0 +1,163 @@
...
+
+let deprecated_functions =
+  String.Map.of_alist_exn
+    [ ("multiply_log", "lmultiply")
+    ; ("binomial_coefficient_log", "lchoose")
+    ; ("cov_exp_quad", "gp_cov_exp_quad")
+    ; ("integrate_ode", "ode_rk45")
+    ; ("integrate_ode_rk45", "ode_rk45")
+    ; ("integrate_ode_bdf", "ode_bdf")
+    ; ("integrate_ode_adams", "ode_adams") ]
```
nhuurre, line 13: `integrate_ode_*` is deprecated but the corresponding `ode_*` is not a simple drop-in replacement. The old interface uses arrays and has tolerances at the end; the new interface uses vectors and has separate `ode_*_tol` function. If these are included in the list then the canonicalizer must not rely on `rename_function`.

Another case of a deprecated function without an easy substitution is `lkj_cov_log`. Stanc2 says
```
Info: the lkj_cov_log() sampling distribution is deprecated.  It will be removed in Stan 3.
Code LKJ covariance in terms of an lkj_corr() distribution on a correlation matrix and independent lognormals on the scales.
```
The canonicalizer ignored this because the required refactoring is too complicated.

src/frontend/Deprecation_analysis.ml
adamhaber: That's a great idea. I can refactor it to something like
```ocaml
let update_suffix name type_ = 
      let open String in
        if is_suffix ~suffix:"_cdf_log" name then drop_suffix name 8 ^ "_lcdf"
        else if is_suffix ~suffix:"_ccdf_log" name then
          drop_suffix name 9 ^ "_lccdf"
        else if Middle.UnsizedType.is_real_type type_ then
          drop_suffix name 4 ^ "_lpdf"
        else drop_suffix name 4 ^ "_lpmf"

let find_udf_suffixes = function
  | { stmt= FunDef {funname= {name; _}; arguments= (_, type_, _) :: _; _}
    ; smeta= _ }
    when String.is_suffix ~suffix:"_log" name ->
      Some (name, type_)
  | _ -> None
```
but then I have list of `(string * UnsizedType.t)` and I can't use `String.Set.of_list`. Is there a way to keep the type information (necessary for `update suffix`) and use a Set? I can think of hacky solutions to this, but is there a nicer way?

src/frontend/Deprecation_analysis.ml
adamhaber: What would be a good way to handle `integrate_ode_*`, then? And do you think we should handle them differently in `emit_warnings` and `canonicalizer`?

src/frontend/Deprecation_analysis.ml
nhuurre: Ah, right, a Hashtable is not a _Set_ but a _Map_. You need `String.Map.of_alist_exn`.

src/frontend/Deprecation_analysis.ml
nhuurre: The canonicalizer cannot handle them at all. I think they need to be special-case in `emit_warnings` (along with `lkj_cov_log`)

adamhaber: Sure. Which warning should be emitted, then? Something along the lines of
```
integrate_ode_x is deprecated and will be removed in the future. Use ode_x instead.
The new interface is slightly different, see:
https://mc-stan.org/users/documentation/case-studies/convert_odes.html
```
?

src/frontend/Deprecation_analysis.ml
adamhaber: OK, I've implemented this and I think it makes more sense now. The only thing I'm not sure about: `String.Map.of_alist_exn` requires the keys to be unique (hence the `dedup` step). This can potentially create issues if we have different `foo_log` UDFs with different signatures, but this shouldn't be possible, right?

rok-cesnovar: UDF overloading is not supported, at least not yet. So it currently is not a problem.

seantalts: (for reference there is also a `multi` series of tables and maps that gives you a map of lists, e.g. `val of_alist_multi : (key * 'a) list -> 'a list t`)

test/integration/canonicalize/canonical.expected
seantalts: You should be able to access the line number as an integer from the start location inside the location span.

adamhaber: I've implemented this, but it comes out (not surprisingly) too similar to [Location.to_string](https://github.com/stan-dev/stanc3/blob/30d1e55227fde791bd0177e997eb099de92aa675/src/middle/Location.ml#L44), which takes optional booleans for the file name and line number, but _not_ the column. Would it makes sense to add an optional `print_column` arg (defaults to true) to Location.to_string and call it, instead?

src/frontend/Deprecation_analysis.ml
rok-cesnovar: That looks ok to me, not sure what to emit for lkj_cov_log, never used it to be honest.

adamhaber:
@rok-cesnovar Jenkins fails on something stanc.js-related?

rok-cesnovar:
The stanc.js tests seem to work fine, fails on building which is not related to this PR. 

@serban-nicusor-toptal when you have a moment, take a look: https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3/detail/PR-726/12/pipeline and https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3/detail/PR-742/1/pipeline

Thanks!

serban-nicusor-toptal:
Hey @rok-cesnovar I'm on it, looks like something must have changed on either docker side or opam...?
`[ERROR] The compilation of opam-installer failed at "/usr/bin/make opam-installer.install".`
It's having a compilation error when trying to install `fmt-windows` but it's the exact same Dockerfile/scripts as master, debugging locally and will push a fix here when done.

serban-nicusor-toptal:
Hey @rok-cesnovar fix is now pushed to master branch, I can't seem to be able to commit here.

What happened ?

Package `opam-file-format` was updated to `2.1.1` since the last build on master and it now fails to build with opam-installer `2.0.7`.  I've checked history of master and it was successful with `opam-file-format` 2.1.0 so I've pinned that version in the installation scripts.

rok-cesnovar:
Thanks @serban-nicusor-toptal for once again acting quickly!

@adamhaber is this ready for review?

adamhaber:
I'll push the different ode warning now. What do you think of https://github.com/stan-dev/stanc3/pull/726#discussion_r524143981?

test/integration/canonicalize/canonical.expected
rok-cesnovar: > Would it makes sense to add an optional print_column arg (defaults to true) to Location.to_string and call it, instead?

To me that makes sense.

seantalts:
A convention that I didn't know about until recently is that `to_blah` functions are generally supposed to be used for a total and reversible mapping between the type and a `blah`, e.g. if you will encode the entire `Location.t` in the string then `to_string` is appropriate, otherwise something like `line_string` might be better if you're just getting the line number formatted. Or just grabbing the line number out and doing the formatting where it's used, if you don't think that format will be shared. 

I know that's not how it currently works but maybe something to aim for going forward?

test/integration/canonicalize/canonical.expected
seantalts: A convention that I didn't know about until recently is that `to_blah` functions are generally supposed to be used for a total and reversible mapping between the type and a `blah`, e.g. if you will encode the entire `Location.t` in the string then `to_string` is appropriate, otherwise something like `line_string` might be better if you're just getting the line number formatted. Or just grabbing the line number out and doing the formatting where it's used, if you don't think that format will be shared. 

I know that's not how it currently works but maybe something to aim for going forward?

adamhaber: I'd be happy to implement `line_string` if you think that's a good solution. Out of curiosity - Is there a place in which we call `Location.to_string` and we actually want the `column ... to column ... ` information? Is there a scenario in which the line number is not "good enough" for the user?

rybern: I do think the column information is useful for reporting errors in long lines when we don't want to give the whole context window, e.g. in some of the pedantic mode warnings. That said:
- I'm in favor of adding that `print_column` option to `Location.to_string` (or something similarly named to follow Sean's convention)
- I'm also in favor of making the `column a to column b` much less verbose, I've been meaning to do that for a while. Alternatives might be `columns a:b` or `characters a:b`, some compilers even just do `a:b`

src/frontend/Deprecation_analysis.ml
rybern: @adamhaber nice job on this fix! Looking good to me

test/integration/canonicalize/canonical.expected
adamhaber: So, do we want something like this:
```ocaml
let rec line_string ?(print_file = true) ?(print_line = true) ?(print_column = false) loc =
  let open Format in
  let file = if print_file then sprintf "'%s', " loc.filename else "" in
  let line = if print_line then sprintf "line %d, " loc.line_num else "" in
  let column = if print_column then sprintf "column %d, " loc.col_num else "" in
  let incl =
    match loc.included_from with
    | Some loc2 -> sprintf "included from\n%s" (line_string loc2)
    | None -> ""
  in
  sprintf "%s%s%s%s" file line column incl
```
If so, do we want to remove the optionals from `Location.to_string` (to enforce Sean's convention)? What about `Location_span.to_string`?

> I'm also in favor of making the column a to column b much less verbose, I've been meaning to do that for a while. Alternatives might be columns a:b or characters a:b, some compilers even just do a:b

I can do this, as well; I assume we need to somehow split to cases in which begin_loc and end_loc differ in their filename/included_from?

rybern: Code looks good, but I think you'll need a little bit more case handling - for example, if `loc.included_from` is `None`, it'll have a trailing comma.

> If so, do we want to remove the optionals from Location.to_string (to enforce Sean's convention)? What about Location_span.to_string?

I'm happy either way w.r.t the naming. I do think whatever's decided for Location will probably also apply to Location_span.

> I can do this, as well; I assume we need to somehow split to cases in which begin_loc and end_loc differ in their filename/included_from?

That would be awesome. Yeah, I think that's right.

IMO this would be reasonable to do in a separate PR

adamhaber:
> @adamhaber is this ready for another review?

Yes, I'll handle all the Location/Location_span printing in a different PR like @rybern suggested.

rok-cesnovar:
Cool.  @rybern, you reviewed this the most. Mind rechecking if everything is in order now?

test/integration/canonicalize/canonical.expected
adamhaber: >Code looks good, but I think you'll need a little bit more case handling - for example, if loc.included_from is None, it'll have a trailing comma.

Is there a clean way to do this, and to avoid all the if-then-else scenarios? I've tried to pad the "%s%s%s%s" string with ", " according to the lengths of the different tags (hope that makes sense), but the type checker didn't like it...

rybern:
Sure thing - either later tonight or tomorrow morning

test/integration/canonicalize/canonical.expected
rybern: One way would be to build a list of strings and then use a library function like `String.join ~sep:", "`
One way to build that list would be to build a list of string options, then use List.filter_opt

adamhaber: I see, so I can skip the whole sprintf part and build the string directly? That'd be easier if that's the case, I think.

src/frontend/Deprecation_analysis.ml
```diff
@@ -0,0 +1,194 @@
...
+    && not
+         ( is_distribution (name ^ "_log")
+         || List.exists ~f:(( = ) name) user_dists )
+  then drop_suffix name 4
+  else name
+
+let warn_deprecated (loc_span, message) =
+  Fmt.pf Fmt.stderr "@[<v>@,Warning: in %s: %s@]@."
+    (Location_span.to_string loc_span)
+    message
```
rybern, line 72: Did you consider using `warn_deprecated` in Errors.ml?

src/frontend/Deprecation_analysis.ml
```diff
@@ -0,0 +1,194 @@
...
+        (warn_deprecated_stmt deprecated_userdefined)
+        (warn_deprecated_lval deprecated_userdefined)
+        (fun l _ -> l)
+        acc stmt
+
+let analyze_udfs program =
+  program.functionblock |> Option.value ~default:[]
+  |> List.filter_map ~f:find_suffixes
+  |> List.dedup_and_sort ~compare:(fun (x, _) (y, _) -> String.compare x y)
+  |> String.Map.of_alist_exn
```
rybern, line 190: I'm not quite clear on the difference between `analyze_udfs` and `userdef_distributions`, they both seem to be collecting user-defined functions with appropriate suffixes. Are they both needed?

rybern (commented):
Looking good in general. I don't really know which of my comments are about this PR vs. what existed before; if the scope of this PR is just to do the refactoring you've done then feel free to ignore those.

I have some suggestions about function names, I leave to your discretion:
- `warn_deprecated_*` could be `collect_deprecated_*` to avoid confusion with the `warn_deprecated` function that does the I/O
- `analyze_udfs` could be something like `collect_udfs` or `collect_userdef_distributions` (see comment on that)
- `find_suffixes` could be something like `find_udf_log_suffix`
- `is_distribution` seems to behave more like `is_deprecated_distribution`

The last thing is that I'm not sure what the state of testing is. Since there's a new implementation here for finding deprecated functions (via derived folds), it might be good to write some test cases with e.g. non-top-level deprecated functions, if those tests don't already exist somewhere.

adamhaber:
> Looking good in general. I don't really know which of my comments are about this PR vs. what existed before; if the scope of this PR is just to do the refactoring you've done then feel free to ignore those.

They're all helpful, thanks! 

> I have some suggestions about function names, I leave to your discretion:

I agree these make more sense, especially after the change to the fold-base approach.

> The last thing is that I'm not sure what the state of testing is. 

Good question - I mostly relied on `test/integration/canonicalize/`, but it might not cover all the cases; I'll verify this.

> Since there's a new implementation here for finding deprecated functions (via derived folds), it might be good to write some test cases with e.g. non-top-level deprecated functions, if those tests don't already exist somewhere.

Not sure what you mean by "non-top-level deprecated functions", can you give an example?

rybern:
> Not sure what you mean by "non-top-level deprecated functions", can you give an example?

Yeah, by that I mean cases where a deprecated function is nested in e.g. an expression in a for-loop in a user-defined function.

adamhaber:
OK, `deprecated.stan` should now cover all cases.

rok-cesnovar:
@adamhaber is this ready for a final look? I am waiting on this PR to fix https://github.com/stan-dev/stanc3/issues/741 afterwards.

adamhaber:
Yes, I renamed the functions as per @rybern 's suggestion, and added some additional tests to `deprecated.stan` - it should cover all cases, I think, but an extra pair of eyes would be appreciated. :-)

rok-cesnovar:
To me, the deprecated.stan tests look good. Will let Ryan give the final green tick.

rybern (approved):
Sorry @adamhaber! Looks good to me.

PR #728, opened 2020-10-21
Fix for reduce_sum with lupdf

### LINKED ISSUE #727: lupdf/lupmf with reduce_sum (variadic functions)
Unnormalized density functions do not work with reduce_sum (and very likely also variadic ode, though I haven't checked that just yet).

The problem is that the reduce_sum functor is defined as

```
struct foo_lpdf_rsfunctor__ {
template <typename T0__>
stan::promote_args_t<T0__>
operator()(const std::vector<T0__>& y_slice, const int& start,
           const int& end, std::ostream* pstream__)  const 
{
return foo_lpdf<false>(y_slice, start + 1, end + 1, pstream__);
}
};
```
so `propto__` is set to false always.

The reduce_sum function does not take in the propto argument. What we could do temporarily is define

```
struct foo_lupdf_rsfunctor__ {
template <typename T0__>
stan::promote_args_t<T0__>
operator()(const std::vector<T0__>& y_slice, const int& start,
           const int& end, std::ostream* pstream__)  const 
{
return foo_lpdf<true>(y_slice, start + 1, end + 1, pstream__);
}
};
```
that would be used in this case and we could get away with this until reduce_sum handles the propto arg. I feel like I am missing something so I am just leaving this here and will think about this a bit more.

Other Q is: ODEs and lpdfs/lpmfs, is that a thing at all?

rok-cesnovar:
Fixes #727 

## Copyright and Licensing

### DIFF
--- a/src/frontend/Semantic_check.ml
+++ b/src/frontend/Semantic_check.ml
@@ -521,11 +521,12 @@ let semantic_check_variable cf loc id =
         |> ok
     | Some ((Param | TParam | GQuant), _) when cf.in_toplevel_decl ->
         Semantic_error.non_data_variable_size_decl loc |> error
-    | Some _ when Utils.is_unnormalized_distribution id.name
-      && not
-           ( (cf.in_fun_def && (cf.in_udf_dist_def || cf.in_lp_fun_def))
-           || cf.current_block = Model )
-    -> Semantic_error.invalid_unnormalized_fn loc |> error
+    | Some _
+      when Utils.is_unnormalized_distribution id.name
+           && not
+                ( (cf.in_fun_def && (cf.in_udf_dist_def || cf.in_lp_fun_def))
+                || cf.current_block = Model ) ->
+        Semantic_error.invalid_unnormalized_fn loc |> error
     | Some (originblock, type_) ->
         mk_typed_expression ~expr:(Variable id)
           ~ad_level:(calculate_autodifftype cf originblock type_)
--- a/src/frontend/Semantic_check.ml
+++ b/src/frontend/Semantic_check.ml
@@ -505,7 +505,7 @@ let semantic_check_postfixop loc op e =
 (* -- Variables ------------------------------------------------------------- *)
 let semantic_check_variable cf loc id =
   Validate.(
-    match Symbol_table.look vm id.name with
+    match Symbol_table.look vm (Utils.stdlib_distribution_name id.name) with
     | None when not (Stan_math_signatures.is_stan_math_function_name id.name)
       ->
         Semantic_error.ident_not_in_scope loc id.name |> error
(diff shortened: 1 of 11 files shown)


nhuurre (changes requested):
Semantic check needs to inspect the context flags `cf` and only allow `_lupdf` where appropriate. For example
 ```stan
generated quantities {
    real ll = reduce_sum(foo_lupdf,x,1);
}
```
should give the same error as 
```stan
generated quantities {
    real ll = foo_lupdf(x| 1, size(x));
}
```

Another thing I noticed is the bogus behaviour when user declares a local variable with a suffix name.
```stan
model {
    real x_lpdf = std_normal_lupdf(x|);
    target += x_lpdf; // ok
    real y_lupdf = std_normal_lupdf(y|);
    target += y_lupdf; // says variable "y_lupdf" not in scope.
}
```
I'd rather forbid such names completely but currently they're allowed.

rok-cesnovar:
Thanks!

> I'd rather forbid such names completely but currently they're allowed.

You mean _lupdf suffixes or all distribution suffixes?

nhuurre:
I was thinking of all special suffixes (`_rng`, `_lpdf`, `_lcdf`, `_lp`). If the name implies special restrictions on _function_  then why don't _non-function_ types also have restrictions? It's confusing, in my opinion. I don't know if weird variable names were intentionally allowed or if it's an accident. It'll get even worse if Stan gets closures because then userdefined functions are first-class objects that can be assigned to local variables. What happens to special names then?

rok-cesnovar:
Good point. Just to make sure, you mean we disallow variables with _rng, _lpdf,... or just use the same restrictions? I am guessing the first as that makes it clearer?

nhuurre:
Yes, I mean disallow them. But only `_lupdf` causes problems at the moment so if the others were allowed on purpose or are currently used in tests then it's fine to leave them.

rok-cesnovar:
> so if the others were allowed on purpose or are currently used in tests then it's fine to leave them.

That was an oversight in https://github.com/stan-dev/stanc3/pull/569
 This has not been release yet so still time to revert or partially revert. You are saying those will cause problems for closures? Then we should not allow them at all.

nhuurre:
> That was an oversight in
No, that only covers namesakes of Stan Math distribution functions. I'm talking about arbitrary names that just have a suffix.
It's not a problem for closures, it's just surprising that function types and primitive types have different naming rules.
Anyway, for this PR let's just say that user cannot declare a function nor variable with name that ends in `_lupdf`.

rok-cesnovar:
> Semantic check needs to inspect the context flags `cf` and only allow `_lupdf` where appropriate. For example

Done, added tests.Thanks.

Also disallowed identifiers with _lupdf/_lupmf suffixes.

nhuurre (approved):
Looks good.

PR #734, opened 2020-10-25
Include transformed parameters in factor graph in unused analysis

### LINKED ISSUE #733: Eight schools model generates three warnings (with --warn-pedantic)
The canonical eight schools model generates three stanc warnings:

```
Warning:
  The parameter mu has no priors.
Warning:
  The parameter tau has no priors.
Warning:
  The parameter theta was declared but was not used in the density
  calculation.
```

This model is used in the RStan and PyStan documentation. It's the first model which new users encounter.

I think it would be better if the first model users encounter did not generate warnings.

Here's the model, for reference:

```stan
// saved as schools.stan
data {
  int<lower=0> J;         // number of schools 
  real y[J];              // estimated treatment effects
  real<lower=0> sigma[J]; // standard error of effect estimates 
}
parameters {
  real mu;                // population treatment effect
  real<lower=0> tau;      // standard deviation in treatment effects
  vector[J] eta;          // unscaled deviation from mu by school
}
transformed parameters {
  vector[J] theta = mu + tau * eta;        // school treatment effects
}
model {
  target += normal_lpdf(eta | 0, 1);       // prior log-density
  target += normal_lpdf(y | theta, sigma); // log-likelihood
}
```

Discussion: https://discourse.mc-stan.org/t/eight-schools-model-generates-three-stanc-warnings/18883

rybern:
Fixes #733

Issue was caused when I changed the default of parameter_names_set to not include transformed parameters


rok-cesnovar (approved):
Thanks!

test/unit/Pedantic_analysis.ml
```diff
@@ -992,3 +992,34 @@ let%expect_test "Function body parameter-dependent control flow" =
...
+  |}
+
+let%expect_test "Missing priors schools warning" =
+  print_warn_pedantic (build_program schools_example) ;
+  [%expect
+    {|
+      Warning:
+        The parameter mu was declared but was not used in the density calculation.
+      Warning:
+        The parameter tau was declared but was not used in the density calculation.
```
rok-cesnovar, line 1024: Nice!

rybern: Still an issue I think - not sure why these aren't missing prior warnings instead

rok-cesnovar: Oh yeah, read too quickly. I just noticed that theta is not listed. In Allen's initial example the both do show correctly:
Warning:
  The parameter mu has no priors.
Warning:
  The parameter tau has no priors.

PR #739, opened 2020-11-10
Remove non scalar if_else

rok-cesnovar:
This PR removes stanc3 support for `if_else()` apart from
```
int if_else(int, int, int)
real if_else(int, real, real)
```
All the other signatures that were allowed in stanc3 did not compile on the C++ level as there is no support in Stan Math: https://github.com/stan-dev/math/blob/3f65abb9f2c61a4b671875964d6d4df2db5c6c44/stan/math/rev/fun/if_else.hpp, nor was it intended: https://mc-stan.org/docs/2_25/reference-manual/if-else-function.html

This is part of a cleanup of expression testing now that all math supports expressions on input https://github.com/stan-dev/math/pull/2186

No release notes as this is a deprecated function.

## Copyright and Licensing


rybern (approved):
Surprised we never had issues with this

rok-cesnovar:
I think its mostly because its a deprecated function and if there are no docs, no one will use it :) 

Thanks!

Issue #614, opened 2020-07-07
Simplify parser


nhuurre:
Simplify parser

The parser distinguishes between `var_decl`, `top_var_decl`, and `top_var_decl_no_assign` even though the AST does not. There's no logical need for the grammar to have three separate rules; semantic check could handle the potential syntax errors.
For instance, Stan does not allow tilde-statements outside of the model block but there's no special `statement_no_tilde` rule in the grammar for other blocks. So this fails to parse
```stan
parameters {
  real x;
  x ~ std_normal(); // only decls allowed
}
```
but this parses successfully and fails in semantic check
```stan
parameters {
  real x;
}
generated quantities {
  x ~ std_normal(); // any statement parses
}
```

Parser errors are defined in `parser.messages`. That file is not easy to work with. Currently it is incomplete because I didn't know I was supposed add some error states there in PR #573 
The parser error messages have no context other than the parser state that failed. [We decided](https://github.com/stan-dev/stanc3/issues/581#issuecomment-647464249) to fix #469 by changing the error message. The message relevant is here:
https://github.com/stan-dev/stanc3/blob/458a7933728debdb8b3ce14e5af6960faef95513/src/frontend/parser.messages#L1392-L1402
The parser state 661 corresponds to _any_ unexpected symbol in data or parameters block. As far as I know the only way to give a specific error for a semicolon is by creating a new parser state with the `UNREACHABLE` hack. On the other hand, if this were handled in Semantic_check the change would be trivial.

So I'm proposing changing the parser to accept any program that AST can represent and moving all other checks to Semantic_check.

@rybern what do you think?

rybern:
Yeah, interesting question.

One thing I should mention is that I did some [refactoring](https://github.com/stan-dev/stanc3/pull/561/files#diff-a455b46ddb916d08703b8e11c5054054) of those declaration rules in #561. Those changes should substantially reduce their redundancy, and certainly we could do more simplifying/clarifying.

I agree with all of your points about the shortcomings of the parser setup, especially error messages. (Side note: we should really add a check to Jenkins for completeness of `parser.messages`).

Ultimately we should put the logic wherever its easiest to read, to update, and to make it obviously correct (plus efficient when that's a concern). To those ends, here are some (all very similar) design principals that I like: 
- [Parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)
[...]

PR #742, opened 2020-11-18
Closures (again)

nhuurre:
The stanc3 part of stan-dev/math#2094

## Copyright and Licensing


rok-cesnovar:
Is this + the math PR everything we need to now review this? Any open issues or questions? Maybe me and @bbbales2 can co-review this? This may be too much frontend for my paygrade but maybe I could manage.

bbbales2:
@rok-cesnovar yeah if you want to look at it please. My goal in the next few days is to get it to build and then try writing a few models and see how it works. If it works good then I'll start pressuring people who know Ocaml to do the review. If that's you great lol.

rok-cesnovar:
Glad to help. I feel bad this is sitting here for so long. You playing with models and me starring at the code sounds like a plan. If I fail we can always bug Ryan :)

rok-cesnovar:
Binaries for this PR are available here: https://github.com/stan-dev/stanc3/actions/runs/385882467

bbbales2:
Built the latest binaries here: https://github.com/stan-dev/stanc3/actions/runs/396298848

Closures in transformed parameters works now so the sir model runs which is cool.

Is the `lupdf`/`lpdf` behavior going to be based off the name still? Was messing with the `reduce_sum` example and was wondering.

This is a model I was using to try out `reduce_sum`. It looks like doing a closure over an array of integers breaks:
```stan
data {
  int N;
  int n_redcards[N];
  int n_games[N];
  vector[N] rating;
}
parameters {
  vector[2] beta;
}
model {
  int grainsize = 1;

  functions
  real partial_sum(int[] slice_n_redcards,
                   int start, int end) {
    return binomial_logit_lpmf(slice_n_redcards |
                               n_games[start:end],
                               beta[1] + beta[2] * rating[start:end]);
  }

  beta[1] ~ normal(0, 10);
  beta[2] ~ normal(0, 1);

  target += reduce_sum(partial_sum, n_redcards, grainsize);
}
```

Errors look like:
```cpp
clang++ -std=c++1y -D_REENTRANT -Wno-sign-compare -Wno-ignored-attributes      -I stan/lib/stan_math/lib/tbb_2019_U8/include   -O3 -I src -I stan/src -I lib/rapidjson_1.1.0/ -I stan/lib/stan_math/ -I stan/lib/stan_math/lib/eigen_3.3.7 -I stan/lib/stan_math/lib/boost_1.72.0 -I stan/lib/stan_math/lib/sundials_5.2.0/include    -DBOOST_DISABLE_ASSERTS        -c -include-pch stan/src/stan/model/model_header.hpp.gch -x c++ -o redcard_lambda.o redcard_lambda.hpp
redcard_lambda.hpp:98:19: error: no matching constructor for initialization of 'std::vector<double>'
  : beta(beta__), n_games(n_games__), rating(rating__),
                  ^       ~~~~~~~~~

```

Data is 
[redcard.dat.txt](https://github.com/stan-dev/stanc3/files/5630515/redcard.dat.txt)

seantalts:
Can you add some information to this about the syntax and design and all that? Does this follow https://github.com/stan-dev/design-docs/blob/master/designs/0004-closures-fun-types.md ? 

Also - let me know if you could use help or review on this. Would love to get closures in.

bbbales2:
@seantalts more conversation over here: https://github.com/stan-dev/math/issues/2197 , @bob-carpenter said he'd do a review for it, so send him a message if you want to do language review.

@nhuurre laid  out the differences in the design doc and the implementation here: https://github.com/stan-dev/math/issues/2197#issuecomment-745249124 . The doc should get updated to reflect what ends up happening.

seantalts:
Sweet, looks good to me. Happy to let Bob or whoever do the review, was just offering a helping hand on the language part and didn't realize that discussion was taking place there. But yeah, in general feel free to tag me if there are lingering PRs or what-have-you.

bob-carpenter:
> Happy to let Bob or whoever do the review

Yikes, I meant to get to this sooner.  @nhuurre --- let me know how I can help.  I'm happy to review PRs or testing plans or whatever.

bbbales2:
@bob-carpenter my attempt at a todo list here: https://github.com/stan-dev/example-models/pull/202, but @nhuurre can tell you more I'm sure

Currently the Math side of this supports ODEs and reduce_sum I think. integrate_1d, the algebra solvers, and map_rect support aren't there yet.

One thing I was curious about was how this implementation stacks up against the original specification. I did not check this. @nhuurre put comments here: https://github.com/stan-dev/math/issues/2197#issuecomment-745249124 .

> Yikes

Ditto. I keep intending to help on the `integrate_1d` and `algebra_solver` front but keep getting unfocused on other stuff. Thanks for the patience @nhuurre 

@nhuurre Draft instructions are [here](https://github.com/stan-dev/example-models/pull/202) for Rok's test cmdstan tarball builder. It can produce a cmdstan tarball from a cmdstan/stanc3/stan/math branch that can be used from cmdstanr.

nhuurre:
@rybern, you mentioned...[uh, back in June](https://github.com/stan-dev/stanc3/pull/570#issuecomment-640076841), that you had some idea that could make the `functions` keyword unnecessary. If you have time could you take a look at this?

Some notes:
* a local variable cannot have function type; the declaration `real(real, real) func = ...;` is forbidden. Instead closure definitions must look just like functions, with the name before the argument list.
* the return type cannot be a function type. This was half-accidental and could be changed. However, if a closure can escape its initial scope then one has to worry about capture-by-reference leaving dangling pointers.
* there are no function arrays. The type `array[] real(vector)` refers unambiguously to a function/closure that takes a vector and returns an array of reals. This is a limitation in the C++ implementation.

nhuurre:
> Draft instructions are here for Rok's test cmdstan tarball builder. It can produce a cmdstan tarball from a cmdstan/stanc3/stan/math branch that can be used from cmdstanr.

I made a tarball here: https://github.com/nhuurre/cmdstan/releases/tag/closures-v0.1

bbbales2:
> I made a tarball here

This should be enough to install the tarball and use it in cmdstanr:

```
cmdstanr::install_cmdstan(release_url = "https://github.com/nhuurre/cmdstan/releases/download/closures-v0.1/cmdstan-stan-closures.tar.gz")
```

I tested it real quick and it worked.

And then use this to go back to the previous cmdstan, delete the `~/.cmdstanr/cmdstan-stan-closures` folder (if you don't delete it cmdstanr might pick it up when it searches for a default cmdstan)

nhuurre:
Thanks @rybern 

IMO the `Ast.index` type should have one more variant to distinguish between `[]` and `[:]`. (That's why you can reject `array[1:]` but not `array[:]`, right?)
I have complained about the complexity of the parser [before](https://github.com/stan-dev/stanc3/issues/614). Maybe we should seriously consider adding a stage between parsing and AST.

Anyway, if it's too hacky then let's just keep the keyword. We can always deprecate it later if we don't need it anymore.
Still, I'd like to at least mention the keyword in any error that results from omitting it. This is especially confusing:
```text
Syntax error in 'err.stan', line 2, column 18 to column 19, parsing error:
     1:  transformed data {
     2:    array[] real foo(array[] real x) {
                           ^
     3:      return x;
     4:    }

Expected non-array type after array in return type.
```

rybern:
> IMO the Ast.index type should have one more variant to distinguish between [] and [:]. (That's why you can reject array[1:] but not array[:], right?)

Bingo, and I agree that's the workaround.

> I have complained about the complexity of the parser before.

Yeah, doing this would make a couple parts of the parser harder to follow. See the branch I linked for roughly how it would look - the main trick is to pack all of the different variants of basic types (unsized, sized, and top-level) into one production, and then specialize them later. I'd want to add some big comments to explain how it works.

> Still, I'd like to at least mention the keyword in any error that results from omitting it.

It's not clear to me how good of an error message we can have without the parser recognizing an unsized type, which is the same difficulty with omitting the keyword. Not sure.

BTW, just IMO, if we do keep a keyword, I'd suggest something like `function` or `func` rather than `functions`, just to make it clear that it's used before one function definition rather than a section.

nhuurre:
Yeah, I guess printing nice error message is no easier.
The potential errors I found are
```
";" or plain assignment is expected after a top-level variable declaration.

Identifier expected after type in top-level variable declaration.

Expected non-array type after array in return type.
```
and I think it's reasonable to append `note: local functions must be declared with keyword` to these.
And yes, `functions` is not ideal, it was just a placeholder.

rok-cesnovar:
I gave this a run, looks great. One minor thing missing is the +1 for start/end of reduce_sum.

nhuurre:
New tarball: https://github.com/nhuurre/cmdstan/releases/download/closures-v0.2/cmdstan-closures-adjoint-ode.tar.gz

* The keyword is now `function`
* `reduce_sum` start/end indexes should be fixed
* `integrate_1d` and the experimental `ode_adjoint_tol`

SteveBronder:
Just to check, this only does closures and not the lambda syntax from [the design doc](https://github.com/stan-dev/design-docs/blob/master/designs/0004-closures-fun-types.md). I'm totally fine with that just wanted to check

test/integration/good/code-gen/closures/closures1.stan
```diff
@@ -0,0 +1,35 @@
...
+model {
+    function
+    real bar(real z, row_vector r) {
+        real rs = sum(r);
+        real ys = sum(y);
+        real vs = sum(v);
+        real pas = sum(pa);
+        real pvs = sum(pv);
+        return z + rs + x + ys + vs + p + pas + pvs;
+    }
```
SteveBronder, line 30: Do you have an example of a failure when there is shadowing in the function and a value is brought in from a higher scope? Like 

```suggestion
    function
    real bar(real z, row_vector r) {
        vector[N] pa = // ... fill this with something
        real rs = sum(r);
        real ys = sum(y);
        real vs = sum(v);
        real pas = sum(pa);
        real pvs = sum(pv);
        return z + rs + x + ys + vs + p + pas + pvs;
    }
```

SteveBronder: Also can you give the test function names that line up with what they intend to check or validate? It makes it easier when looking across the stan/mir/cpp

nhuurre:
> Just to check, this only does closures and not the lambda syntax

Yes, no lambda syntax. I mentioned the differences from the design doc [here](https://github.com/stan-dev/math/issues/2197#issuecomment-745249124). (but that's outdated, the special suffixes `_rng`/`_lpdf`/`_lp` are now allowed on closures.)

test/integration/good/code-gen/closures/closures1.stan
SteveBronder: C++ also allows this in lambdas for `constexpr` defined objects, but that's a very specific type of scoping that's known globally at compile time

nhuurre: The Stan language does not allow shadowing. Closures are no exception.

SteveBronder: Oh good!

SteveBronder: Would you mind adding an example that takes in no arguments and uses just scoped objects? Reading over the gen'd C++ right now and would be easier to parse through that logic.

SteveBronder: It won't let me comment on the C++ file because it's too many changes, for review (or maybe just a good idea) it would be good to print those into a separate closure.expected file

SteveBronder: I'll comment on one here though

```cpp
template <typename T0__, typename T1__, typename T2__, typename T3__,
typename T6__, typename T7__>
stan::promote_args_t<T0__, T1__, stan::value_type_t<T2__>,
stan::value_type_t<T3__>,
T6__, stan::promote_args_t<stan::value_type_t<T7__>>>
bar_L23C9_impl__(const T0__& p, const std::vector<T1__>& pa, const T2__& pv,
                 const T3__& v, const double& x,
                 const std::vector<double>& y, const T6__& z, const T7__& r,
                 std::ostream* pstream__)  ;

template<bool ref__, typename F0__, typename F1__, typename F2__>
class bar_L23C9_cfunctor__ {
  /**
   * When would these be captured by value?
   */
  stan::capture_type_t<F0__, ref__> p;
  stan::capture_type_t<std::vector<F1__>, ref__> pa;
  stan::capture_type_t<Eigen::Matrix<F2__, -1, 1>, ref__> pv;
  const Eigen::Matrix<double, -1, 1>& v;
  stan::capture_type_t<double, ref__> x;
  const std::vector<double>& y;
  public:
  const size_t vars_count__;
  bar_L23C9_cfunctor__(const bar_L23C9_cfunctor__<ref__, F0__, F1__, F2__>&) = default ;
  bar_L23C9_cfunctor__(bar_L23C9_cfunctor__<ref__, F0__, F1__, F2__>&&) = default ;
  bar_L23C9_cfunctor__(const F0__& p__, const std::vector<F1__>& pa__,
                       const Eigen::Matrix<F2__, -1, 1>& pv__,
                       const Eigen::Matrix<double, -1, 1>& v__,
                       const double& x__, const std::vector<double>& y__)
  : p(p__), pa(pa__), pv(pv__), v(v__), x(x__), y(y__),
    vars_count__(count_vars(p__, pa__, pv__, v__, x__, y__)) {}
  template <typename T0__, typename T1__>
  stan::promote_args_t<F0__, F1__, F2__, T0__,
stan::value_type_t<T1__>>
  operator()(std::ostream* pstream__, const T0__& z, const T1__& r)  const 
  {
  return bar_L23C9_impl__(p, pa, pv, v, x, y, z, r, pstream__);
  }
  /**
   * Will any of these be used in places we will also have user defined code?
   *  if not, then no need to double underscore.
   */   
  using captured_scalar_t__ = stan::return_type_t<F0__, F1__, F2__>;
  using ValueOf__ = bar_L23C9_cfunctor__<false, double, double, double>;
  using CopyOf__ = bar_L23C9_cfunctor__<false,
                                        stan::capture_type_t<F0__, false>,
                                        stan::capture_type_t<F1__, false>,
                                        stan::capture_type_t<F2__, false>>;
  /**
   * idt we need these? I think you could just have a function that 
   * returns a tuple of the members like 
   */
   auto get_members() {
     return std::make_tuple(p, pa, pv, v, x, y);
   }
   /**
    * Then on the stan math side we can just run stuff like
    */
    // in math deep_copy_vars.hpp file 
    template <typename F, require_stan_closure_t<F>* = nullptr>
    auto deep_copy_vars(F&& f) {
      return stan::math::for_each([](auto&& x) { return deep_copy_vars(x);}, f.get_members());      
    }
    
  size_t count_vars__() const {
  return vars_count__;
  }
  auto value_of__() const {
  return ValueOf__(eval(value_of(p)), eval(value_of(pa)), eval(value_of(pv)),
                   v, x, y);
  }
  auto deep_copy_vars__() const {
  return CopyOf__(eval(deep_copy_vars(p)), eval(deep_copy_vars(pa)),
                  eval(deep_copy_vars(pv)), v, x, y);
  }
  void zero_adjoints__() {
  stan::math::zero_adjoints(p);
  stan::math::zero_adjoints(pa);
  stan::math::zero_adjoints(pv);
  stan::math::zero_adjoints(v);
  stan::math::zero_adjoints(x);
  stan::math::zero_adjoints(y);
  }
  double* accumulate_adjoints__(double *dest) const {
  return stan::math::accumulate_adjoints(dest, p, pa, pv, v, x, y);
  }
  stan::math::vari** save_varis__(stan::math::vari **dest) const {
  return stan::math::save_varis(dest, p, pa, pv, v, x, y);
  }
  
  
  };
```

nhuurre:
> can you give the test function names that line up with what they intend to check or validate?

I renamed the files and functions to something a bit more explanatory and added [a unit test](https://github.com/nhuurre/stanc3/blob/2ba4cb9ff9b7930a97426d911a1ef8b7d3918c83/test/unit/Stan_math_code_gen_tests.ml#L170). The tests are admittedly rather haphazard.

test/integration/good/code-gen/closures/cpp.expected
```diff
@@ -0,0 +1,4995 @@
...
+                         const std::vector<double>& y, const T3__& z,
+                         const T4__& r, std::ostream* pstream__)  ;
+
+class capture_data_L9C9_cfunctor__ {
+  const Eigen::Matrix<double, -1, 1>& v;
+  double x;
+  const std::vector<double>& y;
+  public:
+  const size_t vars_count__;
+  capture_data_L9C9_cfunctor__(const capture_data_L9C9_cfunctor__&) = default ;
```
SteveBronder, line 66: Does the compiler not handle these automatically?

nhuurre: Probably? I think I was just playing it safe here. Not sure if the constructor is even needed anywhere...

wds15:
I just tried this out and I really like it. While I got it working I was slightly irritated by two things:

- No semicolon after defining the closure.
- No need for a block "functions" within the modeling blocks.

Sorry if this was already discussed and I missed it, but what is the rationale here? The above two points are just what I would have expected naively.

Other than that, it is really cool to use!

nhuurre:
> No semicolon after defining the closure.

Function definitions in the `functions` block don't have semicolons either. Since a lone semicolon is an allowed (no-op) statement it is not an error to put a semicolon after a closure definition. I think it would be a bit weird to require it though.

> No need for a block "functions" within the modeling blocks.

I think a local block would have confusing scope--local variables are not visible outside of the (local) block in which they are declared but the functions from the top-level `functions` block are available in all later blocks--so what exactly is the scope for a _local_ `functions` block?
I'm not opposed to the idea but it didn't seem to add much.

bob-carpenter:
> No semicolon after defining the closure.

This is the way C/C++ does it, too.  I think it's more natural that way.  If you required a semicolon, it'd be like those pesky class-final semicolons in C++.

wds15:
>  think a local block would have confusing scope--local variables are not visible outside of the (local) block in which they are declared but the functions from the top-level functions block are available in all later blocks--so what exactly is the scope for a local functions block?

Everything declared before the block should be in scope. Having a local block would make it consistent with the profiling thing we have, no? Is that a consideration maybe?

I really think we want closures sooner than later, but I also still think a design doc would be great (or have I missed that)? If I missed it, then shame on me! The intent of the design doc is just to make enough people aware of this and see this. I know it's some work, but it is all in scope for 2.28 for sure.

test/integration/good/code-gen/closures/cpp.expected
SteveBronder: word yeah idt these are necessary. If you don't define any of the default generated constructors then it should just make them for you

rok-cesnovar:
With profiling or any local block we have this:
```stan
model {
  profile("test") {
     real a = 5; // this is a local sc
  }
 real b = a + 2; // this would error
}
```
Similarly with local functions 
```stan
model {
  functions {
     real foo() {
          return 0.0;
     }
  }
  real b =   foo(); // foo should be out of scope.
}
```
Allowing foo to work in this case would be a bit confusing to me.

> but I also still think a design doc would be great

A design doc for this was done in 2019 by @bob-carpenter: https://github.com/stan-dev/design-docs/pull/6
This PR deviates from the design doc in some minor implementation details, as Niko noted in a comment above: https://github.com/stan-dev/stanc3/pull/742#issuecomment-862544022

Once this PR is merged I think the important details should probably be added to the design doc so it reflects what was added.

wds15:
Updating the design doc would be great... and thanks for pointing that doc out; I was not sure if that was it.

I haven't put my head too deep in to that doc, so if my comments are late, then sorry for that.

I see your point about that it's odd to declare a block and then use it later on. One possibility I see is to limit the flexibility somewhat, but make it more consistent with the current structure. The way to do that is by simply allowing additional functions block between any other block. So this would be legal:

```stan
// global functions without any scope of other variables
functions {
}
data {
}
// can access data
functions {
}
transformed data {
}
// can access all of the above
functions {
}
//.... and so on
```

This limits the closures to be defined in blocks and it seems as if this would also be consistent with the major deviation here which is no lambdas can be re-assigned.

Going this way would also solve scoping issues which can happen whenever a function is defined within a local blocks of one of the main block

```stan
// ...
model {
   real a =1;
   { // local block
    // define function foo
   }
   // can I use foo here?
}
```

@rok-cesnovar the example you post above could be written like

```stan
model {
  { // local block
  functions {
       real foo() {
            return 0.0;
       }
    } 
    real b =   foo(); // foo should be out of scope.
  }
}
```

So in addition to allowing function blocks between any other blocks, we could also allow them at the start of a local block as above. This would eliminate the restriction that a function could not be defined within a main block.

Thoughts?

(I hope this is helpful... if we deemed all of this set in stone then apologies)

PR #754, opened 2020-12-02
Add OpenCL support for additional distributions

rok-cesnovar:
~~Note: This is waiting for https://github.com/stan-dev/math/pull/2226 and should not be merged prior to that PR.~~ The PR was merged

This PR adds OpenCL support for additional distributions (see release notes for the list). For all the newly added distributions we run all variants on the OpenCL device (apart from the case where all input args are scalars in which case they do not get promoted to matrix_cl or var_value<matrix_cl> anyway).

## Release notes

Added OpenCL support for:
- bernoulli_lpmf
- bernoulli_logit_lpmf
- beta_lpdf
- beta_proportion_lpdf
- binomial_lpmf
- cauchy_lpdf
- chi_square_lpdf
- double_exponential_lpdf
- exp_mod_normal_lpdf
- exponential_lpdf
- frechet_lpdf
- gamma_lpdf
- gumbel_lpdf
- inv_chi_square_lpdf
- inv_gamma_lpdf
- logistic_lpdf
- lognormal_lpdf
- neg_binomial_lpmf
- neg_binomial_2_lpmf
- neg_binomial_2_log_lpmf
- normal_lpdf
- pareto_lpdf
- pareto_type_2_lpdf
- poisson_lpmf
- poisson_log_lpmf
- rayleigh_lpdf
- scaled_inv_chi_square_lpdf
- skew_normal_lpdf
- std_normal_lpdf
- student_t_lpdf
- uniform_lpdf
- weibull_lpdf

## Copyright and Licensing


rok-cesnovar:
@t4c1 would you mind checking the generated C++ here? Once you approve that I can start nagging someone for the OCaml changes. The entire huge model compile to the executable so I am fairly convinced everything is fine. No need to check all 4000 lines as things start repeating.

The variables are named so that the prefix p_ means its a parameters, and d_ means its data. If you see a _opencl__ suffix that means that the variable is already copied to a matrix_cl object (for now this is only used for data variables).

```cpp
      current_statement__ = 22;
      transformed_param_real = bernoulli_logit_lpmf<false>(
                                 d_int_array_opencl__, d_row_vector_opencl__);
      current_statement__ = 23;
      transformed_param_real = bernoulli_logit_lpmf<false>(
                                 d_int_array_opencl__,
                                 to_matrix_cl(p_row_vector));
      current_statement__ = 24;
      transformed_param_real = bernoulli_logit_lpmf<false>(
                                 d_int_array_opencl__, d_real_array_opencl__);
      current_statement__ = 25;
      transformed_param_real = bernoulli_logit_lpmf<false>(
                                 d_int_array_opencl__,
                                 to_matrix_cl(p_real_array));
```

For now this will do all the to_matrix_cl inside the call to the distribution, there are no matrix_cl local variables yet. A follow-up PR will add that.

rok-cesnovar:
Had a chat with @t4c1 about this, we need to do a benchmark to eliminate the obvious cases where OpenCL on a GPU will never be faster (i.e when the transfer cost is higher than the CPU runtime) and then add those restriction. 

Additionally we could then have a "--use-opencl=force" or a separate flag that would use OpenCL signature no matter what. This would be used for when the OpenCL device has a unified memory with the CPU (OpenCL on the CPU or integrated GPU). But this latter things are for the future.

rok-cesnovar:
We ran all distribution benchmarks and identified all signatures for which its not worth it to move to the OpenCL device. The summaries of the benchamarks can be seen [https://github.com/rok-cesnovar/stan-math-opencl-benchmarks/blob/main/summaries/radeonvii_vs_i7_distributions.md](here). There are also figures for all signatures available [here](https://unilj-my.sharepoint.com/:f:/g/personal/rokcesnovar_fri1_uni-lj_si/EgPP9IsCdFJJg0Ygm4_4-HcB9FOozC3FayacTGXeaT6LEg?e=1rll86).

Essentially, it comes down to 4 signatures that have some restrictions: 

- `real bernoulli_lpmf(int[], real)`: this one is not faster in any circumstance (tested up to N = 16 million)
- `real exponential_pdf(vector, real)`: the vector needs to be data in order for this distribution to be faster
- `real std_normal(vector)`: the vector needs to be data in order for this distribution to be faster
- `real uniform_lpdf(vector, real, real)`: the vector needs to be data in order for this distribution to be faster

All other functions and signatures are considerably faster above a certain N and are not a lot slower for smaller N (at least not for those where execution time is non-negligible).

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -3,18 +3,49 @@ open Middle
...
-let opencl_triggers =
+let opencl_trigger_restrictions =
   String.Map.of_alist_exn
-    [ ( "normal_id_glm_lpdf"
-      , [ (* Array of conditions under which to move to OpenCL *)
-          [(1, UnsizedType.UMatrix)]
-        (* Argument 1 is a matrix *)
+    [ ( "bernoulli_lpmf"
+      , [ [ (0, UnsizedType.DataOnly, UnsizedType.UArray UnsizedType.UInt)
+          ; (1, UnsizedType.DataOnly, UnsizedType.UReal) ] ] )
```
rok-cesnovar, line 10: `(1, UnsizedType.DataOnly, UnsizedType.UReal)˙ means that we will not move to OpenCL if the second argument is of type real and is data or a parameter.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -3,18 +3,49 @@ open Middle
...
          ] )
-    ; ("bernoulli_logit_glm_lpmf", [[(1, UnsizedType.UMatrix)]])
-    ; ("categorical_logit_glm_lpmf", [[(1, UnsizedType.UMatrix)]])
-    ; ("neg_binomial_2_log_glm_lpmf", [[(1, UnsizedType.UMatrix)]])
-    ; ("ordered_logistic_glm_lpmf", [[(1, UnsizedType.UMatrix)]])
-    ; ("poisson_log_glm_lpmf", [[(1, UnsizedType.UMatrix)]]) ]
+    ; ( "categorical_logit_glm_lpmf"
+      , [[(1, UnsizedType.DataOnly, UnsizedType.URowVector)]] )
+    ; ( "exponential_lpdf"
+      , [ [ (0, UnsizedType.AutoDiffable, UnsizedType.UVector)
```
rok-cesnovar, line 19: (0, UnsizedType.AutoDiffable, UnsizedType.UVector) means that we will not move to OpenCL if the first argument is a parameter but will if its data.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -31,18 +62,37 @@ let rec switch_expr_to_opencl available_cl_vars (Expr.Fixed.({pattern; _}) as e)
...
+      ~equal:String.equal
+  in
+  let is_fn_opencl_support_restricted f =
+    Map.mem opencl_trigger_restrictions (Utils.stdlib_distribution_name f)
   in
   match pattern with
   | FunApp (StanLib, f, args)
-    when Map.mem opencl_triggers (Utils.stdlib_distribution_name f) ->
+    when is_fn_opencl_supported f && not (is_fn_opencl_support_restricted f) ->
+      {e with pattern= FunApp (StanLib, f, List.map args ~f:to_cl)}
```
rok-cesnovar, line 90: If the functions support is not restricted we do not have to check whether to move to OpenCL or not.

rok-cesnovar (commented):
This is now ready for review.

@seantalts do you have time in the next few days to have a look?

The changes are trivial but I want to get these simpler parts in before we proceed to functions that return non-scalars, which is where it will get trickier. Thanks!

seantalts:
Sure! I'll try to get to this tonight.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -31,18 +62,37 @@ let rec switch_expr_to_opencl available_cl_vars (Expr.Fixed.({pattern; _}) as e)
...
   in
-  let check_type args (i, t) = Expr.Typed.type_of (List.nth_exn args i) = t in
-  let any_req_met args = List.exists ~f:(List.for_all ~f:(check_type args)) in
+  let check_type args (i, ad, t) =
+    Expr.Typed.type_of (List.nth_exn args i) = t
+    && UnsizedType.autodifftype_can_convert
+         (Expr.Typed.adlevel_of (List.nth_exn args i))
+         ad
+  in
+  let any_rest_met args =
```
seantalts, line 71: I would maybe just name this to `restricted` or `is_restricted`

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -31,18 +62,37 @@ let rec switch_expr_to_opencl available_cl_vars (Expr.Fixed.({pattern; _}) as e)
...
+    List.exists ~f:(List.for_all ~f:(fun t -> check_type args t))
```
seantalts, line 72: nit: I think this can be `~f:(check_type args)`

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -31,18 +62,37 @@ let rec switch_expr_to_opencl available_cl_vars (Expr.Fixed.({pattern; _}) as e)
...
   let maybe_map_args args req_args =
-    match any_req_met args req_args with
-    | true -> List.map args ~f:to_cl
-    | false -> args
+    match any_rest_met args req_args with
+    | true -> args
+    | false -> List.map args ~f:to_cl
+  in
+  let is_fn_opencl_supported f =
+    List.mem opencl_supported_functions
```
seantalts, line 80: Maybe we should turn `opencl_supported_functions` into a set? With like `|> String.Set.of_alist` or something.

src/stan_math_backend/Transform_Mir.ml
seantalts: It seems like we should be able to combine this case with the one where there are restrictions and outsource to `any_rest_met`? eg. change to `Map.find opencl_trigger_restrictions` instead of find_exn and have `maybe_map_args` take req_args as an option type that it passes along to `any_rest_met` which just returns `false` if given `None` for the req_args argument?

seantalts (changes requested):

seantalts:
Thanks for doing this! Generally looks good, some small code suggestions.

src/stan_math_backend/Transform_Mir.ml
rok-cesnovar: Agreed. Changed to is_restricted.

src/stan_math_backend/Transform_Mir.ml
rok-cesnovar: Done.

src/stan_math_backend/Transform_Mir.ml
rok-cesnovar: Done.

src/stan_math_backend/Transform_Mir.ml
rok-cesnovar: Ah yes, much cleaner. Done.

rok-cesnovar:
Thank you for a quick review and helpful comments! I incorporated all the suggestions.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -68,31 +69,22 @@ let rec switch_expr_to_opencl available_cl_vars (Expr.Fixed.({pattern; _}) as e)
          (Expr.Typed.adlevel_of (List.nth_exn args i))
```
seantalts, line 69: consider pulling out something like `let arg = List.nth_exn args i in` since it's used twice? not a big deal

seantalts (approved):
LGTM, thanks!!

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -31,18 +63,28 @@ let rec switch_expr_to_opencl available_cl_vars (Expr.Fixed.({pattern; _}) as e)
...
+    List.exists ~f:(List.for_all ~f:(check_type args))
+  in
   let maybe_map_args args req_args =
-    match any_req_met args req_args with
-    | true -> List.map args ~f:to_cl
-    | false -> args
+    match req_args with
+    | None -> List.map args ~f:to_cl
+    | Some x when is_restricted args x -> args
+    | Some _ -> List.map args ~f:to_cl
```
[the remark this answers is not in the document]
rok-cesnovar: Doh, of course. Thanks!

PR #761, opened 2020-12-09
Allow methods that use reader to accept Eigen types

SteveBronder:
The [`reader<>`](https://github.com/stan-dev/stan/blob/develop/src/stan/io/reader.hpp#L37) class was recently update so that it could accept Eigen matrices. This PR updates the generated model code so we still preserve the virtual methods for `write_array()`, `log_prob()`, and `transform_inits()`, but underneath the hood they just call a `*_impl` method. This is nice because now that `reader` can accept an Eigen type we don't have to make temporary `std::vector<>`s and copy back and forth from  eigen vectors to std vectors.

## Release notes

Allows `write_array()`, `log_prob()`, and `transform_inits()` to avoid a vector copy and read/write 

## Copyright and Licensing

### DIFF
--- a/src/stan_math_backend/Stan_math_code_gen.ml
+++ b/src/stan_math_backend/Stan_math_code_gen.ml
@@ -708,16 +708,16 @@ let pp_overloads ppf () =
     {|
     // Begin method overload boilerplate
     template <typename RNG>
-    inline void write_array(RNG& base_rng__,
+    inline void write_array(RNG& base_rng,
                             Eigen::Matrix<double,Eigen::Dynamic,1>& params_r,
                             Eigen::Matrix<double,Eigen::Dynamic,1>& vars,
-                            const bool emit_transformed_parameters__ = true,
-                            const bool emit_generated_quantities__ = true,
+                            const bool emit_transformed_parameters = true,
+                            const bool emit_generated_quantities = true,
                             std::ostream* pstream = nullptr) const {
       std::vector<double> vars_vec(vars.size());
       std::vector<int> params_i;
-      write_array_impl(base_rng__, params_r, params_i, vars_vec,
-          emit_transformed_parameters__, emit_generated_quantities__, pstream);
+      write_array_impl(base_rng, params_r, params_i, vars_vec,
+          emit_transformed_parameters, emit_generated_quantities, pstream);
       vars.resize(vars_vec.size());
       for (int i = 0; i < vars.size(); ++i) {
         vars.coeffRef(i) = vars_vec[i];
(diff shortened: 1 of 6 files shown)


SteveBronder:
I'm having a hard time reading the jenkins build output but it looks like it's failing during the dependency installation?

rok-cesnovar:
I think mergin in master will help there. We had an issue with pinning a package.

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -797,6 +824,7 @@ using stan::model::index_min_max;
 using stan::model::index_multi;
 using stan::model::index_omni;
 using stan::model::nil_index_list;
+using stan::scalar_type_t;
```
rok-cesnovar, line 827: Please use the namespace when using it. We are trying to reduce this using lists rather than expand it :). I am fairly convinced now that this meta template stuff should live in stan::math, but that is a separate story.

SteveBronder: Yeah sure!

The meta template stuff is also being used now in the stan repo now def feel free to hit me up with any issues!

rok-cesnovar: I mostly just get confused about what is in math and what is not. scalar_type to me feels like a Math meta thing but idk. To me, the Stan repo should be just services and algorithms and making sure models compile.

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -631,36 +636,44 @@ let pp_unconstrained_param_names ppf {Program.output_vars; _} =
...
+  pp_method_b ppf "void" "transform_inits_impl" params intro transform_inits
     ~cv_attr
 
 (** Print the `log_prob` method of the model class *)
 let pp_log_prob ppf Program.({prog_name; log_prob; _}) =
-  pf ppf "template <bool propto__, bool jacobian__, typename T__>@ " ;
+  pf ppf
+    "template <bool propto__, bool jacobian__, typename VecR, typename VecI,\n\
+    \     stan::require_vector_like_t<VecR>* = nullptr,\n\
+    \     stan::require_vector_like_vt<std::is_integral, VecI>* = nullptr>@ " ;
```
SteveBronder, line 660: Is there a reason these are red?

rok-cesnovar: Sorry, forgot about this Q. You should just let the formatter handle newlines. Adding manual \n leads to a war with the formatter in my experience :)

rok-cesnovar (approved):

PR #762, opened 2020-12-09
Do not print stdout and stderr in stanc.js

### LINKED ISSUE #741: javascript parser chokes on #-style comments
Doing

```{r}
  stanc_ctx <- V8::v8(typed_arrays = FALSE)
  stanc_ctx$source("https://github.com/stan-dev/stanc3/releases/download/nightly/stanc.js")
  stopifnot(stanc_ctx$validate("stanc"))
  stanc_ctx$call("stanc", "comment", "data { int N; # this is an integer }")
```
fails with a weird error message

> Error in context_eval(join(src), private$context, serialize) : 
  0,248,Failure,-3,

whereas it is legal but deprecated with the OCaml parser, which shows

> Warning: deprecated language construct used in '/tmp/test.stan', line 2, column 9:
     1:  data {
     2:    int N; # this is an integer
                  ^
     3:  }
Comments beginning with # are deprecated. Please use // in place of # for line comments.

rok-cesnovar:
Fixes #741 

Stdout and stderr in stanc.js are rerouted to a ref (is there a more Ocamlish solution?) and the contents of stderr are returned as the warnings element of the return object. Stdout is ignored.

I think this is the more Javascript way of doing it and for use in Rstan we have to avoid printing anyways (see issue).
Currently, in order to [print pedantic warnings](https://rok-cesnovar.github.io/stanc3js-demo/pedantic.html) in a browser, I had to reroute windows.console.error which is not how we want our API to work.

## Release notes

Removed stdout and stderr in stanc.js.

## Copyright and Licensing

### DIFF
--- a/src/stancjs/stancjs.ml
+++ b/src/stancjs/stancjs.ml
@@ -54,7 +54,7 @@ let stan2cpp model_name model_string flags =
     |> Result.map ~f:(fun typed_ast ->
            if is_flag_set "print-canonical" then
              Pretty_printing.pretty_print_typed_program
-                 (Canonicalize.canonicalize_program typed_ast)
+               (Canonicalize.canonicalize_program typed_ast)
            else if is_flag_set "auto-format" then
              match ast with
              | Result.Ok f -> Pretty_printing.pretty_print_program f
@@ -71,7 +71,7 @@ let stan2cpp model_name model_string flags =
                Pedantic_analysis.print_warn_uninitialized mir ;
              if is_flag_set "warn-pedantic" then
                Pedantic_analysis.print_warn_pedantic mir ;
-             cpp)
+             cpp )
 
 let stanc_stdout = ref ""
 let stanc_stderr = ref ""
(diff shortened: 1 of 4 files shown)


src/stancjs/stancjs.ml
```diff
@@ -72,21 +71,26 @@ let stan2cpp model_name model_string flags =
...
+         ; ("warnings", Js.Unsafe.inject (Js.string !stanc_stderr)) |]
   | Error e ->
       Js.Unsafe.obj
         [| ("errors", Js.Unsafe.inject (Array.map ~f:Js.string [|e|]))
-         ; ("warnings", Js.Unsafe.inject Js.array_empty) |]
+         ; ("warnings", Js.Unsafe.inject (Js.string !stanc_stderr)) |]
 
 let stan2cpp_wrapped name code (flags : Js.string_array Js.t Js.opt) =
+  stanc_stdout := "" ;
+  stanc_stderr := "" ;
```
rok-cesnovar, line 91: We have to reset every time because these are global variables. If we don't, we just accumulate warnings from all previous compilations.

src/stancjs/stancjs.ml
```diff
@@ -72,21 +71,26 @@ let stan2cpp model_name model_string flags =
                Pedantic_analysis.print_warn_uninitialized mir ;
              if is_flag_set "warn-pedantic" then
                Pedantic_analysis.print_warn_pedantic mir ;
-             (cpp, []) )
+             cpp )
+
+let stanc_stdout = ref ""
+let stanc_stderr = ref ""
```
rok-cesnovar, line 77: Would it be better to use a list or something?

rok-cesnovar:
@seantalts sorry to keep tagging you with these stanc.js PRs. 

If anyone else feels they can review this, feel free.

wds15:
Let's hope this is the last wall for rstan to cross over... this is getting heroic to make it work.

rok-cesnovar:
The critical stuff should hopefully now be squashed. 

We need to find a better way to support includes but that can be workaround for now.

wds15:
We should make bets on a 2020 or a 2021 rstan 2.25...

rok-cesnovar:
Bump. Anyone up for reviewing this?

seantalts:
Taking a look!

src/stancjs/stancjs.ml
seantalts: yeah, a list would be constant time to append to instead of O(N)

src/stancjs/stancjs.ml
seantalts: A better pattern would be to have the components return warnings and errors rather than printing them. Could we do that instead or would that mess up the timing for a release or something?

src/stancjs/stancjs.ml
```diff
@@ -45,21 +45,20 @@ let stan2cpp model_name model_string flags =
           "Semantic check failed but reported no errors. This should never \
            happen."
   in
-  if is_flag_set "version" then Result.Ok (Fmt.strf "%s" version, [])
```
seantalts, line 48: I think we want to do something to keep the ability for phases of the compiler to return warnings rather than delete these for-now empty lists.

seantalts (changes requested):

src/stancjs/stancjs.ml
rok-cesnovar: That was my first plan, but I quickly realized that is a pretty big refactor. Particulary not sure on how to return from the parser for example.

rok-cesnovar:
Closing in favor of #777

Issue #748, opened 2020-11-21
Post occasioned by stan's new compiler warning msg asking me to post here. Low Priority


PeterLeopold:
While building a model with stan_weibull_survival_model, this message appeared:

> When you compile models, you are also contributing to development of the NEXT
> Stan compiler. In this version of rstan, we compile your model as usual, but
> also test our new compiler on your syntactically correct model. In this case,
> the new compiler did not work like we hoped. By filing an issue at
> https://github.com/stan-dev/stanc3/issues with your model
> or a minimal example that shows this warning you will be contributing
> valuable information to Stan and ensuring your models continue working. Thank you!
> This message can be avoided by wrapping your function call inside suppressMessages()
>  or by first calling rstan_options(javascript = FALSE).
> 
> Semantic error in 'string', line 49, column 12 to column 49:
> 
> Target can only be accessed in the model block or in definitions of functions with the suffix _lp.

None of that means much to me. Line 49 of what, exactly? Not my R script! But I was able to use the resulting build well enough. But as a good citizen of the web, I'm posting this here.
 
```stan
functions { 
  vector sqrt_vec(vector x) { 
    vector[dims(x)[1]] res; 
 
    for (m in 1:dims(x)[1]){ 
      res[m] = sqrt(x[m]); 
    } 
 
    return res; 
  } 
 
  vector bg_prior_lp(real r_global, vector r_local) { 
    r_global ~ normal(0.0, 10.0); 
    r_local ~ inv_chi_square(1.0); 
 
    return r_global * sqrt_vec(r_local); 
  } 
} 
 
data { 
  int<lower=0> Nobs; 
  int<lower=0> Ncen; 
  int<lower=0> M_bg; 
  vector[Nobs] yobs; 
  vector[Ncen] ycen; 
  matrix[Nobs, M_bg] Xobs_bg; 
  matrix[Ncen, M_bg] Xcen_bg; 
} 
 
transformed data { 
  real<lower=0> tau_mu; 
  real<lower=0> tau_al; 
 
  tau_mu = 10.0; 
  tau_al = 10.0; 
} 
 
parameters { 
  real<lower=0> tau_s_bg_raw; 
  vector<lower=0>[M_bg] tau_bg_raw; 
 
  real alpha_raw; 
  vector[M_bg] beta_bg_raw; 
 
  real mu; 
} 
 
transformed parameters { 
  vector[M_bg] beta_bg; 
  real alpha; 
 
[...]

PR #767, opened 2020-12-16
allow lp functions in transformed parameters

nhuurre:
Prompted by issue #748 
One line change because `write_array()` method already had `lp_accum__` variable. The fact that it's not needed for anything other than `_lp` functions suggest that this may have been intented all along.

In any case, this is a backwards compatibility gotcha and [per discussion on the forum](https://discourse.mc-stan.org/t/accessing-target-in-transformed-parameters-stanc2-vs-stanc3/19432/3) something like this is needed.

## Release notes

Allow calling lp functions in transformed parameters block.

## Copyright and Licensing

Copyright holder: Niko Huurre

### DIFF
--- a/src/frontend/Semantic_check.ml
+++ b/src/frontend/Semantic_check.ml
@@ -261,9 +261,7 @@ let semantic_check_unnormalized cf ~loc id =
   Validate.(
     if
       Utils.is_unnormalized_distribution id.name
-      && not
-           ( (cf.in_fun_def && (cf.in_udf_dist_def || cf.in_lp_fun_def))
-           || cf.current_block = Model )
+      && not ((cf.in_fun_def && cf.in_udf_dist_def) || cf.current_block = Model)
     then Semantic_error.invalid_unnormalized_fn loc |> error
     else ok ())
 
--- a/src/frontend/Semantic_check.ml
+++ b/src/frontend/Semantic_check.ml
@@ -234,7 +234,9 @@ let semantic_check_fn_target_plus_equals cf ~loc id =
   Validate.(
     if
       String.is_suffix id.name ~suffix:"_lp"
-      && not (cf.in_lp_fun_def || cf.current_block = Model)
+      && not
+           ( cf.in_lp_fun_def || cf.current_block = Model
+           || cf.current_block = TParam )
     then Semantic_error.target_plusequals_outisde_model_or_logprob loc |> error
     else ok ())
 
--- a/src/frontend/Semantic_error.ml
+++ b/src/frontend/Semantic_error.ml
@@ -364,7 +364,7 @@ module ExpressionError = struct
         Fmt.pf ppf
           "Functions with names ending in _lupdf and _lupmf can only be used \
            in the model block or user-defined functions with names ending in \
-           _lpdf, _lpmf or _lp."
+           _lpdf or _lpmf."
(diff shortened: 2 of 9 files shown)


rok-cesnovar (changes requested):
We need to remove allowing `_lupdf ` in _lp functions as well https://github.com/stan-dev/stanc3/pull/767/files#diff-ac78bfdef130e5ec4b7b3b245cddf78d3e7cb992ea1849fadafb2a70e248a0ffL263

rok-cesnovar:
Hm. And this would also be allowed then:
```
functions {
  void smth_lp(real p) {
    p ~ std_normal();
  }  
}
parameters {
    real y;
}
transformed parameters {
    real r = smth_lp(2.0);
}
model {
    y ~ std_normal();
}
```

nhuurre:
Good catch.
I'd like to keep `target += *_lupdf` as being interchangeable with a tilde statement but I suppose it's not going to work...

rok-cesnovar:
If we are not allowing `_lupdf` then we also shouldn't allow the tilde statement right?

I guess one possible solution is to check that a _lp function used in TP does not use _lupdf/tilde.

nhuurre:
> If we are not allowing _lupdf then we also shouldn't allow the tilde statement right?

lupdf in a tilde statement is not a problem because `write_array()` never reads the value of the accumulator. Although the value may be inconsistent with what was computed in the model block this is not visible to the user.

> I guess one possible solution is to check that a _lp function used in TP does not have _lupdf/tilde but that can get messy and its also pretty tricky.

Yeah, I don't think special restriction based call-site are a good idea.

rok-cesnovar:
I am not sure I completely understand, so allow me another question.

Allowing `_lupdf` would allow:
```stan
functions {
  real lprob_lp(real x) {
    return normal_lupdf(x | 0, 1);
  }
}
parameters {
  real x;
}
transformed parameters {
  real ll = lprob_lp(x);
}
model {
  target += ll;
}
```
With the tilde statement this would only be a problem if someone did this right?

```stan
functions {
  real lprob_lp(real x) {
    x ~ normal(0, 1);
    return target();
  }
}
parameters {
  real x;
}
transformed parameters {
  real ll = lprob_lp(x);
}
model {
  target += ll;
}
```
Which is what stanc2 allows.

nhuurre:
Yes, that looks right. So the inconsistency with unnormalized distributions was present already in stanc2, it was just difficult to find.

rok-cesnovar:
Yeah. So we either:

a) allow _lp in TP, restrict _lupdf and reveal back the difficult to find inconsistency which is present in stanc2
b) not allow _lp in TP and technically break backward compatibility
c) do a check of whether the _lp is inconsistency-proof and allow/not allow based on that

All bad options :) 

a) is I guess the least bad?

nhuurre:
It's also possible to forbid `lupdf` and `target()/get_lp()` but allow `target+=` and tilde in `_lp` function body.

I'm not sure what is the use case for `target()`; the function can't rely on it having a predictable value anyway. If it's only printed immediately for debugging purposes and never stored then the inconsistency isn't too worrying.

nhuurre:
Pushed a change to disallow _lupdfs in _lp functions. With that the only nondeterminism is confined in `target()/get_lp()` and that's probably fine?

rok-cesnovar (approved):
Agree.

Issue #487, opened 2020-03-23
Function argument missing type error on wrong line


bbbales2:
This is the function definition:
```
functions {
  real reduce_func(int start, int end,
                   int[] slice_n_redcards,
                   int[] n_games,
                   vector rating,
                   beta) {
    return binomial_logit_lpmf(slice_n_redcards |
                               n_games[start:end],
                               beta[1] + beta[2] * rating );
  }
}
```
The error is the lack of type defined for beta.

This is the error:
```
--- Translating Stan model to C++ code ---
bin/stanc  --o=redcardt.hpp redcardt.stan

Syntax error in 'redcardt.stan', line 5, column 18 to column 19, parsing error:
   -------------------------------------------------
     3:  		   int[] slice_n_redcards,
     4:  		   int[] n_games,
     5:  		   vector rating,
                            ^
     6:  		   beta) {
     7:      return binomial_logit_lpmf(slice_n_redcards |
   -------------------------------------------------

An argument declaration (unsized type followed by identifier) is expected.


make: *** No rule to make target 'redcardt.hpp', needed by 'redcardt'.  Stop.
```

The caret is pointing to the wrong line. Here is the full model (taken and modified from: https://github.com/rmcelreath/cmdstan_map_rect_tutorial/blob/master/README.md):

```
functions {
  real reduce_func(int start, int end,
                   int[] slice_n_redcards,
                   int[] n_games,
                   vector rating,
                   beta) {
    return binomial_logit_lpmf(slice_n_redcards |
                               n_games[start:end],
                               beta[1] + beta[2] * rating );
  }
}
data {
  int N;
  int n_redcards[N];
  int n_games[N];
  vector[N] rating;
}
parameters {
  vector[2] beta;
}
model {
  beta ~ normal(0,1);

  target += reduce_func(1, size(n_redcards), n_redcards,
                        n_games, rating, beta);
}
```

Issue #463, opened 2020-02-10
Caret placement for keyword variable names error points to end of variable


seantalts:
Points to the end of the keyword instead of the beginning,

```
   -------------------------------------------------
     1:  transformed data {
     2:    int offset = 10;
                     ^
     3:  }
     4:  model { }
   -------------------------------------------------

Expected identifier, but found reserved keyword 'offset'
```

cc @VMatthijs in case he knows how to adjust the pointer

VMatthijs:
It's a hard issue for syntax errors. (For semantic errors, it's easy.) I've tried fixing this before, but no succes.

Issue #329, opened 2019-10-10
Uninformative error message for mistyped variable declaration


nhuurre:
Trying to compile the following invalid model
```stan
parameters {
  // typo
  cholesky_corr[3] x;
}
```
Stan2 gives a "helpful" error message
```
SYNTAX ERROR, MESSAGE(S) FROM PARSER:
 error in 'chol.stan' at line 3, column 2
  -------------------------------------------------
     1: parameters {
     2:   // typo
     3:   cholesky_corr[3] x;
         ^
     4: }
  -------------------------------------------------

PARSER EXPECTED: <one of the following:
  a variable declaration, beginning with type,
      (int, real, vector, row_vector, matrix, unit_vector,
       simplex, ordered, positive_ordered,
       corr_matrix, cov_matrix,
       cholesky_corr, cholesky_cov
  or '}' to close variable declarations>

```
but stanc3 has a less (mis)informative one
```
Syntax error in 'chol.stan', line 1, column 11 to column 12, parsing error:
   -------------------------------------------------
     1:  parameters {
                     ^
     2:    // typo
     3:    cholesky_corr[3] x;
   -------------------------------------------------

Expected top-level variable declaration or "}".
```
Note especially the misplaced caret.

seantalts:
Thanks for reporting, I can verify this happens on master. 

In the Stan 2 error message, isn't that error message telling you that `cholesky_corr` is actually valid? Just checking, haha. 

tagging @VMatthijs in case he sees an easy way out of this - the Menhir state is

```
program: PARAMETERSBLOCK LBRACE WHILE
##
## Ends in an error in state: 605.
##
## parameters_block -> PARAMETERSBLOCK LBRACE . list(top_var_decl_no_assign) RBRACE [ TRANSFORMEDPARAMETERSBLOCK MODELBLOCK GENERATEDQUANTITIESBLOCK EOF ]
##
## The known suffix of the stack is as follows:
## PARAMETERSBLOCK LBRACE
##

Expected top-level variable declaration or "}".
```

So my questions are 
1) is it easy to adjust the line number to the thing that failed parsing instead of the end of the thing that succeeded parsing - if we did that it seems easy enough to add this list to the messages
2) should we try to parse an identifier and then check it in a separate step rather than lexing each of those keywords as tokens? not sure what the tradeoffs there are.

VMatthijs:
We could definitely just change the error message for this state (and the corresponding one for the data block) to

```
[...]

PR #771, opened 2020-12-17
Change reported location of parsing errors

nhuurre:
A couple of issues (#487, #463, #329) complain that parsing error points to the last expression that parsed successfully rather than the first one that didn't. This PR tries to fix that.

I'm not sure if it's correct in all cases. For example the error that says "`Found an expression where we expected a statement`" does not point to the expression.

## Release notes

Move the reported location of parsing errors closer to the problematic section.

## Copyright and Licensing

Copyright holder: Niko Huurre


rybern (approved):
This looks like a good approach to me. I spent some time looking through it and didn't find any issues, and the new error locations do look better, although I agree there might be some adjustments needed for specific error messages like the one you pointed out.

W.r.t the new keyword error setup, it does feel a little bit strange to go string -> token -> string -> pattern match, but I agree that the parser error message system is pretty lousy and this does a better job of locating the error.

I'd be happy for this to me merged when you feel comfortable with the messages like the one you pointed out.

nhuurre:
Thanks for the fast review.

I've decide to leave the error messages as is.
The only one that seems wrong is the aforementioned
```
Found a expression where we expected a statement. Is there a missing semi-colon here?
Or did you mean to use the preceding expression in:
* a function call
* a sampling statement
* the conditional in a for, while, or if statement
* assignment to a variable?
```
Despite what it says I don't think a missing semicolon can trigger it. The only one-character changes that can are
```diff
- x +  1;
+ x += 1;

- foo(x|0,1);
+ foo(x,0,1);
```
(In both cases the caret points to the semicolon.) Seems rather obscure and I don't see a way to point the caret to the correct location.

PR #777, opened 2020-12-27
Refactor compiler phases to avoid printing to std{out, err}

### LINKED ISSUE #741: javascript parser chokes on #-style comments

seantalts:
Fixes #741. @rok-cesnovar you were right, it was a larger refactor, lol. I got started and kept thinking it was almost done. I did a little unrelated cleanup as well and I'm proud to say this PR net deletes 200 lines (though mostly in repeated test code), heh.

## Release Notes
Removed stdout and stderr in stanc.js.
 
## Copyright and Licensing


seantalts:
I think with this in, we could do a bit more refactoring to adjust the filenames at print time and fix #716 in a not too bad way too...

seantalts:
(what I'm trying to do now is add an optional flag to pretty-printing warnings that will replace the filename just when pretty printed)

rok-cesnovar (approved):
Thanks so much! This looks good to me. 

Definitely a way to big refactor for me because it touches a bunch of fronted and the formatter. I was too lazy and opted for the easier route :)

rok-cesnovar:
I expanded the test to make sure someone doesnt reintroduce warnings printing for the current warnings. I pushed directly, hope you dont mind. If you think we can also fix what 716 is trying to that would be great. But can also merge as is.

seantalts:
Oh nice, good catch and test! Thanks and yep pushing directly is great. Seems like your computer and the Jenkins one (and mine) disagree about how node should print those long lines... any idea what's up with that? different versions of node maybe?

for now I will promote and then merge and keep working on a new PR for #716. It already has some cleanup of this one but I wanted to keep them split out into separate PRs to make them easier to review with a single intention for each one.

I've been wanting to do some refactoring in here for a while - it's way too obvious there were at least 3 very different authors in here, haha. Plus we had a big refactor on the MIR that was supposed to be extended to the AST but hasn't been yet, so those look completely different as well. Hopefully over time we can unify to one style!

rok-cesnovar:
> any idea what's up with that? different versions of node maybe? 

I would say so. Wrapped in a stringify to see if that would help.

rok-cesnovar:
It seems that wrapping in JSON.stringify helps. Will merge once tests pass.

PR #780, opened 2020-12-27
Refactor warnings and add filename-in-msg

seantalts:
Fixes #716 without a global variable and refactors Warnings a bunch across deprecations in lexer, in the analysis, and pedantic mode, and unifies output according to some discussion in #726 about ideal message format.

## Release notes
* adds filename-in-msg argument to stancjs and stanc to replace the filename in printed messages.
* Warning messages are made shorter and more often fit on one line. TBD: should we warn every time we see a deprecated language construct?

## Copyright and Licensing


rok-cesnovar:
Just did another merge as some changes were weird. Will take a look now, might take awhile :)

test/stancjs/stancjs.expected
```diff
@@ -31,6 +31,19 @@ model {
   y ~ normal(0, 1);
 }
 
+$ node filename.js
+
+Semantic error in 'string', line 6, column 4 to column 5:
```
rok-cesnovar, line 36: This should be foo right?

test/integration/rstanarm/pretty.expected
```diff
@@ -49,7 +48,7 @@ functions {
     }
     return theta_L;
   }
-  vector make_b(vector z_b, vector theta_L, array[] int p, array[] int l) {
```
rok-cesnovar, line 52: Was this change intended? This seems a bit weird to me?

rok-cesnovar (commented):
Two quick questions before I continue. 

I imagine the change from 

`vector target_(vector y, vector theta, array[] real x_r, array[] int x_i) {
`
back to 

`vector target_(vector y, vector theta, real[] x_r, int[] x_i) {
`
was not intended here? There are a bunch of files touched due to this.

test/stancjs/stancjs.expected
seantalts: Woops, yes

seantalts:
@rok-cesnovar I wasn't sure how to go with the `array[]` thing. Currently, we print out `array[]` when we're canonicalizing and I think also pretty printing but we print the old form for all of the error messages and warnings. I was kind of thinking we'd want to switch them both to the same syntax. I'll fix the error messages and warnings to print out signatures etc. in the new array syntax, how's that sound? I went the other way at first in an attempt to have a small test diff, but I think I ruined that already.

rok-cesnovar:
Ooooh, ok that makes sense then. I think we want people using the new sytnax so lets just print the new syntax.

seantalts:
The diff is a lot larger now, sorry about that! But mostly tests changing to reflect the new array syntax, which we want to do anyway.

rok-cesnovar:
Cool, thanks. Will take a look tomorrow.

rok-cesnovar:
I have gotten through most of the changes and the filename in error messages look good.

If we go with the array[] prints we will need to fix the expressions tests in Stan Math first (these work with --dump-stan-math-signatures on the stanc3 binary). Those currently dont work with array[] and that would break math/stan develop tests.

After thinking about it a bit I am now not 100% sure we want need to print with array[] (sorry for flip-flopping on this). The old syntax isnt really deprecated so we dont need to push the new syntax, but we want the users to use it.

@rybern @nhuurre What do you think? Do we want to print errors with `real[]` or `array[] real`?

rybern:
I think it makes sense to go with the `array[] real` syntax to stay consistent with canonicalization, that's my vote

seantalts:
@rok-cesnovar Is the only thing left to do for review to fix the Math tests? Where do those live? I can take a look at them.

rok-cesnovar:
Yes, I am good with the rest. Those tests live here: https://github.com/stan-dev/math/blob/develop/test/expressions/generateExpressionTests.py

This takes the signature dump and generates gtests on the fly for all signatures with eigen types. The tests check that function can handle input expressions and also work with them efficiently.

rok-cesnovar:
> In general for big refactors I think we should try to get them out quickly since they will break often and break other PRs often as well.

Agree, sorry I dropped the ball here. Got busy with other stuff.

seantalts:
Ah, I think it's just as much on me! We can try to all remind each other :)

seantalts:
Do you or @t4c1 have hints on how best to modify generateExpressionTests.py to work with the new "array[] <type>" syntax? It looks like it's assuming types are always one contiguous word, not sure the best place to fix that.

rok-cesnovar:
> It looks like it's assuming types are always one contiguous word

Yeah.

I think most of the sig handling is in https://github.com/stan-dev/math/blob/develop/test/sig_utils.py

seantalts:
@bgoodri will switching the function signatures to the array[] syntax require changes in RStan as well?

bgoodri:
I am not sure, but if so, I think we can go either way.

rok-cesnovar:
@seantalts I am going to check that expression tests and benchmarks still work locally and then this is good to go. If we want to merge this before the freeze we should proceed I guess.

seantalts:
No pressure on the merge before the freeze from me. Thanks! Can you tell me how to run them locally? :D I didn't see them in any of the normal places (Jenkinsfile, make) but maybe missed them. Was going to let Jenkins tell me haha.

rok-cesnovar:
Expression tests:
`./runTests.py test/expressions --only-function "sin" ` to run for sin() for example.

Benchmarks: 
./benchmarks/benchmark.py "matrix sin(matrix)" --plot="window"

You can place 
STANC3=~/Desktop/stanc3
in make/local to use the local stanc3.

Expression tests run in Stan: https://github.com/stan-dev/stan/blob/develop/Jenkinsfile#L286
My thinking there was stanc3 will at some point be a submodule of Stan and changes to stanc3 should run those. Not sure the reasoning is good.

rok-cesnovar:
> No pressure on the merge before the freeze from me. 

If so, then lets wait a week and merge and we then have 3 months to figure out if it breaks any other interface or script.

Thanks for the math PR!

seantalts:
I can't get the expression tests to see the new signatures. with STANC3 set it builds the right repo and copies the right stanc over but then prints out signatures the old way...

seantalts:
OH the file has a bunch of hard-coded signatures written the old way. Woops.

rok-cesnovar:
Ah yes, the edge cases that have to handled differently... Good call.

seantalts:
Just merged in the upstream test changes, I think this is ready to be reviewed.

rok-cesnovar:
I already reviewed and it looks good. Just need to try locally that this works with https://github.com/stan-dev/math/pull/2323 and will then merge.

rok-cesnovar (approved):

Issue #605, opened 2020-07-04
Cleanup the pre-model generated C++ code


rok-cesnovar:
Currently, the generated code starts with

```cpp
#include <stan/model/model_header.hpp>
namespace eight_schools_ncp_model_namespace {


inline void validate_positive_index(const char* var_name, const char* expr,
                                    int val) {
  if (val < 1) {
    std::stringstream msg;
    msg << "Found dimension size less than one in simplex declaration"
        << "; variable=" << var_name << "; dimension size expression=" << expr
        << "; expression value=" << val;
    std::string msg_str(msg.str());
    throw std::invalid_argument(msg_str.c_str());
  }
}

inline void validate_unit_vector_index(const char* var_name, const char* expr,
                                       int val) {
  if (val <= 1) {
    std::stringstream msg;
    if (val == 1) {
      msg << "Found dimension size one in unit vector declaration."
          << " One-dimensional unit vector is discrete"
          << " but the target distribution must be continuous."
          << " variable=" << var_name << "; dimension size expression=" << expr;
    } else {
      msg << "Found dimension size less than one in unit vector declaration"
          << "; variable=" << var_name << "; dimension size expression=" << expr
          << "; expression value=" << val;
    }
    std::string msg_str(msg.str());
    throw std::invalid_argument(msg_str.c_str());
  }
}


using std::istream;
using std::string;
using std::stringstream;
using std::vector;
using std::pow;
using stan::io::dump;
using stan::math::lgamma;
using stan::model::model_base_crtp;
using stan::model::rvalue;
using stan::model::cons_list;
using stan::model::index_uni;
using stan::model::index_max;
using stan::model::index_min;
using stan::model::index_min_max;
using stan::model::index_multi;
using stan::model::index_omni;
using stan::model::nil_index_list;
using namespace stan::math; 

static int current_statement__ = 0;
static const std::vector<string> locations_array__ = {" (found before start of program)",
                                                      " (in 'eight_schools_ncp.stan', line 8, column 2 to column 10)",
                                                      " (in 'eight_schools_ncp.stan', line 9, column 2 to column 20)",
                                                      " (in 'eight_schools_ncp.stan', line 10, column 2 to column 24)"
```
A couple of ideas: 

- `validate_positive_index` and `validate_unit_vector_index` should be moved to either Stan or Math. Does anyone have a preference where? I think Stan makes more sense here but dont have a strong preference.

[...]

PR #785, opened 2021-01-04
fix constraining transform codegen

### LINKED ISSUE #783: Compile error, possibly a problem for stan-dev/stan
From [this](https://discourse.mc-stan.org/t/error-cannot-create-vector-with-constraints-made-up-of-elements-of-a-cholesky-factor-correlation-matrix/19996) Discouse thread, this model gives a bunch of C++ errors on compile:

```
data {
  int num_refs;
  int nt;
  int NS;
  real z2;
}

parameters {
  ordered[2] a1_m_raw[num_refs];
  vector<lower=0>[2] sd[nt];
  cholesky_factor_corr[2] L_Omega[nt];
  vector<lower=(to_vector(a1_m_raw[ , 1]) - to_vector(a1_m_raw[ , 2]) + sd[1,2]*z2*L_Omega[1,2,2])/(sd[1,1] - L_Omega[1,1,2]*sd[1,2])>[NS] z1; //  Se + Fp > 1                                
}
```

Here's the top of the output from my computer:
```
clang++ -std=c++1y -D_REENTRANT -Wno-sign-compare -Wno-ignored-attributes      -I stan/lib/stan_math/lib/tbb_2019_U8/include   -O3 -I src -I stan/src -I lib/rapidjson_1.1.0/ -I lib/CLI11-1.9.1/ -I stan/lib/stan_math/ -I stan/lib/stan_math/lib/eigen_3.3.7 -I stan/lib/stan_math/lib/boost_1.72.0 -I stan/lib/stan_math/lib/sundials_5.2.0/include    -DBOOST_DISABLE_ASSERTS        -c -include-pch stan/src/stan/model/model_header.hpp.gch -x c++ -o constrain.o constrain.hpp
In file included from constrain.hpp:1:
In file included from stan/src/stan/model/model_header.hpp:15:
In file included from /home/bbales2/cmdstan-develop/stan/src/stan/model/indexing.hpp:9:
/home/bbales2/cmdstan-develop/stan/src/stan/model/indexing/rvalue.hpp:764:12: error: no matching function for call to 'rvalue'
    return rvalue(std::move(v[idxs.head_.n_ - 1]), idxs.tail_, name, depth + 1);
           ^~~~~~
constrain.hpp:490:7: note: in instantiation of function template specialization 'stan::model::rvalue<std::vector<Eigen::Matrix<double, -1, 1, 0, -1, 1>, std::allocator<Eigen::Matrix<double,
      -1, 1, 0, -1, 1> > > &, stan::model::cons_index_list<stan::model::index_uni, stan::model::cons_index_list<stan::model::index_uni, stan::model::nil_index_list> >, nullptr>' requested
      here
      rvalue(L_Omega_in__,
      ^
/home/bbales2/cmdstan-develop/stan/src/stan/model/indexing/rvalue.hpp:54:10: note: candidate function not viable: no known conversion from 'const std::decay_t<cons_index_list<index_uni,
      cons_index_list<index_uni, nil_index_list> > >' (aka 'const stan::model::cons_index_list<stan::model::index_uni, stan::model::cons_index_list<stan::model::index_uni,
      stan::model::nil_index_list> >') to 'const stan::model::nil_index_list' for 2nd argument
inline T rvalue(T&& x, const nil_index_list& /*idx*/, const char* /*name*/ = "",
         ^
/home/bbales2/cmdstan-develop/stan/src/stan/model/indexing/rvalue.hpp:74:10: note: candidate function not viable: no known conversion from 'const std::decay_t<cons_index_list<index_uni,
      cons_index_list<index_uni, nil_index_list> > >' (aka 'const stan::model::cons_index_list<stan::model::index_uni, stan::model::cons_index_list<stan::model::index_uni,
      stan::model::nil_index_list> >') to 'const cons_index_list<stan::model::index_omni, stan::model::nil_index_list>' for 2nd argument
inline T rvalue(T&& x, const cons_index_list<index_omni, nil_index_list>& idxs,
         ^
/home/bbales2/cmdstan-develop/stan/src/stan/model/indexing/rvalue.hpp:112:10: note: candidate function not viable: no known conversion from 'const std::decay_t<cons_index_list<index_uni,
      cons_index_list<index_uni, nil_index_list> > >' (aka 'const stan::model::cons_index_list<stan::model::index_uni, stan::model::cons_index_list<stan::model::index_uni,
      stan::model::nil_index_list> >') to 'const cons_index_list<stan::model::index_omni, cons_index_list<stan::model::index_omni, stan::model::nil_index_list> >' for 2nd argument
inline T rvalue(
```

I think this model should compile, so this might be an issue for `stan-dev/Stan`.

nhuurre:
Fix #783 

When constrained and unconstrained parameters have different types codegen needs to convert the name of the variable being constrained. The problem was that it tried to convert _all_ arguments to the constraining function when only the first one should be converted.

## Copyright and Licensing

Copyright holder: Niko Huurre


test/integration/good/code-gen/cpp.expected
```diff
@@ -20110,6 +20110,573 @@ stan::model::model_base& new_model(
...
+                                    int val) {
+  if (val < 1) {
+    std::stringstream msg;
+    msg << "Found dimension size less than one in simplex declaration"
+        << "; variable=" << var_name << "; dimension size expression=" << expr
+        << "; expression value=" << val;
+    std::string msg_str(msg.str());
+    throw std::invalid_argument(msg_str.c_str());
+  }
+}
```
SteveBronder, line 20130: Is this generated by the compiler? One little optimization trick we use for error checking in the math lib is to tell the compiler that these errors are on the cold path so the compiler/cpu knows there is a very low chance it ever has to care about this code

```suggestion
inline void validate_positive_index(const char* var_name, const char* expr,
                                    int val) {
  if (val < 1) {
      [&]() STAN_COLD_PATH {
        std::stringstream msg;
        msg << "Found dimension size less than one in simplex declaration"
          << "; variable=" << var_name << "; dimension size expression=" << expr
          << "; expression value=" << val;
        std::string msg_str(msg.str());
        throw std::invalid_argument(msg_str.c_str());
    }();
  }
}
```

test/integration/good/code-gen/cpp.expected
```diff
@@ -20110,6 +20110,573 @@ stan::model::model_base& new_model(
...
+          << " variable=" << var_name << "; dimension size expression=" << expr;
+    } else {
+      msg << "Found dimension size less than one in unit vector declaration"
+          << "; variable=" << var_name << "; dimension size expression=" << expr
+          << "; expression value=" << val;
+    }
+    std::string msg_str(msg.str());
+    throw std::invalid_argument(msg_str.c_str());
+  }
+}
```
SteveBronder, line 20149:
```suggestion
  inline void validate_unit_vector_index(const char* var_name, const char* expr,
                                         int val) {
    if (val <= 1) {
      [&]() STAN_COLD_PATH {
        std::stringstream msg;
        if (val == 1) {
          msg << "Found dimension size one in unit vector declaration."
              << " One-dimensional unit vector is discrete"
              << " but the target distribution must be continuous."
              << " variable=" << var_name << "; dimension size expression=" << expr;
        } else {
          msg << "Found dimension size less than one in unit vector declaration"
              << "; variable=" << var_name << "; dimension size expression=" << expr
              << "; expression value=" << val;
        }
        std::string msg_str(msg.str());
        throw std::invalid_argument(msg_str.c_str());
      }();
    }
  }
```

test/integration/good/code-gen/cpp.expected
```diff
@@ -20110,6 +20110,573 @@ stan::model::model_base& new_model(
...
+      L_Omega_in__ = std::vector<Eigen::Matrix<local_scalar_t__, -1, 1>>(nt, Eigen::Matrix<local_scalar_t__, -1, 1>(
+        ((2 * (2 - 1)) / 2)));
+      stan::math::fill(L_Omega_in__, DUMMY_VAR__);
+      
+      current_statement__ = 1;
+      for (int sym1__ = 1; sym1__ <= nt; ++sym1__) {
+        current_statement__ = 1;
+        assign(L_Omega_in__, cons_list(index_uni(sym1__), nil_index_list()),
+          in__.vector(((2 * (2 - 1)) / 2)), "assigning variable L_Omega_in__");
+      }
```
SteveBronder, line 20281: Will this run? Reader has state so idt you can call `in__.vector(N)` multiple times just getting out pieces like this.

 `in__` here is a [`reader`](https://github.com/stan-dev/stan/blob/develop/src/stan/io/reader.hpp#L189) class  which stores a reference to the input data as a serialized vector. When we call `in__.vector(N)` you can see in the link above we grab a block of data from the serialized vector of size N then increment an internal sentinel value so the next call will grab the next chunk of data.

Can this just be generated as

```suggestion
current_statement__ = 1;
assign(L_Omega_in__, nil_index_list(),
  in__.vector(((2 * (2 - 1)) / 2)), "assigning variable L_Omega_in__");
```

test/integration/good/code-gen/cpp.expected
```diff
@@ -20110,6 +20110,573 @@ stan::model::model_base& new_model(
...
+          current_statement__ = 1;
+          assign(L_Omega, cons_list(index_uni(sym1__), nil_index_list()),
+            stan::math::cholesky_corr_constrain(L_Omega_in__[(sym1__ - 1)],
+              2, lp__), "assigning variable L_Omega");
+        } else {
+          current_statement__ = 1;
+          assign(L_Omega, cons_list(index_uni(sym1__), nil_index_list()),
+            stan::math::cholesky_corr_constrain(L_Omega_in__[(sym1__ - 1)], 2),
+            "assigning variable L_Omega");
+        }}
```
SteveBronder, line 20295: Actually if you can nix  `L_Omega_in__` it would be wayy more efficient to use [`in__.cholesky_factor_corr_constrain()`](https://github.com/stan-dev/stan/blob/develop/src/stan/io/reader.hpp#L1184)  like

```suggestion
    if (jacobian__) {
      current_statement__ = 1;
      assign(L_Omega, nil_index_list(),
      in__.cholesky_factor_corr_constrain(((2 * (2 - 1)) / 2), lp__),
       "assigning variable L_Omega");
    } else {
      current_statement__ = 1;
      assign(L_Omega, nil_index_list(), 
        in__.cholesky_factor_corr_constrain(((2 * (2 - 1)) / 2)),
        "assigning variable L_Omega");
    }
```

test/integration/good/code-gen/cpp.expected
```diff
@@ -20110,6 +20110,573 @@ stan::model::model_base& new_model(
...
+        } else {
+          current_statement__ = 2;
+          assign(z1, cons_list(index_uni(sym1__), nil_index_list()),
+            stan::math::lb_constrain(z1[(sym1__ - 1)],
+              rvalue(L_Omega,
+                cons_list(index_uni(1),
+                  cons_list(index_uni(1),
+                    cons_list(index_uni(2), nil_index_list()))), "L_Omega")),
+            "assigning variable z1");
+        }}
```
SteveBronder, line 20323: Is there a reason a lot of the vectorized versions of this code is ignored? Like here again reader has [`vector_lb_constrain()`](https://github.com/stan-dev/stan/blob/develop/src/stan/io/reader.hpp#L1304)

I didn't know this is what actually happened in constrain etc functions within the compiler. For the new matrix type we really need to use those vectorized specializations or this will be a performance killer

SteveBronder (commented):
Couple little comments and one bug catch. I'd really suggest we just use [`reader's`](https://github.com/stan-dev/stan/blob/develop/src/stan/io/reader.hpp#L1304) vectorized functions directly for constraints etc. In general it looks like the for loop scheme actually causes quite a few unnecessary copies/reads/writes.

SteveBronder:
>  I'd really suggest we just use reader's vectorized functions directly for constraints etc. In general it looks like the for loop scheme actually causes quite a few unnecessary copies/reads/writes.

Also if there is anything on the stan backend that can change to make this easier for the compiler lmk. One thing that looks like it may be made easier is if all the signatures are more standardized so the compiler would essentially output something like 

```cpp
assign(X, nil_index_list(), in__.[scalar/matrix/vector/row_vector]_[constrain_fn_name](...))
```

test/integration/good/code-gen/cpp.expected
rok-cesnovar: Yes this is generated by the compiler and should be moved to Stan Math (mentioned in https://github.com/stan-dev/stanc3/issues/605). These are variants of validate_non_negative_index as Niko notes in the issue.

I think we should just move this to Math and optimize it there.

SteveBronder: Wait can't this just use `stan::math::check_positive()`? Like 
```cpp
stan::math::check_positive("", "X", "X", X);
```

rok-cesnovar: Could be. I honestly havent looked into that in detail. It should be moved out of generated C++ in any case. Just not sure if we have time to do that before the release.

rok-cesnovar:
@SteveBronder Your comments were mostly about generated code that wasnt changed in this PR.

I think the fix for the actual issue this PR is addressing is good and would like to just merge as is if you agree.

Your points are very much valid but those should be addressed separately. I updated the already open issues on code-gen cleanup with your points.

SteveBronder:
Yeah it's fine to ignore the comments let me double check them to make sure there wasn't anything directly related to the pr

test/integration/good/code-gen/cpp.expected
SteveBronder: actually nvm about this comment. I misread the code and this is reading in vectors of size `((2 * (2 - 1)) / 2)` `nt` times which is the right behavior

rok-cesnovar:
@SteveBronder I am guessing you are good with merging this? Or did you find anything related to this PR.

rok-cesnovar (approved):

rok-cesnovar:
Had another look and none of Steve’s comments are wrt to the changes here so lgtm!

PR #791, opened 2021-01-12
Math signatures of elementwise and non-elementwise operators for multiplication and division [WIP]

adamhaber:
## Release notes

Fixed inconsistent use of elementwise and non-elementwise operators for multiplication and division

## Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause)

PR #793, opened 2021-01-12
Revert "Math signatures of elementwise and non-elementwise operators for multiplication and division [WIP]"

rok-cesnovar:
Reverts stan-dev/stanc3#791

The math implementations for some of these is missing. We need to revert this and add them.


rok-cesnovar:
Just going to merge as this is blocking develop tests.

PR #794, opened 2021-01-14
Profiling

rok-cesnovar:
A new version for implementing profiling based on discussion in https://github.com/stan-dev/design-docs/pull/31

An example of use:

```stan
data {
  int<lower=1> N;
  real x[N];
  vector[N] y;
}
transformed data {
   
   real sum_y;
   profile("sum") {
      sum_y = sum(y);
   }   
}
parameters {
  real<lower=0> rho;
  real<lower=0> alpha;
  real<lower=0> sigma;
}

model {
   matrix[N, N] cov;
   matrix[N, N] L_cov;
   profile("cov_exp_quad") {
      cov =   gp_exp_quad_cov(x, alpha, rho)
                        + diag_matrix(rep_vector(sigma, N));
   }
   profile("cholesky_decompose"){
      L_cov = cholesky_decompose(cov);
   }   
   profile("multi_normal_cholesky"){
      rho ~ gamma(25, 4);
      alpha ~ normal(0, 2);
      sigma ~ normal(0, 1);

      y ~ multi_normal_cholesky(rep_vector(0, N), L_cov);
   }
}
```

It adds a profiling__ member to the C++ class.

For 

```stan
profile("sum") {
   sum_y = sum(y);
}   
```
it generates

```cpp
{
        profile<local_scalar_t__> profile_0__("sum", const_cast<profile_map&>(profiles__));
        
        current_statement__ = 25;
        sum_y = sum(y);
}
```

## Release notes

Added profiling blocks to the Stan language.

## Copyright and Licensing


src/frontend/parser.messages
```diff
@@ -9416,7 +9416,70 @@ program: TRANSFORMEDDATABLOCK LBRACE WHILE LPAREN UPPER RPAREN UPPER VECTOR
...
+##
+## Ends in an error in state: 299.
+##
+## nested_statement -> PROFILE . LPAREN string_literal RPAREN LBRACE list(vardecl_or_statement) RBRACE [ WHILE VOID VECTOR UPPER UNITVECTOR TRUNCATE TARGET SIMPLEX SEMICOLON ROWVECTOR RETURN REJECT REALNUMERAL REAL RBRACE PROFILE PRINT POSITIVEORDERED PLUS ORDERED OFFSET MULTIPLIER MINUS MATRIX LPAREN LOWER LBRACK LBRACE INTNUMERAL INT INCREMENTLOGPROB IF IDENTIFIER GETLP FOR ELSE COVMATRIX CORRMATRIX CONTINUE CHOLESKYFACTORCOV CHOLESKYFACTORCORR BREAK BANG ARRAY ]
+##
+## The known suffix of the stack is as follows:
+## PROFILE
+##
+
+TODO: PARSER MESSAGE NEEDED HERE.
```
rok-cesnovar, line 9435: Do I just edit that message here? Never did this before so might need help.

rok-cesnovar:
@nhuurre @rybern @seantalts anyone have time for a review of this? Its ready apart from one question I have. Thanks!

src/frontend/parser.messages
nhuurre: Yes, the message goes here. It's delimited by empty lines, which feels a bit weird to me.
The `program:` shows an example of what leads to this error state
```stan
transformed data {
  profile while
          ^ error is here
```
so something like "Expected profile name in parens."

src/analysis_and_optimization/Optimize.ml
```diff
@@ -777,7 +778,7 @@ let dead_code_elimination (mir : Program.Typed.t) =
             && is_skip_break_continue body.pattern
           then Skip
           else For {loopvar; lower; upper; body}
-      | Block l ->
+      | Profile (_, l) | Block l ->
           let l' = List.filter ~f:(fun x -> x.pattern <> Skip) l in
           if List.length l' = 0 then Skip else Block l'
```
nhuurre, line 783: So dead code elimination pass converts profile blocks to ordinary blocks. I guess that's one way to not worry about optimized code profiling the wrong part. Was that intentional?

I think eventually we want optimization to keep profile blocks and not move code across their boundaries. But we can leave the interaction of profiling and optimization unspecified until we get around to addressing #545

rok-cesnovar: Yeah, mostly intentional because I dont know how it would work honestly. With inlining etc. 

I can look into that but not sure there is an easy way.

rok-cesnovar: Also agree with doing this after 545.

src/stan_math_backend/Statement_gen.ml
```diff
@@ -3,9 +3,20 @@ open Middle
...
+let n_profiles = ref 0
 let pp_call_str ppf (name, args) = pp_call ppf (name, string, args)
 let pp_block ppf (pp_body, body) = pf ppf "{@;<1 2>@[<v>%a@]@,}" pp_body body
 
+let pp_profile ppf (pp_body, name, body) =
+  let profile1 =
+    Fmt.strf
+      "profile<local_scalar_t__> profile_%d__(%s, \
+       const_cast<profile_map&>(profiles__));"
+      !n_profiles name
```
nhuurre, line 15: I don't like keeping a global `n_profiles` ref _just_ so you can give every profile a unique name. The C++ name of the profile object should not matter and can be reused even for nested profiles--it's always the first declaration in a new scope, and it's allowed to shadow variables in outer scopes.
If you don't want them all to have the same name you could generate the name from statement's location metadata.

seantalts:
There’s also a gensym construct for this purpose IIRC

src/stan_math_backend/Statement_gen.ml
rok-cesnovar: Ah, of course. That didnt occur to me that its not a problem with nested... Thanks. For the previous language interface we had to be careful because we could have multiple profiles in the same level of nesting. Its not a problem here so `profile__` it is.

test/integration/bad/profiling/stanc.expected
```diff
@@ -0,0 +1,83 @@
...
+
+Syntax error in 'profile-bad5.stan', line 2, column 9 to column 16, parsing error:
+   -------------------------------------------------
+     1:  transformed data {
+     2:      real profile;
+                  ^
+     3:  }
+   -------------------------------------------------
+
+Identifier expected after type in top-level variable declaration.
```
rok-cesnovar, line 67: TBH I dont like this error message but dont know how to 

If this is some other keyword like
```stan
transformed data {
    real while;
}
```
its a more informative message.

Identifier 'while' clashes with reserved keyword.

There is probably no trivial way of keeping profile unreserved?

nhuurre: Keyword tokens remain "unreserved" when listed in `identifier`
https://github.com/stan-dev/stanc3/blob/2c167b64a7eb9078500782ff8c988d562f329287/src/frontend/parser.mly#L132-L136
Although that may create ambiguity in the grammar.
You can change the error by listing the keyword under `decl_identifier` instead
https://github.com/stan-dev/stanc3/blob/2c167b64a7eb9078500782ff8c988d562f329287/src/frontend/parser.mly#L141-L148

rok-cesnovar: I think just a better error message would be fine, especially if unreserving it would cause ambiguity.

rok-cesnovar: Also thanks!

rok-cesnovar:
Added profile to be a reserved keyword. Hopefully that wont cause many issues. I guess grammar ambiguity would be a worse problem. 

As @nhuurre noted, profiling is ignored when `--O` is used, but might be allowed sometime later when we introduce optimization levels.

rok-cesnovar:
The design doc has been approved, so I guess we can proceed here with an official review. This now also has "bad" tests for all the new error messages.

bbbales2:
```
     6:                     vector beta) {
     7:      real lp = 0.0;
     8:      profile(rating) {
                     ^
     9:        lp = binomial_logit_lpmf(slice_n_redcards |
    10:  			       n_games[start:end],

Expected profile name in paranthesis.
```

[Optional] Any chance of saying something like "Expected profile name as string in parenthesis"?

This all seems to work well for me otherwise.

rok-cesnovar:
>Any chance of saying something like "Expected profile name as string in parenthesis"?

Sure, seems reasonable. 

@bbbales2 are you good otherwise with this? If so, then we just need to make sure the OCaml is ocamlish enough.

bbbales2:
This works well for me!

rok-cesnovar:
Anyone up for another review of this before the freeze tomorrow? Thanks?

nhuurre (approved):

rok-cesnovar:
Thanks Niko!

wds15:
This is a 2.26 feature?? Cool!

PR #807, opened 2021-02-01
Cleanup generated code

### LINKED ISSUE #605: Cleanup the pre-model generated C++ code

rok-cesnovar:
Fixes #605 

Requires https://github.com/stan-dev/math/pull/2346 to be merged in sync.

## Release notes

Cleaned up generated C++ of function definitions and using statements.

## Copyright and Licensing


src/stan_math_backend/Locations.ml
```diff
@@ -40,7 +40,8 @@ let pp_globals ppf location_list =
   in
   Fmt.pf ppf
     "@ stan::math::profile_map profiles__;@ static int current_statement__= \
-     0;@ static const std::vector<string> locations_array__ = {@[<hov>%a@]};@ "
+     0;@ static const std::vector<std::string> locations_array__ = @ \
+     {@[<hov>%a@]};@ "
```
rok-cesnovar, line 44: This is to break the long locations array lines:

```cpp
static const std::vector<std::string> locations_array__ = 
{" (found before start of program)",
 " (in 'basic.stan', line 3, column 8 to column 28)",
 " (in 'basic.stan', line 2, column 30 to line 4, column 5)",
```

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -37,16 +37,19 @@ let stanc_args_to_print =
...
 let pp_function__ ppf (prog_name, fname) =
-  pf ppf "static const char* function__ = %S;@ "
-    (strf "%s_namespace::%s" prog_name fname) ;
-  pp_unused ppf "function__"
+  pf ppf {|
+    static const char* function__ = %S;
+    %s
+    |}
+    (strf "%s_namespace::%s" prog_name fname)
+    (strf "%a" pp_unused "function__")
```
rok-cesnovar, line 52: This is so that the `(void) function__;` is now properly indented.

Before this it was:
```cpp
    static const char* function__ = "simple_function_model_namespace::write_array";
(void) function__;  // suppress unused var warning
```

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -509,7 +512,6 @@ let pp_write_array ppf {Program.prog_name; generate_quantities; _} =
     [ "using local_scalar_t__ = double;"; "vars__.resize(0);"
     ; "stan::io::reader<local_scalar_t__> in__(params_r__, params_i__);"
     ; strf "%a" pp_function__ (prog_name, "write_array")
-    ; strf "%a" pp_unused "function__"
```
rok-cesnovar, line 512: This can be removed because pp_function__ calls pp_unused internally. So this was just a duplicate.

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -829,44 +825,6 @@ using stan::model::index_multi;
 using stan::model::index_omni;
 using stan::model::nil_index_list;
 using namespace stan::math;
-using stan::math::pow; |}
```
rok-cesnovar, line 832: - using std::istream; - not used
- using std::string and using std::stringstream; - only one use that now has std::
- using std::pow; and using stan::math::pow; - these were temporary workaround that should be resolved now.
- using stan::math::lgamma; - same here

rok-cesnovar (commented):
Some quick comments. This is trivial stuff.

rok-cesnovar:
Math PR was merged and the tests are passing so this is ready for review.

seantalts:
I started reviewing and saw a bunch of mess I'd made and tried to clean it up a bit. Weird things were happening because `intro` was a list of lines that it would try to print a certain way; Format and Fmt work best compositionally, when you try to always use `%a` and pass formatters around. That way the library can handle the boxing / indentation. So I pushed a commit to clean up that old code a bit, let me know if it doesn't make sense or you're not a fan. 

Thanks again for doing work like this :)

seantalts (approved):

rok-cesnovar:
Looks great! Thanks!

Issue #808, opened 2021-02-02
Adding a --info flag to output meta information of the model


rok-cesnovar:
This is something I have seen users work around with various techniques, there are also packages that do this with regexes and the like.

I understand some of this can be accessed via the generated C++ models but that requires instantiating the model with the data and also requires tightly working with C++ which seems a bit of overkill.

Examples of use: 
- Posteriordb (see https://github.com/stan-dev/posteriordb/issues/198 and https://github.com/stan-dev/posteriordb/issues/198)
- stanbreaker by @jtimonen (parses the transformed mir to get out the information on the data/parameters https://github.com/jtimonen/stanbreaker/blob/main/R/code_analysis.R#L84)

This should not be to difficult to do properly in stanc3. As you can see in the posteriordb issue @mandel has already made a branch with almost everything we need. 

Meta data we could output and use:
- list of included files
- list of input data
- list of transformed parameters and parameters
- list of generated quantities
- dimensionality of all data & parameters
- (optional) does it use reduce_sum/map_rect

There may be other ideas. The output format should probably be pretty printed JSON as its both human readable and nice to work with in other scripts/languages.

@mandel would you be interested in making a PR of your branch?

jtimonen:
An additional wish would be list of user-defined functions and their arguments if that is not too difficult to do. My ultimate goal would be to create a tool that can take an rstantools-based package like rstanarm, which is written with many includes, conditions, etc, and given the input data, turn it into minimal stan code which is much more readable. So this would require removing blocks and loops that will never be accessed, user-defined functions that will not be used, parameters which have dimension zero etc.

mitzimorris:
need to capture type for data and parameters - generated quantities variables can be int as well as real.

> list of input data
you mean input data variables, correct?

rok-cesnovar:
> does this mean the input data variable declarations?

Yes, all data variables and their types excluding transformed data variables (i wrote input data to differentiate from transf. data)

> why?

so we can easily and also programtically check a models dependecies. This can work for cases like taking a model out of a huge database of models and also enable for interfaces to check if any of the included files has changes since last compile.

seantalts:
[...]

PR #813, opened 2021-02-11
The --info option produces a list of used functions and distributions

mandel:
## Release notes

As part of the issue #808, it was discuss to add the list of  "all named functions used". I understood that as the list for functions called from the standard library and the list of used distribution.

This PR adds to the JSON object produced by the `--info` option the two following fields:
- `stanlib_calls`: list of functions of the standard library called in the program
- `distributions`: list of distributions to the right of a `~` statement.

## Copyright and Licensing


mandel:
I do not understand why the changes in this PR cause the following error:
```
--- Translating Stan model to C++ code ---

bin/stanc  --o=examples/bernoulli/bernoulli.hpp examples/bernoulli/bernoulli.stan


clang++-6.0 -march=core2 -std=c++1y -D_REENTRANT -Wno-sign-compare -Wno-ignored-attributes      -I stan/lib/stan_math/lib/tbb_2019_U8/include   -O3 -I src -I stan/src -I lib/rapidjson_1.1.0/ -I lib/CLI11-1.9.1/ -I stan/lib/stan_math/ -I stan/lib/stan_math/lib/eigen_3.3.9 -I stan/lib/stan_math/lib/boost_1.75.0 -I stan/lib/stan_math/lib/sundials_5.6.1/include    -DBOOST_DISABLE_ASSERTS         -c -include-pch stan/src/stan/model/model_header.hpp.gch -x c++ -o examples/bernoulli/bernoulli.o examples/bernoulli/bernoulli.hpp

fatal error: file '/usr/include/linux/futex.h' has been modified since the precompiled header 'stan/src/stan/model/model_header.hpp.gch' was built
```

It looks unrelated to the `--info` option.

rok-cesnovar:
This is unrelated, maybe a one-off error on our test machine or a configuration issue on the machine. In both cases, dont worry about that, we will look into that.

Thanks for the PR! Looks nice.

Two comments/questions:
- I think we dont need to separate distributions? Or do you think there is a use case? Might be easier to just list `dirichlet_lpdf` in stanlib_calls? You  can call this for all tilde statements: 
https://github.com/stan-dev/stanc3/blob/master/src/middle/Utils.ml#L48

- We probably need to add a test for higher-order functions. Example models to use: https://github.com/stan-dev/stanc3/blob/master/test/integration/good/code-gen/reduce_sum_m1.stan or https://github.com/stan-dev/stanc3/blob/master/test/integration/good/ode/ode_rk45_test_model.stan

mandel:
> I think we dont need to separate distributions? Or do you think there is a use case?

I do not have a strong opinion but it is easier to add a post-processing phase that merge them than the converse. Moreover, in our Stan to (Num)Pyro compiler with keep the two forms separated and might have a benefit to have access to this information.

> Might be easier to just list dirichlet_lpdf in stanlib_calls? You can call this for all tilde statements: https://github.com/stan-dev/stanc3/blob/master/src/middle/Utils.ml#L48

It seems that this function does not add the `_lpdf` suffix when called on `"dirichlet"`.

mandel:
> We probably need to add a test for higher-order functions. Example models to use: https://github.com/stan-dev/stanc3/blob/master/test/integration/good/code-gen/reduce_sum_m1.stan or https://github.com/stan-dev/stanc3/blob/master/test/integration/good/ode/ode_rk45_test_model.stan

That is an interesting question. In these examples, the functions in argument of `reduce_sum` and `ode_rk45` are local functions so not reported anyway. But if the function in argument is a function from the Stan standard library it would currently not be present in the list `stanlib_calls`. Would you prefer a different behavior?

seantalts:
> I do not have a strong opinion but it is easier to add a post-processing phase that merge them than the converse. 

+1, I think it's neat to present additional information that could be helpful. I'm not sure if this is better or worse but we could also structure it something like
`functions_called : [{"dirichlet": {"distribution": true}, "log": {"distribution": false}]`

rok-cesnovar:
> I do not have a strong opinion but it is easier to add a post-processing phase that merge them than the converse. 

Very good point. Ignore my previous comment. Splitting these up is good.

>  the functions in argument of reduce_sum and ode_rk45 are local functions so not reported anyway.

Ah yes, I was overthinking this. Obviously these call user-defined functions so the current behaviour is fine. We were discussing functors for Stan lib functions in https://github.com/stan-dev/design-docs/pull/35 but we can address that if we actually do start supporting it.

test/integration/cli-args/info.stan
```diff
@@ -29,12 +29,13 @@ parameters {
...
   {
-    real ignored_tp = r;
+    real ignored_tp = log(r[0][0]);
   }
 }
 
 model {
-    real ignored_model = r;
+    real ignored_model = square(r[0][0]);
+    l ~ dirichlet(c);
```
rok-cesnovar, line 38:
```suggestion
    l ~ dirichlet(c);
    target += normal_lpdf(y| 0, 1);
    target += normal_lcdf(y| 0, 1);
    target += normal_lccdf(y| 0, 1);
```

Also requires adding `real y;` in the parameters.

The functions used in `target +=` are not listed.

rok-cesnovar (commented):
Thanks again for working on this!

Sorry I suggested the wrong function in my comment. This is the example (we only care of the StanLib ones):
```ocaml
let possible_names =
    List.map ~f:(( ^ ) distribution.name) Utils.distribution_suffices
    |> String.Set.of_list
in
if List.exists ~f:(fun (n, _) -> Set.mem possible_names n) ud_dists (* ud_dists is the 'dirichlet' in our case *)
then Fun_kind.UserDefined
else StanLib
```

I am nitpicking here because I think 
```
    l ~ dirichlet(c);
    target += normal_lpdf(y| 0, 1);
```
should both return functions either with or without the suffix. I prefer with suffix but would be fine without it as well as long as its consistent.

Also please add a test for Stan lib functions in user-defined functions:

```stan
functions {
  real foo(real a) {
    return sin(a);
  }
  real goo_lpdf(real a) {
    return a;
  }
}
```

If I then add 
```stan
    y ~ goo();
```
goo should not be listed.

mandel:
Thanks @rok-cesnovar. I have implemented most of your suggestions. But for now we still distinguish distributions with and without post-fix. 

If you do not like the current solution, I also prefer the solution where the post-fix is added since it is always easy to remove the post-fix later. 

But in this case, I think that it would be good to expose the function `dist_name_suffix` (https://github.com/stan-dev/stanc3/blob/master/src/frontend/Ast_to_Mir.ml#L501-L511) such that this code is not duplicated. I don't know if it should be added to the interface of `Ast_to_Mir` or moved to `Middle.Utils`?

rok-cesnovar:
Thanks!

I think we want to have the _lpdf/_lpmf suffixes. My reasoning is that these two models are equivalent, produce the exact same C++ code and should also produce the same output with --info.
```stan
parameters {
    real y;
}
model {
    y ~ std_normal();
}

parameters {
    real y;
}
model {
    target += std_normal_lpdf(y);
}
```

> But in this case, I think that it would be good to expose the function dist_name_suffix (https://github.com/stan-dev/stanc3/blob/master/src/frontend/Ast_to_Mir.ml#L501-L511) such that this code is not duplicated. I don't know if it should be added to the interface of Ast_to_Mir or moved to Middle.Utils?

I would say Middle.Utils but not sure. Do you have a preference @seantalts ? Utils has all the other distribution suffix utilities, idk.

test/integration/cli-args/info.stan
```diff
@@ -24,17 +33,23 @@ parameters {
...
   }
 }
 
 model {
-    real ignored_model = r;
+    real ignored_model = square(foo(square(r[0][0])));
+    l ~ dirichlet(c);
+    target += normal_lpdf(y| 0, 1);
+    target += normal_lcdf(y| 0, 1);
+    target += normal_lccdf(y| 0, 1);
```
rok-cesnovar, line 51:
```suggestion
    target += normal_lccdf(y| 0, 1);
    target += std_normal_lupdf(y)
```

This one should be listed as `std_normal_lpdf` (target += lupdf is the same as using the sampling statement).

For this rename you can just call Utils.stdlib_distribution_name on all distributions.

rok-cesnovar (commented):
My above comment got me thinking about another test case. Sorry I missed this earlier.

mandel:
In fact, it is not possible to move `dist_name_suffix` in `Middle.Utils` because it introduces a cyclic dependency with `Middle.Stan_math_signatures`. So should it be in `Middle.Stan_math_signatures`?

rok-cesnovar:
> So should it be in Middle.Stan_math_signatures?

I think so yes.

I would also rename `stanlib_calls` to `functions`. I think then its good to go.

rok-cesnovar (approved):
Thank you Louis!

mandel:
Your welcome!

PR #825, opened 2021-02-24
Minor testing refactor

rok-cesnovar:
- removes "Run all models end-to-end" as that was used to compare with Stanc2 which is no longer a part of Cmdstan.
- adds model compile stage
  A test that compiles all models int test/integration/good and https://github.com/stan-dev/example-models to executables.
  Given that this only takes ~20 minutes I think its a good price to pay to have this tested always.
 
## Copyright and Licensing


rok-cesnovar:
This is ready but needs https://github.com/stan-dev/example-models/pull/203 in order for tests to pass.

Jenkinsfile
```diff
@@ -96,74 +96,6 @@ pipeline {
...
-                            cd performance-tests-cmdstan
-                            git show HEAD --stat
-                            echo "example-models/regression_tests/mother.stan" > all.tests
-                            cat known_good_perf_all.tests >> all.tests
-                            echo "" >> all.tests
-                            cat shotgun_perf_all.tests >> all.tests
-                            cat all.tests
-                            echo "CXXFLAGS+=-march=core2" > cmdstan/make/local
-                            cd cmdstan; git show HEAD --stat; STANC2=true make -j4 build; cd ..
-                            CXX="${CXX}" ./compare-compilers.sh "--tests-file all.tests --num-samples=10" "\$(readlink -f ../bin/stanc)"
```
bbbales2, line 145: Is this some sort of compiler benchmark we just aren't doing anymore?

bbbales2 (commented):
Review!

Jenkinsfile
rok-cesnovar: Hm, yeah, it previously ran stanc2 tests that all failed since 2.25.

But comparing nightly vs PR compiler is something we should do, just not stanc2. Maybe we can add this as a separate stage.

rok-cesnovar:
Ready now. See Jenkins run for the new stage organization.

bbbales2 (approved):
Looks good

rok-cesnovar:
Thanks, I temporarily commented out compiling the example models because upon upgrading the example models submodule in performance-tests-cmdstan to the latest version I discovered that there are now a few models added recently that are broken (invalid syntax, undefined variables) and it makes no sense to delay this.

PR #830, opened 2021-02-28
promote types in array constructors

### LINKED ISSUE #827: Mixed type array fails to compile to exe
This model
```stan
transformed parameters {
  real td_arr33[3,3]
      = { {1 , 2 , 3 }, 
          {1 , 2., 3 }, 
          {1., 2., 3 } };
}
```
fails on the C++ level because its generated as

```
assign(td_arr33, nil_index_list(), std::vector<std::vector<double>>{
        std::vector<int>{1, 2, 3}, std::vector<double>{1, 2., 3},
        std::vector<double>{1., 2., 3}}, "assigning variable td_arr33");
```

but should be (see double instead of int).

```
assign(td_arr33, nil_index_list(), std::vector<std::vector<double>>{
        std::vector<double>{1, 2, 3}, std::vector<double>{1, 2., 3},
        std::vector<double>{1., 2., 3}}, "assigning variable td_arr33");
```

nhuurre:
Fix #827

The fix applies `stan::math::promote_scalar` to all array constructor arguments that are not the expected type. The math library's `array_builder` uses `stan::math::promote_elements`  but those seem identical expect that the template argument to `promote_scalar` is a scalar type while `promote_elements` takes an array type.

## Release notes

Fix incorrect codegen for mixed-type array expressions.

## Copyright and Licensing

Copyright holder: Niko Huurre


rok-cesnovar (approved):
Double checked that the test models compile to executables now. Thanks!

PR #832, opened 2021-03-01
Building ARM binaries in the pipeline

serban-nicusor-toptal:
## Release notes

Added build stage for ARM in Jenkinsfile.  
I have added a new EC2 Fleet for ARM with 0 instances so it will spin up only when we need it, this way we can now easily build ARM for other projects too.  

## Copyright and Licensing

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -281,7 +281,6 @@ pipeline {
 
                             opam update; opam install -y js_of_ocaml-compiler.3.4.0 js_of_ocaml-ppx.3.4.0 js_of_ocaml.3.4.0
                             opam update; bash -x scripts/install_build_deps.sh
-                            opam update; bash -x scripts/install_dev_deps.sh
                             opam update; bash -x scripts/install_js_deps.sh
                         """)
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -272,6 +272,34 @@ pipeline {
                     post {always { runShell("rm -rf ./*")}}
                 }
 
+                stage("Build & test a static Linux ARM binary") {
+                    agent { label "arm-ec2" }
+                    steps {
+                        script {
+                            checkoutRepository("sh")
+                        }
+
+                        runShell("""
+                            eval \$(opam env)
+                            opam update || true
+                            bash -x scripts/install_build_deps.sh
+                            dune subst
+                            dune build @install --profile static
+                        """)
+
+                        echo runShell("""
+                            eval \$(opam env)
+                            time dune runtest --verbose
+                        """)
+
+                        sh "mkdir -p bin && mv `find _build -name stanc.exe` bin/arm-stanc"
+                        sh "mv _build/default/src/stan2tfp/stan2tfp.exe bin/arm-stan2tfp"
+
+                        stash name:'arm-exe', includes:'bin/*'
+                    }
+                    post { always { runShell("rm -rf ./*") }}
+                }
+
                 // Cross compiling for windows on debian
                 stage("Build & test static Windows binary") {
                     agent {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -272,6 +272,34 @@ pipeline {
...
+                            dune build @install --profile static
+                        """)
+
+                        echo runShell("""
+                            eval \$(opam env)
+                            time dune runtest --verbose
+                        """)
+
+                        sh "mkdir -p bin && mv `find _build -name stanc.exe` bin/arm-stanc"
+                        sh "mv _build/default/src/stan2tfp/stan2tfp.exe bin/arm-stan2tfp"
```
rok-cesnovar, line 296:
```suggestion
                        sh "mkdir -p bin && mv `find _build -name stanc.exe` bin/linux-arm-stanc"
                        sh "mv _build/default/src/stan2tfp/stan2tfp.exe bin/linux-arm-stan2tfp"
```

Jenkinsfile
```diff
@@ -272,6 +272,34 @@ pipeline {
...
+
+                        stash name:'arm-exe', includes:'bin/*'
```
rok-cesnovar, line 298:
```suggestion
                        stash name:'linux-arm-exe', includes:'bin/*'
```

Jenkinsfile
```diff
@@ -313,6 +341,7 @@ pipeline {
                 unstash 'windows-exe'
                 unstash 'linux-exe'
                 unstash 'mac-exe'
+                unstash 'arm-exe'
```
rok-cesnovar, line 344:
```suggestion
                unstash 'linux-arm-exe'
```

rok-cesnovar (changes requested):
Looks great, maybe we rename this to linux-arm?

serban-nicusor-toptal:
Thanks @rok-cesnovar ! I've renamed that locally but will wait before pushing to see if the pipeline is working as expected.
I have added a new `EC2 Fleet` for ARM with `0` instances so it will spin up only when we need it, this way we can now easily build ARM for other projects too.
I've seen that the [binaries workflow](https://github.com/stan-dev/stanc3/blob/master/.github/workflows/binaries.yml) is used so I will make that build ARM too if that's okay with you.

rok-cesnovar:
> I've seen that the binaries workflow is used so I will make that build ARM too if that's okay with you.

Definitely!

serban-nicusor-toptal:
Argh looks like Github Actions [do not offer](https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners#about-github-hosted-runners) ARM based VMs as a service, only if we self-host an ARM based VM and use it for builds.
I can make the Jenkins ARM instance to self-register to GitHub  ( to avoid having one 24/7 ) when it boots but this is only working if we start a Jenkins job first, if it's the other way around GitHub won't have a machine available. Now it's a matter of usability I guess, if we only need it for releases it's enough in Jenkins, if we want to have arm in gh actions then we should go for it. I'm up for any that fits us best.
What do you think @rok-cesnovar ?

rok-cesnovar:
Ah yes, they do not. 

In that case its just Jenkins for ARM on releases.The point of the GHA builder is that anyone can clone the repo and build the binaries for themselves so those cant rely on our instances/Jenkins.

serban-nicusor-toptal:
Hey Rok, any idea why it's failing [here](https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3/detail/PR-832/13/pipeline/) ?

rok-cesnovar:
This seems to be caused by slight numerical differences (maybe due to different CPU architecture). Just comment out `eval $(opam env) time dune runtest --profile static --verbose` as it isnt really needed for building binaries. We can make this data gen resistant to this numerical differences at a later point.

rok-cesnovar (approved):

rok-cesnovar:
Oh right, that would be better. So you now have to spawn the arm instance manually so it builds? Or it just takes awhile to spawn?

serban-nicusor-toptal:
It should spawn by itself when needed if everything works as expected.

rok-cesnovar:
Cool, then lets apply the fix you suggest.

serban-nicusor-toptal:
For some reason it's taking way longer than expected:
image
I need to find a fix for this later, looks like it doesn't like our region :(

PR #835, opened 2021-03-02
Remove example models from test/integration 

rok-cesnovar:
Removes test/integration/example-models and adds compiling example-models to executables to Jenkins tests.

I think it makes no sense to duplicate these models here. It made a ton of sense during early development I imagine, but I dont think its required anymore. We can just check those models compile by cloning the repo in Jenkins tests. That way we always use the updated example models.

## Copyright and Licensing


bbbales2 (approved):
Good!

bbbales2:
(this removes the tests from stanc3 but the tests are still run through the cmdstan performance tests via the change in the Jenkins file)

PR #675, opened 2020-08-13
[Outdated] Add tuples to the language

rybern:
This PR adds support for tuple types and values to Stan. It builds on PRs #669 and #670.

This won't be ready to merge (it's a draft PR) until we can test against the backend. I'm currently generating code for std::tuple.

I also have a few more things to do like figure out JSON I/O.

The following compiles:
```
functions {
  // Nested array/tuple unsized types
  (int, array[] (int, real))
    f(int x, array[,] (array[] (int, int), array[,] int) x2)
  {}
}
transformed data {
  // nested array/tuple sized types and literals
  (int, real, array[2] (int, int)) a = (1, 2.5, {(1,2), (3,4)});

  // simple indexing
  int b = (1, 2.5).1;

  // sized/unsized types unify
  array[2,2] (array[2] (int, int), array[2,2] int) c;
  (int, array[3] (int, real)) d = f(2, c);

  // complex nesting indexing
  array[5] (array[10,5] (int, array[1,2,3] real), real) e;
  real f = e[5].1[10,5].2[1,2,3];
}
```

Issue #786, opened 2021-01-04
[FR] Use Vectorized Versions of reader member functions


SteveBronder:
Given code like this

```stan
data {
  int nt;
  int NS;
}

parameters {
  cholesky_factor_corr[2] L_Omega[nt];
  vector<lower=L_Omega[1,1,2]>[NS] z1; //  Se + Fp > 1
}
```

The compiler currently outputs code for the constraints like 

```cpp
      Eigen::Matrix<local_scalar_t__, -1, 1> z1;
      z1 = Eigen::Matrix<local_scalar_t__, -1, 1>(NS);
      stan::math::fill(z1, DUMMY_VAR__);
      current_statement__ = 2;
      z1 = in__.vector(NS);
      current_statement__ = 2;
      for (int sym1__ = 1; sym1__ <= NS; ++sym1__) {
        current_statement__ = 2;
        if (jacobian__) {
          current_statement__ = 2;
          assign(z1, cons_list(index_uni(sym1__), nil_index_list()),
            stan::math::lb_constrain(z1[(sym1__ - 1)],
              rvalue(L_Omega,
                cons_list(index_uni(1),
                  cons_list(index_uni(1),
                    cons_list(index_uni(2), nil_index_list()))), "L_Omega"),
              lp__), "assigning variable z1");
        } else {
          current_statement__ = 2;
          assign(z1, cons_list(index_uni(sym1__), nil_index_list()),
            stan::math::lb_constrain(z1[(sym1__ - 1)],
              rvalue(L_Omega,
                cons_list(index_uni(1),
                  cons_list(index_uni(1),
                    cons_list(index_uni(2), nil_index_list()))), "L_Omega")),
            "assigning variable z1");
        }}
```

There's two things here

1. this is a separate issue but `z1` is a parameter and so will always have data in the `reader` (`in__`) associated with it. Right now the current reading in of parameters takes 

```cpp
      Eigen::Matrix<local_scalar_t__, -1, 1> z1; // matrix creation
      z1 = Eigen::Matrix<local_scalar_t__, -1, 1>(NS); // Allocation of memory of size NS
      stan::math::fill(z1, DUMMY_VAR__); // writing to memory of size NS
      current_statement__ = 2;
      z1 = in__.vector(NS); // Read from `in__` of size NS and Writing to memory again of size NS 
```

This really should just be

```cpp
      // Read from `in__` of size NS 
[...]

Issue #736, opened 2020-10-28
Future Request: Using arena_t in the model class


SteveBronder:
Wanted to make a note before I forgot. Once we can pass around expressions in Stan math we'll also be able to pass around data and variables that we know are already allocated by adding `arena_matrix` types to the model class and it's functions

I'll use the model below that gives [this](https://gist.github.com/SteveBronder/e14ad2588fa433ba59dfb9b3810e661f) c++ code

```stan
data {
  int N;
  int M;
  vector[N] y;
  matrix[N, M] x;
}
parameters {
  vector[M] betas;
  real alpha;
  real<lower=0> sigma;
}
model {
  target += normal_id_glm_lpdf(y | x, alpha, betas, sigma);
}
```

So we have this new meta type in stan math now called `arena_t<EigenType>` that makes an eigen matrix (or vector) whose dynamic memory is fully allocated by our stack allocator. This is really nice because, since we know all the memory was created by our allocator, we can guarantee it will exist until we manually clean up that memory. So we can do a shallow copy (copy a pointer) vs a deep copy (allocate a whole new matrix for the reverse pass)

 for example for the function below

```cpp
template <typename VarMatrix, typename ArithMatrix>
inline auto some_func(const VarMatrix& a, const ArithMatrix& b) {
  // if arena_t is input, do shallow copies (copy pointer) 
  //   else make a deep copy
  arena_t<VarMat> arena_a(a);
  arena_t<ArithMatrix> arena_b(b);
  arena_t<Eigen::MatrixXd> ret(a.val() + b);
  // add a callback for computing the reverse pass
  reverse_pass_callback(
    [ret, arena_a, arena_b]() mutable {
      arena_a.adj() += ret.adj() + arena_b; 
  });
  return ret;
}
```

In the above, if the input `a` or `b` is an `arena_t<matrix>` then when we make a copy for the reverse pass we can do a shallow copy instead of a deep copy for either/or.

There's two ideas we can use this with

1. Inside of `log_prob` we can read directly from the reader for types defined in `parameters` and instead of assigning them to an eigen matrix we assign them to an `arena_t<matrix>`. For instance the below

```cpp
    stan::io::reader<local_scalar_t__> in__(params_r__, params_i__);
    local_scalar_t__ DUMMY_VAR__(std::numeric_limits<double>::quiet_NaN());
    (void)DUMMY_VAR__;  // suppress unused var warning

    try {
     // HERE
[...]

PR #837, opened 2021-03-03
Refactoring and improving constraint codegen

rybern:
There are some issues with how the constraint codegen works. See #783 and discussion under #675. My goals for this refactor are:
 * Simplify the code - it's quite challenging to understand current
 * Implement either one or both of @SteveBronder's suggestions in #783. I might start with just #1
 * Make the code flexible enough to be able to handle tuples

### DIFF
--- a/src/frontend/Ast_to_Mir.ml
+++ b/src/frontend/Ast_to_Mir.ml
@@ -464,8 +464,11 @@ let rec constrain_decl (st : Expr.Typed.t SizedType.t)
         | _ -> ([], decl_id, SizedType.to_unsized st)
       in
       unconstrained_decls
-      @ [ Stmt.Helpers.assign_indexed ut decl_id smeta var_constraint_funapp
-            decl_var ]
+      @ [ Stmt.Fixed.
+            { meta= smeta
+            ; pattern=
+                Assignment ((decl_id, ut, []), var_constraint_funapp decl_var)
+            } ]
   | _, (SVector d | SRowVector d), _ ->
       [Stmt.Helpers.mkfor d (fun e -> recur_block SReal e) decl_var smeta]
   | _, SMatrix (d1, d2), _ ->
--- a/src/frontend/Ast_to_Mir.ml
+++ b/src/frontend/Ast_to_Mir.ml
@@ -362,30 +362,97 @@ let remove_possibly_exn pst action loc =
         [%message
           "Error extracting sizedtype" ~action ~loc:(loc : Location_span.t)]
 
-let constrain_decl st dconstrain t decl_id decl_var smeta =
-  let mkstring = mkstring (Expr.Typed.loc_of decl_var) in
-  match Option.map ~f:(constrain_constraint_to_string t) dconstrain with
-  | None | Some "" -> []
-  | Some constraint_str ->
-      let dc = Option.value_exn dconstrain in
-      let fname = constrainaction_fname dc in
+(* let constrain_decl_ st dconstrain_opt t decl_id decl_var smeta =
+ *   match
+ *     Option.map dconstrain_opt ~f:(fun dc ->
+ *         (constrain_constraint_to_string t dc, dc) )
+ *   with
+ *   | None | Some ("", _) -> []
+ *   | Some (constraint_str, dconstrain) ->
+ *       (\* Make an MIR string literal at this location *\)
+ *       let mkstring str = mkstring (Expr.Typed.loc_of decl_var) str in
+ *       (\* Get appropriate function name for the constraint action, e.g. FnConstrain *\)
+ *       let fname = constrainaction_fname dconstrain in
+ *       let extra_args =
+ *         match dconstrain with
+ *         | Constrain -> extra_constraint_args st t
+ *         | _ -> []
+ *       in
+ *       (\* The argument expressions for the constraint action function *\)
+ *       let args var =
+ *         (var :: mkstring constraint_str :: extract_transform_args var t)
+ *         @ extra_args
+ *       in
+ *       let var_constraint_funapp var =
+ *         { Expr.Fixed.pattern= FunApp (CompilerInternal, fname, args var)
+ *         ; meta= var.meta }
+ *       in
+ *       let unconstrained_decls, decl_id, ut =
+ *         let ut = SizedType.to_unsized (param_size t st) in
+ *         match dconstrain with
+ *         | Unconstrain when t <> Identity ->
+ *             ( [ Stmt.Fixed.
+ *                   { pattern=
+ *                       Decl
+ *                         { decl_adtype= DataOnly
+ *                         ; decl_id= decl_id ^ "_free__"
+ *                         ; decl_type= Sized (param_size t st) }
+ *                   ; meta= smeta } ]
+ *             , decl_id ^ "_free__"
+ *             , ut )
+ *         | _ -> ([], decl_id, SizedType.to_unsized st)
+ *       in
+ *       unconstrained_decls
+ *       @ [ (constraint_forl t) st
+ *             (fun var ->
+ *               Stmt.Helpers.assign_indexed ut decl_id smeta
+ *                 var_constraint_funapp var )
+ *             decl_var smeta ] *)
+
+let rec constrain_decl (st : Expr.Typed.t SizedType.t)
+    (dconstrain_opt : constrainaction option)
+    (t : Expr.Typed.t Program.transformation) decl_id decl_var smeta =
+  let recur_block st var =
+    Stmt.Fixed.
+      { pattern=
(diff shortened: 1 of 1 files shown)


rybern:
Here's my understanding of how we currently handle constrains:

1. In Ast_to_Mir, for a constrained variable `v`, we generate:
   * `v_free__` declaration (with the same type as `v`) ([here](https://github.com/stan-dev/stanc3/blob/d26f65e891b4a13ca1b96486bfcd85d530a4583b/src/frontend/Ast_to_Mir.ml#L393))
   * nested for-loops to assign each element of either `v` or `v_free__` to an internal function ([here](https://github.com/stan-dev/stanc3/blob/d26f65e891b4a13ca1b96486bfcd85d530a4583b/src/frontend/Ast_to_Mir.ml#L365)), e.g.:
     `v = fn_constrain__(v[..], "ub", ..)`
2. In Transform_Mir, we generate:
   * read in parameters in genquant, log_prob, transform_inits, and prepare_data ([here](https://github.com/stan-dev/stanc3/blob/d26f65e891b4a13ca1b96486bfcd85d530a4583b/src/stan_math_backend/Transform_Mir.ml#L571)):
     * when a variable's constrained and unconstrained types are different, create a new variable `_in__` and read to that; otherwise read to `v` ([here](https://github.com/stan-dev/stanc3/blob/d26f65e891b4a13ca1b96486bfcd85d530a4583b/src/stan_math_backend/Transform_Mir.ml#L186))
   * Search through every statement and find instances where `fn_constrain__` is called on a variable that has a `_in__` variant, and change it to use the `_in__` version on the RHS ([here](https://github.com/stan-dev/stanc3/blob/d26f65e891b4a13ca1b96486bfcd85d530a4583b/src/stan_math_backend/Transform_Mir.ml#L394))
   * nested for-loops to call each element of either `v` or `v_free__` with an internal function ([here](https://github.com/stan-dev/stanc3/blob/d26f65e891b4a13ca1b96486bfcd85d530a4583b/src/stan_math_backend/Transform_Mir.ml#L346)):
     `fn_write_param__(v)`
3. In Expression_gen, we re-parse the fn_ calls via string matching and then translate them into actual functions ([here](https://github.com/stan-dev/stanc3/blob/d26f65e891b4a13ca1b96486bfcd85d530a4583b/src/stan_math_backend/Expression_gen.ml#L369)), for example: 
   `fn_constrain__(v, "ub", ..)` -> `FnConstrain` -> `constrain_ub(v, ...)`

rybern:
I think the main issue with the current setup is the separation of concerns, in two ways:
- The details of how we handle read and constraints with the Stan math backend are split across those three different parts of the compiler (especially between Ast_to_Mir and Transform_Mir), e.g. we manually create _free__ variables in Ast_to_Mir and then again in Transform_Mir
- Reading and constraints should actually be tightly coupled tasks, so they should be done together, rather than constraints being partially generated in Ast_to_Mir and reads generated in Transform_Mir

So it seems like the strategy should be either:
- Move the logic for applying constraints from ast_to_mir into transform_mir
- Move the logic for reading (and writing?) parameters from transform_mir to ast_to_mir. From what I've seen of transform_mir, this would probably be harder. Also, it seems more appropriate for that to be in the "backend" code since it's Stan math-specific

@seantalts it would be great to get your thoughts

seantalts:
> Reading and constraints should actually be tightly coupled tasks, so they should be done together, rather than constraints being partially generated in Ast_to_Mir and reads generated in Transform_Mir

Can you say more about this? I think the MIR should represent code that is portable across a variety of backends. In the ones I can think of (other than Stan Math), reading and constraining are different tasks.

rybern:
So one relevant issue is #786 

If I understand correctly Steve's new deserializer prefers constrains and reads to happen in one call: https://github.com/stan-dev/stan/pull/3013

@nhuurre gave a good example where we'd like reading and constraining int the same for-loop, here: https://github.com/stan-dev/stanc3/pull/675#issuecomment-789806615

They also seem coupled because reads need to know about \_free__ variables

rybern:
I agree that ideally the MIR should be as platform independent as possible prior to Transform_mir, so I'm leaning toward handing this in Transform_mir

seantalts:
What does it look like moving it to Transform_mir? 

A related opinion of mine is that I am fully in favor of vectorizing everything possible in the MIR now - maybe that makes all of the mentioned issues happy? I was against it originally thinking it wasn't the level we wanted for backends, but I discovered that ~all backends have vectorized ~everything :D 

I currently lean against adding a MIR node that is a combined Read/Constrain or Read/Unconstrain, but perhaps with vectorized versions of Read and Constrain we can get everything we want pretty easily?

SteveBronder:
> If I understand correctly @SteveBronder's new deserializer prefers constrains and reads to happen in one call:

Yep!

Are there issues with the free variables used in `transform_init()`? On the backend we could write a context that has similar functions to the deserializer class for call the equivalent free functions. So the call would look something like 

```cpp
double x = context.get_free_lb("x");
```

> A related opinion of mine is that I am fully in favor of vectorizing everything possible in the MIR now - maybe that makes all of the mentioned issues happy? I was against it originally thinking it wasn't the level we wanted for backends, but I discovered that ~all backends have vectorized ~everything :D

Yeah in general I think it's a good idea for the compiler to dish-out to the backends as much as possible. I think that would let the compiler work on higher level optimizations instead of things the C++ compiler / stan math / eigen already does. V unrelated but it would be nice if the stan compiler could see a loop like

```stan
vector[N] x;
vector[N] y = // ...
for (i in 1:N) {
 x[i] = exp(y[i]);
}
```

and know there's a vectorized version of `exp()` so it can just call `x = exp(y)`. I'd also guess the one liner would be easier for more advanced compiler optimizations like @VMatthijs 's reverse mode compiler stuff

rybern:
> What does it look like moving it to Transform_mir?

I probably won't have time to fully think this out until Monday-ish, but:
What if we don't add FnConstrains to the MIR prior to Transform_mir? I _think_ all of the necessary information is already added in `output_vars`, because we're using it in Transform_mir to re-derive the \_free__ variables. We're already adding e.g. the FnReads and FnWrites to the appropriate blocks in Transform_mir, so it wouldn't be a huge refactor.

rybern:
@SteveBronder Could you give example code of reading e.g. an `array[5,5] simplex` using deserializer? Do you need to loop, or is that what the vecsize argument is for?

SteveBronder:
> @SteveBronder Could you give example code of reading e.g. an array[5,5] simplex using deserializer? 

Yes if you check out the comment [here](https://github.com/stan-dev/stanc3/issues/736#issuecomment-792354200) where I have a stan program with a few different types in using deserializer


For instance [here](https://gist.github.com/SteveBronder/4fceb869bccff2926674bdecbdf5ce0a#file-proposed_model-hpp-L510) is for an array of correlation matrices.

> Do you need to loop, or is that what the vecsize argument is for?

vecsize!

I don't have simplex there but for something like 

```stan
parameters {
  simplex[N] xx[K];
}
```

That would become

```cpp
// just making alias here so line is easier to read
using stdvec_vec_t = std::vector<Eigen::Matrix<local_scalar_t__, -1, 1>> ;
stdvec_vec_t xx = in__.read_simplex<stdvec_vec_t, jacobian__>(lp__, K, N);
```

All of the constraints have the signature

```cpp
read_{constrain_type}<{return_type}, {bool_jacobian}>({constrain_params}, lp__, {top-to-bottom dims});
```

And by top-to-bottom dims there I mean the outer dimensions are passed as the first arg, then inner args recursively. So an  

```stan
parameters {
  simplex[N] xx[K, J];
}
```

becomes 

```cpp
// just making alias here so line is easier to read
using stdvec_stdvec_vec_t = std::vector<std::vector<Eigen::Matrix<local_scalar_t__, -1, 1>>>;
stdvec_stdvec_vec_t xx = in__.read_simplex<stdvec_stdvec_vec_t, jacobian__>(lp__, J, K, N);
```

On the backend it just recursively calls the std::vector specialization of `read_simplex` until it needs to call the specialization for the eigen vector. Also the function always takes in `lp__` but it only ever does stuff with it if `jacobian__` is `true`

bbbales2:
> stdvec_stdvec_vec_t xx = in__.read_simplex<stdvec_stdvec_vec_t, jacobian__>(lp__, J, K, N);

Should be K, J, N? I think with `simplex[N] xx[K, J];` `k` is the first index (the outer-most `std::vector`)

SteveBronder:
Oh shoot yes, sorry that should be flipped on the dims so in to out is `(lp__, K, J, N);`

rybern:
I think I'm going to leave the refactors for another day/PR, because the new deserializer interface is even simpler than I planned for. New PR at #856

PR #842, opened 2021-03-08
Make locations_array constexpr and move curr_statement__ to function scope

SteveBronder:
## Summary

This does three things

1. Makes `locations_array__` into a `static constexpr std::array<const char*, {size_of_error_list}>`
2. Moves `current_statement__` to be created inside of local function scopes.
3. Adds spaces to the templates for the `_impl` functions so they are a little easier to read.

(1) is just a nice optimization. That array is only read from and we can deduce the size in the compiler so we just make it a `static constexpr` array of `const char*` so the compiler knows it's size and `static` so that it knows that global can only be accessed inside this file.

(2) is related to [#897](https://github.com/stan-dev/cmdstan/pull/987) which let's a Stan program run with multiple chains in one program. Having `current_statement__` as a global variable means that when we call log prob in that etc. in that PR we hit a bunch of weird race conditions since each thread can be at a different point in each function (or running seperate functions that update `current_statement__`. This also produces [false sharing](https://en.wikipedia.org/wiki/False_sharing) in loops. Like say a user wrote

```stan
for (i in 1:N) {
  lp[i] = log_mix(x, xx[i], yy[i]);
  lp2[i] = log_mix(x2, xx2[i], yy2[i]);
}
```

In the c++ that becomes
```cpp
// pseudocode
for (int sym1__ = 1; sym1__ <= N;  ++sym1__) {
  current_statement__ = 5;
  lp[i] = log_mix(x, xx[i], yy[i]);
  current_statement__ = 6;
  lp2[i] = log_mix(x2, xx2[i], yy2[i]);
}
```

Since we are writing to a global if `current_statement__`, if the scalars used in the function and `current_statement__` are used on the same cache line then the cpu marks that cache line as invalid and forces the entire cache line to be re-read back in from memory. By making it defined locally each thread has their own version of `curr_statement__` and so there is no false sharing

So we just make current statement locally above the try block for each function that uses it. ~Thinking now I need to see how this will work with user defined functions.~ I think I sorted that out by just making it locally in the udf

(3) is just a nice thing to be able to read the generated C++

## Release notes

Make locations_array constexpr and move curr_statement__ to function scope

## Copyright and Licensing

Copyright Holder: Steve Bronder

### DIFF
--- a/src/stan_math_backend/Stan_math_code_gen.ml
+++ b/src/stan_math_backend/Stan_math_code_gen.ml
@@ -496,10 +496,10 @@ let pp_method_b ppf rt name params intro ?(outro = nop) ?(cv_attr = ["const"])
 let pp_write_array ppf {Program.prog_name; generate_quantities; _} =
   pf ppf
     "template <typename RNG, typename VecR, typename VecI, typename VecVar, @ \
-     \x20 stan::require_vector_like_vt<std::is_floating_point, VecR>* = \
-     nullptr, @ \x20 stan::require_vector_like_vt<std::is_integral, VecI>* = \
-     nullptr, @ \x20 stan::require_std_vector_vt<std::is_floating_point, \
-     VecVar>* = nullptr> @ " ;
+     stan::require_vector_like_vt<std::is_floating_point, VecR>* = nullptr, @ \
+     stan::require_vector_like_vt<std::is_integral, VecI>* = nullptr, @ \
+     stan::require_std_vector_vt<std::is_floating_point, VecVar>* = nullptr> \
+     @ " ;
   let params =
     [ "RNG& base_rng__"; "VecR& params_r__"; "VecI& params_i__"
     ; "VecVar& vars__"; "const bool emit_transformed_parameters__ = true"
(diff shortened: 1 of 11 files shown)


src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -493,10 +495,11 @@ let pp_method_b ppf rt name params intro ?(outro = nop) ?(cv_attr = ["const"])
...
   pf ppf
-    "template <typename RNG, typename VecR, typename VecI, typename VecVar, \
-     stan::require_vector_like_vt<std::is_floating_point, VecR>* = nullptr, \
-     stan::require_vector_like_vt<std::is_integral, VecI>* = nullptr, \
-     stan::require_std_vector_vt<std::is_floating_point, VecVar>* = nullptr>" ;
+    "template <typename RNG, typename VecR, typename VecI, typename VecVar, @ \
+     \x20 stan::require_vector_like_vt<std::is_floating_point, VecR>* = \
+     nullptr, @ \x20 stan::require_vector_like_vt<std::is_integral, VecI>* = \
+     nullptr, @ \x20 stan::require_std_vector_vt<std::is_floating_point, \
+     VecVar>* = nullptr> @ " ;
```
rok-cesnovar, line 502: I am not completely sold on this `@ \x20` thing. What you are trying to do is just limit the number of characters in a line right? If so, we can do this with less intervention.

Also dont really like the space before the comma in (see jacobian)
```
template <bool propto__, bool jacobian__ , typename VecR, typename V
```

rok-cesnovar (changes requested):
1 and 2 look great and make sense to me. Are you trying to limit the number of characters in a line for 3? Or are you interested in actual spaces?

src/stan_math_backend/Stan_math_code_gen.ml
SteveBronder: I just thought it was nice looking to have an indentation there but I removed them

rok-cesnovar: Oh indentation, I get it. Not an expert on this printing stuff, but I think there is a way to do that without the use of hex. Will give it a shot, else just merge if that is ok with you?

SteveBronder: Yeah it's fine as is

rok-cesnovar (approved):

rok-cesnovar:
Indeed. That was fast :)

Issue #779, opened 2020-12-27
Upgrade dune to version >2.1


seantalts:
It's been a while since we upgraded and there are some good changes. I'm particularly excited about using [this](https://discuss.ocaml.org/t/announcing-dune-deps-produces-a-project-centric-dependency-graph/5451/2) to graph project dependencies, as it's been a bit tricky to piece it all together myself.

Issue #710, opened 2020-10-04
upgrade to OCaml compiler 4.08+


seantalts:
I'd love to start using [the new `let+` etc operators](https://jobjo.github.io/2019/04/24/ocaml-has-some-new-shiny-syntax.html) to streamline monadic code :) 

I would expect this to not require much in the way of changes to our code, though who knows. Hopefully we can also make some kind of upgrade / repin script that will set the version numbers of all of our important dependencies to the correct new versions.

PR #627, opened 2020-07-10
Update ocamlformat, change style to match Jane Street style guide

rybern:
We've been pinned to ocamlformat 0.8, which is a couple years old now. I suggest we update to 0.13.

When I looked at the diff for the 0.13, I actually didn't like it much. When I switched the style to janestreet I liked it much much better. I think it makes sense for us to follow Jane Street anyway. Have a look at the diff and see what you think.

Also, could someone that uses Opam please check that ocamlformat 0.13 is available?

PR #843, opened 2021-03-08
Compatibility with OCaml 4.12.0

mandel:
## Release notes

Update the code of the compiler to make it compatible with:
- OCaml 4.12.0
- dune 2.8.2
- core_kernel v0.14.1 
- menhir 20201216 
- ppx_deriving 5.2.1
- fmt 0.8.9
- yojson 1.7.0

These changes are compatible with the current requirements.

Upgrading to `ocamlformat 0.17.0` changes the formatting of the code and thus is not retro-compatible with `ocamlformat 0.8`.

This PR is useful for issues  #779 and #710. 

## Copyright and Licensing


rok-cesnovar:
Thanks!!

Also helpful for https://github.com/stan-dev/stanc3/pull/627

Will push the relevant CI changes if others are good with this @seantalts @rybern @nhuurre

rok-cesnovar:
We will very likely have to remain at 4.11 until ocaml-cross-windows is updated (that is required for building Windows binaries).

mandel:
I still have not identified the problem, but it seems that it is in `Optimize.lazy_code_motion`.

SteveBronder:
I had some time to look at this today and I think the issue is that a lot of the stuff in the optimizer uses if's and && and || and those are causing some weird bad things to happen in the ocaml

rok-cesnovar:
I am not sure we will be able to make it work with 4.12.0 due to https://github.com/ocaml-cross/opam-cross-windows that we use to make windows executables. They state that the latest supported version is 4.11.1

SteveBronder:
I have a branch of this that sort of works for 4.11.1 @mandel if you can take a look. 

What I can't figure out is why we need to import `=` from stdlib now? Is there something we need to change with how we import stuff? Feels counterintuitive to have to import the equality operators

https://github.com/stan-dev/stanc3/compare/master...SteveBronder:update/4.11.1?expand=1

mandel:
Sorry for the late answer!

We have to import the comparison operators now due to a change in the Core library. They are masking the polymorphic equality  of the standard library (https://discuss.ocaml.org/t/removing-polymorphic-compare-from-core/2994). So to recover it, we have to re-import it after importing Core.

@SteveBronder Your branch looks good to me! The only thing that I do not understand is why you have to do the `open Stdlib` in `Ast.ml`. How do you want to proceed? Do you want to commit your changes in this branch or do a new PR?

mandel:
@SteveBronder Sure you can merge here!

mandel:
> icic okay so your code makes sense now. I'm looking at this thread and it just seems odd that ocaml doesn't allow you to have your own overloaded comparison operators defined for a given type, but whatever. imo I'd prefer we just write equals functions for each of our types instead of importing the polymorphic one from stdlib. But if that's a lot of hassle then I'm fine with just have the polymorphic import.

I agree that it can be nice to use type specific equals but it is a big amount of work. I though that it can be done in multiple stages. First have the code working with the polymorphic equal (like the 4.07 version). Then switch the equals.

mandel:
I renamed the branch https://github.com/mandel/stanc3/tree/update-4.12

mandel:
The development of this is now in https://github.com/stan-dev/stanc3/pull/885

Issue #731, opened 2020-10-23
cmdstan eigen error vs rstan informative error


spinkney:
Having a udf with a 0 index compiles fine in both rstan and cmdstan. At run time they produce different errors and I believe cmdstanr should produce a better error.

## cmdstanr error
```
> mod_test$sample(data = list(N = 100, y = rnorm(100)),
+                chains = 2,
+                parallel_chains = 2
+ )
Running MCMC with 2 parallel chains...

Chain 1 test: stan/lib/stan_math/lib/eigen_3.3.7/Eigen/src/Core/DenseCoeffsBase.h:162: Eigen::DenseCoeffsBase<Derived, 0>::CoeffReturnType Eigen::DenseCoeffsBase<Derived, 0>::operator[](Eigen::Index) const [with Derived = Eigen::Matrix<double, -1, 1>; Eigen::DenseCoeffsBase<Derived, 0>::CoeffReturnType = const double&; Eigen::Index = long int]: Assertion `index >= 0 && index < size()' failed.
```
## rstan error
```
> out_test <- sampling(mod_test_rstan,
+                 data = list(N = 100, y = rnorm(100)),
+                 chains = 2,
+                 iter = 1000)

SAMPLING FOR MODEL 'test' NOW (CHAIN 1).
Chain 1: Unrecoverable error evaluating the log probability at the initial value.
Chain 1: Exception: Exception: []: accessing element out of range. index 0 out of range; expecting index to be between 1 and 100; index position = 1x  (in 'model5a82748ab0ae_test' at line 7)
  (in 'model5a82748ab0ae_test' at line 19)
```

```
functions {
  real does_nothing (vector x){
    return x[0];
  }
}
data {
  int<lower=0> N;
  vector[N] y;
}
parameters {
  real mu;
  real<lower=0> sigma;
}
transformed parameters {
  real z = does_nothing(y);
}
model {
  y ~ normal(mu, sigma);
}
```

#### versions
```
cmdstan_version()
[1] "2.24.0"

stan_version()
[1] "2.21.0"
```

rok-cesnovar:
There was one attempt at fixing this: https://github.com/stan-dev/stanc3/pull/521

But that proved to cause a performance regression. We should give this another go once the expression rework is done in /stan.

rok-cesnovar:
The issue covering this is also still open: https://github.com/stan-dev/stanc3/issues/489

We can merge the two issues I guess.

Issue #847, opened 2021-03-09
Compiler allows out-of-bounds indexing, Stan programs seg-faulting


mitzimorris:
Stan programs are segfaulting if user gets the indexing wrong.
Here's an example:

```
data {
  int<lower=0> N;               // number of items
  int<lower=0> M;               // number of predictors
  int<lower=0,upper=1> y[N];           // outcomes
  row_vector[M] x[N];      // predictors
}
transformed data {
  print(x[200,200]);
}
```

Given input where N=128 and M=2, print statement segfaults.  The generated code should be checking bounds sized N,M before allowing access.

2 recent reports coming from CmdStanPy users confused by return code -11 - 
https://discourse.mc-stan.org/t/cmdstanpy-error-code-11/21051/3
https://github.com/stan-dev/docs/issues/228#issuecomment-792357073

bbbales2:
I think this is the same bug as https://github.com/stan-dev/stanc3/issues/776.

mitzimorris:
agreed, same bug as https://github.com/stan-dev/stanc3/issues/776

Issue #776, opened 2020-12-24
Indexing out-of-bounds not always handled properly


syclik:
```
$ ./bin/stanc --version
stanc3 9df5448 (Unix)
```

When indexing a Stan array or vector with an index that's out-of-bounds, the Stan program terminates with different error conditions:

* I would expect the program to catch the error, print a message, and terminate. This happens under some conditions.
* Eigen assertion is triggered.
* segmentation fault
* no error -- looks like it's reading off the end of memory.

### Examples

This triggers an Eigen assertion:
```stan
parameters {
  real theta;
}
transformed parameters {
  vector[0] x;
}
model {
  theta ~ normal(x[1], 1);
}
```

This seg faults (may not consistently seg fault):
```stan
transformed data {
  vector[0] x[2];
}
parameters {
  real theta;
}
model {
  theta ~ normal(x[0][1], 1);
}
```

This looks like it's reading off the end of memory:
```stan
transformed data {
  vector[0] x[2];
}
parameters {
  real theta;
}
model {
  theta ~ normal(x[3], 1);
}
```

I believe @bbbales2 has looked at the generated C++ and confirmed the issue.

syclik:
edit: @SteveBronder has looked at the generated code.

bbbales2:
I think this is the same as the issue here: https://github.com/stan-dev/stanc3/issues/489, that got fixed here: https://github.com/stan-dev/stanc3/pull/521, which got reverted here cause it caused a performance regression: https://github.com/stan-dev/stanc3/pull/656

Given that it causes segfaults, I kinda like the check back even if it causes noticeable slowdowns this time.

seantalts:
Ah, interesting! Is it the case that this only happens on index literals like in the example here and in the first linked issue? Those we could check at compile time with no runtime performance cost. That would be neat.

rok-cesnovar:
> Is it the case that this only happens on index literals (eg y[0], not y[i]) like in the example here and in the first linked issue?

Unfortunately not.

```stan
data {
    int N;
}
transformed data {
  vector[N] x[2];
}
parameters {
  real theta;
}
model {
  theta ~ normal(x[N][1], 1);
}
```
suffers from the same issue.

seantalts:
Only if N = 0?

rok-cesnovar:
Indeed.

bbbales2:
I ran into someone having this problem today. They were getting the error:

```
[...]

PR #849, opened 2021-03-10
Adding back in rvalue checks on 1d indexes

### LINKED ISSUE #489: Eigen out of bounds access error with simple model

bbbales2:
## Release notes

This is a revert of #656 to put the fix in #521 back in to fix #489, #776, #847.

The revert in #656 was cause there was a substantial performance hit (like 30%) of including this fix and we were in code freeze and the simplest thing was revert.

I think it would be good to put this fix back in even without an immediate fix to the performance thing. We can re-evaluate the performance thing (lots of stuff has changed between error checks and indexing) and fix that later, but at least then the segfaults are out of the way.

## Copyright and Licensing

### DIFF
--- a/src/stan_math_backend/Expression_gen.ml
+++ b/src/stan_math_backend/Expression_gen.ml
@@ -56,6 +56,11 @@ let minus_one e =
 
 let is_single_index = function Index.Single _ -> true | _ -> false
 
+let dont_need_range_check = function
+  | Index.Single Expr.Fixed.({pattern= Var id; _}) ->
+      not (Utils.is_user_ident id)
+  | _ -> false
+
 let promote_adtype =
   List.fold
     ~f:(fun accum expr ->
@@ -499,7 +504,7 @@ and pp_expr ppf Expr.Fixed.({pattern; meta} as e) =
       when Some Internal_fun.FnReadData = Internal_fun.of_string_opt f ->
         pp_indexed_simple ppf (strf "%a" pp_expr e, idx)
     | _
-      when List.for_all ~f:is_single_index idx
+      when List.for_all ~f:dont_need_range_check idx
            && not (UnsizedType.is_indexing_matrix (Expr.Typed.type_of e, idx))
       ->
         pp_indexed_simple ppf (strf "%a" pp_expr e, idx)
--- a/test/integration/good/code-gen/cl.expected
+++ b/test/integration/good/code-gen/cl.expected
@@ -1258,11 +1258,15 @@ class optimize_glm_model final : public model_base_crtp<optimize_glm_model> {
         current_statement__ = 174;
         lp_accum__.add(
           normal_id_glm_lpdf<false>(y_v_d_opencl__,
-            to_matrix_cl(X_d_a[(1 - 1)]), alpha, to_matrix_cl(beta), sigma));
+            to_matrix_cl(
+              rvalue(X_d_a, cons_list(index_uni(1), nil_index_list()),
+                "X_d_a")), alpha, to_matrix_cl(beta), sigma));
         current_statement__ = 175;
         lp_accum__.add(
           normal_id_glm_lpdf<propto__>(y_v_d_opencl__,
-            to_matrix_cl(X_d_a[(1 - 1)]), alpha, to_matrix_cl(beta), sigma));
+            to_matrix_cl(
+              rvalue(X_d_a, cons_list(index_uni(1), nil_index_list()),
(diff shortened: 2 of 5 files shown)


bbbales2:
Looks like @rok-cesnovar has a reproducible example here: https://github.com/stan-dev/stanc3/pull/656#issuecomment-663522849

Let me start a couple cmdstan tarball builds and I'll compare:
1. Stanc3 develop
2. Stanc3 here
3. 2.24rc-1
4. 2.24

bbbales2:
Alright the timings are (for N = 10000 in the model above and using the same seed and verifying the last samples are the same):

```
2.24.0:     4.4s
2.24.0-rc1: 5.0s
```

And then:

```
develop: 4.8s
this pull: 5.1s
```

So this change looks like less of a performance hit now, though it looks like everything overall is slower but that's a different issue.

The binaries I'm using are:
[2.24.0](https://github.com/stan-dev/cmdstan/releases/tag/v2.24.0)
[2.24.0rc-1](https://github.com/stan-dev/cmdstan/releases/tag/v2.24.0-rc1) with the stanc binaries [here](https://github.com/stan-dev/stanc3/actions/runs/639435369)
[develop](https://github.com/stan-dev/cmdstan/actions/runs/639341882), and
[this pull](https://github.com/stan-dev/cmdstan/actions/runs/639343362)

SteveBronder:
I'm 50/50 on this. Why don't we wait for @rok-cesnovar 's cleanup of rvalue and assign before doing this?

@wds15 I think we should just use the `-DNDEBUG` macro to turn off these checks at the stan C++ level. That macro is used to turn off indexing checks in the std so I think it makes sense.

bbbales2:
Does the rvalue cleanup fix this? I was looking at the generated code and it doesn't seem like it to me (for instance here: https://github.com/stan-dev/stanc3/blob/9b5e9d4739d7a6106a337583761a6f0bf601f6bf/test/integration/good/code-gen/cl.expected#L340).

If it does we can wait, but otherwise better to just get this issue out of the way and then the focus can be on speeding things up and there's only one thing to worry about (rather than having performance and a segfault hanging over out heads -- just performance, which is sort of business as usual).

If there's a change needed in the rvalue refactor pull, it will presumably be like one or two lines (there's only 5 lines of new code here and the rest is changed generated code).

rok-cesnovar:
> Does the rvalue cleanup fix this? 

No, just cleans up thing a bit. I think the order in which we merge does not make a difference.

I agree that the baseline should be no-segfault, then worry about performance. This same check was present in stanc2. I propose merging this and working on the no-debug stuff. 

@SteveBronder this would have to be a stancflag or do you propose avoiding the range check based on C++ compiler flags? I am fine with the C++ flag since its already "a thing" in other cases.

I was debating this with @t4c1 at the time of the revert and I think the conclusion was there is no performant way of doing this for std::vector.

SteveBronder:
Cool cool we can add the NDEBUG thing

rok-cesnovar:
Also fixes #731

rok-cesnovar (approved):

ryan-richt:
Just a user comment, the no-debug flag to recover the performance is critical to our use cases as Stan users. We have profiled a ~15% slowdown due to these bounds checks in the older version (in models that otherwise are fully performance optimized). Having access to the "production" flag to turns these checks off is very important to us.

wds15:
Maybe open an issue for that?

ryan-richt:
@wds15 I have opened said ticket here: https://github.com/stan-dev/stanc3/issues/853 please do edit for correctness

Issue #523, opened 2020-04-21
Assignment to vector of different size currently allowed


bbbales2:
Here is a test model:

```stan
parameters {
  real a;
}

model {
  vector[3] l1;
  vector[2] l2;
  l1 = rep_vector(1, 2);
  l2 = rep_vector(1, 3);
  print("Length 3: ", l1);
  print("Length 2: ", l2);
  
  a ~ normal(0, 1);
}
```

It produces output:
```
Chain 1 Length 3: [1,1] 
Chain 1 Length 2: [1,1,1]
```

It should print errors like (this is from rstan 2.19.3):
```
SAMPLING FOR MODEL 'm3' NOW (CHAIN 1).
Chain 1: Unrecoverable error evaluating the log probability at the initial value.
Chain 1: Exception: assign: Rows of left-hand-side (3) and rows of right-hand-side (2) must match in size  (in 'model57e03fe8a5ab_m3' at line 8)
```

I made this in issue in stanc3 because it looks like the tests for this are in Math: https://github.com/stan-dev/math/blob/develop/test/unit/math/prim/fun/assign_test.cpp#L54

Edit: This came from: https://discourse.mc-stan.org/t/speeding-up-and-reducing-divergences-for-multivariate-hiearchical-stochastic-factor-model/14366

bbbales2:
Pinging the compiler crew, @rok-cesnovar @rybern @nhuurre. Could one of you investigate if this is a stanc3 issue? I don't think it's a math issue. It might be a stan issue.

rok-cesnovar:
Well it's kind of both, but the issue is in Stan.

cmdstan develop with stanc2 outputs something similar:

stanc2 assignment:
```cpp
current_statement_begin__ = 8;
stan::math::assign(l1, rep_vector(1, 2));
current_statement_begin__ = 9;
stan::math::assign(l2, rep_vector(1, 3));
```
As you can see this uses the stan math assign from https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/assign.hpp

In stanc3 its:

```cpp
current_statement__ = 4;
assign(l1, nil_index_list(), rep_vector(1, 2),
          "assigning variable l1");
current_statement__ = 5;
assign(l2, nil_index_list(), rep_vector(1, 3),
          "assigning variable l2");
```
Which in itself is fine. This assign is from the Stan repo: https://github.com/stan-dev/stan/blob/develop/src/stan/model/indexing/lvalue.hpp

[...]

PR #856, opened 2021-03-12
Call new deserializer backend for constrained reads

rybern:
Replacing #837

This is meant to be a minimal PR to call the new backend for constrained reads (merged backend PR here: https://github.com/stan-dev/stan/pull/3013)

For example, what was
```
      sigma = in__.scalar();
      current_statement__ = 4;
      if (jacobian__) {
        current_statement__ = 4;
        sigma = stan::math::lb_constrain(sigma, 0, lp__);
      } else {
        current_statement__ = 4;
        sigma = stan::math::lb_constrain(sigma, 0);
      }
```
is now
```
      sigma = in__.read_lb<local_scalar_t__, jacobian__>(0, lp__);
```

The strategy is just to:
 * Stop generating FnConstrain in Ast_to_Mir
 * Pack the information necessary for the new constrained reads into FnRead, which is built in Transform_Mir
 * Change the codegen for FnRead to match the new interface

Work still to do:
 * Remove FnConstrain from Internal_fun
 * Make sure stanc3 test diffs are okay (I'll note a couple odd ones in line comments)
 * End-to-end compile tests


rok-cesnovar:
> End-to-end compile tests

Can you explain a bit more what you mean by this? Do you mean test that models compile? If so we have that test running in our Jenkins pipeline. The top one in the image below compiles all integration/good and all models in the stan-dev/example-models repo.

image

seantalts:
I think the MIR should retain the separation of constrain and read
operations - if we need to merge them for the Stan Math backend, we can do
that in that back-end. Other backends don't combine those operations.

SteveBronder:
Looking over the code now, one note, you need to change the line [here](https://github.com/stan-dev/stanc3/blob/9f6836b1d2fcaa2ed1eb5b3857faa7c40696ce7a/src/stan_math_backend/Stan_math_code_gen.ml#L513)

to
```
      ; "stan::io::deserializer<local_scalar_t__> in__(params_r__, params_i__);"
```

test/integration/good/code-gen/cl.expected
```diff
@@ -579,60 +579,64 @@ class optimize_glm_model final : public model_base_crtp<optimize_glm_model> {
       stan::math::fill(alpha_v, DUMMY_VAR__);
       
       current_statement__ = 1;
-      alpha_v = in__.vector(k);
+      assign(alpha_v, nil_index_list(),
+        in__.read<Eigen::Matrix<local_scalar_t__, -1, 1>, jacobian__>(lp__,
+          k), "assigning variable alpha_v");
```
SteveBronder, line 584: So I can change this if it's not generic enough but this line

```
in__.read<Eigen::Matrix<local_scalar_t__, -1, 1>, jacobian__>(lp__, k);
```

Should just be

```
in__.read<Eigen::Matrix<local_scalar_t__, -1, 1>>(k);
```

As jacobian calcs are only needed when calling constrains I made the read methods not take in the jacobian template value or lp.


Also we can shorten all of this up to remove the fill etc. for parameters and just have

```cpp
      Eigen::Matrix<local_scalar_t__, -1, 1> alpha_v = in__.read<Eigen::Matrix<local_scalar_t__, -1, 1>>(k);
```

rybern: Ah I missed that thanks, shouldn't be hard.

Yeah, we should get rid of the fills, but I think that'll be a separate solution

SteveBronder: Yeah that could be a different matter. I think can check for `Decl` and do something with that

rybern:
Hi @seantalts - couple of points, IMO:

The transformed MIR is already backend-specific, so it's okay if we tailor it to the Stan Math backend. Other backends can read the transformations from out_vars and generate constraints in their own way.

Also, I think it's much easier to read a single MIR node and produce multiple codegen statements than it is to parse multiple MIR nodes back together into a single codegen statement (though again, this shouldn't be a problem because they're inserted in transform_mir).

rybern:
Thanks @rok-cesnovar, I hoping that was the case

seantalts:
> The transformed MIR is already backend-specific, so it's okay if we tailor it to the Stan Math backend. Other backends can read the transformations from out_vars and generate constraints in their own way.

Agreed :) I feel like there are a few major projects underway in that area but once things settle down I'd be happy to take a crack at adding a Stan Math LIR form. 

I was (over?)reacting to 
> Stop generating FnConstrain in Ast_to_Mir

How will constraints be represented in the MIR pre-Stan_math_backend.Transform_mir?

test/integration/good/code-gen/opencl/cpp.expected
```diff
@@ -6729,38 +6744,43 @@ class restricted_model final : public model_base_crtp<restricted_model> {
       p_real = std::numeric_limits<double>::quiet_NaN();
       
       current_statement__ = 1;
-      p_real = in__.scalar();
+      p_real = in__.read<local_scalar_t__>(lp__);
```
SteveBronder, line 6747: Close! Just gotta get rid of the `lp__` for the `read()` methods

rybern: woooooops

rybern:
Yeah an LIR might be really nice - I wonder if we could get rid of the CompilerInternal calls altogether

> How will constraints be represented in the MIR pre-Stan_math_backend.Transform_mir?

I think it's enough to represent constraints as the `transformations` in `output_vars`, I don't think they need to be statements in untransformed MIR

src/stan_math_backend/Expression_gen.ml
```diff
@@ -418,9 +418,20 @@ and pp_compiler_internal_fn ad ut f ppf es =
...
+    | {Expr.Fixed.pattern= Lit (Str, constraint_string); meta= emeta}
+      :: {Expr.Fixed.pattern= Lit (Int, n_constraint_args_str); _} :: args ->
+        let n_constraint_args = int_of_string n_constraint_args_str in
+        let constraint_args, dims = List.split_n args n_constraint_args in
+        let lp_expr = Expr.Fixed.{pattern= Var "lp__"; meta= emeta} in
+        let arg_exprs = constraint_args @ [lp_expr] @ dims in
+        let maybe_constraint, maybe_jacobian =
+          if String.is_empty constraint_string then ("", "")
+          else ("_" ^ constraint_string, ", jacobian__")
+        in
```
SteveBronder, line 430: I think up here for not having `lp__` for plain reads `deserializer.read<>(dims...)` you want to check if `constrain_string` is blank (or if constrain_args is size 0? either or) and if not then don't include `lp__` as an arg

SteveBronder: Also for here @bbbales2 and I are talking about changing the methods names to

```
read() : just plain read
read_{constrain_name}() -> read_constrain_{constrain_name}() : reading in variables that need a constrain
```

to match the signature for the read free functions

```
read_free_{constrain_name}() : reading in variables that need a free transform
```

But that would just involve changing this line to
```suggestion
          if String.is_empty constraint_string then ("", "")
          else ("_constrain_" ^ constraint_string, ", jacobian__")
        in
```

We should have that sorted by today

rybern: Sounds good, lemme know

seantalts:
> I think it's enough to represent constraints as the transformations in output_vars, I don't think they need to be statements in untransformed MIR

In most backends we'd actually call a function here. In general the MIR is supposed to be lowering us towards procedural code as much as possible while being general across backends. It can retain higher level information but not rely on it out-of-band like that - you want everything you need to generate code to be there at your MIR node.

rybern:
I see where you're coming from. In this case, I think we're being as low level as possible while being backend agnostic - it wouldn't be backend agnostic to put in a separate Constrain call, because the Stan Math backend would have to undo that work.

In a sense, it's the Decl nodes that represent the constraints in the MIR statements, because we're expanding the Decls just like we've been doing for reads. Just like we've been doing for reads, we're expanding parameter Decls with information from output_vars.

To put more of the information inline into the MIR, what do you think about adding `transformation option` to `Decl`? Or maybe further, `(constrainaction * transformation) option`, so that backends don't need to know where to put Check/Constrain/Unconstrain? Then backends are free to generate whatever read or constrain calls they need to without looking at `output_vars`. On the one hand, I like that because the information is inline but can't be separated from the declaration, on the other hand, I don't want to do away with `output_vars` but I also don't like redundant representations that can get out of sync...

SteveBronder:
@rybern fyi I pushed up the just changing `reader` to `deserializer` in the gen code and promoting the tests

rybern:
@SteveBronder great - I'll be free to help finish up this afternoon

test/integration/good/code-gen/cpp.expected
```diff
@@ -19567,54 +19010,20 @@ class param_constraint_model final : public model_base_crtp<param_constraint_mod
...
-              2, lp__), "assigning variable L_Omega", index_uni(sym1__));
-        } else {
-          current_statement__ = 1;
-          assign(L_Omega,
-            stan::math::cholesky_corr_constrain(L_Omega_in__[(sym1__ - 1)], 2),
-            "assigning variable L_Omega", index_uni(sym1__));
-        }}
+      assign(L_Omega,
+        in__.template read_constrain_cholesky_factor_corr<std::vector<Eigen::Matrix<local_scalar_t__, -1, -1>>, jacobian__>(
+          lp__, nt, 2, 2), "assigning variable L_Omega");
```
SteveBronder, line 19016: Alright so fixing this bit here got it to `param-constrain.stan` to compile for me (after we add the little fix in this PR [here](https://github.com/stan-dev/stan/pull/3024) since we forgot to include the deserializer header lol)

The model is 

```stan
data {
  int nt;
  int NS;
}

parameters {
  cholesky_factor_corr[2] L_Omega[nt];
  vector<lower=L_Omega[1,1,2]>[NS] z1;
}
```

Is making the `nt, 2, 2)` where the deserializer signature the dimensions to just be `nt, 2)` ([here](https://github.com/stan-dev/stan/blob/develop/src/stan/io/deserializer.hpp#L790)) since then we grab a lower triangular of data with # of cells equal to `(K * (K - 1)) / 2)`. I think this is caused by get_dims in the expression gen stuff for `let read = ...`

https://github.com/stan-dev/stanc3/pull/856/files#diff-a1f5ff85d2b828c19a91e01d5c1e8caaf75c47e8af9bb205a3e0ed486a7e064fR233

bbbales2: Oooo, it's possible that we could have a problem somewhere. `cholesky_factor_cov` supports both: `cholesky_factor_cov[M]` and `cholesky_factor_cov[M, N]` (according to [here](https://mc-stan.org/docs/2_26/reference-manual/vector-and-matrix-data-types.html)). I think it's the only matrix parameter like this (that has a non-square version), so watch out.

bbbales2: And I say it's possible there could be a problem not cause I know -- just cause it is a little different and but sounds related to what you're talking about.

SteveBronder: We also do this pattern for `read_constrain/free_cholesky_factor_corr`, `read_constrain/free_cov_matrix`, and `read_constrain/free_corr_matrix`.

SteveBronder: Aight I just pushed something up that I think will gen the thing we want

https://github.com/stan-dev/stanc3/pull/856/commits/3eaab9eef69c9867ce963d4b857d5111d4e2605d

SteveBronder: Oh ooff Ben just seeing your comment. I think that's fine? The signature for `read_constrain_cholesky_factor_cov` is 

```cpp
read_constrain_cholesky_factor_cov(LP& lp, Eigen::Index M, Eigen::Index N) 
```

And in the thing I wrote we handle that the normal way, so hopefully should be fine

SteveBronder:
Aight that header PR is in I think we just need to wait for cmdstan's develop to update and we can restart the tests

SteveBronder:
Also @rybern it might be good to do the `free` things in a separate PR as `transform_inits()` is the only thing that uses the free functions and those don't need to work with the new matrix type. Then I'm pretty sure once this PR is in @rok-cesnovar can add the new matrix stuff

rybern:
@SteveBronder Agreed about the separate PR

SteveBronder:
@serban-nicusor-toptal does the Jenkins message

```
‘Jenkins’ doesn’t have label ‘linux’; ‘LinuxFleet i-0cd2e1706467eff90’ is offline; ‘WindowsFleet i-09e8f7736d5c66b99’ doesn’t have label ‘linux’; ‘gelman-group-mac’ doesn’t have label ‘linux’; ‘gelman-group-win-new’ doesn’t have label ‘linux’; ‘gelman-group-win2’ doesn’t have label ‘linux’; ‘old-imac’ doesn’t have label ‘linux’
```

Just mean to wait for `LinuxFleet i-0cd2e1706467eff90` to spin up?

test/integration/good/code-gen/cl.expected
```diff
@@ -1298,54 +1304,65 @@ class optimize_glm_model final : public model_base_crtp<optimize_glm_model> {
...
+      phi = in__.template read<local_scalar_t__>();
       Eigen::Matrix<double, -1, -1> X_p;
       X_p = Eigen::Matrix<double, -1, -1>(n, k);
       stan::math::fill(X_p, std::numeric_limits<double>::quiet_NaN());
       
       current_statement__ = 7;
-      X_p = in__.matrix(n, k);
+      assign(X_p,
+        in__.template read<Eigen::Matrix<local_scalar_t__, -1, -1>>(n, k),
+        "assigning variable X_p");
```
rok-cesnovar, line 1349: Do these have to go through assigns?

rok-cesnovar:
One of the code-gen changes seems to be causing models to compile quite a bit longer. I have restarted tests a few times to make sure its not a testing glitch, but it seems its legit 2x+ slower.

I guess its fine if we can not avoid longer compile times, but it would be great to understand why.

rok-cesnovar:
Hmm, it actually took 2h32m (as opposed to 23min on master) so definitely something to look at.

serban-nicusor-toptal:
Hey @SteveBronder that's it yea, we have on-demand agents and they scale to 0 if there's no usage, so as demand grows it will spin up new ones by itself, usually takes a few minutes.

SteveBronder:
Oof yeah that is concerning. I'm surprised though since we use this pattern a decent amount in the math library. My guess is that it's the SFINAE for things like `read()`. I think we could have partial specializations in deserializer for these so that it can at least avoid those lookups

SteveBronder:
What flags do we have on when we compile for those tests? At home I use `-march=native -mtune=native -O3` and I'm seeing for compiling mother with this

Develop cmdstan:
```
real 69.73
user 67.91
sys 1.80
```

```
real 68.42
user 66.61
sys 1.78
```

If the tests compile with O1 or O2 then I could see SFINAE etc. taking more time. But since we normally are compiling at O3 I think the overhead of optimizations is going to overshadow any extra compilation time here

rok-cesnovar:
We have the default flags everywhere which is also what most users have I would say. 

Can you try with: https://github.com/stan-dev/stanc3/blob/02faf205aa4368f910824f2285aef377645e3b59/test/integration/good/function-signatures/distributions/univariate/continuous/gamma/gamma_lpdf.stan

These are typically harder for the C++ compiler. It might be some corner case but we need to know. what is up. Is it maybe just more RAM demanding and this is causing swap usage with compiling 15 parallel models?

SteveBronder:
For gamma I'm getting 

PR:
```
real 110.57
user 108.02
sys 2.50
```

cmdstan master
```
real 110.33
user 108.38
sys 1.92
```

Can you try these locally and see if you can replicate it? This could be the RAM thing, though that feels like a very convenient excuse. If you can check the results locally and get what I'm getting for mother and gamma then I think it could be that. I think the mother model is the one I'd be more concerned about since it hits a lot of the recursion and parameter pack expansions for setting up arrays of arrays

SteveBronder:
This is the model I ballpark high overhead from this change

```stan
data {
  int N;
}

parameters {
  real x[N];
  matrix[N, N] x2[N, N];
  matrix[N, N] x4[N, N, N];
  matrix[N, N] x5[N, N, N, N];
  matrix[N, N] x6[N, N, N, N, N];
  matrix[N, N] x7[N, N, N, N, N, N];
  matrix[N, N] x8[N, N, N, N, N, N, N];
  matrix[N, N] x9[N, N, N, N, N, N, N, N];
  vector[N] y2[N, N];
  vector[N] y4[N, N, N];
  vector[N] y5[N, N, N, N];
  vector[N] y6[N, N, N, N, N];
  vector[N] y7[N, N, N, N, N, N];
  vector[N] y8[N, N, N, N, N, N, N];
  vector[N] y9[N, N, N, N, N, N, N, N];
  real z2[N, N];
  real z4[N, N, N];
  real z5[N, N, N, N];
  real z6[N, N, N, N, N];
  real z7[N, N, N, N, N, N];
  real z8[N, N, N, N, N, N, N];
  real z9[N, N, N, N, N, N, N, N];
}


model {
  target += normal_lpdf(1 | 1, 1);
}
```

But running that locally I'm also seeing 

PR:
```
real 32.08
user 31.13
sys 0.95
```

Master:
```
real 30.29
user 29.28
sys 1.00
```

So 2 extra seconds

test/integration/good/code-gen/cl.expected
SteveBronder: I'd like to avoid it but idk how. Since we are guaranteed that the deserializer has enough values in it for each parameter this could be 

```
      Eigen::Matrix<local_scalar_t__,, -1, -1> X_p = in__.template read<Eigen::Matrix<local_scalar_t__, -1, -1>>(n, k);
```

The assignment is generated [here](https://github.com/stan-dev/stanc3/pull/856/files#diff-a1f5ff85d2b828c19a91e01d5c1e8caaf75c47e8af9bb205a3e0ed486a7e064fR253). Would we need to change that `Fixed.pattern` to a `Decl`? I can't figure out how to do deceleration and assignment on the same line. But doing that would save both memory and computational time

rok-cesnovar:
I tried a few models locally and it looks good on my end too. Also looked at RAM usage and it looks to be in the same ballpark, give or take a few hundred megs (highly scientific measurements as you can see). But certainly not something that would cause 5x longer compilation. 

Its easily repeatable on Jenkins so I am going to spin off a separate PR to debug this a bit over the weekend. 2h30 mins for compile tests is something we want to avoid. Its also not a Jenkins glitch as Stan integration tests are keeping the ~20min pace.

Maybe its clang6?

SteveBronder:
Yeah I can't put my finger on this. I just ran the model I posted above with all the recursive things and I'm getting 

PR:
```
real 54.76
user 54.24
sys 0.47
```

Master:
```
real 63.34
user 62.69
sys 0.58
```

Which is faster??? That doesn't add up, looking at the mem usage they seemed to be in the same ballpark

test/integration/good/code-gen/cl.expected
rok-cesnovar: I pushed a change that removes that, feel free to remove it if you dont like it.

test/integration/good/code-gen/cl.expected
```diff
@@ -427,7 +423,7 @@ class optimize_glm_model final : public model_base_crtp<optimize_glm_model> {
       y_vi_d = std::vector<int>(n, std::numeric_limits<int>::min());
       
       current_statement__ = 189;
-      assign(y_vi_d, context__.vals_i("y_vi_d"), "assigning variable y_vi_d");
+      y_vi_d = context__.vals_i("y_vi_d");
```
rok-cesnovar, line 426: These go through assign on master but I am guessing they dont have to as well?

SteveBronder: Essentially I think anywhere we were doing an `assign()` without any form of indexing we can just go right through to an `=`

SteveBronder:
@serban-nicusor-toptal is the ARM ec2 instance down?

> All nodes of label ‘arm-ec2’ are offline

serban-nicusor-toptal:
Hey @SteveBronder looks like it didn't scale by itself, I've increased the limit to `1` so it should spin up in a few moments.

SteveBronder:
Noice ty!

Jenkinsfile
```diff
@@ -156,7 +155,8 @@ pipeline {
                                     text:"O=0\nCXX=${CXX}")
                             sh """
                                 cd performance-tests-cmdstan
-                                cd cmdstan; make -j${env.PARALLEL} build; cd ..
+
+                                cd cmdstan; make -j6 build; cd ..
```
SteveBronder, line 159: @rok-cesnovar I set this down to 6 because I looked the job was running out of memory. Should we set it to just like half the available `env.PARALLEL`?

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -510,13 +510,14 @@ let pp_write_array ppf {Program.prog_name; generate_quantities; _} =
...
-      ; "stan::io::reader<local_scalar_t__> in__(params_r__, params_i__);"
+      ; "stan::io::deserializer<local_scalar_t__> in__(params_r__, params_i__);"
       ; "double lp__ = 0.0;"
       ; "(void) lp__;  // dummy to suppress unused var warning"
       ; "int current_statement__ = 0; "
       ; "stan::math::accumulator<double> lp_accum__;"
       ; "local_scalar_t__ \
-         DUMMY_VAR__(std::numeric_limits<double>::quiet_NaN());" ]
+         DUMMY_VAR__(std::numeric_limits<double>::quiet_NaN());"
+      ; "constexpr bool jacobian__ = false;" ]
```
SteveBronder, line 520: This is just here because I'm lazy, I'm fine with leaving this but we should probably just have a `false` in the constrain calls for write_array

Jenkinsfile
rok-cesnovar: This was resolved by #858

test/integration/good/code-gen/cl.expected
rok-cesnovar: Not really, only when we we know size of lhs is equal to rhs (see https://github.com/stan-dev/stanc3/issues/523). With context__ and in__ I think we can avoid it right?

SteveBronder:
@rybern you think this is ready for review?

rybern:
I thiink so. Theres some code in ast_to_mir that doesnt really make sense anymore, but it makes more sense to fix that after the free PR

rok-cesnovar:
Who do we get to review this? 

I think @seantalts had some thoughts on this. Do you want to do it Sean?

All models compile so in terms of C++/codegen parts its looking great.

SteveBronder:
Pinging @seantalts or @nhuurre would one of you mind taking a look at this?

seantalts:
I can take a look this weekend, also happy for nhuurre to take a look if
he’s interested:)

SteveBronder:
Also it looks like the jenkins is waiting for docker?

seantalts:
Did the compile times end up getting fixed? What was the issue?

rok-cesnovar:
The issue was `-j${env.PARALLEL}` missing from:

`./runPerformanceTests.py --runs=0 ../test/integration/good`

which meant `-j 32` and that uses too much RAM. See https://github.com/stan-dev/stanc3/blob/master/Jenkinsfile#L161

The compile tests now run for ~23 min, where ~1++ minutes are spent compiling the huge models for exp_mod_normal, pareto_type_2, wiener, skew_normal and student_t dist.. Models like this: https://github.com/stan-dev/stanc3/blob/master/test/integration/good/function-signatures/distributions/univariate/continuous/student_t/student_t_ccdf_log_1.stan

I feel these are a bit unnecessary actually and we are wasting resources here as these signatures are already tested in both distribution as well as expression tests. But on the other hand 17minutes is not a ton.

seantalts:
Is it true that using these new constructs makes compilation take more RAM than previously? Still something to note and maybe add to the release notes or something, even if the tests we're using are somewhat large and arbitrary.

rok-cesnovar:
I havent been able to reproduce a noticeable increase of RAM. Definitely not in a range of a 20% increase or anything like that which we would have to mention in release notes.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -161,56 +161,97 @@ let data_read smeta (decl_id, st) =
...
+      match constrain_transform with
+      | Program.CholeskyCorr | Correlation | Covariance ->
+          constrain_get_dims st
+      | _ -> SizedType.get_dims st
+    in
+    (* this is an absolute hack
+
+       I need to unpack the constraint arguments and the dimensions in codegen, but we pack them all together into a fake function FnReadParam as expressions
+       So, I'm packing in the number of constraint arguments to read as an int expression
+       To avoid this hack, keep internal functions as a variant type instead of a normal funapp
```
seantalts, line 240: I think that seems like a fine thing to do now if you think it's a good idea (I think I was against it before, apologies!)

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -161,56 +161,97 @@ let data_read smeta (decl_id, st) =
...
+    match c with Check -> "" | Constrain | Unconstrain -> "offset_multiplier" )
+  | Identity -> ""
+
+let constrain_constraint_to_string t (c : constrainaction) =
+  match t with
+  | Program.CholeskyCorr -> "cholesky_factor_corr"
+  | _ -> check_constraint_to_string t c
+
+let default_multiplier = 1
+let default_offset = 0
```
seantalts, line 198: Nice, I like encoding this explicitly here.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -161,56 +161,97 @@ let data_read smeta (decl_id, st) =
...
+    let n_args_expression = Expr.Helpers.int (List.length transform_args) in
+    let read =
+      Expr.(
+        Helpers.(
+          internal_funapp FnReadParam
+            ( Expr.Helpers.str
+                (constrain_constraint_to_string out_trans Constrain)
+            :: n_args_expression
+            :: (transform_args @ read_constrain_dims out_trans cst) ))
+          Typed.Meta.{decl_var.meta with type_= ut})
```
seantalts, line 251: Would you mind adding an expect test for this Transform_mir step? I just realized it's only kind of implicitly tested and I'd really love to see the actual reified MIR before and after to help me see how this step works. It doesn't need to be super complicated but it'd be good if it had a constrained parameter. A thing that would be great would be picking some example model and adding a test that just runs the compiler up to right before the Transform_mir step, prints that out, then runs Transform_mir and prints it out again.

seantalts (changes requested):
This looks good, thanks! I think I have some feedback/answer to some of the questions you had inline + a few small suggestions. Thank you!

src/frontend/Ast_to_Mir.ml
```diff
@@ -815,7 +815,9 @@ let trans_block ud_dists declc block prog =
           if Utils.is_user_ident decl_id then
             let constrain_checks =
               match declc.dconstrain with
-              | Some Constrain | Some Unconstrain ->
+              | Some Constrain ->
```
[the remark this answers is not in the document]
rybern: Yep I agree this is weird. It's left like this because the plan is to do unconstrains in a separate PR, so a lot of cleanup should happen then

src/analysis_and_optimization/Pedantic_analysis.ml
```diff
@@ -75,12 +75,14 @@ let list_multi_twiddles (mir : Program.Typed.t) :
 let var_deps info_map label ?expr:(expr_opt : Expr.Typed.t option = None)
     (targets : string Set.Poly.t) : string Set.Poly.t =
   (* Labels of dependencies *)
-  let dep_labels =
+  (* let expr_vars_opt = Set.Poly.map ~f:fst (expr_var_set expr) in *)
```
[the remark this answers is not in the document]
rybern: Nope thanks

Jenkinsfile
```diff
@@ -156,6 +155,7 @@ pipeline {
                                     text:"O=0\nCXX=${CXX}")
                             sh """
                                 cd performance-tests-cmdstan
+
```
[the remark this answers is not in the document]
rybern: Didn't mean to touch this file..

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -161,56 +161,97 @@ let data_read smeta (decl_id, st) =
...
-  | URowVector -> "row_vector"
-  | UReal -> "scalar"
-  | UInt -> "integer"
-  | UArray t -> base_ut_to_string t
-  | t ->
-      raise_s
-        [%message "Another place where it's weird to get " (t : UnsizedType.t)]
+type constrainaction = Check | Constrain | Unconstrain [@@deriving sexp]
+
+let check_constraint_to_string t (c : constrainaction) =
```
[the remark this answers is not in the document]
rybern: I left it because we're leaving unconstrains for another PR, but I can still probably clean it up

src/analysis_and_optimization/Pedantic_analysis.ml
```diff
@@ -75,12 +75,14 @@ let list_multi_twiddles (mir : Program.Typed.t) :
...
+  let dep_labels, expr_vars =
     match expr_opt with
-    | None -> node_dependencies info_map label
+    | None -> (node_dependencies info_map label, Set.Poly.empty)
     | Some expr ->
-        let vars = Set.Poly.map ~f:fst (expr_var_set expr) in
-        node_vars_dependencies info_map vars label
+        let vvars = Set.Poly.map ~f:fst (expr_var_set expr) in
```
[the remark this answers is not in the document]
rybern: It's a wrapper around string to represent variables, to make refactoring easier in case variables were ever not just strings. It's just used in the very oldest parts of code analysis. I might actually keep it around in order to handle indexed tuple expressions because they behavior like variables but aren't strings

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -161,56 +161,97 @@ let data_read smeta (decl_id, st) =
...
+        raise_s
+          [%message "LowerUpper is really two other checks tied together"]
+    | Constrain | Unconstrain -> "lub" )
+  | Offset _ | Multiplier _ | OffsetMultiplier _ -> (
+    match c with Check -> "" | Constrain | Unconstrain -> "offset_multiplier" )
+  | Identity -> ""
+
+let constrain_constraint_to_string t (c : constrainaction) =
+  match t with
+  | Program.CholeskyCorr -> "cholesky_factor_corr"
```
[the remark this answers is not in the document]
rybern: sure is, it used to be different and I didn't clean it up, thanks

src/stan_math_backend/Transform_Mir.ml
rybern: Sure, maybe I'll see if it's relatively easy to do here and otherwise a separate PR. Would be nice to avoid the silly packing thing I did here

src/stan_math_backend/Expression_gen.ml
```diff
@@ -417,9 +417,23 @@ and pp_compiler_internal_fn ad ut f ppf es =
...
   | Some FnReadParam -> (
     match es with
-    | {Expr.Fixed.pattern= Lit (Str, base_type); _} :: dims ->
-        pf ppf "@[<hov 2>in__.%s(@,%a)@]" base_type (list ~sep:comma pp_expr)
-          dims
+    | {Expr.Fixed.pattern= Lit (Str, constraint_string); meta= emeta}
+      :: {Expr.Fixed.pattern= Lit (Int, n_constraint_args_str); _} :: args ->
+        let n_constraint_args = int_of_string n_constraint_args_str in
+        let constraint_args, dims = List.split_n args n_constraint_args in
+        if String.is_empty constraint_string then
```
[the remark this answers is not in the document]
rybern: I think I'd slightly favor "" (if it has to be a string), because 1) it's one less special constant value that we have to keep track of throughout the code, and 2) in all other cases that string represents part of the `read_constrain_` function name, so it makes some kind of sense to be empty string because it's not calling any constraint. I agree "identity" is more informative, but at this point it's generating backend-specific code that's packed into a pretty unreadable state anyway. This would also be solved by refactoring internal functions into variant types

src/stan_math_backend/Transform_Mir.ml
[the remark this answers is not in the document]
rybern: I think this'll actually be a bit annoying to clean up since Ast_to_mir shouldn't depend on Transform_mir. I'd prefer to clean this up with everything else in the unconstrain PR

rybern:
Thanks @SteveBronder forgot to push that fix. I've got an MIR update to fix that constrain function hack I'll try to push tonight

rybern:
I did that refactor to add information to Internal_fun.t, and the hack is somewhat better now. I didn't actually end up using the refactor to the extent that I meant to but fitting expressions into the FnReadParam is more of a headache than I want tonight. Anyway I think that should address some of your points @seantalts

src/stan_math_backend/Transform_Mir.ml
rybern: This should be much simpler now, the MIR arguments just reflect how they're called in codegen. It's lousy that I added a an lp__ variable in the MIR but it's better than it was.

test/unit/Optimize.ml
```diff
@@ -2339,7 +2336,7 @@ model {
         data vector[3] y_vector;
         data matrix[2, 2] x_cov;
         data vector[3] x_cov_free__;
-        x_cov_free__ = FnUnconstrain__(x_cov, "cov_matrix");
+        x_cov_free__ = (FnUnconstrain cov_matrix)__(x_cov);
```
rybern, line 2339: I should fix this

SteveBronder:
@seantalts lil' bump!

SteveBronder (approved):
@seantalts both @rok-cesnovar and I looked this over and we think it looks good. If you have time to look it over tonight one more time that would be super rad but this pr being merged is holding up some other PRs so we'd like to merge it tmrw if you don't have time to look at it

seantalts:
Thanks @rybern, this is looking good, I like the refactors and think this is going in a good direction.
I understand how we got here and I see the value in the PR as it is today, but I don't get a good feeling from this PR having a couple of half-refactors in it. @rybern (or @SteveBronder) are you guys committed to cleaning up after? I think if this were in the old compiler or Stan Math someone would be asking for completeness and removal of old code. For the record, this form of compromise is how we ended up with the MIR using a radically different pattern from the AST, and I think we can all agree that the half-refactored state is the worst of both worlds.

At the minimum I think this should have the MIR test I asked for - I can add that this weekend if need be.

seantalts:
I'm trying to figure out how hard it would be to just do the unconstrain param refactor at the same time - @SteveBronder do you know if the Stan Math part of that is ready? Ideally we wouldn't leave stanc3 with constrains and unconstrains having totally separate code paths, MIR representations, and backend calls with all of the duplicated stuff that implies.

rok-cesnovar:
I am guessing this adds the serializer: https://github.com/stan-dev/stan/pull/3019

src/analysis_and_optimization/Pedantic_analysis.ml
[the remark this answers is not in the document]
rybern: Honestly don't remember why it's called vvar, it might just have been to avoid colliding with var. I think it's just used in the code analysis sections and it should be replaced with lvalues when we can handle array indices. I just renamed it here to make it clearer that the type of this variable is vvar, nothing changed semantically

rybern:
@seantalts Thank you for putting time into this!

I totally agree that it's uncomfortable to merge it in it's half state. This way just made sense given the plan for a followup PR.

The unconstrain piece is next up whether it's in this PR or a new PR, I just didn't want to write it as a new PR until this one got the all-clear. I don't think it'll take long especially if we have examples of what the output should look like.

Also, good call on making Fun_kind non-parametric for now. I imagine the internal function type could be a parameter of MIR later on so that backends can provide their own, but no need for now.

SteveBronder:
> I'm trying to figure out how hard it would be to just do the unconstrain param refactor at the same time - @SteveBronder do you know if the Stan Math part of that is ready? Ideally we wouldn't leave stanc3 with constrains and unconstrains having totally separate code paths, MIR representations, and backend calls with all of the duplicated stuff that implies.

All the serializer and deserializer code is in Stan. The reason I wanted to break this up into two is because right now `transform_inits()` takes in a `context__` and queries it for the data it needs (which stanc then has to unfold into the matrices, vectors, etc. So to use the serializer we need to change the transform inits signature to accept just a vector of data which is then passed to `deserializer()` (and then placed into `serializer()` using the free functions). But that also means changing things in the C++ to pass that correctly so we wanted to do the dual merge scheme where we merge stanc and the stan repo at the same time like we did with `rvalue()` and `assign()`.

An alternative to the dual merge thing is to hard code the current transform inits method like we do with `log_prob` etc. In that method we would take all the data we need out of the context and put it into a vector, then pass that vector to the new transformed inits (we could just call it `transform_inits_impl()`). So putting 

```cpp
inline void transform_inits(const stan::io::var_context& context,
                         Eigen::Matrix<double, Eigen::Dynamic, 1>& params_r,
                         std::ostream* pstream = nullptr) const final {
        std::vector<std::string> constrain_param_names;
        context.names_r(constrained_param_names);
        std::vector<double> serialized_val_r();
        serialized_val_r.reserve(context.total_vals_r_size(), 0);
        size_t sentinal_iter = 0;
        for (size_t i = 0; i < constrained_param_names.size(); ++i) {
          std::vector<double> init_val = context.vals_r(constrained_param_names[i]);
          for (size_t j = 0; j < init_val.size(); ++j) {
            serialized_val_r[sentinal_iter] = init_val[i];
            sentinal_iter++;
          }
     transform_inits_impl(serialized_val_r, params_i, vars, pstream);
}
```

into the model. This would let us have backwards compatibility.

I'm fine with doing it either way, I have a branch of the Stan repo with the changes we would need to make for the dual merge [here](https://github.com/stan-dev/stan/compare/feature/new-transform-inits#diff-5091b05a2df83b13e7ddbcff8ca8ea520d005fa714d5883204830ff8823fa83a)

seantalts:
> Also, good call on making Fun_kind non-parametric for now. I imagine the internal function type could be a parameter of MIR later on so that backends can provide their own, but no need for now.

I realized that's what you were probably thinking after I pushed it up, haha. I think I agree it's slightly better to add the parameter when it's time to use it. Also I'm hoping I or someone else can take a crack at a LIR for Stan Math sometime, in which case it probably just has its own types.

>  So to use the serializer we need to change the transform inits signature

Doesn't that change the API and require bumping Stan's major version number? Or do we have some backwards-compatible route planned?

seantalts:
Sorry for the double post - what does "dual merge" refer to here? Splitting `constrain` and `unconstrain` across two PRs?

SteveBronder:
> Doesn't that change the API and require bumping Stan's major version number? Or do we have some backwards-compatible route planned?

Oh I didn't know that would cause a version bump. I think we can do the backwards compatible version I posted above. 

> Sorry for the double post - what does "dual merge" refer to here? Splitting constrain and unconstrain across two PRs?

I didn't know what to call it so I just made that term up lol. It's like, you know when we have a PR in one of our repos that causes something to fail in the upstream repo so we make a branch in each repo, test both of those against one another, then accept and merge both. Like a good example is the `rvalue()` and `assign()`that prettied those up by removing the `cons_index_list()` stuff. We made a branch in Stan with the change, then Rok made a branch of the compiler, both were tested against one another, and then both were merged in. That's what I mean by "dual merge"

seantalts:
> Oh I didn't know that would cause a version bump.

I think that transform_inits is part of the API used by RStan, PyStan, et al. I think this is basically true of all model classes on Stan. Perhaps some day we'll all agree to have the API be that exposed by cmdstan, but I think for now that's still a pretty public facing endpoint. That said I haven't been to meetings a year or so, maybe things have changed. But it's good you have a backwards-compatible version ready.

I think re: "dual merge" we used to call them cross-repo features or PRs.

Is there anything going on in this space for Stan constraint checks? e.g. data that is declared as lower=0. Does anyone know if there are vectorized versions of that as well? Would be nice to delete that too from the MIR.

seantalts:
PS re: `sentinel_iter`: an "iterator sentinel" is a different thing than a loop index - a sentinel is some special value in your type that signals something, in this case that the end of iteration has been reached. I think this is a C++20 feature you can read about here: https://foonathan.net/2020/03/iterator-sentinel/. In the integers, negative numbers or MIN_INT etc are often used as a sentinel. This was more common when you didn't have access to a complex enough type system to represent things like `option` types.

rok-cesnovar:
> Is there anything going on in this space for Stan constraint checks? e.g. data that is declared as lower=0. Does anyone know if there are vectorized versions of that as well? Would be nice to delete that too from the MIR.

Yes, there is, see https://github.com/stan-dev/stanc3/pull/552

Also see the test file: https://github.com/stan-dev/stanc3/blob/master/test/integration/good/code-gen/transform.stan

seantalts (approved):

seantalts:
I'm fine with merging this now. Looking forward to the unconstrain version. Thanks all!

PR #858, opened 2021-03-20
Set -j PARALLEL in compile tests

rok-cesnovar:
The compile tests in #856 seem to be taking way longer than before. Last master run they took a little under 24 minutes, on PR they take 2-3hours.


rok-cesnovar:
@SteveBronder this should help fix the issue in #856 . -j PARALLEL for runPerformanceTests was not set, which meant it ran with -j32.

rok-cesnovar:
Ready for review.

SteveBronder (approved):
Good catch thanks for investigating this!

PR #864, opened 2021-03-27
Adds type annotation to functions in semantic_check.ml

SteveBronder:
## Release notes

Adds type annotation in `Semantic_check.ml`

This PR shouldn't change any functionality. I often found that when looking at the code in `vscode` the gui would dump out some ginormous type whenever we actually had a specialized type or it would put `'a` in for a type that was actually the same from all the function call sites. This PR adds type annotations to the functions in semantic checking that I thought could reasonably use it. A good example is the `cf` argument which was some `'a` but actually is just a `context_flag_record` This also helps clarify some cases where we use the same argument name in multiple functions but they have different types in each of those functions. `name` is a common example where sometimes it is an `Ast.identifier` and other times it's a `string`

## Copyright and Licensing

license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause) Steve Bronder

### DIFF
--- a/src/frontend/Ast.ml
+++ b/src/frontend/Ast.ml
@@ -62,6 +62,9 @@ type typed_expr_meta =
 type typed_expression = (typed_expr_meta, fun_kind) expr_with
 [@@deriving sexp, compare, map, hash, fold]
 
+type 'a abstract_typed_expression = (typed_expr_meta, 'a) expr_with
+[@@deriving sexp, compare, map, hash, fold]
+
 let mk_untyped_expression ~expr ~loc = {expr; emeta= {loc}}
 
 let mk_typed_expression ~expr ~loc ~type_ ~ad_level =
--- a/src/frontend/Semantic_check.ml
+++ b/src/frontend/Semantic_check.ml
@@ -17,7 +17,8 @@ module Validate = Common.Validation.Make (Semantic_error)
    on (1) a global symbol table vm, and (2) structure of type context_flags_record
    to communicate information down the AST. *)
 
-let check_of_compatible_return_type rt1 srt2 =
+let check_of_compatible_return_type (rt1 : UnsizedType.returntype)
+    (srt2 : Ast.statement_returntype) : sexp_bool =
   UnsizedType.(
     match (rt1, srt2) with
     | Void, NoReturnType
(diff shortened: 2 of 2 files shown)


seantalts:
@SteveBronder I think there might be something wrong with your merlin setup? The examples I checked from the PR and the ones you list in the description (`context_flags_record` and `name`) are showing up as the right type in my editor. Crucially, one has to have run `dune build` recently for the types to show up. 

Does that fix things for you?

SteveBronder:
Odd I thought I rebuilt and now `cf` is showing up correctly but I'm still seeing some which don't use the named type like `typed_statement`'s in  `semantic_check_fundef_return_tys` is showing up as 

`(typed_expression, stmt_typed_located_meta, typed_lval, fun_kind) statement_with`.

 And in `semantic_check_functions_have_defn` the type is coming up as a 

`('a, stmt_typed_located_meta, 'b, 'c) statement_with sexp_list sexp_option` 

when it's actually a `typed_statement sexp_list sexp_option`

Is there a reason not to include type annotation? I know folks suggest having separate .ml and .mli files with definitions in the .mli and impls in .ml. Personally I kind of see that like how in C++ folks separate out their function definitions from impls into separate header and source files (`.hpp` and `.cpp`. That can help compile times when linking multiple packages but I personally find that harder to work with and all of these modules are private within stanc. I know at one point there was discussion on breaking up each of these private libraries into public libraries. But tmk that conversation happened a while ago and so far hasn't gained any traction.

seantalts:
I think you're right that `.mli` files are similar to header and source files in cpp, though I think in both cases it's more than just for compile times - these files authoritatively define the interface for those modules. So you're right that these internal functions probably wouldn't appear in this module's .mli file because they shouldn't be used elsewhere.

I think this style of inline type definition for functions is pretty rarely used in practice and can be somewhat cluttersome, but I don't feel that strongly. 

Another option that may work well here would be to define some additional helper modules inline and give them a signature:
```ocaml
module Helpers : sig 
  val type_of_expr_typed : typed_expression -> UnsizedType.t
end = struct
  let type_of_expr_typed ue = ue.emeta.type_
end
```

(possibly a bad example function as it says the types in the name, but you get the idea). This is more verbose but should constrain them and make the interface easy to look at at a glance.

I like this less, but it's also possible to write a somewhat cleaner version of inline types:
```ocaml
let lub_rt : Location_span.t -> UnsizedType.returntype -> UnsizedType.returntype -> UnsizedType.returntype Validate.t = fun loc rt1 rt2 -> ...
```

Just a nit - we ~never want to specify `sexp_bool` or `sexp_*` in general as the type - the fact that Merlin displays those is sort of a bug / artifact of the Core + Merlin version we're using. You can just write `bool` and `list` etc.

PR #867, opened 2021-04-02
Add a transformed_mir_diff expect test for #856 to diff against

seantalts:
Just adding an expect test so we can track changes in the transformed MIR as well as the pre-transform MIR.

(no release notes)

## Copyright and Licensing


rok-cesnovar (approved):
Thanks Sean!

seantalts:
@rok-cesnovar while you're here, what do you think of this idea of adding the transformed MIR expect test as a diff from the pre-transform one? Is that too complicated? That's what's currently up - it's about 15000 lines as a diff, and about 30,000 lines not as a diff. Also, the diff logic uses bash so may not be portable - waiting for Jenkins to yell. I have the non-diff version ready to push if it doesn't work or you like the non-diff version better. I get a little confused thinking about future PRs comparing a diff of a diff...

rok-cesnovar:
Honestly, 30k is not that big and it makes is simpler to understand I guess? There are a bunch of distributions with 35k+ expected lines and those are not nearly as important as this.

seantalts:
Sounds good. Also, apparently `diff` has different behavior on mac, lol. Pushing up the simpler version, thank you!

PR #869, opened 2021-04-03
Handle int division using Warnings module

### LINKED ISSUE #824: Int division warning should not be printed to stdout
```stan
data {
}
transformed data {
}
parameters {
  real mu;
  real sigma;
}
model {
  target += normal_lpdf(mu | 0, 1) + 1/2;
}
```
will print 
```
Info: Found int division at 'test.stan', line 9, column 37 to column 38:
  1 / 2
Values will be rounded towards zero. If rounding is not desired you can write
the division as
  1.0 / 2
```
to stdout. 

It should instead be hadled by the Warnings module to print this to stderr and to avoid printing to stderr directly (a problem for stanc.js).

adamhaber:
## Release notes

Fixes #824

Int division is now handled by the Warnings module to print this to stderr and to avoid printing to stderr directly (a problem for stanc.js).

## Testing

Not 100% sure how to test this - I've verified that the warning isn't printed to stdout anymore, but not sure how to verify that's it's going where it should...

## Copyright and Licensing

Adam Haber


seantalts:
Hey, thanks for this! I think we probably want to avoid using the stateful Warnings.warnings list here - that basically just exists because the lexer isn't very flexible and needs a place to store global state. In some sense it probably doesn't belong in this module and should instead be a separate module that has that global state just for that purpose. You could pull that out in this PR if you like - basically just move the stateful warnings, init, and collect to another module maybe called something like Lexer_warnings.ml. But that's optional. 

For this one, I think maybe transitioning Semantic_check (or at least this part of it) to also return a list of Warning.ts would be a more OCaml-y way to do this. Currently the main signature is:
```ocaml
val semantic_check_program :
     Ast.untyped_program
  -> (Ast.typed_program, Middle.Semantic_error.t list) result
```

and I think we'd basically want to change that to 
```ocaml
val semantic_check_program :
     Ast.untyped_program
  -> ((Ast.typed_program * Warnings.t list), Middle.Semantic_error.t list) result
```

There may be some better type that encapsulates successful output, warnings, and error types than this, I'll look into a bit, but this should be fine. The implementation of this would be I think very difficult if you wanted to try to change Validation.ml to deal with the warnings as well; instead I'd suggest for now just copying the same global state pattern into Semantic_check.ml - Initialize some global warnings list ref when you start type-checking a program, use `add_warning` throughout Semantic_check.ml to add warnings to it, and then when returning the results `collect` them into a list and put it in the right place in the result tuple. 

Does that make sense? This isn't the only way to do it but this seems like a reasonable one to me that avoids reusing the lexer's global state from another module and encapsulates it inside Semantic_check so that it's not visible (or ruinable) from outside the module. Happy to hop on a call or chat more about other implementation ideas, hope this found a good middle ground between overly prescribed and overly vague.

adamhaber:
Thanks @seantalts, that makes perfect sense and sounds like a good OCaml exercise for me. :-) I'll try to work on this next week, will let you know if I get stuck.

riddell-stan:
This affects PyStan too. Glad to see it's being addressed!

WardBrian:
The updated test outputs suggest to me that this change appears to result in the warning not being emitted at all. I believe this is because the warnings are only "collected" when the parser is called.

Following @seantalts suggestion and changing changing the return type of `semantic_check_program` is the best way around this, I believe. You would then need to call Warnings.pp/pp_warnings at all the call sites of that function if it succeeds, and the warnings should re-appear in the test files.

adamhaber:
Apologies for the tardy response, will get back to working on this later today!

adamhaber:
@seantalts it this roughly what you were aiming for?

Not sure if you were suggesting adding the warnings list ref at the beginning of the `Semantic_check.ml` file or locally within `semantic_check_program` and pass the list around. I went with the former since it seemed simpler - but perhaps it's less "clean"? I'd be happy to change that if you prefer the other approach (or something else altogether :-))...

Also, I'm not sure how to move from `(Ast.typed_program * Middle.Warnings.t list)` (the output of `attach_warnings`) to `((Ast.typed_program * Middle.Warnings.t list), Middle.Semantic_error.t list) result` (the desired return type). Do I need to change `Validation.get_with`?

src/frontend/Semantic_check.ml
```diff
@@ -1926,7 +1934,5 @@ let semantic_check_program
   Validate.(
     ok mk_typed_prog |> apply_to ufb |> apply_to udb |> apply_to utdb
     |> apply_to upb |> apply_to utpb |> apply_to umb |> apply_to ugb
-    |> check_correctness_invariant_validate
-    |> get_with
-         ~with_ok:(fun ok -> Result.Ok ok)
-         ~with_errors:(fun errs -> Result.Error errs))
+    |> check_correctness_invariant_validate)
+    |> attach_warnings
```
WardBrian, line 1938: I believe the way you would get the correct return type here would be 

```ocaml
|> get_with
      ~with_ok:(fun ok -> Result.Ok (attach_warnings ok))
      ~with_errors:(fun errs -> Result.Error errs))
```

adamhaber: Thanks!

adamhaber:
Thanks @WardBrian , that was exactly what I was missing.

So now `semantic_check_program` returns the desired output. Other than `stancjs`, in which `semantic_check_program` is called directly, `semantic_check_program` is only called via `Frontend_utils.typed_ast_of_string_exn` or `Frontend_utils.type_ast_or_exit`. I've change them so that they emit the warnings using `Warnings.pp_warnings` - hope this makes sense!

If this looks good, I'll try to fix `stancjs` accordingly - for some reason, VSCode is having a hard time inferring types in `stancjs` which makes refactoring challenging...

WardBrian:
That looks great to me! Fixing up the final call site in stanjs would be great, and resolving the merge conflict in warnings.ml would also be nice. I after this is finished I’ll submit a separate PR to implement the rest of @seantalts suggestions re:moving the existing warning stateful behavior to something called input-warnings or lexer-warnings

adamhaber:
OK, I tried to fix stancjs but not sure whether `dune runtest` actually tests these or not - will let Jenkins be the judge. :-)

Also, not sure why I'm getting all the newlines in the expected files, or if this is problematic or not.

WardBrian:
The newlines appear to be generated by `Warnings.pp_warnings` when a blank list is supplied. It would be trivial to fix, but I'm not sure if it is a big concern. Want to weigh in @seantalts?

Other than that this is looking good to me

rok-cesnovar:
@adamhaber to run the stancjs tests you only need to run `dune build @runjstest`. Also run `opam pin -y js_of_ocaml-ppx 3.4.0` prior to that to install the dependency.

adamhaber:
Thanks! This is probably ready for review, then.

seantalts:
I can take a look tonight or tomorrow

seantalts (approved):
Looks great to me, thank you! Made two tiny tweaks - the first a slight style preference and the second to remove "Info:" from the message that now always starts with "Warning:"

WardBrian (approved):
Looks good to me as well! I’m going to go ahead and merge this. Thank you!

PR #872, opened 2021-04-05
Refactor unconstraining to use deserializer interface

rybern:
This PR moves the codegen for unconstraining variables to use to new deserializer backend interface (see e.g. https://github.com/stan-dev/stan/pull/3018)

This is follows up on PR #856.

The approach is to create a new internal function, ReadUnconstrainData, that generates a function call from the new interface.


rybern:
@SteveBronder This is a WIP, but can you tell close the codgen is? Should transform_init_impl be getting passed a `deserializer` object now?

SteveBronder:
It looks pretty darn close! I can take a deeper dive into it a bit more today and tmrw

SteveBronder:
> Should transform_init_impl be getting passed a deserializer object now?

It should just get handed a vector and the deserializer can be made in the `impl` I think we could make deserializer and pass it to the impl but I always kind of worry about passing things into functions that hold references. I can write up the bit for how things should be passed to the impl

SteveBronder:
So I think what we need to do is take the hardcoded function here in transform_inits

```cpp
    inline void transform_inits(const stan::io::var_context& context,
                                std::vector<int>& params_i,
                                std::vector<double>& vars,
                                std::ostream* pstream = nullptr) const final {
      transform_inits_impl(context, params_i, vars, pstream);
    }
```

and serialize all the data in the context into a one flat std::vector. I could do some funny stuff in this PR but it might actually be easier if I add a `flatten()` method to the context that returns back an `std::vector<double>` of the data so then we can just change this to 

```cpp
    inline void transform_inits(const stan::io::var_context& context,
                                std::vector<int>& params_i,
                                std::vector<double>& vars,
                                std::ostream* pstream = nullptr) const final {
      std::vector<double> flat_context = context.flatten();
      transform_inits_impl(flat_context, params_i, vars, pstream);
    }
```

I can take a shot at just doing it in the compiler here and if it doesn't look too bad we will do that and otherwise I'll add the flatten() method

and we can also just make a 

```cpp
    inline void transform_inits(std::vector<double>& params_r,
                                std::vector<int>& params_i,
                                std::vector<double>& vars,
                                std::ostream* pstream = nullptr) const final {
      transform_inits_impl(params_r, params_i, vars, pstream);
    }
```

So future API points don't have to pay for the flattening

rybern:
That sounds good Steve, let me know if I can help. I'll do some cleanup in the mean time..

SteveBronder:
Alright so I'm using this as the example program and got everything compiling

```stan
data {
  int<lower=0> N;
  int M;
}
parameters {
  matrix<lower=0>[N, N] x1;
  matrix<lower=0, upper = 1>[N, N] x2;
  matrix[N, N] x3;
  matrix<lower=0, upper = 1>[N, N] x4[M];
}
```

Here's the C++ right now

https://gist.github.com/SteveBronder/ad4004530f9f9126f23ec76269c47f12

And we can use the serializer to do this

https://gist.github.com/SteveBronder/d02822989daf434a54cba4e3f3aa51e7

Does that work for you? I think we can do something similar in `write_array` as well if we want to clean that up

SteveBronder:
Also notice the signature change to have a name `out_vars__` as the output we write to, we make a local `vars__` vector filling it with zeros, fill that up using `serializer` then we just assign `vars__` to `out_vars__` at the end

SteveBronder:
Also a quick thing to take note of. We need to make sure the vector that goes into serializer is fully initialized aka for our case it has the correct size and it's values are all initialized to zero

rybern:
This all makes sense. Nice looking code! I'll work on that output

rybern:
Here's now: https://gist.github.com/rybern/a74ba8bc87a7682217a20e5cbf3bb886

rybern:
Looks like it'll be easy to do this for write_array as well, if you want to handle the signatures again

SteveBronder:
Nice! Yeah if it's easy peasy then lets do it. Only thing we need to change with serializer is to do

```cpp
    std::decay_t<VecVar> vars__(params_r__size(), 0);
```

instead of 

```cpp
    std::decay_t<VecVar> vars__;
    vars__.reserve(params_r__.size());
```

I can write the scheme for write_array as well. I'll put it on a separate branch so you can work on this one and merge it in when it's done.

For write array we let's use 

```stan
data {
  int<lower=0> N;
  int M;
}
parameters {
  matrix<lower=0>[N, N] x1;
  matrix<lower=0, upper = 1>[N, N] x2;
  matrix[N, N] x3;
  matrix<lower=0, upper = 1>[N, N] x4[M];
}
transformed parameters {
  matrix<lower=0>[N, N] x1_tp = multiply(x1, x1);
  matrix<lower=0, upper = 1>[N, N] x2_tp = multiply(x2 x2);
  matrix[N, N] x3_tp = multiply(x3, x2)
  matrix<lower=0, upper = 1>[N, N] x4_tp[M];
}
```

Since write array works on both parameters and transformed parameters. For parameters we'll do the same serialize scheme but for td we need to do the ops on the transformed parameters and then make the call to serial

rybern:
Okay, changed decay_t. You've welcome to use this branch if you want, you can just comment this paragraph [here](https://github.com/stan-dev/stanc3/blob/acf5c8e63820afd2ca4983e89b49e82b59151a23/src/stan_math_backend/Transform_Mir.ml#L386) and uncomment the paragraph above it to switch over to the serializer version in write_arrays

SteveBronder:
Got everything look nice, but I was wrong about being able to inline `deserialize()` calls into serialize, goofed and totally forgot you can have parmaeters with constraints depending on other parameters so this fails

```cpp
parameters {
  real p_real;
  real<lower=p_real> p_upper;
}
```
But I think that's in then we are good!

rybern:
Good call, I put back the assignments and did a little cleanup. Could probably stand to do some more cleanup when we're sure things are settled.

Side note, maybe I'm just tired - how does the following work?
```
parameters {
  real<lower=5> p_real;
  real<lower=p_real> p_upper;
}
```
Won't we be passing the unconstrained version of p_real as the constraint argument of p_upper, but we should be passing the constrained version?

SteveBronder:
Ah shoot, okay looking at this I think what we need is something like the below

https://gist.github.com/SteveBronder/a575c5e1089c8a40ea9a3c4df4480ca9#file-ex_new-hpp

In my brain I thought free'ing variables would be carried forward to other parameters constraints but that doesn't make sense lol. That gist has how we currently do it, how this one does it, and what we should be doing. I think I can just add the free methods to the serializer which I should be able to do today and then this should work just fine. Essentially it's just doing the `read()` on the input variable, then for the serializer calling the free equivalent

```
        p_real = in__.read<local_scalar_t__>();
        out__.write_free_lb(5, p_real);
```

Then the constraints are correct for parameters depending on other parameters.

@seantalts can you check out the above link to make sure that makes sense?

seantalts:
Sure thing - the `ex_curr` and `ex_new` examples look like they're doing the right thing at a high level to me (I didn't look into implementations).

rybern:
@SteveBronder Output should match your new example now, let me know what you think of the tests

test/integration/good/code-gen/cpp.expected
```diff
@@ -8229,61 +8229,60 @@ class mother_model final : public model_base_crtp<mother_model> {
...
       
-      p_lower = in__.read<local_scalar_t__>(p_upper);
+      p_lower = in__.read<local_scalar_t__>();
       out__.write_free_ub(p_upper, p_lower);
       std::vector<local_scalar_t__> offset_multiplier;
       offset_multiplier = std::vector<local_scalar_t__>(5, DUMMY_VAR__); 
       
-      assign(offset_multiplier,
-        in__.read<std::vector<local_scalar_t__>>(1, 2, 5),
+      assign(offset_multiplier, in__.read<std::vector<local_scalar_t__>>(),
```
SteveBronder: For this we still need to have the `1, 2, 5` dims

test/integration/good/code-gen/cpp.expected
```diff
@@ -1078,14 +1078,14 @@ class eight_schools_ncp_model final : public model_base_crtp<eight_schools_ncp_m
       local_scalar_t__ tau;
       tau = DUMMY_VAR__; 
       
-      tau = in__.read<local_scalar_t__>(0);
+      tau = in__.read<local_scalar_t__>();
```
SteveBronder: Though these ones are correct and should be empty

SteveBronder:
Noice! I'm waiting on someone to review that Stan PR for free in the serializer then this should be ready

test/integration/good/code-gen/cpp.expected
rybern: d'oh, I was looking too hard at the example

rybern:
I'll do some cleanup as well when we get the end-to-end tests passing

rybern:
@SteveBronder any reason not to use deserializer for reads in the constructor as well? Maybe troublesome to change the signature for the `params_r__` argument? The bonus motivation is that we currently have two versions of `FnDataRead`

rybern:
Would it be super easy to add a corresponding interface for `check` functions to the backend?

SteveBronder:
> @SteveBronder any reason not to use deserializer for reads in the constructor as well? Maybe troublesome to change the signature for the params_r__ argument? The bonus motivation is that we currently have two versions of FnDataRead

I think we could have a seperate signature for the constructor since right now it takes in a `var_context` which is virtual so we can't do our template trick we did here

rybern:
I did some cleanup as promised. The main change I can finally make is that `Internal_fun` variants can take structured arguments including expressions. For example, FnWriteParam takes an optional unconstrain `Transformation.t`, a list of dimension expressions, and an expression to write ([code here](https://github.com/stan-dev/stanc3/blob/100f3533972190c153929e6e80c27637bf1c5858/src/middle/Internal_fun.ml#L13)). I think this is better for a few reasons:
 * Structured APIs are easier to read and write without errors
 * We avoid the packing and re-parsing pattern we had with some of the `Internal_fun`s (like packing a dummy "offset_multiply" string expression inside a FnCheck call)
 * The string translation and argument packing for e.g. FnCheck in `Ast_to_Mir` was all specific to the Stan Math backend, it makes more sense for those details to be in codegen

So the current state of checks, constrains and unconstrains:
 * FnChecks holding `Transformation.t`s are inserted in `Ast_to_Mir` (often inside big loops)
 * FnConstrains and FnUnconstrains are gone, we instead use FnRead{Param,Data} and FnWriteParam that take `Transformation.t`s
 * Reads and writes are inserted in Transform_mir (I'm not as against keeping them in Ast_to_Mir with the new Internal_fun types, but I think they make more sense in Transform_mir)
 * The string translation (e.g. OffsetMultiply -> "offset_multiply") and argument packing (e.g. [arg1, arg2, lp__, dim1, dim2] and Offset x -> OffsetMultiply x 1) is now handled in Stan_math_backend instead of Ast_to_Mir
 * There are two versions of data reads: FnReadData and FnReadDataSerializer, because we don't use the serializer interface in prepare_data

rybern:
Whoops reopenning

SteveBronder:
Nice! @rok-cesnovar  can you lmk when the Jenkins stuff is good to go and I'll fix these conflicts

rok-cesnovar:
For stanc3 side its all good.

SteveBronder:
Alright @rybern I think I got this all working. Just had to remove the dims from the `serializer.write()` function since we are passing in an object which knows it's dimensions already

SteveBronder:
@rok-cesnovar I think their may be something weird still going on, unless we've also blown up the size of stan binary files. In one of the tests I'm seeing

https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3/detail/PR-872/31/pipeline

```
Iteration:  1 / 20 [  5%]  (Warmup)
Iteration: 11 / 20 [ 55%]  (Sampling)
Exception: no more storage available to write (in '../example-models/bugs_examples/vol3/data_cloning/seeds.stan', line 40, column 4 to column 27)
```

rybern:
Done I think @SteveBronder

SteveBronder:
Oh whoops actually the above error is in this PR somewhere @rok-cesnovar ignore the above

SteveBronder:
Okay I have an idea of what's going on here. The above error `no more storage available to write ` (which extremely needs to change to something more meaningful) is telling us that the serializer is out of space so it's throwing an error. That shouldn't happen in `transform_inits` and the fix I just pushed should handle that, but I think I need to think of a way for that not to happen in `write_array`. Is it possible to get the size of everything in transformed parameters and generated quantities? Because for serializer I want to allocate space in the vector it's passed equal to 

```
num_params + (emit_transformed_parameters * num_transformed) + (emit_generated_quantities * num_gen_quantities)
```

So in the non-impl `write_array()` `vars_vec` becomes 

```
std::vector<double> vars_vec(num_params + (emit_transformed_parameters * num_transformed) + (emit_generated_quantities * num_gen_quantities));
```

SteveBronder:
Oh I also need the number of transformed params for transform_init as well

rybern:
That sounds possible to me, I'll have a look

Out of curiousity, why wasn't this needed before?

SteveBronder:
> Out of curiousity, why wasn't this needed before?

Before we didn't need to because we would pass in a vector which would then call `push_back()` which will dynamically allocate more memory as needed. Now we just pass in a vector to the impl we expect to be of the right size

SteveBronder:
For transformed params, could we just sum the dims of the decls in `log_prob`, then subtract the number of parameter values?

rybern:
Sorry for the delay. So to clarify, we need the total number of parameters (which is the product of the unconstrained dimensions) in transformed parameters and in generated quantities?

I thiiink it's okay to just generate that expression from the dimensions in `output_vars`. That's what `get_dims` does. I'll have a quick go at that.

rybern:
@SteveBronder lmk if that's the right direction

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -753,8 +773,11 @@ let pp_overloads ppf () =
                             const bool emit_transformed_parameters = true,
                             const bool emit_generated_quantities = true,
                             std::ostream* pstream = nullptr) const {
-      std::vector<double> vars_vec;
-      vars_vec.reserve(vars.size());
+      int num_transformed = %a;
+      int num_gen_quantities = %a;
+      std::vector<double> vars_vec(num_params
```
SteveBronder, line 778:
```suggestion
      std::vector<double> vars_vec(num_params_r__ 
```

SteveBronder:
@rybern fix was actually v simple we just forgot to do the size thing for the `write_array` that accepts vectors _and_ the `write_array()` that accepts Eigen matrices

SteveBronder:
Ack there must be an off by one error in one of these I'll have a look

SteveBronder:
Okay I've spotted a difference, but I'm not sure who's right here. For the [`magnesium`](https://github.com/stan-dev/example-models/tree/6f0ed2269f0b8a39022ac1202e7622fd644904fe/bugs_examples/vol1/magnesium) model, and any model that uses arrays of arrays, we are doing a row major order access pattern in the current `transform_inits` but a column major order access pattern in this PR.

[here](https://gist.github.com/SteveBronder/2b8e16755c32cbe5f0a4dd646f8495ee#file-magnesium_new-hpp-L553) in the new C++ we are doing 

```cpp
      assign(theta,
        in__.read<std::vector<std::vector<local_scalar_t__>>>(N_priors,
          N_studies), "assigning variable theta");
```

Which is going to fill things in like 

```
theta[1, 1]
theta[1, 2]
theta[1, 3]
...
theta[2, 1]
theta[2, 2]
theta[2, 3]
```

But in the [current code](https://gist.github.com/SteveBronder/2b8e16755c32cbe5f0a4dd646f8495ee#file-magnesium_orig-hpp-L576) we are doing 

```cpp
      {
        std::vector<local_scalar_t__> theta_flat__;
        current_statement__ = 2;
        theta_flat__ = context__.vals_r("theta");
        current_statement__ = 2;
        pos__ = 1;
        current_statement__ = 2;
        for (int sym1__ = 1; sym1__ <= N_studies; ++sym1__) {
          current_statement__ = 2;
          for (int sym2__ = 1; sym2__ <= N_priors; ++sym2__) {
            current_statement__ = 2;
            assign(theta, theta_flat__[(pos__ - 1)],
              "assigning variable theta", index_uni(sym2__),
                                            index_uni(sym1__));
            current_statement__ = 2;
            pos__ = (pos__ + 1);
          }
        }
      }
```

Which if I'm reading the `sym2__` and `sym1__` correctly is actually going to go

```
theta[1, 1]
theta[2, 1]
theta[3, 1]
...
theta[1, 2]
theta[2, 2]
theta[3, 2]
```

They are both assigning from a flattened vector, but the current one is assigning those values by rows and the new one by columns. Does the init file come in column or row major order? It's not clear to me whether this is a bug or intended.

The `magnesium` init file is not terribly helpful to sort this out as all the values are the same

https://github.com/stan-dev/example-models/blob/6f0ed2269f0b8a39022ac1202e7622fd644904fe/bugs_examples/vol1/magnesium/magnesium.init.R

What's also confusing is that when we go to write to the output in the current code we then switch back to writing things in column major order (aka going from `N_studies` as the outer loop and `N_prior` as the inner loop to `N_prior` as the outer loop and `N_studies` as the inner loop.)

```cpp
      for (int sym1__ = 1; sym1__ <= N_priors; ++sym1__) {
        for (int sym2__ = 1; sym2__ <= N_studies; ++sym2__) {
          vars__.emplace_back(theta[(sym1__ - 1)][(sym2__ - 1)]);
        }
      }
```

Tagging @bob-carpenter, was this the behavior in the old compiler as well?

SteveBronder:
ftr I can fix this just fine in the new code to do what the old code does, but wanted to check that this is intended behavior

SteveBronder:
Actually reading this I think I totally misread things, I think the current behavior is correct since everything comes in column major order and it wants to write the columns first

SteveBronder:
Alright I think I know how to handle this in not too bad of a way but need to check if matrices are handled in the init file in the same way as well

SteveBronder:
Ack actually I tried to fix this and my brain hurt real bad. @rybern so the issue is that everything comes in as column major order, even arrays of arrays. The deserializer assumes the data it receives is coming in for arrays of arrays as if we serialized an `std::vector<std::vector<local_scalar>>` aka if this was a matrix we assume it came in row major order where each row is the inner vector. So in transform inits, the inits for like a `std::vector<std::vector<std::vector<local_scalar>>>` of size {2, 3, 2} come in like 

[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]

and we want to assign from the serialized array of arrays to the array of arrays like (as zero indexing)

```
0 -> (0, 0, 0)
1 -> (1, 0, 0)
2 -> (0, 1, 0)
3 -> (1, 1, 0)
4 -> (0, 2, 0)
5 -> (1, 2, 0)
6 -> (0, 0, 1)
7 -> (1, 0, 1)
8 -> (0, 1, 1)
9 -> (1, 1, 1)
10 -> (0, 2, 1)
11 -> (1, 2, 1)
```

aka ticking by the first index, then second index, then third index. That lets us to an output of arrays that looks like the below where the `{}` represent an array

```
{{{1, 7},
  {2, 8},
  {3, 9}}, 
 {{4, 10}, 
  {5, 11},
  {6, 12}}}
```


If we still have the code for writing these arrays as loops, then for an array like 

```stan
  real test_var_arr3[N_priors, N_studies, N_dim1];
```

Would it be fine with you to generate the loop still like

```cpp
      {
        for (int sym1__ = 1; sym1__ <= N_dim1; ++sym1__) {
          for (int sym2__ = 1; sym2__ <= N_studies; ++sym2__) {
            for (int sym3__ = 1; sym3__ <= N_priors; ++sym3__) {
              assign(test_var_arr3, in__.read<local_scalar_t>(),
                     "assigning variable test_var_arr3", index_uni(sym3__),
                     index_uni(sym2__), index_uni(sym1__));
            }
          }
        }
      }
```

We only need to loop over the array portion. So like for arrays of matrices like 

```stan
  matrix<lower=0>[N_priors, N_studies] test_mat_arr[N_dim1, N_dim2];
```

since matrices are column major order already then we would do

```cpp
      {
        for (int sym1__ = 1; sym1__ <= N_dim2; ++sym1__) {
          current_statement__ = 4;
          for (int sym2__ = 1; sym2__ <= N_dim1; ++sym2__) {
            current_statement__ = 4;
            assign(test_mat_arr, in__.read<Eigen::Matrix<local_scalar_t__, -1, -1>>(N_priors, N_studies),
                   "assigning variable test_mat_arr", index_uni(sym2__), index_uni(sym1__));
            current_statement__ = 4;
            pos__ = (pos__ + 1);
          }
        }
      }
```

Does that work for you?

rybern:
@SteveBronder yep I think that should be fine. We have to write loops anyway for tuples

rybern:
@SteveBronder Lmk if that looks right

SteveBronder:
Ack so looking at [`seeds.stan`](https://github.com/stan-dev/example-models/blob/6f0ed2269f0b8a39022ac1202e7622fd644904fe/bugs_examples/vol3/data_cloning/seeds.stan) ([hpp files](https://gist.github.com/SteveBronder/94ca437d5765bb1572bc5e39fb82b0e3)) we do also need to loop over the matrices and arrays in vectors in reverse order as well. I misread what the original code was doing. So like for `b` in `transform_inits` in the original it actually iterates over I and then K

```cpp
      {
        std::vector<local_scalar_t__> b_flat__;
        current_statement__ = 6;
        b_flat__ = context__.vals_r("b");
        current_statement__ = 6;
        pos__ = 1;
        current_statement__ = 6;
        for (int sym1__ = 1; sym1__ <= K; ++sym1__) {
          current_statement__ = 6;
          for (int sym2__ = 1; sym2__ <= I; ++sym2__) {
            current_statement__ = 6;
            assign(b, b_flat__[(pos__ - 1)],
              "assigning variable b", index_uni(sym2__), index_uni(sym1__));
            current_statement__ = 6;
            pos__ = (pos__ + 1);
          }
        }
      }
```

instead of doing I then K for these cases

```
      for (int sym1__ = 1; sym1__ <= I; ++sym1__) {
        assign(b, in__.read<Eigen::Matrix<local_scalar_t__, -1, 1>>(K),
          "assigning variable b", index_uni(sym1__));
      }
```

And I think we need to do the same thing in in `write_array`. Like looking at [`SIR`](https://github.com/stan-dev/stat_comp_benchmarks/blob/64b6597cd17de77e91dcfb1a9977758407017d20/benchmarks/sir/sir.stan) ([hpp files](https://gist.github.com/SteveBronder/f3faf5eb9b6b10be188848cd06c540f1) It looks like it uses the iterators starting from out to in here as well.

One alt, we could go back to just doing `out__.write(y);` for everything, but we add a template parameter to write like `RowMajor | ColMajor` and call it like `out__.write<RowMajor>(y);` which just tells it to write in whatever order we want. That would be pretty easy for me to post up (though has to wait till the 2.27 rc is finished to merge it)

SteveBronder:
I think it kind of makes sense to change the serializer because it doesn't seem to do the pattern we want

SteveBronder:
@rybern I finally got this all sorted out! Sadly we can't really use serializer like we do elsewhere because the data for transform inits comes in with a weird flattened array style for arrays. Essentially it flattens multi-dimensional matrices so that the outermost index is the one that changes the most, which causes us to have to do the reverse loop style.

But I think this is good! I'll give the ocaml one more read through and then I think we can merge!

https://github.com/stan-dev/stan/blob/develop/src/stan/io/dump.hpp#L63

SteveBronder (approved):

Issue #743, opened 2020-11-19
Pull out function suffixes and related utilities


rybern:
Stan has a pretty big zoo of function suffixes. Lots have subtly different meanings, some are deprecated, some get transformed away at various stages of compilation. Currently, these suffixes are spread out and duplicated throughout the code as hardcoded strings and `String.is_suffix` calls. For some examples, see [Canonicalize.ml](https://github.com/stan-dev/stanc3/blob/master/src/frontend/Canonicalize.ml), or [Ast_to_Mir.ml](https://github.com/stan-dev/stanc3/blob/fd30e88be447d4988521c3c270471dd462966a99/src/frontend/Ast_to_Mir.ml#L115), or [Semantic_check.ml](https://github.com/stan-dev/stanc3/blob/fd30e88be447d4988521c3c270471dd462966a99/src/frontend/Semantic_check.ml#L83), or [code gen](https://github.com/stan-dev/stanc3/blob/fd30e88be447d4988521c3c270471dd462966a99/src/stan_math_backend/Expression_gen.ml#L107), [etc etc.](https://github.com/stan-dev/stanc3/search?q=%22_lpmf%22).

It would be great to encapsulate knowledge of function suffixes into one place and provide a nice interface.

adamhaber:
Now that #726 is almost done, I can try to take this. Can you elaborate a bit on what you have in mind? Would creating a Function_suffix module with different `is_foo`, `to_foo` etc make sense?  How should this relate to `type fnkind` from Stan_math_signatures?

nhuurre:
Distribution suffix tools are already implemented in [Utils.ml](https://github.com/stan-dev/stanc3/blob/fd30e88be447d4988521c3c270471dd462966a99/src/middle/Utils.ml). These should be used more consistently instead of being re-implemented in other modules, such as `is_user_dist` function in `Expression_gen` which essentially just duplicates `Utils.is_distribution_name`.

You could also survey all uses of `_lp`, `_rng` and `_lcdf` and expand `Utils` to handle those too.

adamhaber:
@rybern is this still relevant?

rok-cesnovar:
I think this might all have been addressed by #889? Will let @nhuurre close.

nhuurre:
Yes, most of the `String.is_suffix` calls have been removed. The remaining ones are in deprecation analysis and other frontend modules where they cannot be standardized.
Also a few unused functions were left behind, e.g. `Expression_gen.is_stan_math`.

Issue #324, opened 2019-10-05
Represent distributions in MIR, refactor backend


seantalts:
The backend is getting pretty complicated and in many places we're still relying on function name string suffices to determine things like which arguments need to be passed to which, and the logic is getting kinda crazy. I'd like to see some separate variants made that line up more closely with the platonic ideal of distributions (with their various grouped functions like _rng, _lpdf, _ccdf, etc).

bob-carpenter:
This'd be good as the argument signatures should also be tied together.

Issue #215, opened 2019-06-28
Make partial evaluator deal with unnormalized distributions


VMatthijs:
Currently the partial evaluator only does algebraic rewrites on normalized distributions. We should make sure the same rewrites also work on unnormalized ones.

nhuurre:
Now that #889 has been merged normalized and unnormalized distributions are handled in a uniform way.

PR #889, opened 2021-04-26
Add function suffixes to MIR

nhuurre:
See issues  #324 and #743 for background. This adds `rng`, `lp` and `lpmdf/lupmdf` suffixes to the MIR function types which simplifies the backend codegen a bit (and should also solve issue #215).  No special handling for `lcdf` suffixes, though.

## Copyright and Licensing
Copyright holder: Niko Huurre


nhuurre:
The only change in the generated code is the one suggested [here](https://github.com/stan-dev/stanc3/pull/562#issuecomment-639106513). It's closer to what Stanc2 used to do.

src/middle/Fun_kind.ml
```diff
@@ -1,11 +1,29 @@
 open Core_kernel
 
+type 'a suffix = FnPure | FnRng | FnLpdf of 'a | FnTarget
```
SteveBronder, line 3: idk if `FnPure` is the right word for this, every function for reverse mode has the side effect of adding a callback to the global callback stack for calculating the adjoints when calling `grad()`. Mby just `FnDefault`?

https://gcc.gnu.org/onlinedocs/gcc-4.7.2/gcc/Function-Attributes.html

> Many functions have no effects except the return value and their return value depends only on the parameters and/or global variables. Such a function can be subject to common subexpression elimination and loop optimization just as an arithmetic operator would be. These functions should be declared with the attribute pure. For example,

>          int square (int) __attribute__ ((pure));
     
> says that the hypothetical function square is safe to call fewer times than the program says.

> Some of common examples of pure functions are strlen or memcmp. Interesting non-pure functions are functions with infinite loops or those depending on volatile memory or other system resource, that may change between two consecutive calls (such as feof in a multithreading environment). 

I think we reverse mode does not follow the line here

> except the return value and their return value depends only on the parameters and/or global variables.

Since the inputs will also depend on global variables (the global call stack we put the gradient calculations on)

nhuurre: They're pure functions in that the autodiff callback stack is not a user-visible side effect. In particular, common subexpression elimination is a legal optimization.
That being said, I'm not a huge fan of either `FnPure` or `FnDefault`. I considered a plain `Fn` but it feels _too_ short.

rybern: Maybe `FnBasic` or `FnPlain`?

src/middle/Fun_kind.ml
```diff
@@ -1,11 +1,29 @@
...
+[@@deriving compare, sexp, hash, map]
+
 type t =
-  | StanLib of string
+  | StanLib of string * bool suffix
   | CompilerInternal of Internal_fun.t
-  | UserDefined of string
+  | UserDefined of string * bool suffix
```
rybern, line 9: This looks awesome. Maybe you could add a comment saying what the bool means (normalization it looks like), since it doesn't have a name? A little comment or a link to the docs at the definition of the `suffix` variants might be helpful too.

src/middle/Fun_kind.ml
SteveBronder: `FnPlain`? tbh `Fn` is a bit short but fine by me.

SteveBronder: ^lol I think we commented within a few minutes of each other

nhuurre: I'm fairly sure you both "came up with the same idea" only because I literally said "a plain `Fn`" but sure `FnPlain` is as good as anything else I can think of.

src/middle/Fun_kind.ml
nhuurre: I renamed the type parameter on `suffix` to `'propto`. Does it need more docs?
The only reason it's even parametrized is because I want to use the same suffixes for function definitions and those don't make a distinction between normalized and unnormalized lpdfs.

rybern: I do think a sentence or two would improve clarity, but this is good enough for me

src/analysis_and_optimization/Pedantic_analysis.ml
```diff
@@ -261,16 +263,14 @@ let compiletime_value_of_expr
...
-      | Some dname ->
-          let params = parameter_set mir in
-          let data = data_set mir in
-          let args =
-            List.map ~f:(compiletime_value_of_expr params data) arg_exprs
-          in
-          Some {name= dname; loc= expr.meta.loc; args}
-      | _ -> None )
+    | Expr.Fixed.Pattern.FunApp (StanLib (fname, FnLpdf true), arg_exprs) ->
+        let fname = chop_dist_name fname |> Option.value_exn in
```
rybern, line 267: It's interesting that we use value_exn now because the representation is redundant. I wonder if we should cut the suffix off of the string when we parse the `suffix` type? What do you think?

rybern: I guess that would probably be too complicated to translate back and forth for e.g. codegen and signature lookup. Hmm

test/integration/good/code-gen/cpp.expected
```diff
@@ -194,6 +194,8 @@ class _8_schools_ncp_model final : public model_base_crtp<_8_schools_ncp_model>
     using local_scalar_t__ = double;
     vars__.resize(0);
     stan::io::deserializer<local_scalar_t__> in__(params_r__, params_i__);
+    static constexpr bool propto__ = true;
+    (void) propto__;
```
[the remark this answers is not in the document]
nhuurre: Good question. The original plan was no but the first CI run failed due to an accidental `_lp` (aka `FnTarget`) function codegen change. (These aren't in any `cpp.expected` test but luckily Jenkins tries to compile all example models.)

The change is this: at C++ level `_lp` functions, like `_lpdf`s, require a propto template argument. Until now stanc3 has always generated `<false>` for that argument, in contrast with stanc2 that used to always have `<true>`. I thought we [had already changed that](https://github.com/stan-dev/stanc3/pull/562#issuecomment-639106513) and (without reading the previous code) I made this branch to always generate `<propto__>`. I could have changed that back after CI caught it but I like this better.

The two lines added here ensure that `<propto__>` is available the `write_array` method. That's required for an `_lp` function call in the transformed parameters block.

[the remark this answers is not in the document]
rybern: Gotcha, makes sense

rybern (approved):
Looks good to me

PR #897, opened 2021-05-09
Clear read/write event for global matrix_cls

### LINKED ISSUE #857: Clear read/write event  for the global matrix_cls
We should add a `matrix.wait_for_read_write_events()` for all data matrix_cls in the log_prob_impl. 

Right now, a kernel call to a data matrix adds a read event to the read event vector: https://github.com/stan-dev/math/blob/develop/stan/math/opencl/matrix_cl.hpp#L50
These are never cleared between log_prob_impl calls which means the events add on, causing a rising memory consumption over time. 

Adding `matrix.wait_for_read_write_events()` for all data matrices in log_prob_impl will clear these vectors.

rok-cesnovar:
Fixes #857 

## Release notes

Added clearing of read/write events for global matrix_cls.

## Copyright and Licensing

### DIFF
--- a/src/middle/Internal_fun.ml
+++ b/src/middle/Internal_fun.ml
@@ -21,6 +21,7 @@ type t =
   | FnResizeToMatch
   | FnNaN
   | FnDeepCopy
+  | FnReadWriteEventsOpenCL of string
 [@@deriving sexp, hash, compare]
 
 let to_string x = Sexp.to_string (sexp_of_t x) ^ "__"
--- a/src/stan_math_backend/Statement_gen.ml
+++ b/src/stan_math_backend/Statement_gen.ml
@@ -224,6 +224,7 @@ let math_fn_translations = function
   | FnValidateSize -> Some ("validate_non_negative_index", [])
   | FnValidateSizeSimplex -> Some ("validate_positive_index", [])
   | FnValidateSizeUnitVector -> Some ("validate_unit_vector_index", [])
+  | FnReadWriteEventsOpenCL x -> Some (x ^ ".wait_for_read_write_events", [])
   | _ -> None
 
 let trans_math_fn fname =
--- a/src/stan_math_backend/Transform_Mir.ml
+++ b/src/stan_math_backend/Transform_Mir.ml
@@ -627,6 +627,14 @@ let trans_prog (p : Program.Typed.t) =
          [log_prob; generate_quantities])
     |> String.Set.to_list
   in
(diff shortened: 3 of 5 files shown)


SteveBronder (approved):
lgtm! The only thing I'd add here though is that when we have multi-chain programs in one process I think this won't play as nicely with the async and sending things over multiple threads.

Personally I'd like to add a boolean template parameter to `matrix_cl<>` like 

```
template <typename, bool ReadOnly = true typename = void>
class matrix_cl;
```

Then adding read and write events becomes 

```cpp
  inline void add_read_event(cl::Event new_event) const {
   if (ReadOnly) {
     this->read_events_.push_back(new_event);
   }
  }
```

And for copying to and from the device we'd force it to wait till it was finished. Since for async the only time we care about when a read happens is when we need to make a write wait. But for the data here we only write once and then after that we don't have to worry about writing again.

t4c1:
I agree - this should be fine. The only downside is that one thread could be waiting on events that represent a kernel from another thread. Although as far as I know a GPU usually only runs one kernel at once anyway so this is only a minor downside.

Anyway regarding multiple chains in single program - how are data races avoided when adding or removing events on matrices from multiple threads?

rok-cesnovar:
> how are data races avoided when adding or removing events on matrices from multiple threads?

The only matrices shared between the threads are these data matrices. The rest of them are local to the log_prob function.

SteveBronder:
Also ftr I'm v open to other approaches to solving this. If yinz both think the things here won't have the race conditions then I'm fine with dropping the idea of the template thing.

t4c1:
I am pretty sure this is a race condition, but I am not sure if it can do any harm in this particular use case (Even if OpenCL is thread safe, `std::vector` is not. Not adding an event does not hurt us in any way, but if two events are added at the same time it might result in uninitialized event. OpenCL might be fine with that, but if one thread does realocation the other one could segfault. Although all these race conditions are extremely unlikely to actually happen.

> But for a longer term solution what do you both think of the above with the template parameter?

It seems a lot of work, because than all functions need to support both template argument values and we need to do conversions between them without copying ...

I would prefer to make all `matrix_cl` objects thread local (sharing only the underlying `cl::Buffer` between threads).

wds15:
I am out of context… still I see thread local and openCL… thread local variables are usually a slow thing and are avoided. As I recall, the TBB has some integration with openCL / maybe that can be explored as an alternative.

PR #898, opened 2021-05-11
[WIP] Adding a var matrix type for the model

SteveBronder:
## Summary

This is a WIP trying to allow the model to utilize the `var_value<Eigen::Matrix>` type. It's about halfway there and most of what I need help on is design decisions around where I should be doing some parsing and transformations.

## Background

For a full background on the difference between `var_value<Eigen::Matrix>` and `Eigen::Matrix<var>` see the design doc [here](https://github.com/stan-dev/design-docs/blob/master/designs/0005-static-matrices.md). Essentially this is a pretty classic optimization to swap between an Array of Structs (AoS) and Struct of Arrays (SoA). The speed improvements come from the fact that for matrix operations we tend to want whole chunks of the `value` and `adjoint` of an autodiff type. So having a struct with two arrays let's us grab those value and adjoint arrays in big chunks to enable better vectorization and promote SIMD instruction usage.

To do this in stanc I think there are two things 

1. Having `Decl`s in log_prob know they can allocate a `var_value<Eigen::Matrix>` (SoA) instead of an `Eigen::Matrix<var>` (AoS).
2. There are some functions that we can't deduce whether we should return a `var_value<Eigen::Matrix>` (SoA) instead of an `Eigen::Matrix<var>` (AoS) such as `rep_vector(parameter A, 10)`. For these cases in the math library we made these functions overloadable by the return type, so the stanc compiler can generate something like

```cpp
var xx = 5.4;
var_value<Eigen::Matrix> A = rep_vector<var_value<Eigen::Matrix>>(xx, 10);
```
(the brackets <> after the function name tell the C++ compiler what we want the return type to be)

But in order to do that we need `StanLib` functions to also know that they return back a `var_value<Eigen::Matrix>` (SoA) instead of an `Eigen::Matrix<var>` (AoS).

The requirements to be able to swap a `Eigen::Matrix<var>` (AoS) for a `var_value<Eigen::Matrix>` (SoA) in the `Decl` is

1.  Every function that is called using the object in the `Decl` must have a signature in Stan math that accepts a `var_value<Matrix>` type.
   - The rule we have setup in Stan math is that if a function accepts a `var_value<Matrix>` in it's signature it supports it for all arguments in the signature. This is nice as we don't have to check back and forth between things like binary functions for both of the argument types. If we fail on one side of the signature we will fail in all the other parts.
   - This also cuts out using these functions in user defined functions (and thereby the ode solver, integrators, solvers, reduce_sum, etc.). This is just a restriction for the first impl but later impls can loosen this by investigating user defined functions to make sure all the stan math functions they are calling have valid signatures.

2. The object must not be accessed by single cells inside of a for loop.
  - This is an optimization catch. Accessing a struct of arrays by single cells causes a lot of overhead as the single cell value has to be actually constructed and kept track of by peeling off the indexed values in the area to make a `var` and then pushing a callback onto the callback stack to keep track of updating adjoints. So in this case it's actually better to have an array of structs.

So what I have so far is

1. Added a `mem_pattern` type to `Common.Helpers` that just defines the two memory patterns like 

```ocaml
type mem_pattern = AoS | SoA
```

These names might be too short and I'd be happy to change them to something longer like `ArrayofStruct` / `StructofArray`

2. Changed the hash table in `Stan_math_signature` so it's values are now a list of 

```ocaml
UnsizedType.returntype * (UnsizedType.autodifftype * UnsizedType.t) list * Common.Helpers.mem_pattern
```

When writing the math signatures I'm doing somthing a little odd. If a function can accept a struct of arrays I tag is just with `SoA`. Since we just use this table to check if a signature is valid, and since every signature that can accept an `SoA` accepts an `AoS` so when checking these in the below steps this value is only needed to verify whether a signature can accept an `SoA`.

An alternative here is to have a separate set of functions in the hash table that pass for the struct of arrays scheme, but since we seem to just use this hash table for validation I think it is okay to use the ordering thing I have above.

3. Changed `SizedType` so that it's constructor takes in a `Common.Helpers.mem_pattern`.

4. Changed `StanLib` function kinds so that they take in a `Common.Helpers.mem_pattern`, which will be used for the above to deduce the return type for certain functions that the C++ compiler can't deduce.

At this point there are two ways forward and I'm not really sure which one is better. @rybern has suggested a method that manipulates the AST and does the deduction during the symantic checking and conversion from AST-> Typed AST. The other scheme is to create a new optimization pass that transforms the MIR.

Either way, each scheme assumes we are starting with a `var_value<Matrix>` (SoA), then because of the rule in the math library above we are able to iterate over each of the `matrix/vector/row_vector` types defined in parameters and transformed paramaters one by one, so either scheme we pick will be `O(N)`

### AST -> Typed Ast Modification

In this process we first tag each `SizedType` in the AST `Decl` as `SoA` so that by default we assume we are working with a `var_value<Eigen::Matrix>`. 

```ocaml
sized_basic_type:
  | INT
    { grammar_logger "INT_var_type" ; (SizedType.SInt, Identity) }
  | REAL
    { grammar_logger "REAL_var_type" ; (SizedType.SReal, Identity) }
  | VECTOR LBRACK e=expression RBRACK
    { grammar_logger "VECTOR_var_type" ; (SizedType.SVector (SoA, e), Identity) }
  | ROWVECTOR LBRACK e=expression RBRACK
    { grammar_logger "ROWVECTOR_var_type" ; (SizedType.SRowVector (SoA, e) , Identity) }
  | MATRIX LBRACK e1=expression COMMA e2=expression RBRACK
    { grammar_logger "MATRIX_var_type" ; (SizedType.SMatrix (SoA, e1, e2), Identity) }
```

Then during the semantic checking in `type_ast_or_exit` we check for a new semantic error `VarMatrixFailure` which will check for the two failure conditions above. If the semantic checking fails with a `VarMatrixFailure` we rewrite the AST, swapping out the `SoA` above and set the matrix to an `Eigen::Matrix<var>` (`AoS`). That would look in pseudocode like

```ocaml
let rec type_ast_or_exit ast =
  try
    match Semantic_check.semantic_check_program ast with
    | Result.Ok prog -> prog
    | Result.Error (VarMatrixFailure (matrix_name, location) :: _) ->
      let new_ast = remove_var_matrix_from_ast ast matrix_name, location
      type_ast_or_exit new_ast
    | Result.Error (error :: _) ->
        Errors.pp_semantic_error Fmt.stderr error ;
        exit 1
    | Result.Error [] ->
        Printf.eprintf
          "Semantic check failed but reported no errors. This should never \
           happen." ;
        exit 1
  with Errors.SemanticError err ->
    Errors.pp_semantic_error Fmt.stderr err ;
    exit 1
```

This is kind of nice as for the case below where `rep_vector` is used for an vector that then is used in a non-compatiable function.

```stan
parameters {
  real a;
}
transformed parameters {
  vector[N] X = rep_vector(a, 10);
  vector[N] Y = not_a_var_func(X);
}
```

Here we would fail at `not_var_var_func`, rewrite the AST to make `X` an `Eigen::Matrix<var>` (`AoS`), and then everything is cool and good. 

But the downside to this is that it's quite different from other manipulations we've done in the past which tend to happen in the MIR. This also stops us from doing optimization tricks like rolling up for loops to vectorized functions and then letting those functions use a `var_value<Eigen::Matrix>`.

### Optimization Step (MIR Transformation)

This will be a bit more involved, but what we want to do here is iterate over the list of MIR statements looking for each `Decl` Then for each `Decl` we search the rest of the MIR for functions where any of the inputs matches up with the name of the `Decl` and any for loops where the `Decl` object is single cell accessed. If we fail then we have to go up and change the `Decl` so that it is an `Eigen::Matrix<var>` (`AoS`) and then rewrite the rest of the MIR transforming `StanLib`'s so that they use an `AoS` instead of an `SoA`. 

This will be a bit more expensive and involved, but this follows with how we would normally do things and it also allows for us to utilize optimzations (I think writing this will make it exponentially easier to figure out how to write the for loop to vectorized function optimization step.)

So if anyone has comments on which of the two approaches are best or if anything smells really weird/funky here I would love to hear it. Also does anyone have papers / books I can lookup on the different schemes in the optimization steps? There's some stuff in there like the `flowgraph` and that look like they would be very useful for some parts of this.

## Copyright and Licensing

### DIFF
--- a/src/analysis_and_optimization/Mem_pattern.ml
+++ b/src/analysis_and_optimization/Mem_pattern.ml
@@ -328,6 +328,41 @@ and query_initial_demotable_funs (in_loop : bool) (kind : Fun_kind.t)
   | CompilerInternal (_ : Internal_fun.t) -> Set.Poly.empty
   | UserDefined ((_ : string), (_ : bool Fun_kind.suffix)) -> all_eigen_names
 
+let find_args Expr.Fixed.({meta= Expr.Typed.Meta.({type_; adlevel; _}); _}) =
+  (adlevel, type_)
+
+let is_fun_soa_supported name exprs =
+  let fun_args = List.map ~f:find_args exprs in
+  Stan_math_signatures.query_stan_math_mem_pattern_support name fun_args
+
+let rec query_bad_assign2
+    Expr.Fixed.({pattern; meta= Expr.Typed.Meta.({adlevel; _})}) : bool =
+  if adlevel = UnsizedType.DataOnly then true
+  else
+    let query_expr = query_bad_assign2 in
+    match pattern with
+    | FunApp (kind, (exprs : Expr.Typed.Meta.t Expr.Fixed.t list)) ->
+        query_bad_assign_fun2 kind exprs
+    | Indexed (expr, (_ : Typed.Meta.t Fixed.t Index.t sexp_list)) ->
+        query_expr expr
+    | Var (_ : string) | Lit ((_ : Expr.Fixed.Pattern.litType), (_ : string))
+      ->
+        true
+    | TernaryIf _ -> false
+    (*I think we can just return true for this?*)
+    | EAnd (lhs, rhs) | EOr (lhs, rhs) -> query_expr lhs && query_expr rhs
+
+and query_bad_assign_fun2 (kind : Fun_kind.t)
+    (exprs : Typed.Meta.t Expr.Fixed.t list) : bool =
+  match kind with
+  | CompilerInternal (Internal_fun.FnMakeArray | FnMakeRowVec) -> false
+  | UserDefined ((_ : string), (_ : bool Fun_kind.suffix)) -> false
+  | CompilerInternal (_ : Internal_fun.t) -> true
+  | Fun_kind.StanLib (name, (_ : bool Fun_kind.suffix), _) -> (
+    match name with
+    | "check_matching_dims" -> true
+    | _ -> is_fun_soa_supported name exprs )
+
 let rec query_bad_assign
     Expr.Fixed.({pattern; meta= Expr.Typed.Meta.({adlevel; _})}) : bool =
   if adlevel = UnsizedType.DataOnly then false
@@ -350,13 +385,6 @@ let rec query_bad_assign
     | TernaryIf (predicate, texpr, fexpr) ->
         query_expr predicate || query_expr texpr || query_expr fexpr
     | EAnd (lhs, rhs) | EOr (lhs, rhs) -> query_expr lhs || query_expr rhs
-
-and find_args Expr.Fixed.({meta= Expr.Typed.Meta.({type_; adlevel; _}); _}) =
-  (adlevel, type_)
-
-and is_fun_soa_supported name exprs =
-  let fun_args = List.map ~f:find_args exprs in
-  Stan_math_signatures.query_stan_math_mem_pattern_support name fun_args
 
 and query_bad_assign_fun (kind : Fun_kind.t)
     (exprs : Typed.Meta.t Expr.Fixed.t list) : bool =
@@ -430,7 +458,7 @@ let rec query_initial_demotable_stmt (in_loop : bool)
           ( UnsizedType.contains_eigen_type type_
           , adlevel = UnsizedType.AutoDiffable )
         with
-        | true, true -> query_bad_assign rhs
+        | true, true -> query_bad_assign rhs || not (query_bad_assign2 rhs)
         | _ -> false
       in
       if Set.Poly.length check_rhs > 0 || check_bad_assign then


SteveBronder:
Also I made a gist [here](https://gist.github.com/SteveBronder/b15e65c2e1fa975615b93c5a3224902b) of all the functions in stan math that support `var<Matrix>`. This doesn't include any of the probability functions as we have all of those working and that would just clutter up the list a bit

SteveBronder:
And the other good news is that I think this scheme should also work with the OpenCL stuff! For the signatures table we'll make the `Common.Helpers.mem_pattern = AoS | SoA | OpenCL` or something like that. Then if the OpenCL flag is set by the user we do everything first for OpenCL and then for `SoA`.

rybern:
> Every function that is called using the object in the Decl must have a signature in Stan math that accepts a var_value<Matrix> type.

What about the case like:
```
matrix x;
f(g(x));
```
where `g` is an SoA function but `f` is not. IIUC, if `x` is SoA then `g(x)` is SoA and `f` fails, but this won't be caught by just looking for functions where `x` is used as an argument.

When is it okay to cast between SoA and AoS? Is it okay for e.g. an SoA matrix to assign to an AoS matrix? If not, that seems like another constraint

> The rule we have setup in Stan math is that if a function accepts a var_value<Matrix> in it's signature it supports it for all arguments in the signature.

Is it safe to assume that an SoA function returns an SoA structure when any of the arguments are SoA?

> @rybern has suggested a method that manipulates the AST and does the deduction during the symantic checking and coversion from AST-> MIR

This would be during semantic checking, so Untyped AST -> Typed AST rather than MIR

> So if anyone has comments on which of the two approaches are best or if anything smells really weird/funky here I would love to hear it.

The main plus for the AST version would be that it already implements type checking, so we could add the janky type inference on top of it. Otherwise we'll have to re-implement some of the logic that's in the type checker into the optimization framework.

But like you said, if we want this to interact with the other optimizations, we've gotta go with the MIR option.

> Optimization Step (MIR Transformation)

> This will be a bit more involved, but what we want to do here is iterate over the list of MIR statements looking for each Decl Then for each Decl we search the rest of the MIR for functions where any of the inputs matches up with the name of the Decl and any for loops where the Decl object is single cell accessed. If we fail then we have to go up and change the Decl so that it is an Eigen::Matrix<var> (AoS) and then rewrite the rest of the MIR transforming StanLib's so that they use an AoS instead of an SoA.

I think we can do this with a monotone framework instance similarly to how we do AD optimization. Happy to help design that when I'm up to speed on the details

> Also does anyone have papers / books I can lookup on the different schemes in the optimization steps?

Yeah, we followed the "monotone framework" from Principles of Program Analysis by Nielson and Nielson. I can't quite remember how readable it was but it explains the design

nhuurre:
> When is it okay to cast between SoA and AoS? Is it okay for e.g. an SoA matrix to assign to an AoS matrix?

No, I don't think so. The constraints are described in [the design doc](https://github.com/stan-dev/design-docs/blob/master/designs/0005-static-matrices.md#stanc3-implementation).

> Changed StanLib function kinds so that they take in a Common.Helpers.mem_pattern, which will be used for the above to deduce the return type for certain functions that the C++ compiler can't deduce.

Would it be enough to get the return type from the expression's meta field? I was thinking maybe `UnsizedType.autodifftype` could be `DataOnly|AutoDiffable|ADMatrix` and no other changes to MIR.

SteveBronder:
>> Every function that is called using the object in the Decl must have a signature in Stan math that accepts a var_value type.

> What about the case like:

```
matrix x;
f(g(x));
```

> where g is an SoA function but f is not. IIUC, if x is SoA then g(x) is SoA and f fails, but this won't be caught by just looking for functions where x is used as an argument.

So right now what I'm doing is when I match to a function I look at all of it's inner expressions and if _all_ of them are able to be SoA as well then we keep the function as SoA. But that's too aggressive in switching back to AoS. If we had something like the binary function below, as long as one of those functions accepts an SoA input then the outer function can be SoA as well.


```
matrix x(SoA);
matrix y(AoS);
f(g(x), h(y));
```

I think the easiest way to check this is to have a match for unary functions and then a second match for functions with multiple inputs.

> When is it okay to cast between SoA and AoS? Is it okay for e.g. an SoA matrix to assign to an AoS matrix? If not, that seems like another constraint

Yes that's also illegal, not because it's not possible but because it's almost always very inefficient.

>> The rule we have setup in Stan math is that if a function accepts a var_value in it's signature it supports it for all arguments in the signature.

> Is it safe to assume that an SoA function returns an SoA structure when any of the arguments are SoA?

Yep!

>> @rybern has suggested a method that manipulates the AST and does the deduction during the symantic checking and coversion from AST-> MIR

> This would be during semantic checking, so Untyped AST -> Typed AST rather than MIR

Fixed!

>>  So if anyone has comments on which of the two approaches are best or if anything smells really weird/funky here I would love to hear it.

> The main plus for the AST version would be that it already implements type checking, so we could add the janky type inference on top of it. Otherwise we'll have to re-implement some of the logic that's in the type checker into the optimization framework.

> But like you said, if we want this to interact with the other optimizations, we've gotta go with the MIR option.

Yeah I ended up going with the MIR version. tbh it's partly because that's the part of the compiler I feel more comfortable progrmaming over, but also because I'd like to be able to use those extra optimization steps. 

> I think we miiiight be able do this with a monotone framework instance similarly to how we do AD optimization, depending on how assignments work

> Yeah, we followed the "monotone framework" from Principles of Program Analysis by Nielson and Nielson. I can't quite remember how readable it was but it explains the design

I'm currently doing an incredibly janky thing where I just keep rewriting parts of the list of expressions (see [`rewrite_soa_to_aos`](https://github.com/stan-dev/stanc3/blob/e9a1b0d99088def1b28587db1989084399430a9d/src/middle/Mem_pattern.ml#L292) it's very hmm but is not as slow as I thought it would be. @VMatthijs sent me over that book so I'm going to take a look!

--------------------------------------------------------

@nhuurre 

>> Changed StanLib function kinds so that they take in a Common.Helpers.mem_pattern, which will be used for the above to deduce the return type for certain functions that the C++ compiler can't deduce.

> Would it be enough to get the return type from the expression's meta field? I was thinking maybe UnsizedType.autodifftype could be DataOnly|AutoDiffable|ADMatrix and no other changes to MIR.

So yeah this is what led to #899 because tbh I had no idea what meta was or what it did. I think what I'd like to have is `RevAd` in `autodifftype` and then I'd also like to have a completely separate log_prob that works just on reverse mode. I think for a lot of things we want to do in the future (like embedding reverse mode in the compiler) we will want to do this anyway. That would also make debugging this a lit easier since then I can look at the normal version of `log_prob` and the rev version in one file. We also currently have to do some odds in the C++ to support the generic `log_prob` like `conditional_var_value_t<scalar, Matrix>` for things that should be `Eigen::Matrix<double>` when we are doing log_prob with doubles or forward mode AD, but should be `var_value<Eigen::Matrix<double>>` when working in reverse mode.

SteveBronder:
Alright so I think I'm going to go with @nhuurre 's suggestion of putting this in meta. I have a branch (with compare below) that does this and I think it's much less invasive to the rest of the MIR. I'm pretty sure I should be able to modify the current thing I'm doing to check for the mem pattern. I'm going to go deeper into the below tomorrow because I think I need to be more diligent in when I tag things as SoA or AoS so I'm not wasting time reading through exprs and stmts that I don't need to

https://github.com/stan-dev/stanc3/compare/master...SteveBronder:feature/var-matrix-via-meta?expand=1

SteveBronder:
Ack actually I think I misread @nature's comment. I think what we might want is a return type like AosAutoDiffable and SoAAutoDiffable and add those to UnsizedType.autodifftype

nhuurre:
A `mem_pattern` field in meta also makes sense. I was thinking that `mem_pattern` is irrelevant for pure data so a separate field is sometimes redundant.
I think `optimize_ad_levels` assumes that if `adlevel` is not `AutoDiffable` then it must be `DataOnly` so if you expand `autodifftype` optimizer needs updating too.

SteveBronder:
Thinking about this more I'm not sure I like the `mem_pattern` being part of meta. That means that everything is going to be tagged either `SoA` or `AoS`, even in places where it's not really applicable like scalar operations. It's def more invasive, but I kind of like them being part of just the matrix / vector / row vector sized types and the StanLib functions since we can use that along with the return type and autodifftype to deduce the return in situations we need to

rybern:
> > When is it okay to cast between SoA and AoS? Is it okay for e.g. an SoA matrix to assign to an AoS matrix? If not, that seems like another constraint
> 
> Yes that's also illegal, not because it's not possible but because it's almost always very inefficient.

Gotcha. I'm not sure how greedy of an algorithm will be good here. Another question, say you have `x` that has to be AoS, and `y` and `z` that can be SoA. Then in `x = f(y, z)`, `f(y, z)` will need to return AoS, right? Does that `y` and `z` have to be AoS too?

SteveBronder:
> Another question, say you have x that has to be AoS, and y and z that can be SoA. Then in x = f(y, z), f(y, z) will need to return AoS, right? Does that y and z have to be AoS too?

Yes! Good catch I need to check for that as well

SteveBronder:
Ooof actually I think with what @rybern pointed out, we do need the mem pattern in the meta type. Because I need to fail assignments of `AoS = SoA` I need to know the Assigned types memory pattern. The two ways I can think of doing that are to add an `mem_pattern` Unsized type or to add an `mem_pattern` to the meta part of the stmts record. It feels like adding a `mem_pattern` to the unsized type is going to be very invasive, while adding it to the meta will be less work for the rest of the codebase it's def going to require being clever when as I rewrite the MIR on.

Orr, another way I could do this that's a bit weirder is, everytime the check fails for a declaration, I keep a running list of the names of the objects that failed. Then when I match an assign as I'm checking if another object can still be SoA, I check if I'm assigning to any of the objs that failed and if so then I also mark that as a failure

rybern:
> Because I need to fail assignments of AoS = SoA I need to know the Assigned types memory pattern. The two ways I can think of doing that are to add an mem_pattern Unsized type or to add an mem_pattern to the meta part of the stmts record.

I think the type route makes more sense. I don't think it's a good idea to add what's basically type information about a variable to statement metadata. What would that field mean when the statement doesn't even have an expression in it, like `Block`?

I'm not sure what implementing you've done so far. I think a simple greedy algorithm similar to the iterative typechecking is probably fast enough, and it should be optimal as long as changing one variable to AoS never lets another be SoA.

We could implement that directly, but I'm also 95% sure we could generalize the monotone framework for AD levels without much effort. AD levels and SoA/AoS are both trying to find a minimal set of variables (autodiffable and AoS) under similar rules, we'd just have to swap out the rules. Using the monotone framework would be more consistent with the other optimizations and would reuse a lot of code, though it might be a tad slower than writing it directly.

Happy to chat if that would be helpful!

SteveBronder:
> I think the type route makes more sense. I don't think it's a good idea to add what's basically type information about a variable to statement metadata. What would that field mean when the statement doesn't even have an expression in it, like Block?

Yeah I feel the same way

All the logic is in `Mem_pattern.ml` it's a bit brutalist and just iterates over the list of statements in log problems, but it does seem to do its job of querying and modifying the code. I'm afk this weekend but locally I have a version that does the list thing to keep track of failed AoS assignments. 

> We could implement that directly, but I'm also 95% sure we could generalize the monotone framework for AD levels without much effort. AD levels and SoA/AoS are both trying to find a minimal set of variables (autodiffable and AoS) under similar rules, we'd just have to swap out the rules. The main changes would be to replace Mir_utils.update_expr_ad_levels with e.g. update_SoA_levels and to add the function/for-loop rules to Monotone_framework.autodiff_level_fwd1_transfer. That way we'd be more consistent with the other optimizations and we could reuse a lot of code, but it might be a tad slower than writing it directly.

I'm down for whatever! I mostly did the list version because I wanted to focus on constructing all the logic we'd need without really caring how slow it is / having it be a bit weird 

> Happy to chat if that would be helpful

That would be rad I'll send you an email!

SteveBronder:
Thanks man! I'm afk till Monday but I'll check this out then!

SteveBronder:
nvm I think I got it! At least sort of! I made a separate branch that I'll probably merge into here once I get everything working. Though I do still have to recursivly look through the stmts in `query_stmt_functions`, which means I have to pass in the `flowgraph_to_mir` and then do a lookup on the map based on keys in sub stmts of stmts. I'm not sure if that's right, but I based it on the adtype promotion stuff and it looks like it's changing all my previous `var<matrix>` types to `Eigen::Matrix<double>`'s so I'm at least sort of where I want to be

https://github.com/SteveBronder/stanc3/compare/feature/var-matrix-types...SteveBronder:feature/var-matrix-monotone-framework?expand=1

rybern:
I haven't had a chance to look at your branch yet, but I don't think you should need to do any manual recursion on the statements here (except maybe to match the for loops). In gen_variables you should just be able to look at one statement at a time, the framework handles the whole list of statements for you

SteveBronder:
Mkay let me play around with this a bit more then I'll clean it up and ping you if I hit anything too weird

SteveBronder:
Alright so I got a version of that monotone framework branch up that partially works and is able to compile models! I used the [diamonds](https://github.com/stan-dev/posteriordb/blob/master/posterior_database/models/stan/diamonds.stan) model as an example, though I modified it to use `normal` instead of `normal_id_glm` because I still need to go through and cleanup some things for the new matrix type with the glm functions.

But things seem good! With the current stanc the model took 196.21 seconds and with this PR it took 105.35 seconds, so about a 2x speedup for a pretty simple model which is nice.

I think the last major piece I need to sort out is user defined functions and the logic on when functions of functions can use an Array of Structs vs. Struct of Arrays. @rybern I think the monotone framework stuff is pretty ready for review at this point minus the logic missing that I just mentioned. Would it be easier if I open up a PR of that branch into this branch so it's just the monotone framework stuff we are looking at?

rybern:
@SteveBronder Nice!! Yeah doing a separate PR sounds good

SteveBronder:
Word! I put it up here, note the huge line change num is because I'm pretty out the transformed mir from all the tests as a new test that I'm using for debugging.

https://github.com/SteveBronder/stanc3/pull/4/files

I still have a bit of the logic to sort out but the main complication is how to work with user defined functions. I think for now it would be better to have UDFs always just be Array of Structs and have anything that uses a UDF not be a Struct of Arrays. I'm not sure how else to do them, Is there a way I can lookup a functions body defined in the functions block while I'm doing the check during the monotone framework? I think with that and some C++ stuff I can sort out something to get UDFs to work with all this

rok-cesnovar:
>  I think for now it would be better to have UDFs always just be Array of Structs

That is how we had it planned in the design doc as well. I would not worry about this for the first implementation.

rybern:
> Is there a way I can lookup a functions body defined in the functions block while I'm doing the check during the monotone framework?

Yeah, you should be able to get the function definitions out of `mir` in your top-level `optimize-soa` function and then use it inside `gen_aos_variables`

Running function inlining first should help sometimes

You can also run the optimization on UDFs, but you'd have to know the set of AoS inputs. Maybe you could just cover the common case by checking if all of the arguments and output can be SoA? Probably too much for this PR

rybern:
@SteveBronder Made some comments, but mainly just that you can lean more heavily on the framework. Happy to hop on a call today if you'd like, I'm gonna be away next week

SteveBronder:
@rybern much appreciated! I got a bit of stuff to do today but let me read over your comments and dive into those books a bit more. I'm def not using the framework exactly as it should be, think I just need to get more familiar with the schema

wds15:
@SteveBronder the benchmarks you post look amazing! When you run them next time could you also run along the glm version (without varmat stuff)? I would expect that the glm version is already doing a good job at vectorizing things efficiently. So if the varmat stuff even kills the glm performance, then that‘s super exciting. Don‘t get me wrong in that I don‘t want to say we could just use glm functions and drop varmat…no, no! Varmat is the right thing, but beating glm would be really amazing.

(This is all optional, of course… I am just curious; so never mind if too busy)

SteveBronder:
> @SteveBronder the benchmarks you post look amazing! When you run them next time could you also run along the glm version (without varmat stuff)? I would expect that the glm version is already doing a good job at vectorizing things efficiently. So if the varmat stuff even kills the glm performance, then that‘s super exciting. Don‘t get me wrong in that I don‘t want to say we could just use glm functions and drop varmat…no, no! Varmat is the right thing, but beating glm would be really amazing.

So that `diamonds` example is very specific in that it actually uses the glm stuff at first and I swap it out for a multiply + add. I need to go look at the glm functions as I think somewhere we are forcing a copy to happen and that's removing the performance improvement from the new matrix type. I'll run a benchmark this week, but last time I did it the new matrix type was just a hair faster

SteveBronder:
Alright looks like this is failing because I need to add a bunch of `check_*` functions that accept `var<Matrix>` types and it's having trouble finding the correct `assign()` statement for `var<Matrix>`

PR #885, opened 2021-04-19
Update to ocaml 4.11.1 

SteveBronder:
This updates the setup scripts and code to work with ocaml 4.11.1. This PR just has the changes necessary to 4.11.1 but has not been formatted so Jenkins will fail. It does in fact have several differences in the MIR and generated C++ that we should investigate before doing all the formatting stuff as that will make some things harder to sort out. 

Most of the code changes are very small, things like using `let ( = ) = Stdlib.( = )` as ocaml by default in these versions no longer allows for polymorphic equality comparison operators by default. Should we just do away with any code that does `=`, `<>`, etc. and just have `equals a b`, `not_equals a b` etc. functions for each type? Or is the above safe? tbh I just don't understand the implications of performing the above import vs. writing our own.

All the other changes are imported from #843 except for the `run_*_on_args.ml` files in the tests which use Core's unix and channels stuff.

The changes to the MIR and the C++ are related to the optimizer. My guess is that something in the optimizer is hitting something weird during the lazy code motion phase that deals with comparison operators

## Release notes

Updates Ocaml version to 4.11.1 along with dependencies

## Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause)

PR #899, opened 2021-05-13
make meta part of records visible in mir

SteveBronder:
## Summary

I never knew what the meta part of a record was because when I printed it out in the mir all I got back was `<opaque>`. Turns out, sometimes they have useful things in them. This PR makes is so that the meta parts of a record are visible in the mir. It makes for a lot more code that's printed out but imo I'd rather have it printed out so I can see what's there. 

## Release notes

Makes meta records visible in mir instead of `<opaque>`

## Copyright and Licensing


rybern:
It seems like most of the bloat in the printouts comes from locations, but I'm guessing that's not the parts that's useful for you. Can we keep only the location opaque?

SteveBronder:
Yeah I'm fine with that. One weird thing, I can just put an `opaque_list` there but in #885 ocaml 4.12 throws a warning / error for these. Is there some other way to suppress these besides `opaque_list`? If not then I'm fine doing it anyway and we can sort out how to fix that when we bump up versions

rybern:
Not sure about that, is `opaque_list` an alternative to `[@sexp.opaque]`?

SteveBronder:
Honestly idk lol I'm gonna just put it on the line number and we can sort it when we update to 4.12

SteveBronder:
Huh it looks like doing that just puts us back to where we started. Maybe I was just confused since I kept seeing opaque everywhere and we were actually printing the meta in places. Going to close this

PR #911, opened 2021-06-03
Fix testing error

rok-cesnovar:
This PR just merges the release branch (contains release notes) and attempts to fix the issues we were having in testing where end-to-end tests occasionally threw

```
fatal error: file '/usr/include/errno.h' has been modified since the precompiled header 'stan/src/stan/model/model_header.hpp.gch' was built

note: please rebuild precompiled header 'stan/src/stan/model/model_header.hpp.gch'
```
which means that somewhere a `make clean-all` was not happening properly.

## Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause)

PR #916, opened 2021-06-13
Fix Jenkins tests

rok-cesnovar:
We have been experiencing some issues with tests, mostly as issues with PCH files. Opening a PR to debug and find a solution.

## Copyright and Licensing

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,6 +3,7 @@ import org.stan.Utils
 
 def utils = new org.stan.Utils()
 def skipExpressionTests = false
+def buildingAgentARM = "linux"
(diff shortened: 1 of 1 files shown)


rok-cesnovar:
This is ready for review now. 

This fixes another issue similar to what https://github.com/stan-dev/stanc3/pull/911 tried to resolve, this time for another stage.
In addition I changed the logic so that Jenkins uses a regular linux agent for the arm stage when we are going to skip it.

serban-nicusor-toptal (approved):
LGTM.
Very nice approach with the ARM when not using it, so it won't wait for ARM just to pass.

Issue #905, opened 2021-05-26
empty models translated without warning?


wds15:
There is no warning at all when translating an empty model:

```
[13:04:52][weberse2@C02XK2AGJHD2:~/.cmdstanr/cmdstan-2.26.1]$ touch empty.stan
[13:04:56][weberse2@C02XK2AGJHD2:~/.cmdstanr/cmdstan-2.26.1]$ make empty

--- Translating Stan model to C++ code ---
bin/stanc  --o=empty.hpp empty.stan

--- Compiling, linking C++ code ---
```

I think there should be at least a warning. This is an obvious user fault which the user should be made aware of, I think.

rok-cesnovar:
Agreed, this should definitely warn. There is already an issue open for this https://github.com/stan-dev/stanc3/issues/164

wds15:
Oh...then we can close this one. Sorry, I did not see the other one.

rok-cesnovar:
No problem, its hidden way back on page 5 I think. Thats like google search results on page 2 :)

nhuurre:
I don't think it's a good idea to close duplicates of "page 5 issues", unless the original is marked as `wontfix`. Those deeply buried issues are never going to get looked at.
Even if someone goes through all of our open issues looking for something easy to fix, the fact that after _two years_ the original report has _no discussion what-so-ever_ suggests that no-one cares and it's not worth your time.

rok-cesnovar:
Sure, lets close that issue then instead I guess? I am fine with that as well.

WardBrian:
Should the warning only appear on completely empty programs (like empty.stan in the example), or would the following also elicit a warning?

mostly_empty.stan:
```stan
// todo
```

I think this changes where in the compiler the warning would need to be handled

PR #919, opened 2021-06-22
Produce a warning for empty files

WardBrian:
This is addressing #905. 
The new output is:

```
$ touch empty.stan
$ ./_build/default/src/stanc/stanc.exe empty.stan 
Warning: Empty file 'empty.stan' detected; this is a valid stan model but likely unintended!
```

The change is done in the parser rather than the lexxer so that a file of only comments still produces the warning. This may (or may not) want to be changed based on #93, but it is a simple enough fix for now. One test was added, integration/bad/empty.stan, and a few existing test outputs were updated to include this warning.

## Release notes

A warning is now produced if an empty file is compiled.

## Copyright and Licensing

### DIFF
--- a/src/frontend/parser.mly
+++ b/src/frontend/parser.mly
@@ -89,6 +89,13 @@ program:
     EOF
     {
       grammar_logger "program" ;
+      (* check for empty programs*)
+      let () =
+        match (ofb, odb, otdb, opb, otpb, omb, ogb) with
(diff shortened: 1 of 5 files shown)


rok-cesnovar (approved):
Thank you @WardBrian! This looks great to me.

To answer your question on the issue, I think warning for a comment-only .stan file is the correct behavior.

PR #901, opened 2021-05-18
Do NA initialization and declarations on one line

SteveBronder:
This PR takes the separate declarations and NA filling C++ code and combines them into one line. I was working on my forever quest of removing NA filling for parameters and thought this would be good to slice off into it's own PR.

The change is pretty simple, we're moving from multiple lines to one line that just does the declaration of the object and assignment of the object's value in one line. For Eigen matrices we just use `Eigen::MatrixType::Constant(dims..., DUMMY_VAR__)` to make a matrix type with a constant value in all cells. For vectors it's just using the constructor which accepts a default value. For scalars it's just `scalar_type blah = init_nan`

## Release notes

Do NA initialization and declarations on one line

## Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause)

PR #930, opened 2021-07-23
Remove initialization from objects that are immediately assigned to

SteveBronder:
## Summary

This adds an optimization that marks declarations in the transformed MIR such that the code gen can know whether or not the Decl needs to be initialized with NA values after construction.

Stan needs to fill objects with NA values whenever a user defines an object. This is so that when slices of that object happen badly they get back NA values to know something is going wrong

```stan
matrix[N, M] X;

X[1,1] = 3; // Whoops the only assignment that ever happens :(
```

But often users can write inline assignments after the decl, which has no need for initialization.

```stan
matrix[big_N, big_M] X = cholesky_decompose(OtherX);
```

So as long as the users first assignment to an object fully assigns to it, we can remove the code generated to fill the object with NA values. This PR adds an optimization step to do that which I'd like to have on by default. `allow_uninitialized_decls` searches through each section of the MIR for Decls, once it finds one it then searches for the first assignment. If the assignment is to the entire object we flip the new `initialize` flag in the `Decl` to let the code-gen know it's safe to not initialize that object. When going from the AST->MIR `initialize` is always set to true. The only places where `initialize` is flipped to `false` are in `allow_uninitialized_decls` and `data_read` whenever the `_flat__` vectors are created to read in values from the `context`.

I wanted to use the monotone framework for this, however it wasn't clear to me whether I could use it since I'm not doing a scan over the whole program, but stopping the scan after the first assignment.

## Release notes

Remove initialization of NA values from objects that are immediately assigned to

## Copyright and Licensing


rok-cesnovar:
Does this require #901 or is in anyway dependant on that?

I am guessing @rybern would be most suitable for reviewing?

SteveBronder:
> Does this require #901 or is in anyway dependant on that?

Nope!

@rybern review would be great if you have time, though I think I'm going to need you soon for the new matrix stuff. @rok-cesnovar would you have time to look at this?

rok-cesnovar:
Sure, will do over the weekend.

rok-cesnovar:
@SteveBronder do you have an example model with which you tested this with? I am trying to play with this locally and cant get it to do anything.

```stan
transformed data { # or transformed parameters
  real r = 6.0;
  real p = r * 5.0;
}
```
produces

```cpp
      current_statement__ = 1;
      r = std::numeric_limits<double>::quiet_NaN();
      
      current_statement__ = 1;
      r = 6.0;
      current_statement__ = 2;
      p = std::numeric_limits<double>::quiet_NaN();
      
      current_statement__ = 2;
      p = (r * 5.0);
```
Do I need to set `--O`? Am I missing something completely silly?

SteveBronder:
LoL oh whoops that's because everything is in `Blocks` of `Slist` statements and I wasn't recursing through those. If you try it now it should work

test/integration/good/code-gen/cpp.expected
```diff
@@ -4269,7 +4243,7 @@ f12(const int& a1, const std::vector<int>& a2,
   local_scalar_t__ DUMMY_VAR__(std::numeric_limits<double>::quiet_NaN());
   (void) DUMMY_VAR__;  // suppress unused var warning
   try {
-    current_statement__ = 760;
+    current_statement__ = 749;
```
SteveBronder, line 4246: Hmm, I need to look into why it's changing these current statements. Seems weird?

rok-cesnovar:
Before I dig deeper, a quick example that I think isnt currently working correctly.

```stan
parameters {
  real y;
}
transformed parameters {
  real p;
  real r = y * 5.0;
  real t = p * 5.0;
  r = 6.0;
  p = 6.0;
}
```

No optimization:
```cpp
local_scalar_t__ y;
y = DUMMY_VAR__;

current_statement__ = 1;
y = in__.template read<local_scalar_t__>();
local_scalar_t__ p;
p = DUMMY_VAR__;

local_scalar_t__ r;
r = DUMMY_VAR__;

current_statement__ = 3;
r = (y * 5.0);
local_scalar_t__ t;
t = DUMMY_VAR__;

current_statement__ = 4;
t = (p * 5.0);
```
With this PR:
```cpp
local_scalar_t__ y;
      
      
current_statement__ = 1;
y = in__.template read<local_scalar_t__>();
local_scalar_t__ p;


local_scalar_t__ r;


current_statement__ = 3;
r = (y * 5.0);
local_scalar_t__ t;


current_statement__ = 4;
t = (p * 5.0);
```
It works properly if I dont set `p` later on. It should still fill `p` with dummy vars for this model.
Minor, but I would also maybe not print a newline in place of where the assignment was previously.

SteveBronder:
Oh very good catch! Yeah that's pretty bad. I wonder if we should actually catch 'use before assigned transformed parameters' in the semantic checking? Like, I can't think of a time where it would be correct to perform something like 

```stan
real t;
real p;
t = (p * 5.0);
```

rok-cesnovar:
You get a warning with pedantic mode: 
```
bin/stanc --warn-pedantic --o=examples/test/test.hpp examples/test/test.stan
Warning in 'examples/test/test.stan', line 7, column 11: The variable p may not have been assigned a value before its use.
Warning: The parameter y was declared but was not used in the density calculation.
```

I think its fine if this stays under pedantic mode, though we could at some point move some of the more clear-cut checks from pedantic mode to the "default mode".

SteveBronder:
Okay if that's fine to just have in pedantic mode then I know how to fix this in this PR so it still produces the NA value

rok-cesnovar:
Yeah, I think thats fine. Its also output if you use the `--warn-uninitialized` flag. Which seems like something we should recommend to users.

SteveBronder:
Hmm actually, there's only two ways I can think of fixing this right now

1. In this optim, only check the very next line. So that would cover stuff like 

```stan
matrix[N, M] X = some_func(...);
```

Though we would miss out on a good few other places, this is very safe (and tbh I was really just looking to do this for parameters any way which the above effectively covers)

2. Turn `--warn-uninitialized` on by default (and make it an error?). @rybern how do you feel about that? imo I'd really like this approach as I think 99/1 odds a user did not write this scheme on purpose

rok-cesnovar:
I like 1, its safe and testable and clean.

SteveBronder:
Mkay! I set this up so that it only does (1)

test/integration/good/code-gen/cpp.expected
```diff
@@ -4544,9 +4431,9 @@ binomialf(const T0__& phi_arg__, const T1__& theta_arg__,
     lpmf = Eigen::Matrix<local_scalar_t__, -1, 1>(1);
     stan::math::fill(lpmf, DUMMY_VAR__);
     
-    current_statement__ = 784;
+    current_statement__ = 773;
     assign(lpmf, 0.0, "assigning variable lpmf", index_uni(1));
-    current_statement__ = 785;
+    current_statement__ = 774;
```
SteveBronder, line 4436: @seantalts I just want to safety check, I'm looking at the `locations_array__` and these seem to match up (773 pointing to an error on line 311 and 774 lining up with an error on line 312). Did we previously tick up the current statement for filling the object with NA values? If so then these current statement changes make sense, but wanted to double check

SteveBronder:
@rok-cesnovar waiting to hear back from Sean but otherwise I think this is ready for review!

seantalts:
I don’t think we did anything special for those types of statements; if
we’re no longer emitting some statements for filling NAs then it makes
sense that we aren’t emitting their locations either. Does that answer your
question?

SteveBronder:
Yep ty!

rok-cesnovar:
@SteveBronder is this ready for another look?

SteveBronder:
Yep!

rok-cesnovar:
Will do tomorrow! Sorry for the delay.

test/integration/good/code-gen/cl.expected
```diff
@@ -568,56 +566,32 @@ class optimize_glm_model final : public model_base_crtp<optimize_glm_model> {
     
     try {
       Eigen::Matrix<local_scalar_t__, -1, 1> alpha_v;
-      alpha_v = Eigen::Matrix<local_scalar_t__, -1, 1>(k);
-      stan::math::fill(alpha_v, DUMMY_VAR__);
```
rok-cesnovar, line 572: Is this removed  because these are parameters and we know they will be read in immediately?

The model is [here](https://github.com/stan-dev/stanc3/blob/58b05aaf3ae02441ca986fc0b280ce2075e6ec9b/test/integration/good/code-gen/optimize_glm.stan#L27)

I guess there is never a case where we generate code for parameters without reading in right?

test/integration/good/code-gen/dune
```diff
@@ -22,7 +22,7 @@
 
 (rule
  (targets transformed_mir.output)
- (deps (package stanc) (:stanfiles mother.stan))
+ (deps (package stanc) (:stanfiles mother.stan unenforce_initialize.stan))
```
rok-cesnovar, line 25: Nitpicking alert.

Could we just add those example to the mother model?

rok-cesnovar (commented):
Looks good to me, just a question to confirm and then we are good to go. Sorry for the delay.

test/integration/good/code-gen/cl.expected
SteveBronder: Yeah so parameters this optimization will always happen since they are immediately read

test/integration/good/code-gen/dune
SteveBronder: Ummm yeah we can I'm fine with that

test/integration/good/code-gen/cl.expected
rok-cesnovar: Cool cool. Resolved.

test/integration/good/code-gen/dune
rok-cesnovar: Cool, move to mother so we only have one model and the we are good to go.

SteveBronder: Actually I think we have a lot of examples where this is used in the tests already so I'm just going to remove this example since it already is used by mother.stan

rok-cesnovar: Yeah, agreed.

PR #940, opened 2021-08-10
Complex numbers scalar type

nickdidio:
## Summary

Implemented a complex number type in Stan. The documentation for complex numbers can be found at https://github.com/stan-dev/docs/pull/378. The implementation was originally based off the design doc found at https://github.com/stan-dev/design-docs/pull/41 but has varied from some of the original ideas due to implementation constraints. As well as this, this PR varies from the design doc in that it only contains complex scalars implementation, not complex vectors or complex matrices.

## Goal

The goal of this feature is to allow users to code models involving complex numbers. The reason we want to extend Stan to complex numbers is that many statistical problems are most naturally formulated in the complex domain. For example, in imaging applications such as magnetic resonance imaging (MRI) and cryo electron microscopy (cryo-em), we do not observe an image, but only the magnitude (or modulus) of its discrete Fourier transform. This is naturally a problem that requires a complex FFT operation. Similarly, eigendecomposition of asymmetric matrices is fundamentally a complex operation, as is Schur decomposition. Stationarity conditions for time-series models are often stated in terms of conditions on the magnitudes of complex roots.

## Testing

A test file for complex numbers can be found in /stanc3/test/integration/good/code-gen/complex_scalar.stan. This test file contains many different instantiations of complex numbers, complex arrays, complex math functions, complex user functions, type promotion, for loops, etc.

## Release notes

- Implemented complex type functionality

## Copyright and Licensing


bob-carpenter:
@SteveBronder:  Thanks for dealing with the merge conflicts.  Will this be ready to go if it passes tests?  If so, any suggestions for reviewers?

SteveBronder:
I think I can review a good bit of it, might need some help for @rybern for the semantic checking but I have a pretty good handle of a lot of the other parts of the repo

SteveBronder:
Word I'll read it over today!

SteveBronder:
fyi reviewing this right now, oddly the tests seem to be taking forever to run on my machine?

WardBrian:
I sometimes have an issue with dune where one of the test jobs gets stuck and I need to cancel/restart

SteveBronder:
Yeah I thought it was that but once it hits test 3230+ it just comes to a total crawl. I'll let this run for an hour to see what happens

SteveBronder:
@WardBrian can you try a `make clean` and then `dune runtest`? For me it's stopping at `3233/3242` when I run the tests

WardBrian:
Mine works fine. Drops down to 1 job at 2339/3242 and takes a little while but eventually finishes

SteveBronder:
Oh that's weird I can't even run the tests on the master branch. Let me see what's going on locally

SteveBronder:
Got it! Just needed to update opam? Making some changes I'll push up in a min

src/middle/SizedType.ml
```diff
@@ -37,33 +39,55 @@ let collect_exprs st =
...
+  | SMatrix (_, dim1, dim2) -> [dim1; dim2]
+  | SArray (t, dim) -> dim :: get_dims_io t
 
 let rec get_dims st =
   match st with
-  | SInt | SReal -> []
+  | SInt | SReal | SComplex -> []
   | SVector (_, d) | SRowVector (_, d) -> [d]
   | SMatrix (_, dim1, dim2) -> [dim1; dim2]
   | SArray (t, dim) -> dim :: get_dims t
```
SteveBronder, line 93: A small note here, we essentially need two ways to query dimensions. In one case we treat the complex type as a dual number when we are talking to the I/O to tell it how many parameters we have. In the other we think about complex types from the users perspective where we treat the complex type as a scalar. To differentiate between the two I just made a `get_dims_io` and a `get_dims` as seperate methods. Is that cool?

SteveBronder (approved):
@bob-carpenter @WardBrian @nickdidio One small comment but if tests pass this is good to go!

src/middle/SizedType.ml
WardBrian: I think that’s a good approach (cleaner than what I had done), I would just add to the comment calling out that the primary difference is in Complex types so it’s clear at a glance. 

Also, it’s nit picky but can we use Expr.Helpers.int in the _io version? 

Otherwise LGTM!

SteveBronder: Yes I like both of those!

WardBrian (approved):
Tests ran into the PCH issue, but once they are green I think we can merge this

SteveBronder:
@bob-carpenter I'll start reviewing the docs this week. @WardBrian we should chat soon about how to add complex matrices/vectors. Bob are there any blockers in the math library for that?

bob-carpenter:
> Bob are there any blockers in the math library for that? [complex matrices]

We need all the mixed matrix ops to work and I don't think we ever finished the Eigen overloads for mixed `complex` and `double` matrix operations.  There are a couple of ways to go about that---either with their meta-config or with our own functions `multiply`, etc.  I think @sbronder may have created an issue for these at some point.

SteveBronder:
>  don't think we ever finished the Eigen overloads for mixed complex and double matrix operations.

I think as long as `multiply()` and `add()` in Stan math are supported for complex Eigen matrices then it should be okie dokie since that's what we use in the compiler.

bob-carpenter:
It's the mixed cases among `var`, `double`, `complex<var>` and `complex<double>` that will need to work.  That means

* making sure templates resolve in `prim/fun/multiply.hpp`,
* making sure underlying `operator*` properly overloaded in Eigen.

I'll start at the `prim` level by adding the tests for `unit/math/prim/fun/multiply_test.hpp` and then move on to `mix/fun/multiply_test.hpp`.

PR #944, opened 2021-08-17
Adds mem_pattern needed for SoA and AoS types

SteveBronder:
This is a subset of #898. There's a lot of stuff going on in that PR so I think it's smarter to do it in smaller pieces.

This PR adds the `Common.Helpers.mem_pattern` type and uses it in the necessary types that we need for optimization of the tmir and printing the C++ the uses `SoA` (Struct of Arrays). The largest change is to `Stan_math_signatures`, where the `StanLib` types there now have an extra space in their tuple for a `mem_pattern`.

Everything here is set to `AoS` (Array of Structs) and so while the mir and tmir have changed, the C++ should not. That said, the C++ from the optimizer has changed slightly, though only the names of the symbol ticks have changed while the actual code has not changed at all.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [ ] If a user-facing facing change was made, the documentation PR is here: <LINK>
    - [x] OR, no user-facing changes were made

## Release notes

Adds types used for deducing memory patterns

## Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause)

PR #952, opened 2021-08-24
Properly escape string literals with non-ascii characters

WardBrian:
Currently, string literals are escaped with `%S` formatting, which turns non-printing characters into a 3-long escape sequence like `\012`. However, this escaping is [in _decimal_](https://ocaml.org/manual/lex.html#escape-sequence), while C++ expects an escape sequence `\123` to be in[ _octal_](https://en.cppreference.com/w/cpp/language/escape). By copying most of the [ocaml stdlib's escaping code](https://github.com/ocaml/ocaml/blob/trunk/stdlib/bytes.ml), but replacing the relevant section to output octal sequences instead of decimal, string literals will now properly pass to the C++ code.

In theory this allows you to write print/reject statements in non-English character sets, in practice I only tested it on some emojis and Cyrillic characters. The existing behavior led to some very strange results being printed, but after this change it printed as expected (on a UTF-supporting terminal, anyway)

This was mostly a joke between @bob-carpenter and I -- he said if I added emojis to stan, it would get a blog post. But, it is a relatively simple change and doesn't require any upkeep/support, because strings so limited in their use.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/388

## Release notes

Provide rough support for non-ASCII characters in string literals.

## Copyright and Licensing


WardBrian:
I'm not sure what is causing the test failures here. Those models compile for me locally, and in particular _none_ of the existing cpp output is changed by this PR

rok-cesnovar:
That is unrelated to this PR, the error happens on master too. See https://github.com/stan-dev/stanc3/pull/944#issuecomment-903123575

Its been fixed but it needs a few more hours for the Math changes to propagate up to Cmdstan. Sorry about that.

WardBrian:
Ah, I'm just glad to hear it wasn't caused by this - that would have been deeply confusing

WardBrian:
I've updated the doc PR to more or less say that and avoid the question of 'what is a character'

WardBrian:
Docs have been merged so this should be good to go

nhuurre (changes requested):
Looks good but could you add a test file `test/integration/good/code-gen/print_unicode.stan` with a couple of examples.

WardBrian:
> test file `test/integration/good/code-gen/print_unicode.stan`

Done! Here is what I added:

```stan
transformed data {
  print("test: Љ😃");
  print("λ β ζ π");
}
```
(file encoded in UTF-8)

and the expected output segment:

```cpp
      current_statement__ = 1;
      if (pstream__) {
        stan_print(pstream__, "test: \320\211\360\237\230\203");
        stan_print(pstream__, "\n");
      }
      current_statement__ = 2;
      if (pstream__) {
        stan_print(pstream__, "\316\273 \316\262 \316\266 \317\200");
        stan_print(pstream__, "\n");
      }
```

You can check that those are the correct escape sequences, and running the program prints as expected on my UTF-8 supporting terminal

nhuurre (approved):

PR #959, opened 2021-09-02
Add imaginary literal syntax

### LINKED ISSUE #941: imaginary number literals
R allows users to write a complex number as `1.2 + 3.7i`, where `3.7i` is understood to be an imaginary literal.  Python and MATLAB use `j` in place of `i`.  The Stan language currently requires this to be written as `to_complex(1.2, 3.7)`.  

Stan could support imaginary literals of the form `<fp>i` where `<fp>` is a floating-point literal.  Examples include`1i`, `3.7i`, and `2.9876e-12i`.  The imaginary unit must be written `1i` rather than just `i`.    That is, the postfix `i` acts much like the prefix `0x` marks the literal `238` as hexadecimal in the C++ literal `0x238` or the postfix `L` in `237L` marks `237` as a long integer in C++.

The whole expression `1.2 + 3.7i` would consist of a real literal and imaginary literal, which when added together, produce the same value as `std::complex(1.2, 3.7)`.   The number 3.7 would be instantiated as the complex number `std::complex(0, 3.7)` with real part 0 and imaginary part 3.7.    Thus `1.2 + 3.7i` would compile to `1.2 + std::complex(0, 3.7)`.   This has the same behavior as promoting 1.2 to complex, i.e., `std::complex(1.2, 0) + std::complex(0, 3.7)`, but can be accomplished without explicitly promoting 1.2 to a complex number because `std::operator+()` has appropriate overloads for mixed real and complex arithmetic.  

The complex constructor `std::complex()` is `constexpr` and so is `std::operator+(U, V)` where either of the `U` or `V` is complex.    That means the C++ compiler should be able to evaluate the complete complex literal `1.2 + 3.7i` to a complex object at compile time.

WardBrian:
This is a follow on to #940 and closes #941. It adds the syntax `(realconstant | intconstant) i` for representing imaginary numbers. Examples:

```
  complex zi = 1 + 3.14i;
  zi = zi * 0i;
  real x = get_real(3i - 40e-3i);
```

The implementation just generates a call to `to_complex` with the first argument set to zero **at parse time.** This will be constexpr'd away by the C++ compiler, and means we don't need to treat these any different. The one special case is done in pretty-printing, where I preserve this syntax. 

This will _not_ pretty-print any existing calls to `to_complex` as the new syntax [1]. For example, this string will be pretty printed as itself:
`  complex yi = to_complex(0, 1.1) + to_complex(0.0, 2.2) + to_complex();`

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: 
    This will follow on with https://github.com/stan-dev/docs/pull/378

## Release notes

Added the ability to write imaginary numbers as literals, such as "3.1i"

## Copyright and Licensing

----

[1] If you're wondering how, the parser generates the call to to_complex as having the first argument `RealNumeral "0"`. This is semantically valid, but can't be generated anywhere else - you'll either have `IntNumeral "0"` or `"RealNumeral "0."` etc.  Therefore, we can treat this as a sort of sentinel value. This was a happy accident in my first attempt that I think is worth preserving.


src/frontend/parser.mly
```diff
@@ -546,6 +547,20 @@ common_expression:
...
     {  grammar_logger ("realnumeral " ^ r) ; RealNumeral r }
+  | z=IMAGNUMERAL
+    { grammar_logger ("imagnumeral " ^ z); 
+    (* we fake complex literals through a FunApp. C++ should constexpr this away *)
+    FunApp ( (), 
+             (build_id "to_complex" $loc), 
+             [ { expr= RealNumeral "0"
+               ; emeta={loc= Location_span.of_positions_exn $loc}
+               }
+             ; { expr= RealNumeral (String.sub z ~pos:0 ~len:((String.length z)-1) )
```
nhuurre, line 558:
```suggestion
             ; { expr= RealNumeral (String.drop_suffix z 1)
```

src/frontend/parser.mly
```diff
@@ -565,6 +580,7 @@ common_expression:
          && List.exists ~f:(fun x -> String.is_suffix ~suffix:x id.name) Utils.conditioning_suffices
        then CondDistApp ((), id, args)
        else FunApp ((), id, args) }
+
```
nhuurre, line 583: Where did this space come from?

nhuurre (approved):
Not sure I like a "`to_complex()` with a sentinel" instead of a proper `ImagNumeral` AST node but ok.

WardBrian:
We could generate an `ImageNumeral` in the AST, but then it's just another thing to match on for the entire rest of the compiler which we never touch, and the thing we generated in the C++ would look identical after we special-cased the `Lit` rule in codegen. 

I'm happy to write the PR the other way if we decide to, it just seemed to me like adding complexity while this stays pretty local

nhuurre:
I didn't say MIR needs an `ImagNumeral`. I meant for the frontend only (semantic check, debug data gen, canonicalizer, pretty printer) and `Ast_to_Mir` would convert it to a `to_complex()` call.
Although, come to think of it, the pedantic mode works with the MIR. What does the warning about large numeric literals say? I guess it reports the number, just without the trailing `i`.
Anyway, it's not that big of a deal which is why I already approved.

WardBrian:
Here's the alternative if you want to take a look:

https://github.com/stan-dev/stanc3/compare/master...WardBrian:imag-literal-other?expand=1

WardBrian:
I believe the pedantic size warning is only done for integers, so it wouldn't apply (to either implementation)

WardBrian:
Actually @nhuurre - I think for any backend other than stan math this PR is not the way to go, so we should go with the other version.

PR #962, opened 2021-09-09
Allow certain C++ reserved words in stan programs through mangling 

### LINKED ISSUE #961: Mangle C++ reserved words and allow their use
Currently, we have a pretty long list of reserved words that are only reserved because they're used in C++:
https://mc-stan.org/docs/2_27/reference-manual/variables-section.html#reserved-names-from-c

We don't need to reserve these if we instead mangle them, like is done in the TFP backend:

https://github.com/stan-dev/stanc3/blob/a1740dbd9929f86106948300144320512e5dc94c/src/tfp_backend/Transform_mir.ml#L5-L22

We could also do this for `var, fvar, STAN_MAJOR, STAN_MINOR, STAN_PATCH,
STAN_MATH_MAJOR, STAN_MATH_MINOR, STAN_MATH_PATCH`, so that our reserved words list is **only** those words which appear in actual stan programs.

My proposal is to mangle with either `_stan_`  as a prefix or postfix,  or use something dependent on the model name.

WardBrian:
Closes #961. The remaining reserved words in the language at present are:
```
for, in, while, repeat, until, if, then, else, true, false, target, int, real, complex, void, 
vector, simplex, unit_vector, ordered, positive_ordered, row_vector, matrix, 
cholesky_factor_corr, cholesky_factor_cov, corr_matrix, cov_matrix, functions, model, 
data, parameters, quantities, transformed, generated, profile, return, break, continue, 
increment_log_prob, get_lp, print, reject, typedef, struct, var, export, extern, static, auto 
```

The last several were saved for potential future use in the stan type system.

Interfaces like CmdStanPy et al may want to intercept the _stan_ prefix and rename those variables back to their intended name, but this is probably optional.

I did add a test file to code-gen and manually confirm that it compiles with gcc. Would we need a separate PR to add this to the things that are actually compiled with a C++ compiler during testing?

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Allow most reserved words in C++ to appear in a stan program; they will be changed by the compiler to `_stan_<reserved word>` in the generated C++.

Fixed a bug where the lexer was allowing illegal variable names.

## Copyright and Licensing


WardBrian:
Note: Most of the diff is deleting old tests, it's not really 86 files that need review

WardBrian:
This is ready for review, not sure we'd want to sneak it in to 2.28 though

rok-cesnovar:
I can take a look, but lets add this for the next release.

PR #963, opened 2021-09-09
Alternative: Add imaginary literal syntax

### LINKED ISSUE #941: imaginary number literals

WardBrian:
This is an alternative implementation to #959 which I think I prefer more now that I look at it.

This is a follow on to #940 and closes #941. It adds the syntax `(realconstant | intconstant) i` for representing imaginary numbers. Examples:

```
  complex zi = 1 + 3.14i;
  zi = zi * 0i;
  real x = get_real(3i - 40e-3i);
```

The implementation generates a call to `to_complex` with the first argument set to zero during the codegen phase. This will be constexpr'd away by the C++ compiler. For most of the compiler, this is treated by mirroring the way RealNumeral is in terms of type/semantic checking. 

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: 
    This will follow on with https://github.com/stan-dev/docs/pull/378

## Release notes

Added the ability to write imaginary numbers as literals, such as "3.1i"

## Copyright and Licensing


src/stan_math_backend/Expression_gen.ml
```diff
@@ -540,6 +540,7 @@ and pp_expr ppf Expr.Fixed.({pattern; meta} as e) =
   match pattern with
   | Var s -> pf ppf "%s" s
   | Lit (Str, s) -> pf ppf "%S" s
+  | Lit (Imaginary, s) -> pf ppf "to_complex(0, %s)" s
   | Lit (_, s) -> pf ppf "%s" s
```
nhuurre, line 544: [optional] I prefer avoiding wildcards when there aren't too many variants.
```suggestion
  | Lit ((Int | Real), s) -> pf ppf "%s" s
```

nhuurre (approved):
Looks good, one question.
What's the status of complex numbers in the TFP backend? The complex scalar PR didn't add any TFP tests but if `to_complex()` works then this PR should do something there too. Python has imaginary literals with `j` suffix but the backend does a type cast for floats so I'm not sure what exactly is the correct output.

WardBrian:
Complex numbers currently fail in the TFP backend because their functions weren't added to the mapping of stanlib functions:

https://github.com/stan-dev/stanc3/blob/a1740dbd9929f86106948300144320512e5dc94c/src/tfp_backend/Transform_mir.ml#L35

Though this mapping is also missing most basic math functions, like abs, exp, etc, so even adding simple translations for to_complex, get_real, and get_imag still doesn't allow the complex_scalar.stan test to work on that backend. I think that is worth a separate PR later

WardBrian:
@nhuurre - just to confirm is this good to merge?

nhuurre:
Yes. No reason to worry about imaginary literals in TFP if the complex numbers are unsupported anyway.

PR #947, opened 2021-08-18
[WIP] Composable transforms

WardBrian:
## Overview
This PR refactors our internal representation of transformations to include a concept of 'chaining' several in a row. The idea is roughly based on [TFP's Chain bijector](https://www.tensorflow.org/probability/api_docs/python/tfp/bijectors/Chain). 

The code generation is based on turning something which currently would be codgen'd as
```c++
      y = in__.template read_constrain_lb<local_scalar_t__, jacobian__>(0,  lp__);
```
into
```c++
      y = lb_constrain<jacobian__>(in__.template read<local_scalar_t__>(),
                                                     0, lp);
```
with the obvious extension that you can add additional reassignments to add additional constraints, e.g.
```c++
      y = lb_constrain<jacobian__>(
              offset_multiplier_constrain<jacobian__>(
                   in__.template read<local_scalar_t__>(), 1, 3, lp__),
              0, lp__);
```

The same is done (in reverse order) for freeing.

There are several areas of the code which are not yet completed (hence the WIP). They are all marked with the string "TR TODO" in either a comment or a failwith statement.

-----
## Remaining work here
- [x] Variables with transforms can also be 'checked' that they are properly constrained. The way this is currently implemented is not correct in general for multiple composed transforms, as you need to unconstrain the outermost value (peel off a transform) before applying the next check
- [x] Much of the code in `Transform_Mir`, `Mir_Utils`, and `Ast_to_Mir` makes assumptions about sizes/dimensions based on the transform associated with the declaration, which still needs to be expanded or rethought
- [x] stan::math functions for constraining need to take a Jacobian template parameter, like those in stan::io::deserializer. This should be an easy change which @SteveBronder has agreed to do
 
-----
## Future work
As a follow on PR, the parser will be updated to parse declarations like
```stan
real <offset=1, multiplier=3, lower=0> variable;
```
As brought up in #659. Further extensions to allow more general compositions are left for others to consider.

-------

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [ ] If a user-facing facing change was made, the documentation PR is here: <LINK>
    - [x] OR, no user-facing changes were made

## Release notes

Internal refactor to allow composing multiple transforms of a single variable

## Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause)

Issue #659, opened 2020-07-31
Can't mix offset/multiplier with upper/lower bounds


andrjohns:
The following code:
```
data { 
  int<lower=0> N; 
  int<lower=0,upper=1> y[N];
} 
parameters {
  real<multiplier=1,lower=0,upper=1> theta;
} 
model {
  theta ~ beta(1,1);  // uniform prior on interval 0,1
  y ~ bernoulli(theta);
}
```

Gives the error:
```
   -------------------------------------------------
     4:  } 
     5:  parameters {
     6:    real<multiplier=1,offset=0,lower=0,upper=1> theta;
                                     ^
     7:  } 
     8:  model {
   -------------------------------------------------

Expected '>' after multiplier expression.
```

Should this be legal syntax?

nhuurre:
Related to #546
It would be nice to allow this but we need to first decide if offset/multiplier should be on the constrained or unconstrained scale. Probably makes most sense to have a constrained value as the offset and an unconstrained scale multiplier? IIRC correlation matrix unconstraining functions are not autodiffable so corr_matrix<offset=x> might not work.

bob-carpenter:
We definitely need to figure out how to compose transformations (what TensorFlow Probability calls "bijectors").    I'd prefer to have the transformations separated with:

```
real<multiplier = m, offset = o><lower = l, upper = u> theta;
```

being read left to right as:

* `real`:  start with unconstrained real
* `real<multiplier = m, offset = o>`:  start with unconstrained, multiply by `m` then add `o`
* `real<multipler = m, offset = o><lower = l, upper = u>`:  start with unconstrained, multiply by `m` then add `o`, then finally apply inverse logit transform designated by `<lower = l, upper = u>`.  

That way, the subsequences of the type expression denote reasonable types, and each additional transform gets added to the end.

@nhuurre --- any idea how hard this would be to do?  It'd be really useful for things like non-centered lognormal parameterizations.

> IIRC correlation matrix unconstraining functions are not autodiffable so corr_matrix<offset=x> might not work.

It's the constraining functions that need to be autodiffable, and they are here (it's an inverse logit plus stick breaking process to get each unit-length row with positive element on the diagonal;  the details are in the ference manual).    So we could compose an offset and correlation matrix constraint, but it'd only make sense doing the offset/multiplier first as otherwise the result won't be a correlation matrix.  But I don't see an application for it.   

[...]

PR #971, opened 2021-09-24
Allow scale to compose with transformations

WardBrian:
This is a much more limited alternate take on #947. Basically, separate out offset and multiplier from the idea of a transformation, and allow them to be composed, but not arbitrary other compositions. This aligns more directly with #659

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Offset/Multiplier scaling can now be applied to constrained data types, like `real<lower=0><offset=1>`. When constraining, the scaling is done first so that the result is truly of the constrained type. 

## Copyright and Licensing


WardBrian:
@SteveBronder - This is the proof of concept of the alternative version. If you ignore the multiarch build problem ATM it compiles all the models fine, including the following:

```stan
data {
   int<lower=0> N;
}

parameters {
    array[3] real<lower=0><offset=10> loweroffset;
    real<lower=0, upper=1><offset=1, multiplier=10> allfour;
    simplex<offset=1>[10] offset_simplex;
    positive_ordered<multiplier=10>[N] multiplied_ordered; 
}
```

We can throw this into cmdstan and start doing some actual modeling tests next week

WardBrian:
The implementation here should be good to go, @bob-carpenter and I need to work on some example models and the doc for it.

SteveBronder:
Is this waiting on review?

WardBrian:
There was a lot of discussion in #659 about whether this was the right thing to do or not, which is the bigger blocker I believe

WardBrian:
This has fallen too far behind, so I'm going to close it for now

PR #936, opened 2021-07-27
Multiarch Stanc3 Building

andrjohns:
## Summary

This PR adds automatic building of Stanc3 for multiple linux architectures:

  - `arm64`
  - `armhf`
  - `armel`
  - `mips64el`
  - `ppc64el`
  - `s390x`

This is done using Docker images that are run through QEMU and don't require separate instances for each architecture. The base Dockerfile and instructions for building on each architecture are also included

## Release notes

Adds build process for arm64, armhf, armel, mips64el, ppc64el and s390x linux architectures

## Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause)

PR #972, opened 2021-09-24
Fix multiarch stanc3 builds

andrjohns:
The current multiarch builds are failing because one of the files are no longer available to download. This PR fixes that by replacing the use of a dockerfile which is built and run, with a pre-built docker image that has the necessary files included

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fix automated builds of non-x86 architectures

## Copyright and Licensing

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -277,8 +277,8 @@ pipeline {
                 }
                 stage("Build stanc.js") {
                     agent {
-                        dockerfile {
-                            filename 'docker/debian/Dockerfile'
+                        docker {
+                            image 'andrjohns/stanc3-building:static'
                             //Forces image to ignore entrypoint
                             args "-u root --entrypoint=\'\'"
                         }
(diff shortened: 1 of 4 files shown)


WardBrian:
Should there be a stan-dev dockerhub account for these images, rather than hosting them on your personal account?

rok-cesnovar:
@serban-nicusor-toptal do we have a stan-dev account on dockerhub maybe?

WardBrian:
I've found both  https://registry-ui.mc-stan.org/ and https://registry.hub.docker.com/u/stangroup

rok-cesnovar:
Great, the latter was probably created by @syclik (at least that is my guess), the former @serban-nicusor-toptal should have access to. 

I would go ahead and merge this once tests pass so we have things ready for the release candidate at least a day in advance. We can then move the images. @WardBrian does that work for you?

rok-cesnovar (approved):

andrjohns:
I definitely agree about using a stan repo rather than my own. We already had a chat about it in #936, but there were access issues so we left it pulling from mine in the interim

PR #977, opened 2021-09-28
Call eval on conditional return values

### LINKED ISSUE #975: Stanc3 ternary not compatible with expressions
Stan models using the `? :` ternary operator fail to compile if either return value is an expression, as the two operands have different types.

Reproduce using:
```r
mod = "
data {
  real y;
}
parameters {
  vector[2] a;
  vector[2] b;
}
transformed parameters {
  vector[2] c = sum(a) > 1 ? add(multiply(2.0, a), b) : multiply(2.0, a);
}
model {
}
"

> cmdstan_model(write_stan_file(mod))
Compiling Stan program...
/tmp/RtmpSEBcre/model-1df683f412af8.hpp: In member function ‘void file1df6868475bf0_model_namespace::file1df6868475bf0_model::write_array_impl(RNG&, VecR&, VecI&, VecVar&, bool, bool, std::ostream*) const’:
/tmp/RtmpSEBcre/model-1df683f412af8.hpp:178:32: error: operands to ‘?:’ have different types ‘Eigen::CwiseBinaryOp<Eigen::internal::scalar_sum_op<double, double>, const Eigen::CwiseBinaryOp<Eigen::internal::scalar_product_op<double, double>, const Eigen::CwiseNullaryOp<Eigen::internal::scalar_constant_op<double>, const Eigen::Matrix<double, -1, 1> >, const Eigen::Matrix<double, -1, 1> >, const Eigen::Matrix<double, -1, 1> >’ and ‘Eigen::CwiseBinaryOp<Eigen::internal::scalar_product_op<double, double>, const Eigen::CwiseNullaryOp<Eigen::internal::scalar_constant_op<double>, const Eigen::Matrix<double, -1, 1> >, const Eigen::Matrix<double, -1, 1> >’
  178 |         (logical_gt(sum(a), 1) ? add(multiply(2.0, a), b) : multiply(2.0, a)),
```

I think the outputs are going to have to be wrapped in `eval()` or similar when being transpiled

WardBrian:
Closes #975. This is the simplest solution which simply calls eval on both halves. It adds @andrjohns' model as a compile test

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fixed a bug with the conditional operator `?` and the use of vector or matrix expressions.

## Copyright and Licensing


src/stan_math_backend/Expression_gen.ml
```diff
@@ -567,7 +567,10 @@ and pp_expr ppf Expr.Fixed.({pattern; meta} as e) =
           Expr.Typed.(local_scalar (type_of t) (adlevel_of t))
           pp_expr e
       in
-      let tform ppf = pf ppf "(@[<hov 2>@,%a@ ?@ %a@ :@ %a@])" in
+      let tform ppf =
+        pf ppf
+          "(@[<hov 2>@,%a@ ?@ stan::math::eval(%a)@ :@ stan::math::eval(%a)@])"
+      in
       if types_match et ef then tform ppf pp_expr ec pp_expr et pp_expr ef
       else tform ppf pp_expr ec promoted (e, et) promoted (e, ef)
```
SteveBronder, line 575: Beat me to it! How about something like this

```ocaml
      let tform ppf = pf ppf "(@[<hov 2>@,%a@ ?@ %a@ :@ %a@])" in
      let eval_pp ppf a =
        if UnsizedType.contains_eigen_type meta.type_ then
          pf ppf "stan::math::eval(%a)" pp_expr a
        else pf ppf "%a" pp_expr a
      in
      if types_match et ef then tform ppf pp_expr ec eval_pp et eval_pp ef
      else tform ppf eval_pp ec promoted (e, et) promoted (e, ef)
```

So that if the statement doesn't contain an eigen type we don't bother having to call `eval()`

WardBrian: Did you mean `UnsizedType.is_eigen_type`? If so I've done this exactly

SteveBronder (approved):
Good!

WardBrian:
@rok-cesnovar - will merging this fix break the RC cycle or should we get it in 2.28?

rok-cesnovar:
This is a bugfix so lets definitely get it in 2.28.

If you think this requires us to build RC3 we can do that. Based on my quick look I dont think we need to as the changes are not tightly coupled with anything but will leave it up to you to decide.

WardBrian:
IMO its not necessary to have an RC3 just for this unless @SteveBronder feels strongly.

SteveBronder:
Imo this and the pystan release would be good

rok-cesnovar:
Cool, will let Nic know once its all in.

PR #980, opened 2021-09-29
Allow target += in TParams

### LINKED ISSUE #979: Inconsistency of allowing target+=
Consider two stan programs:

prog1:
```stan
transformed parameters {
   target += 1;
}
```

prog2:
```stan
functions {
   void foo_lp(real x){
     target += x;
   }
}

transformed parameters {
   foo_lp(1);
}
```

prog1. stan produces a type error:
```
Semantic error in 'prog1.stan', line 2, column 3 to column 15:
     1:  transformed parameters {
     2:     target += 1;
            ^
     3:  }
Target can only be accessed in the model block or in definitions of functions with the suffix _lp.
```

prog2.stan compiles just fine. 

This is because we have this check for target+= statements:
https://github.com/stan-dev/stanc3/blob/095b8ef6e93c3fe029f77ad56c88d650fbd60019/src/frontend/Semantic_check.ml#L1112-L1113

And this check for the calling of _lp functions:
https://github.com/stan-dev/stanc3/blob/095b8ef6e93c3fe029f77ad56c88d650fbd60019/src/frontend/Semantic_check.ml#L238-L246

WardBrian:
This closes #979 by allowing target += and `~` statements directly in transformed parameters, since function calls allowed them already.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [ ] If a user-facing facing change was made, the documentation PR is here: Will be addressed by https://github.com/stan-dev/docs/issues/413

## Release notes

Fixed inconsistent behavior with accessing target in transformed parameters.

## Copyright and Licensing


spinkney:
I'm pretty excited about this! Shouldn't we allow target += syntax in functions that aren't `_lp` then? And probably deprecate `_lp` functions as they're not used very much

WardBrian:
If we wanted to allow target+= in functions other than lp, we wouldn't need to deprecate _lp, just make it not have any meaning (people could still call a function _lp if they wanted to). We'd still need some way of telling side-effecting functions from normal functions, so that we don't try to edit target in generated quantities or anything crazy, so we probably don't want to remove this restriction.

That said, that feels like more of a design question than a consistency patch. IMO, I like the fact that our typechecker constrains side effects so that you can tell based on the name of the function, but it's definitely valid to discuss.

WardBrian:
Closing this for now, it should be re-considered with something like a `jacobian += ` statement

PR #915, opened 2021-06-10
Add newline after warnings are printed

rybern:
We don't emit a newline after printing warnings:
```
stanc3$ ./run pedantic-mode-example.stan --warn-pedantic
Warning: The parameter c has no priors.
Warning: The parameter b has no priors.
Warning: The parameter a has no priors.stanc3$
```
I'm guessing that's causing this issue: https://discourse.mc-stan.org/t/bugs-in-pedantic-mode/22961/3
But even if it's not, that's some weird formatting. The new behavior is:
```
stanc3$ ./run pedantic-mode-example.stan --warn-pedantic
Warning: The parameter c has no priors.
Warning: The parameter b has no priors.
Warning: The parameter a has no priors.
stanc3$
```

PR #982, opened 2021-09-30
Always flush after semantic and lexical errors

WardBrian:
Currently, parsing errors always have newlines at the end, added by Menhir. Our semantic error messages and lexing error messages didn't, so they would result in a command line prompt like
```
Invalid character found.[brian@Flatiron stanc3 on master [!?]] $ 
```
This change simply adds "@." (the format newline/flush command) to the end of the non-parsing errors. Note that this doesn't effect the captured test output, so the diff doesn't have to edit every single fail test in the project.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Error messages should now always have a trailing newline.

## Copyright and Licensing

### DIFF
--- a/src/frontend/Frontend_utils.ml
+++ b/src/frontend/Frontend_utils.ml
@@ -9,7 +9,7 @@ let untyped_ast_of_string s =
   res
 
 let emit_warnings_and_return_ast (ast, warnings) =
-  if List.length warnings > 0 then Warnings.pp_warnings Fmt.stderr warnings ;
+  Warnings.pp_warnings Fmt.stderr warnings ;
   ast
 
 let typed_ast_of_string_exn s =
--- a/src/frontend/Preprocessor.ml
+++ b/src/frontend/Preprocessor.ml
@@ -20,7 +20,7 @@ let rec try_open_in paths fname pos =
         (Middle.Errors.SyntaxError
            (Include
               ( "Could not find include file " ^ fname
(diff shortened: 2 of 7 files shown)


WardBrian:
@rok-cesnovar both this and #981 are minor tweaks to error messages that I think we should get into 2.28

rok-cesnovar:
Yeah, agreed. Thanks for these cleanups. Will review shortly once I am back at my computer.

rok-cesnovar (approved):
Looks good. 

I found another example where a newline is produces but did not find where. With `--warn-pedantic` or `--warn-uninitialized` on a file that produces no warnings a newline is printed to the stderr.

```
./_build/default/src/stanc/stanc.exe --warn-pedantic test/integration/good/assignments.stan 2>test.txt && wc test.txt && rm test.txt
1 0 1 test.txt
```

No need to fix it here though.

WardBrian:
My guess is that was introduced by #915, it seems like the way that code is formatted we would always print a trailing newline even if the list is empty. We added a conditional to avoid doing the same thing in #869, but didn't add it to the existing usages

rok-cesnovar:
> My guess is that was introduced by #915

Can confirm this added it yes.

WardBrian:
Might as well fix alongside this. Give me a moment

rok-cesnovar:
Thanks! Verified this removes the newline. Merge once tests pass.

PR #984, opened 2021-09-30
Update README

WardBrian:
The desugaring phase is long gone at this point, and our advice for connecting VSCode to WSL has been superseded. I've updated it to the current best practice, in the future [wslg](https://github.com/microsoft/wslg) may make even this obsolete 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Updated stanc3 README

## Copyright and Licensing

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Stanc3 has 3 main src packages: `frontend`, `middle`, and `stan_math_backend`.
 ![top-level stanc3 architecture](docs/img/architecture.png)
 
 The Middle contains the MIR and currently any types or functions used by the two ends.
-The entrypoint for the compiler is in `src/stanc/stanc.ml` which sequences the various components together.  
+The entrypoint for the compiler is in `src/stanc/stanc.ml` which sequences the various components together.
 
 ### Distinct stanc Phases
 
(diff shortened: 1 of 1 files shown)


WardBrian:
@serban-nicusor-toptal do we have something like a `[ci skip]` flag we can include in commit messages to prevent Jenkins from running? PRs like this which only change non-code definitely don't need our full CI

serban-nicusor-toptal:
Sadly we don't have that right now, a quick fix would have been this [plugin](https://plugins.jenkins.io/scmskip/) but it doesn't work with multibranch pipelines so I'll need to create a PR and implement it ourselves.
We achieve this in repositories with a "[source code whitelist](https://github.com/stan-dev/stanc3/blob/master/Jenkinsfile#L47)" which we pass to [a function](https://github.com/stan-dev/jenkins-shared-libraries/blob/master/src/org/stan/Utils.groovy#L56) that verifies changes made and decides if to run all/specific jobs.
Example usage in [cmdstan ](https://github.com/stan-dev/cmdstan/blob/develop/Jenkinsfile#L78)for this purpose. I'll note this down and create a PR in stanc3 so we can define what is source code is and just skip the rest. Let me know your pov if you think of another way to do it.

rok-cesnovar (approved):
Looks good.

PR #989, opened 2021-10-04
Display lexer warnings the same as others

WardBrian:
This is somewhat of a follow on to #982. If we don't get it in 2.28 that's probably okay.  
We were outputting input warnings with a piece of code that duplicated `Warnings.pp_warnings` but not as well. The result was that the newlines were not properly handled. This was especially noticeable if you had a warning and error in the same file, as the error would just get appended to the line rather than following it. 

See changed test output for examples

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Display errors on separate lines from warnings if both exist

## Copyright and Licensing


rok-cesnovar (approved):
Thanks!

I am good with this going in 2.28 as I dont see us needing to do a RC. The binaries will have to be rebuilt anyways to include version numbers. If you agree go ahead and merge.

Issue #992, opened 2021-10-06
Issue with simplex_free in 2.28.0


rok-cesnovar:
This issue was reported to me by @spinkney and it seems to have made its way to the 2.28.0 release.

The model: 

```stan
data {
  int<lower=0> K;
  vector[10] y;
}
parameters {
  simplex[K] s;
  real<lower=0> lambda;
}
model {
  y ~ exponential(lambda);
}
```
If you sample without specifying any initial values, it samples fine, but if you specify the initial value only for `lambda` it outputs:
```
Rejecting initial value:
  Error evaluating the log probability at the initial value.
Exception: stan::math::simplex_free: Simplex variable is not a valid simplex. sum(Simplex variable) = 2.384263259, but should be 1 (found before start of program)
Rejecting initial value:
  Error evaluating the log probability at the initial value.
Exception: stan::math::simplex_free: Simplex variable is not a valid simplex. sum(Simplex variable) = 2.274321496, but should be 1 (found before start of program)
```
This however works fine with 2.27.0 (sampling finishes with 0 warnings or errors). 

It also fails with cmdstan 2.27.0 and the 2.28.0 stanc3 binary which leads me to believe this is a stanc3 issue.

Here is a folder with all the required files to reproduce:
[reprex.zip](https://github.com/stan-dev/stanc3/files/7293409/reprex.zip)

Reproduce simply by running:
```
./model sample data file=data.json init=init.json
```

You can also run with cmdstanr:
```r
library(cmdstanr)

code <- "
data {
  int<lower=0> K;
  vector[10] y;
}
parameters {
  simplex[K] s;
  real<lower=0> lambda;
}
model {
  y ~ exponential(lambda);
}
"
stan_file <- write_stan_file(code)
mod_test <- cmdstan_model(stan_file)

fit_opt <- mod_test$sample(
  data = list(K = 3, y = rexp(n = 10, 4)),
  init = list(list(lambda = 2)),
  chains = 1
)
```

SteveBronder:
Yes I'm looking right now, I have can see what's happening and am working on a fix

rok-cesnovar:
Thank you good sir!

rok-cesnovar:
Yes, definitely worthy of 2.28.1.

PR #993, opened 2021-10-06
fix transform inits input and output sizes

SteveBronder:
Fix for transform inits so that the input vector is the same as the number of constrained parameters while the output vector is the size of the number of unconstrained params

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Copyright and Licensing


WardBrian:
OCaml looks fine to me, can't comment much on the generated C++/whether this resolves the issue. I assume you tried it out?

SteveBronder:
Yeah this fixes the issue in #992, tested out that model. As the summary says that goof happened because of the difference in the number of constrained and unconstrained parameters for things like simplex and unit vectors

rok-cesnovar (approved):

PR #641, opened 2020-07-16
Add signatures for elementwise pow operator and matrix power

andrjohns:
This PR adds the signatures and tests for the new elementwise ```.^``` operator, as well as for the ```matrix_power``` function.

There's a good chance that I haven't taken the most efficient route for adding the ```.^``` operator, as well as defining the signatures, so let me know if there's a better way that I should be doing this.

Issue #604, opened 2020-07-04
Disambiguating vectorised pow operator from matrix pow operator


andrjohns:
With the merging of [this pull](https://github.com/stan-dev/math/pull/1917) in the math library, the ```pow``` function is now vectorised to work elementwise with matrix and nested array inputs, one of the possible signatures being:
```
pow(matrix, real)
```

With the relevant syntactic sugar being:
```
matrix operator^(matrix x, real y)
```

However, the Math library also has a ```matrix_power``` function [here](https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/matrix_power.hpp), which (AFAIK) isn't exposed in the Stan language. Once that function is exposed it would have the same operator signature.

A simple fix would be to have the vectorised (elementwise) ```pow``` operator have the signature:
```
matrix operator.^(matrix x, real y)
```
But then there's also the likelihood that a user would call ```matrix^real``` expecting an elementwise power, rather than the matrix power. So it might be worth having a compiler warning (like with integer division). Thoughts?

nhuurre:
In theory multiplication has the same issue but no-one worries about that.

Integer division has a special operator. The warning is only for backward-compatibility. If you're worried that people think `^` means element-wise then let's not allow it at all. Instead everyone must use either `.^` or `matrix_pow`, whichever they need.

Btw, `.*` doesn't work with arrays, and we don't even have `.+`. If `pow` is vectorized for arbitrary arrays does that mean the plan is to extend `.* ` too? Actually, that does get confusing for arrays of matrices...

andrjohns:
> Instead everyone must use either .^ or matrix_pow, whichever they need.

That sounds good to me.

> If pow is vectorized for arbitrary arrays does that mean the plan is to extend .* too?

I hadn't thought about that, to be honest. Personally speaking, I don't see much of a problem with things like ```real .* matrix[]``` or ```matrix[] .^ real```, but I'm not sure how consistent that is with Stan's conceptualisation of the difference between arrays and vector/row-vector/matrix. I'm also not particularly wedded to any kind of implementation, so if it keeps things consistent I can always leave the signatures at:
```
real pow(real, real)
vector pow(real, vector)
vector pow(vector, real)
row_vector pow(real, row_vector)
row_vector pow(row_vector, real)
matrix pow(real, matrix)
matrix pow(matrix, real)
```

bob-carpenter:
[...]

PR #994, opened 2021-10-06
Deprecate ".^" operator

WardBrian:
See the discussion in #604 

This marks `.^` as deprecated in favor of `^` everywhere. These two have had the exact same semantics since the former's introduction in #641 

#### Submission Checklist

- [x] Run unit tests
- Documentation
    
## Release notes

Deprecated `.^` operator - this has always had the same meaning as `^`, use that instead.

## Copyright and Licensing


nhuurre:
I thought the plan was to deprecate `^` (not `.^`) for non-scalars because `matrix .^ 2` is obviously elementwise but `matrix ^ 2` could be mistaken to mean `matrix_power(matrix, 2)`.

bob-carpenter:
Good points.  

Now that I think more about it, I like @nhuure's suggestion to use `.^` for elementwise matrix powers.  That would let us reserve `matrix^int` for the multiply-a-matrix-by-itself-multiple-times version.

The alternative seems to be having `^` be elementwise everywhere and require `matrix_pow(matrix, int)` for the multiply-by-self version.

nhuurre:
> I have a hard time imagining a way to safely redefine the meaning of matrix ^ 2.

We don't need to redefine it, we already have `matrix_power(_, _)`.
It's going to be the same situation as currently with `int / int`, i.e. the operator `^` retains its current meaning indefinitely but emits a warning that advises people to use either `.^` or `matrix_power()` instead.

WardBrian:
If it's indefinite, why do it? (for the record, I think the `%/%` warning is also a bad idea outside pedantic mode)

nhuurre:
> If it's indefinite, why do it?

Because `matrix ^ int` is confusing to readers. Arguably yes, these warnings should be limited to pedantic mode. But the main reason for "indefinitely" is because that's what the schedule for deprecations has been in the past. If we start actually removing deprecated features then I like to simply forbid `matrix ^ int` and `int / int`.

WardBrian:
> a deprecation warning for uses of matrix^real suggesting .^ for elementwise or matrix_pow for self-multiply instead.

This warning would be for matrix^real and matrix^int right @bob-carpenter?

That seems fair enough. I don't think we could actually define operator^ to use matrix_power without a version 3.0 level bump, but issuing a warning that it might not do what you want is good.

Issue #998, opened 2021-10-12
Ints and reals not promoted properly in complex function signatures


WardBrian:
Currently, this does not compile (throws a signature error):

```stan
generated quantities {
   complex x = add(1, 3i);
}
```

This creates a C++ error:

```stan
real x = norm(1.0);
```
* [x] SignatureMismatch does not recognize promotion from int/real to complex as valid (see #999 )
* [ ] C++ does not promote `double` to `std::complex<T>` in function calls 
* [ ] Update tests/integration/good/code-gen/complex_scalar.stan to verify all complex functions, use of doubles and ints in function calls; remove explicit casts

PR #999, opened 2021-10-13
Allow complex promotion in functions, not just operators

WardBrian:
This address the first portion of #998. We still need to do static casting in the code generation so the C++ will compile.

Currently, things like 

```stan
    complex foo_two(complex z, complex z2){
      return add(z,z2);
    }
   
    foo_two(1, 3.5);
```

throw signature errors because we're not properly allowing promotion for functions.

Partially this was caused by the fact that we have different systems for checking functions and checking operators, but this does not fix that issue, just repair this symptom for now.

However, this now fails at C++ time. 

So does something like `proj(1.0)`.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fixed an issue with calling functions that expect complex arguments with reals or integers.

## Copyright and Licensing


SteveBronder:
Will do! Trying to finish up some of the adjoint ODE fix today then I will look at this

SteveBronder:
>> Also I just realized idt we can generate functions that take in an array of array of complex types?

> Why not?

How do you write the signature? I think the C++ always gens `const std::vector<std::complex<T1__>>&` no? Maybe I just don't know the syntax

I'll take a stab at the compiler internal promotion type tmrw. That feels like the cleanest way imo

WardBrian:
Does `void foo(array[,,] complex zs){...}` not codegen at the moment? That sounds like a whole other kind of issue if so

SteveBronder:
Oh whoops nvm!

SteveBronder:
The backend for this is pretty much the same whether I do the big or small thing so I'm just going to the full promotion thing

WardBrian:
Array literal promotion for complex works in Stan, I’ll need to check that the C++ compiles and that we have it in the test

WardBrian:
Superceded by #1004

Issue #991, opened 2021-10-04
Canonicalizer should turn single-statement branches into blocks


WardBrian:
This may be controversial, but I think if we're defining a canonical form when it comes to parenthesis etc, we should also do it with brackets around statements like

```stan
if(cond)
   # result
```
and turn them into 

```stan
if (cond) {
    # result
}
```

adamhaber:
This should be doable by adding an `IfElse` case for
https://github.com/stan-dev/stanc3/blob/763421b9493ea1f7f7446392176435d85055afdb/src/frontend/Canonicalize.ml#L5-L18
and making sure it contains a single statement, right?

WardBrian:
You could do it there or during the parens mapping. It would be pretty simple though - match on if/else, for, and while, and if they contain a statement that is not a block make it the body of a new block.

PR #1003, opened 2021-10-14
Feature/add block single stmt

adamhaber:
Addresses #991 - canonicalizer now adds brackets around statements like

```stan
if(cond)
   # result
```
and turns them into

```stan
if (cond) {
    # result
}
```
and similarly for the else branch, and single-statement for/while loops.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/423

## Release notes

Canonicalizer now adds brackets around single statements in if-else/for/while.

## Copyright and Licensing

### DIFF
--- a/test/integration/canonicalize/blocks.stan
+++ b/test/integration/canonicalize/blocks.stan
@@ -5,44 +5,69 @@ data {
 model {
     int j = 0;
 
-    if (N>5) y ~ std_normal();
-    else y ~ std_normal();
-
-    if (N>5) y ~ std_normal();
-
+    // test for - with / without parens
     for (i in 1:N) target += i;
+    for (i in 1:N) {target += i;}
     
+    // test for - with / without parens
     while(j<N) j = j+1;
+    while(j<N) {j = j+1;}
+    while((j<N)) j = j+1;
 
-    while(j<N) if (N>5) y ~ std_normal(); else y ~ std_normal();
+    // test if - with / without parens
+    if (N>5) y ~ std_normal();
+    if (N>5) {y ~ std_normal();}
+
+    // test if-else - with / without parens
+    if (N>5) 
+      y ~ std_normal();
+    else
+      y ~ std_normal();
+
+    if (N>5) 
+      {y ~ std_normal();}
+    else 
+      y ~ std_normal(); 
+
+    if (N>5) 
+      y ~ std_normal();
+    else 
+      {y ~ std_normal();}
+
+    if (N>5) 
+      {y ~ std_normal();}
+    else 
+      {y ~ std_normal();}
 
     ///////////////////////////////
     // nested if else combinations
     ///////////////////////////////
-    if (N>5) 
-      y ~ std_normal(); 
-    else if (N>5) 
-      y ~ std_normal(); 
-    else if (N>5) 
-      y ~ std_normal(); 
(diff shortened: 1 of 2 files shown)


src/frontend/Canonicalize.ml
```diff
@@ -159,12 +167,21 @@ let rec parens_stmt {stmt; smeta} =
           ; identifier
           ; initial_value= Option.map ~f:no_parens init
           ; is_global }
+    | While (e, s) -> While (e, parens_stmt (stmt_to_block s))
+    | IfThenElse (e, s1, s2) -> (
+      match s2 with
```
WardBrian, line 172: I think this is a good opportunity to use `Option.map`

src/frontend/Canonicalize.ml
```diff
@@ -159,12 +167,21 @@ let rec parens_stmt {stmt; smeta} =
...
+      | Some x ->
+          IfThenElse
+            ( e
```
WardBrian, line 175: I think we need to propagate the parens normalizing on this expression. Notice the extra parenthesis on one of the unit tests, shouldn’t happen

test/integration/canonicalize/canonical.expected
```diff
@@ -1,3 +1,35 @@
...
+  real mu;
+  real<lower=0> sigma;
+}
+model {
+  int j = 0;
+  
+  if (N > 5) {
+    y ~ normal(mu, sigma);
+  }
+  else {
```
WardBrian, line 16: Any idea why this doesn’t pretty-print such that the else is on the same line? I think we do this for normal if/else

test/integration/canonicalize/canonical.expected
```diff
@@ -123,8 +155,9 @@ model {
   for (i in 1 : (N - 3)) {
     n[i] = m[i : (i + 3), i]';
   }
-  if ((b < x) && (x < 1)) 
+  if (((b < x) && (x < 1))) {
```
WardBrian, line 158: This test line has the extra parents

test/integration/canonicalize/blocks.stan
```diff
@@ -0,0 +1,20 @@
...
+  vector[N] y;
+}
+parameters {
+   real mu;
+   real<lower=0> sigma;
+}
+model {
+    int j = 0;
+
+    if (N>5) y ~ normal(mu, sigma);
```
WardBrian, line 12: Can you add some more complicated examples here? What happens if you have nested ifs/fors, or a comment between the if and the statement, or an if/else that has brackets for the if but not for the else?


Feel free to make the test file pretty ugly to really stress this feature out

src/frontend/Canonicalize.ml
adamhaber: Do you mean `no_parens e`? This seems to solve the extra parens on the unit test below.

test/integration/canonicalize/canonical.expected
WardBrian: Huh. My guess is this is something in the special logic in the pretty printer, but I suspect it should be looked into. I’ll look later today if it’s an easy change or not, but I don’t think you need to worry for this PR

PR #1005, opened 2021-10-18
Add arg function signature

### LINKED ISSUE #1000: Expose documented complex function `arg`
https://mc-stan.org/docs/2_28/functions-reference/complex-special.html lists a function `arg` which is not in Stan_math_signatures.

WardBrian:
This exposes `arg`, a function which was already documented but not in the signatures file. It closes #1000 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Expose complex function `arg`

## Copyright and Licensing

### DIFF
--- a/src/middle/Stan_math_signatures.ml
+++ b/src/middle/Stan_math_signatures.ml
@@ -849,7 +849,6 @@ let () =
   List.iter
     ~f:(fun x -> add_unqualified ("add", ReturnType x, [x; x], AoS))
     bare_types ;
-  (* add_unqualified ("add", ReturnType UComplex, [UComplex; UReal], AoS) ; *)
   add_unqualified ("add", ReturnType UVector, [UVector; UReal], AoS) ;
   add_unqualified ("add", ReturnType URowVector, [URowVector; UReal], AoS) ;
   add_unqualified ("add", ReturnType UMatrix, [UMatrix; UReal], AoS) ;


src/middle/Stan_math_signatures.ml
```diff
@@ -849,7 +849,6 @@ let () =
   List.iter
     ~f:(fun x -> add_unqualified ("add", ReturnType x, [x; x], AoS))
     bare_types ;
-  add_unqualified ("add", ReturnType UComplex, [UComplex; UReal], AoS) ;
```
rok-cesnovar, line 852: We are removing these due to the automatic conversion issue? Same Q for norm.

rok-cesnovar (commented):
One simple question. I would love to see a model that uses the arg function as well. Can be a separate model or just a call to arg() with data and parameters as input in an existing model.

src/middle/Stan_math_signatures.ml
WardBrian: These are related to #998. I think that they were added because the PR misunderstood which conversions should be automatic.

Norm is being removed for a different reason: g++ crashes at the moment if you actually call it with a real input. 

Rewriting the complex tests is part of the PRs related to the above. I can add it separately here if you'd like, but part of the testing is testing it on real/int inputs which wouldn't work yet

rok-cesnovar: Ok, I am good with that.

PR #1006, opened 2021-10-20
Reorganize repo to support developer documentation

WardBrian:
This starts the process of using [odoc](https://github.com/ocaml/odoc) to document our internals, similar to what we have for stan-math.

I've uploaded the current version of what gets built here:
https://brianward.dev/stanc3-doc/stanc/index.html

Draft for a few reasons:

- [x] This works kind of like doxygen, so we will want a similar build process to what stan-dev/math uses for their developer documentation. I'll need @serban-nicusor-toptal's help for this. Here's what the odoc repo itself uses: [publish.sh](https://github.com/ocaml/odoc/blob/master/doc/publish.sh)
     - [x] It would be great if we also set up Jenkins to ignore changes outside of test/ and src/ 
- [x] We should clean up a lot of our doc comments to actually have the documentation contain more than just the signatures
- [ ] We may want to move or duplicate parts of the notes folder and github wiki into these docs.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Create internal developer documentation

## Copyright and Licensing

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,14 @@
 # A New Stan-to-C++ Compiler, stanc3
 This repo contains a new compiler for Stan, stanc3, written in OCaml. To read more about why we built this, see this [introductory blog post](https://statmodeling.stat.columbia.edu/2019/03/13/stanc3-rewriting-the-stan-compiler/). For some discussion as to how we chose OCaml, see [this accidental flamewar](https://discourse.mc-stan.org/t/choosing-the-new-stan-compilers-implementation-language/6203).
-We're testing [these models](https://jenkins.mc-stan.org/job/stanc3/job/master/)(listed under Test Results) on every pull request and think we are currently up to par and mostly backwards compatible with the previous Stan compiler (see [this wiki](https://github.com/stan-dev/stanc3/wiki/changes-from-stanc2) for a list of minor differences).
+We're testing [these models](https://jenkins.mc-stan.org/job/stanc3/job/master/)(listed under Test Results) on every pull request. Since version 2.26, this has been the default compiler for Stan. See [this wiki](https://github.com/stan-dev/stanc3/wiki/changes-from-stanc2) for a list of minor differences between this compiler and the previous Stan compiler.
 
 [![Build Status](http://d1m1s1b1.stat.columbia.edu:8080/job/stanc3/job/master/badge/icon)](http://d1m1s1b1.stat.columbia.edu:8080/job/stanc3/job/master/)
+
+## Documentation
+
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -567,7 +567,45 @@ pipeline {
...
+                deleteDir()
+                retry(3) { checkout scm }
+                withCredentials([usernamePassword(credentialsId: 'a630aebc-6861-4e69-b497-fd7f496ec46b',
+                                                  usernameVariable: 'GIT_USERNAME', passwordVariable: 'GIT_PASSWORD')]) {
+                    sh """#!/bin/bash
+
+                        set -e
+
+                        dune build @doc
+                        dune build @docgen
```
WardBrian, line 583: We don't want to run `@docgen` for our uses (it is a custom alias for them)

Jenkinsfile
```diff
@@ -567,7 +567,45 @@ pipeline {
...
+
+                        cd ../_build/default/doc
+                        ocaml-mdx test driver.md
+                        cd ../../..
```
WardBrian, line 587: These three are also not useful for us

Jenkinsfile
```diff
@@ -567,7 +567,45 @@ pipeline {
...
+
+                        mkdir -p doc
+                        rsync -av _build/default/doc/html/odoc/ doc/
```
WardBrian, line 590: for us it will be `_build/default/_doc/_html/`

WardBrian:
@serban-nicusor-toptal Something is definitely going wrong during the verify-changes step - my last commit edited many source files but didn't trigger a rebuild

serban-nicusor-toptal:
Checking, it might have something to do with the PR coming from an outside repo.

serban-nicusor-toptal:
Yes, it's using `git diff` see [function](https://github.com/stan-dev/jenkins-shared-libraries/blob/master/src/org/stan/Utils.groovy#L56)

```
+ git checkout master
Already on 'master'
Your branch is up-to-date with 'origin/master'.

Comparing differences between current 3618c13847a91f45c442225d8f2c1a7a315d88d2  and target master

+ git diff 3618c13847a91f45c442225d8f2c1a7a315d88d2 master -- src
+ git diff 3618c13847a91f45c442225d8f2c1a7a315d88d2 master -- test
There aren't any differences in the source code, CI/CD will not run.
```

If `git diff` returns empty when there are no diffs.
I'm debugging but more than sure we need an `if` and to checkout differently for outside PRs

serban-nicusor-toptal:
Yes that's correct

WardBrian:
Good to know. Do you think you've resolved the Jenkins issue? If so this is ready for review

PR #1007, opened 2021-10-21
Use the opam file to manage dependencies

gbdrt:
#### Submission Checklist

- [X] Run unit tests
- Documentation

## Release notes

Use the opam file to manage dependencies.

To install all the dependencies and the compiler: `opam pin -k /path/to/local/clone` or even `opam pin stanc https://github.com/stan-dev/stanc3`. The `stanc` executable should then be in the path.

To install the dev dependencies add the `--with-test` option, i.e., `opam pin -k /path/to/local/clone --with-test` (or reinstall the `stanc` package with this option, i.e., `opam install stanc --with-test`)

## Copyright and Licensing

### DIFF
--- a/stanc.opam
+++ b/stanc.opam
@@ -3,18 +3,20 @@ maintainer: "The Stan Development Team"
 authors: "The Stan Development Team"
 homepage: "https://github.com/stan-dev/stanc3"
 bug-reports: "https://github.com/stan-dev/stanc3/issues"
-synopsis: "The Stan transpiler (from Stan to C++ and beyond)."
+synopsis: "The Stan transpiler (from Stan to C++ and beyond)"
 depends: [
   "ocaml" {= "4.07.0"}
   "dune" {= "1.11.4"}
-  "ocamlformat" {= "0.8"}
   "core_kernel" {= "v0.11.1"} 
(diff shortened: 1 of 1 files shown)


WardBrian:
I have actually already set up something similar in #1006! The with-test is nice though. I believe that manually writing .opam is discouraged by dune nowadays, but I'm not sure how to recreate that part with the dune package information

gbdrt:
> I have actually already set up something similar in #1006! The with-test is nice though. I believe that manually writing .opam is discouraged by dune nowadays, but I'm not sure how to recreate that part with the dune package information

Ah sorry, I did not realised this.
It's true that dune can automatically generate opam files, but I don't know how to separate the dev deps...
And a opam file was already present in the repo, so it was the simplest path.

WardBrian:
Seems like you and I figured that out at the same time: https://github.com/stan-dev/stanc3/pull/1006/commits/18ae1e8c1453e2698898915656ec56215e5d59e1

I'm not sure if we would want to use this for CI etc. I think our .sh scripts work fine (not to mention this doesn't cover the extra libraries needed to build Windows binaries), but it's definitely useful for individual users

PR #1011, opened 2021-10-25
Fix issue with function shadowing

### LINKED ISSUE #997: Issue with shadowing and binary functions
This is a continuation of the discussion with @WardBrian here: https://github.com/stan-dev/stanc3/pull/995#discussion_r724759499

Long story short, this model's generated code fails to compile:

```stan
transformed data {
   real multiply;
   real a = multiply(0, 1);   
}
```

with 

```
examples/test/test.hpp: In constructor ‘test_model_namespace::test_model::test_model(stan::io::var_context&, unsigned int, std::ostream*)’:
examples/test/test.hpp:66:24: error: expression cannot be used as a function
   66 |       a = multiply(0, 1);
      |                        ^
```

We allow shadowing and it works with nullary and unary functions. For example this model works fine:
```
transformed data {
   real log;
   real a = log(10);
   real e;
   real f = e();
   real positive_infinity;
   real g = positive_infinity();  
}
```

Solutions to this are:
a) add `stan::math` namespace to a function if there is a variable with the same name used in the model
b) add `stan::math` namespace to all Stan Math function calls (this probably doesnt work as we may fallback to std:: in some cases)
c) if a function is shadowed it can no longer be used as a function in that model (@WardBrian mentioned this is what Python does)
d) disallow shadowing for binary functions (we have to check functions with 3+ args as well)

This is not a new issue for this release, its been present since we allowed shadowing.

WardBrian:
Closes #997. 

This PR moves to qualifying all non-user functions with either `stan::math` or `stan::model` as appropriate. This means that the C++ will compile even if the user defines a variable of the same name. 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fixed an issue that arose during C++ compilation of models that used a variable with the same name as a Stan library function and called that function.

## Copyright and Licensing


WardBrian:
I should note for posterity's sake that the alternative to this change (qualifying everything not in userspace with it's namespace) is to mangle everything from userspace, which could reuse the machinery from #962 

I personally lean toward this solution as it leaves the C++ easier to read, IMO

WardBrian:
I believe I have run down every compilation issue that isn't a result of the stan-math signatures @rok-cesnovar. When that PR gets merged this should be green, but we could also try to test before that to make sure

WardBrian:
@serban-nicusor-toptal - is there any easy way to run these tests with @rok-cesnovar's stan-math branch?

serban-nicusor-toptal:
I was about to suggest parametrizing the checkout step but @rok-cesnovar (thanks for the quick action!) already acted on it here #1064.

rok-cesnovar:
Started a run with the Math PR branch: https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3/detail/PR-1011/39/pipeline/

WardBrian:
@rok-cesnovar I don't understand the failures in that run. It claims that hearts.stan from the bugs example models generates code with a `pow` call that isn't qualified by `stan::math` but this simply isn't true

WardBrian:
Okay, we define overloads of operators in stan::math, so my plan to completely remove the `using namespace stan::math;` directive won't work 100% (or we have to add `using stan::math::operatorX;` for 8 operators).

That doesn't seem like 100% of the issues but we will start there

rok-cesnovar:
Oh man, that test isn't actually doing what we (or at least I) thought it should be doing. It was originally thought of as a comparison with stanc2. It's currently just comparing nightly with nightly. And its not trivial to force ./compare-compiler.sh to do what it should (compare nightly with PR stanc3 binary).

Do you mind if we remove the test and make an issue?

WardBrian:
Sure, it seems like this PR has some other issues to sort out in the meantime anyway

WardBrian:
@serban-nicusor-toptal - now getting a weird failure during verify changes https://jenkins.mc-stan.org/blue/organizations/jenkins/stanc3/detail/PR-1011/43/pipeline

serban-nicusor-toptal:
Looks like a permission issue, checking.

WardBrian:
@rok-cesnovar - still a problem with `pow`:

```
../../test/integration/good/code-gen/complex_scalar.hpp:1958:39: error: call to 'pow' is ambiguous
      stan::model::assign(gq_complex, stan::math::pow(z, y),
                                      ^~~~~~~~~~~~~~~
/mnt/nvme1n1/workspace/stanc3_PR-1011/performance-tests-cmdstan/cmdstan/stan/lib/stan_math/stan/math/rev/fun/pow.hpp:213:26: note: candidate function [with T = double, $1 = void]
inline std::complex<var> pow(const std::complex<var>& x,
                         ^
/mnt/nvme1n1/workspace/stanc3_PR-1011/performance-tests-cmdstan/cmdstan/stan/lib/stan_math/stan/math/rev/fun/pow.hpp:251:26: note: candidate function [with T = double, $1 = void]
inline std::complex<var> pow(std::complex<T> x, const std::complex<var>& y) {
                         ^
/mnt/nvme1n1/workspace/stanc3_PR-1011/performance-tests-cmdstan/cmdstan/stan/lib/stan_math/stan/math/rev/fun/pow.hpp:199:26: note: candidate function
inline std::complex<var> pow(const std::complex<var>& x,
                         ^
../../test/integration/good/code-gen/complex_scalar.hpp:1961:39: error: call to 'pow' is ambiguous
      stan::model::assign(gq_complex, stan::math::pow(z, gq_r),
                                      ^~~~~~~~~~~~~~~
```

rok-cesnovar:
Hm, that would be complex pow. Ok, that i didnt actually account for in my PR.

WardBrian:
@rok-cesnovar this still needs math changes right? #1067 isn't enough?

rok-cesnovar:
Yeah, pow(complex) and friends are missing for primitive inputs. Sorry, was a busy week and havent gotten back to this.

WardBrian:
No worries, just wanted to ask before trying to run the tests again!

rok-cesnovar:
All models compiled 🎉  Finally.

WardBrian:
@rok-cesnovar what's the necessary merge order between this and the stan-math PR?

rok-cesnovar:
> what's the necessary merge order between this and the stan-math PR?

stanc3 PR should be merged first and then the Math one should be merged as soon as possible. No other Math/Stan/Cmdstan PR will pass between merging the stanc3 PR and the Math PR.

The Math PR is ready, it won't pass the upstream tests until we merge this PR. The next step would be to get both PRs approved then I can handle the merging procedure.

rok-cesnovar:
If Steve is not available or busy, happy to review this one. This is in my Ocaml/stanc3 comfort zone :)

src/stan_math_backend/Expression_gen.ml
```diff
@@ -223,75 +216,68 @@ and pp_scalar_binary ppf scalar_fmt generic_fmt es =
 
 and gen_operator_app = function
   | Operator.Plus ->
-      fun ppf es -> pp_scalar_binary ppf "(%a@ +@ %a)" "add(@,%a,@ %a)" es
+      fun ppf es ->
+        pp_scalar_binary ppf "(%a@ +@ %a)" "stan::math::add(@,%a,@ %a)" es
```
rok-cesnovar, line 220: I am not sure this special casing is needed at all anymore, as we also have a definition of add for scalars: https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/add.hpp

The same goes for all operators I think. If any of them are missing its probably just oversight rather than intention.

rok-cesnovar (approved):
Looks good to me. One minor comment on the special casing operators.

src/stan_math_backend/Expression_gen.ml
WardBrian: I think you are right, with a few exceptions like transpose which don't seem to have overloads. That said, it ends up being very much worth it IMO, to avoid things like

` pos - 1` turning into `stan::math::minus(pos,-1)` in the generated code. I've cleaned up this code a little though

rok-cesnovar: Oh yeah, that would be verbose and ugly. Good call.

rok-cesnovar:
All good. Now we need the Math PR approved and we should be good to go.

Issue #974, opened 2021-09-25
Move the docker images used in CI to a Stan account


rok-cesnovar:
We currently use the images on @andrjohns's personal account on Dockerhub. We should move those to a
Stan account. We probably need to first create a Stan account on Dockerhub.

serban-nicusor-toptal:
I managed to get access to some passwords in LastPass, will check today to see if I have access to [stangroup](https://registry.hub.docker.com/u/stangroup)

serban-nicusor-toptal:
I've checked `stangroup` which is something I don't have access to, the emails that I have do not own it. I'm not entirely sure who to ask about it. @rok-cesnovar any idea who might have set that up?
In case we don't figure that out I've created [stanorg](https://registry.hub.docker.com/u/stanorg) that we can use.

PR #1017, opened 2021-10-27
Move docker images used in CI to stanorg account

serban-nicusor-toptal:
#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Replaces all references to Docker Hub for andrjohns images with stanorg.
#974 

## Copyright and Licensing


serban-nicusor-toptal:
I'll build and push the rest to `stanorg`.
Does it make any sense to set up and CI to build the images when there are changes to `dockerfile/windows` for example? I would do it so we have less to worry about.

WardBrian:
If we’re building the images would we still want to maintain a docker folder here? I think just the scripts would be enough them?

serban-nicusor-toptal:
I would say we need it as it defines the build steps of the docker build process, else I would either separate them into another repo or template them on the fly.

WardBrian:
Feels like they could live in the ci-scripts repo if we moved to that system. It can be a separate PR though

serban-nicusor-toptal:
`debian` and `debian-windows` pushed to Docker Hub.
If these images are used only in the CI, we can surely move them to ci-scripts ( will do another PR ).

WardBrian:
Great!

The follow on PR can also remove the comment in stanc.ml which is great.

WardBrian:
Any merit to leaving that comment in during the merge to test all the non-x86 binaries?

serban-nicusor-toptal:
Well we can do it to be extra sure that everything works right after the merge.

WardBrian:
If we're doing a follow on PR to remove the dockerfiles (them being moved to ci-scripts) we can remove it then, yes. Otherwise this seems good to me.

PR #1021, opened 2021-11-03
Better error messages for incomplete probability calls

### LINKED ISSUE #987: descriptive error messages for incomplete probability library calls
# Description

Our probability-distribution functions come in familes, e.g., `normal_X()`, where `X` can be any of
`lpdf`, `lupdf`, `cdf`, `lcdf`, `lccdf`, `rng`.  But not all probability distributions implement each expected probability function.  When a user calls `foo_X` for a probability distribution `foo` that exists, but `foo_X` doesn't exist, we should have a more specific message informing them of that so they don't think they've made a typo or waste time looking in the doc.

# Motivation 

Quoted from @betanalpha's comment https://github.com/stan-dev/design-docs/pull/42#issuecomment-932368377:

> Most of the probability libraries in R, Python, Mathematica, etc have a pretty uniform coverage of the exposed representations.  One you know the name of a family of distributions you know how to call the corresponding density, CDF, quantile function, rng, etc.  At the point the coverage in Stan is pretty uniform, but it’s becoming less and less uniform as more obscure families are added without all of the functionality.  If a user is assuming that everything will be uniformly available but gets an error message then can become frustrating assuming that there’s a spelling error or other typo somewhere and not just a non-implemented functionality.

> It would be great to add functionality to the compiler to fill in these gaps so that if a user calls for some functionality that has not yet been implemented then they get a precise error message noting that gap and not the default error message noting the the function being requested just doesn’t exist.

adamhaber:
Fixes #987 . 

Not sure what would be the best way to test this. I've built and ran stanc on 
```stan
data { 
}
model {  
    // known family, known suffix, not implemented  
    target += von_mises_cdf(1|0,1);
    target += von_mises_lcdf(1|0,1);
    target += von_mises_lccdf(1|0,1);

    target += dirichlet_cdf([0.2, 0.1, 0.7]|[1,1,1]);
    target += dirichlet_lcdf([0.2, 0.1, 0.7]|[1,1,1]);
    target += dirichlet_lccdf([0.2, 0.1, 0.7]|[1,1,1]);
}
```
and I'm getting the right error message but only on the first line.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Improved error messages for incomplete probability library calls.

## Copyright and Licensing


rok-cesnovar:
> and I'm getting the right error message but only on the first line.

Isn't that what happens also if you use 3 non-existent functions? Won't it only error for the first one?

src/middle/Semantic_error.ml
```diff
@@ -168,6 +169,13 @@ module TypeError = struct
           "A returning function was expected but an undeclared identifier \
            '%s' was supplied."
           fn_name
+    | ReturningFnExpectedUndeclaredDistSuffixFound (prefix, suffix) ->
+        Fmt.pf ppf
+          "A function was expected but an unknown identifier %s was recieved. \
+           This appears to be part of the %s family of distributions, for \
+           which the %s suffix in not implemented."
```
rok-cesnovar, line 176: @bob-carpenter @betanalpha or anyone else with more experience working with end-users. What would be the best way to describe this error of a missing implementation for a function of a distribution? Not saying this one is not good, just wanted to get more input.

adamhaber:
> Isn't that what happens also if you use 3 non-existent functions? Won't it only error for the first one?

Definitely! Just wanted to ask what would be a good way to test, in light of this (since I can't throw all the different cases in one large file) :-)

rok-cesnovar:
I guess a few short test files, each with a faulty signature in a subfolder should be fine.

src/frontend/Semantic_check.ml
```diff
@@ -290,9 +290,30 @@ let semantic_check_fn_normal ~is_cond_dist ~loc id es =
     | Some _ ->
         (* Check that Funaps are actually functions *)
         Semantic_error.returning_fn_expected_nonfn_found loc id.name |> error
-    | None ->
-        Semantic_error.returning_fn_expected_undeclaredident_found loc id.name
-        |> error)
+    | None -> (
+      match List.rev (String.split id.name ~on:'_') with
+      | (("lupdf" | "cdf" | "lcdf" | "lccdf" | "rng") as suffix) :: tl ->
```
rok-cesnovar, line 295: This should probably be another constant in https://github.com/stan-dev/stanc3/blob/b0f2bd3de4e19ed9fb57e556cc4ba539af234195/src/middle/Utils.ml#L11

I guess none of the ones currently there are useful.

You can also skip `lupdf`. If there is an lpdf/lpmf implementation, that automatically means a lupdf/lupmf is available. If it weren't, we also couldn't use the lpdf/lpmf in a sampling statement (with the tilde).

rok-cesnovar:
A question on which I don't have an answer to, is if we envision a case where one of the non-lpdf/lpmf suffixed functions is implemented, but we don't have an lpdf/lpmf function implementation. So, things like there is a foo_rng but no foo_lpdf. 

We don't right now and that would probably be bad anyway, so I don't think we need to handle it.

rok-cesnovar (approved):
Outside of the error message, for which I would like someone else's input, this all looks good.

src/middle/Semantic_error.ml
nhuurre: It's not really an unknown identifier when it matches the expected pattern. How about simply
```ocaml
"Function '%s_%s' is not implemented for distribution '%s'." prefix suffix prefix
```

nhuurre:
> a case where one of the non-lpdf/lpmf suffixed functions is implemented, but we don't have an lpdf/lpmf function implementation.

That would be very bad, I agree. It would also impact twiddle statements.

I think it's worth special-casing `_lpdf` vs `_lpmf` so that if e.g. someone writes `binomial_lpdf` then the error says you must use `binomial_lpmf` instead.

rok-cesnovar:
> I think it's worth special-casing _lpdf vs _lpmf so that if e.g. someone writes binomial_lpdf then the error says you must use binomial_lpmf instead.

Oh nice, yeah. That would also be great, also include lupdf vs lupmf if we add that.

WardBrian:
Our of curiosity, what happens if there’s a user defined function `bar_lpdf` and a user tries to call `bar_rng`?

Also, I think the original issue suggested listing the suffixes which _are_ available, do we still want that?

nhuurre:
Oh and one more thing before I forget: `Stan_math_signatures.distributions` list does not contain all the Stan math distributions.
Take for example `categorical`.
Ok, `categorical_lpmf` is from `distributions`
https://github.com/stan-dev/stanc3/blob/b0f2bd3de4e19ed9fb57e556cc4ba539af234195/src/middle/Stan_math_signatures.ml#L222-L223
However, note that the above suggest `categorical_rng` does not exist but it is actually added later 
https://github.com/stan-dev/stanc3/blob/b0f2bd3de4e19ed9fb57e556cc4ba539af234195/src/middle/Stan_math_signatures.ml#L977-L983
where we also have `categorical_logit_glm` which is completely absent in `distributions`.

Many more distribution functions are scattered around the file. These should all be consolidated in the `distributions` list if at all possible.

adamhaber:
> Our of curiosity, what happens if there’s a user defined function bar_lpdf and a user tries to call bar_rng?

For:
```stan
functions {
    real bar_lpdf(real x) {
        return(x);
    }
}
model {  
    target += bar_lcdf(1|0,1);
}
```
I get:
```
Semantic error in './tmp.stan', line 7, column 14 to column 29:
     5:  }
     6:  model {  
     7:      target += bar_lcdf(1|0,1);
                       ^
     8:  }

A returning function was expected but an undeclared identifier 'bar_lcdf' was supplied.
```

> Also, I think the original issue suggested listing the suffixes which are available, do we still want that?

Another option would be to add a link to the functions-reference docs (I personally like compiler errors that do that), but I'd be happy to add whatever people think is right.

src/frontend/Semantic_check.ml
```diff
@@ -290,9 +290,45 @@ let semantic_check_fn_normal ~is_cond_dist ~loc id es =
     | Some _ ->
         (* Check that Funaps are actually functions *)
         Semantic_error.returning_fn_expected_nonfn_found loc id.name |> error
-    | None ->
-        Semantic_error.returning_fn_expected_undeclaredident_found loc id.name
-        |> error)
+    | None -> (
+        let gen_prefix l = String.concat ~sep:"_" (List.rev l) in
+        match List.rev (String.split id.name ~on:'_') with
+        | (("lpmf" | "lupmf") as suffix) :: tl
```
WardBrian, line 296: Following #995 (which I'm hoping gets merged today) I think this can be made simpler with this:

https://github.com/WardBrian/stanc3/blob/31e04c11a9b84c42955656d477849f7e2fff3741/src/middle/Utils.ml#L22-L28

I think this also avoids the need for the new lists in `Utils`

src/frontend/Semantic_check.ml
```diff
@@ -290,9 +290,45 @@ let semantic_check_fn_normal ~is_cond_dist ~loc id es =
...
+          when Utils.is_distribution_name
```
WardBrian, line 297: I think this excludes the (deprecated) `cdf_log` forms? Which I think we should include. A test to add:

```stan
model {
    target += von_mises_ccdf_log(1, 0,1);
}
```

test/integration/bad/missing_dist_suffix/non_existing_distribution_suffix4.stan
```diff
@@ -0,0 +1,5 @@
+data { 
+}
+model {  
+    target += von_mises_notasuffix(1|0,1);
```
WardBrian, line 4:
```suggestion
    target += von_mises_notasuffix(1,0,1);
```

This is testing a different error message which involves the use of `|`, we should try to trigger the new behavior (currently this change leads to the test outputting the old "An undecleared identifier" message

WardBrian (changes requested):
This will be very nice for users I think. A few comments, and there are a few tests I'd like added:

user_defined:
```stan
functions {
   real bar_lpdf(real x){
     return 1.0;
   }
}

model {
   target += bar_lpmf(19.2);
}
```
user_defined_no_rng
```stan
functions {
   real bar_lpdf(real x){
     return 1.0;
   }
}

generated quantities {
   real x = bar_rng(10);
}
```
(also, if there are any builtin distributions without RNG's we should test one. Not sure)

user_defined_no_cdf:
```stan
functions {
   real bar_lpdf(real x){
     return 1.0;
   }
}

model{
  target += bar_cdf(1|0);
}
```

and something that tests the older/deprecated _cdf_log/_ccdf_log, like

```stan
model {
    target += von_mises_ccdf_log(1, 0,1);
}
```


These currently output a variety of things, and we should discuss what we want them to say

WardBrian:
Here's a test case for a distribution w/o rng suffix:
```stan
generated quantities {
   real x = multi_gp_cholesky_rng([[0,0],[0,0]], [1,0]);
}
```

WardBrian:
Partially because of @nhuurre's comment above and partially because of the output for user-defined functions, I wonder if we just shouldn't test if the function is a 'known' distribution family if it has a known suffix? It would mean userdefined functions would also get the output, but maybe that's good?

nhuurre:
> if we just shouldn't test if the function is a 'known' distribution family if it has a known suffix?

From the original issue
> can become frustrating assuming that there’s a spelling error or other typo somewhere and not just a non-implemented functionality.

It would be confusing the other way too, if a legit typo gets reported as unimplemented functionality.

WardBrian:
Good point. I do wonder if the user-defined family should be handled though, like if someone defines bar_lpdf should we give this error when they try to use bar_lcdf?

src/frontend/Semantic_check.ml
adamhaber: OK, will wait!

adamhaber:
@WardBrian added the tests you've requested, I _think_ the behaviour is expected but would appreciate an extra look at this.

WardBrian:
#995 is in, which hopefully lets you handle the suffixes a bit nicer. Looking at the tests now

test/integration/bad/missing_dist_suffix/stanc.expected
```diff
@@ -1,3 +1,14 @@
...
+Semantic error in 'deprecated_suffix.stan', line 4, column 14 to column 40:
+   -------------------------------------------------
+     2:  }
+     3:  model {
+     4:      target += von_mises_ccdf_log(1, 0,1);
+                       ^
+     5:  }
+   -------------------------------------------------
+
+A returning function was expected but an undeclared identifier 'von_mises_ccdf_log' was supplied.
```
WardBrian, line 11: This should display the same sort of message it does for `von_mises_lccdf`, IMO

test/integration/bad/missing_dist_suffix/stanc.expected
```diff
@@ -58,9 +79,42 @@ Semantic error in 'non_existing_distribution_suffix4.stan', line 4, column 14 to
...
+Semantic error in 'user_defined.stan', line 7, column 13 to column 27:
+   -------------------------------------------------
+     5:  }
+     6:  model {
+     7:     target += bar_lpmf(19.2);
+                      ^
+     8:  }
+   -------------------------------------------------
+
+Function 'bar_lpmf' is not implemented for distribution 'bar', use 'bar_lpdf' instead.
```
WardBrian, line 98: It's interesting to me that this one is caught but the other user defined functions are not. What happens if you test this same model but without the functions block, just the model block. Does it still tell you to use the (now nonexistant) `bar_lpdf`?

src/frontend/Typechecker.ml
```diff
@@ -394,7 +394,40 @@ let check_fn ~is_cond_dist loc tenv id es =
...
+          let is_known_family s =
+            List.mem known_families s ~equal:String.equal
+          in
+          match suffix with
+          | ("lpmf" | "lumpf")
+            when Utils.is_distribution_name (prefix ^ "_lpdf") ->
+              Semantic_error.returning_fn_expected_wrong_dist_suffix_found loc
+                (prefix, suffix)
+          | ("lpdf" | "lumdf")
+            when Utils.is_distribution_name (prefix ^ "_lpmf") ->
```
WardBrian, line 413: This conditional is too weak. See this test model (should add to tests):

completely_undefined:
```stan
model {
  target += foo_lpdf(1);
}
```
Current test output:
```
Semantic error in 'test.stan', line 2, column 12 to column 23:
     1:  model {
     2:    target += foo_lpdf(1);
                     ^
     3:  }

Function 'foo_lpdf' is not implemented for distribution 'foo', use 'foo_lpmf' instead.
```

I think `Utils.is_distribution_name` is checking if that is a well-formed name, _not_ checking that the function is defined. For that, the simplest check is to use `Env.mem`. That will still get fooled by something like

```stan
data {
   int foo_lpmf;
}
model {
  target += foo_lpdf(1);
}
```

but IMO this is probably okay, the user is being intentionally obtuse here.

WardBrian (commented):
Looks much cleaner now! I think there is one remaining issue we should fix before merging

src/frontend/Typechecker.ml
adamhaber: Added

WardBrian (approved):
Thanks @adamhaber !

PR #1022, opened 2021-11-03
Support .stanfunctions as an extension for files containing only function definitions/declarations

WardBrian:
This was spawned by a discussion with @spinkney and @rok-cesnovar on Slack a few weeks ago. In particular, our ability to pretty-print code is really nice, and it would be great to use to format standalone Stan function files, which are often used in programs like 
```stan
functions {
  //possibly include decls for clarity
  #include<functions.stanfunctions>
}
```

The changes required to support these .stanfunctions are very minimal. The parser interface allows you to do it with just one extra nonterminal symbol, and the rest of the compiler can treat it like it is more or less a standalone functions block. The only special casing needed is in the pretty-printer, which now has the ability to print a 'bare' block (without the title of the block and {, }).

While it is a simple change, I think it's debatable how useful it would be so some discussion would be good. 

While I was working on this, I also extended `--auto-format` and `--print-canonical` to work with the `--o` flag, so you can now run

`stanc --auto-format prog.stan -o prog.stan`, which I think is useful even outside the context of this extension

(NB this currently fails if you do `stanc --auto-format prog.stan > prog.stan`)

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/434

## Release notes

The compiler can now compile or format standalone function definitions in a `.stanfunctions` file. These are compiled as if a normal Stan program was compiled with `stanc3 --standalone-functions`.

## Copyright and Licensing


rok-cesnovar:
I want to reiterate that I think this feature is useful for projects with many included files, which are also the same projects that may want to enforce formatting.

> stanc --auto-format prog.stan -o prog.stan, which I think is useful even outside the context of this extension

This is great and useful!

WardBrian:
@bgoodri This is the stanc3 PR mentioned in the meeting today. Does `"functions-only"` work for you as a flag name for this behavior?

Here is the StancJS test file I just created showing the behavior:

```js
var stanc = require('../../src/stancjs/stancjs.bc.js');
var utils = require("./utils/utils.js");

// NB: no "functions { ... }"
var functions_only = `
real my_log1p_exp(real x) {
    return log1p_exp(x);
}

real array_fun(real[] a)
{
    return sum(a);
}

real int_array_fun(int[] a)
{
    return sum(a);
}

vector my_vector_mul_by_5(vector x)
{
    vector[num_elements(x)] result = x * 5.0;
    return result;
}

int int_only_multiplication(int a, int b) {
    return a*b;
}

real test_lgamma(real x) {
    return lgamma(x);
}

// test special functions
void test_lp(real a) {
    a ~ normal(0, 1);
}

real test_rng(real a) {
    return normal_rng(a, 1);
}

real test_lpdf(real a, real b) {
    return normal_lpdf(a | b, 1);
}
`
// compiles as if 'standalone-functions' was set on a normal model
let basic = stanc.stanc("functions_only", functions_only, ["functions-only"]); 
utils.print_result(basic)
utils.print_error(basic)

// just parses and pretty-prints the same style (e.g. no "functions {")
let basic_fmt = stanc.stanc("functions_only", functions_only, ["functions-only", "auto-format"]);
utils.print_result(basic_fmt)

// parses, type checks (prereq for canonicalizing), prints canonical (once again, no "functions {")
let basic_canon = stanc.stanc("functions_only", functions_only, ["functions-only", "print-canonical"]);
utils.print_result(basic_canon)
```

If this sounds good to you I can start writing the doc for this behavior and then we can merge this

WardBrian:
Docs PR available here: https://github.com/stan-dev/docs/pull/434. I could only really think of two places in the SUG to document the feature, happy to add more if people can think of them.

bgoodri:
I think this is fine, particularly since we already have the functionality for exposing user-defined Stan functions in R, so allowing parsing to be done without an explicit `functions` block seems reasonable. As far as using it in an automated way, I think R would grep for any `#include` statements and if so, call this thing on that line's path. Then, if all of those succeed, call `stanc` in the regular way after calling the preprocessor on the parent Stan program. AFAICT, that would be straightforward with this feature.

WardBrian:
I think what you describe would work inside the functions block itself, if you see an include. In general, includes could also contain whole blocks etc, but those should be able to be parsed as just a mini program

test/stancjs/stancjs.expected
```diff
@@ -53,6 +53,448 @@ Identifier 'z' not in scope.
 Semantic error in 'string', line 6, column 4 to column 5: 
 Identifier 'z' not in scope.
 
+$ node functions-only.js
```
rok-cesnovar, line 56: I tried to avoid adding generated C++ in the Javascript tests because that essentially makes `node` a dependency for all stanc3 developers and we should then just make JS tests run on dune runtest.

I am guessing just the fact that this doesn't error out is enough? 

If you think we can't avoid it, then I guess we roll with it.

rok-cesnovar (changes requested):
One thing around stancjs tests, otherwise LGTM.

test/stancjs/stancjs.expected
WardBrian: Yeah it would be enough. I'll remove it

rok-cesnovar (approved):

WardBrian:
@rok could you also look at the docs PR https://github.com/stan-dev/docs/pull/434 ?

rok-cesnovar:
You tagged the wrong Rok :), but yes definitely. Will look after lunch.

rok:
LGTM :)

PR #1026, opened 2021-11-04
Warn for matrix^pow, suggest .^ or matrix_power()

### LINKED ISSUE #604: Disambiguating vectorised pow operator from matrix pow operator

WardBrian:
Replacement for #994, closes #604 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Emit warning for `matrix^scalar` and point users to `.^` and `matrix_power()`.

## Copyright and Licensing

### DIFF
--- a/src/frontend/Typechecker.ml
+++ b/src/frontend/Typechecker.ml
@@ -603,7 +603,7 @@ and check_expression cf tenv ({emeta; expr} : Ast.untyped_expression) :
                  operator %/%."
             in
             add_warning x.emeta.loc s
-        | UMatrix, (UInt | UReal), Pow ->
+        | (UArray UMatrix | UMatrix), (UInt | UReal), Pow ->
             let s =
               Fmt.strf
                 "@[<v>@[<hov 0>Found matrix^scalar:@]@   @[<hov \
@@ -614,7 +614,7 @@ and check_expression cf tenv ({emeta; expr} : Ast.untyped_expression) :
                  warning by using elementwise operator .^"
                 Fmt.text
                 "If you intended matrix exponentiation, use the function \
-                 matrix_power() instead."
+                 matrix_power(matrix,int) instead."
             in
             add_warning x.emeta.loc s
         | _ -> ()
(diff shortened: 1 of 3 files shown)


src/frontend/Typechecker.ml
```diff
@@ -603,9 +603,23 @@ and check_expression cf tenv ({emeta; expr} : Ast.untyped_expression) :
                  operator %/%."
             in
             add_warning x.emeta.loc s
+        | UMatrix, (UInt | UReal), Pow ->
```
nhuurre, line 606: I'd at least add `UArray UMatrix` here. Emitting a good warning in every case is hard because `^` is fully vectorized so that even `matrix ^ matrix` compiles.

nhuurre (changes requested):

src/frontend/Typechecker.ml
WardBrian: I don't believe `matrix_power()` is overloaded for arrays of matrices (if it is, it isn't documented as such). So the suggestion wouldn't be 100% a drop-in replacement in that case.

nhuurre: It's not available for `UReal` exponent either. The warning still guards against misinterpretation.

WardBrian: Good point. I'll extend the case, and specifically say "function matrix_power(matrix,int)" for maximum clarity in the hint.

nhuurre (approved):
Looks good.

PR #1027, opened 2021-11-08
Function overloading

WardBrian:
This PR enables function overloading of user defined functions.

Current status:

- [x] Overloading of UDFs
- [x] Overloading a stan_math function name: Needs the changes in #1011
- [x] Passing overloaded functions to map_rect, ODEs, etc: Should work but needs testing
- [x] Properly sort to fewest promotions when there are multiple options

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/473

## Release notes

User defined functions can now be overloaded. Multiple definitions of the same function name are allowed if the arguments are different in each definition.

## Copyright and Licensing


spinkney:
Hey Brian, when do you expect this pr to be merged?

WardBrian:
After the holidays we should be able to merge #1011 and then I’ll take this back up

spinkney:
👍 thanks! I think I want to wait until announcing the "grand opening" of the helpful repo until this is in. It will be nice to have this multiple dispatch for vectorized versions of functions

WardBrian:
Code generation for this PR should now be in a good place (thanks @rok-cesnovar for all the C++ help). 

There is a very tricky issue remaining on the typechecker level, which has to do with the higher order functions like `reduce_sum`. At the moment, a call

`reduce_sum(foo, ....)` 

just naively looks up the identifier `foo` in the function table and then typechecks it. This is a problem when there are multiple values for `foo`, some of which are valid and others are may not be, and it is tricky to resolve. Currently it uses the most recently defined signature, but obviously this isn't desired behavior. The typechecker gives a pretty uninformative error too, reporting that the wrong number of arguments has been provided.

It's obviously best if I can make it so this 'just works', but if not I need to gracefully disallow overloaded functions as higher-order arguments

WardBrian:
I've added a test which shows the problem at the moment. If you disable the typechecker and compile the model, you get this .hpp:

https://gist.github.com/WardBrian/9c947f405a641df664e4bb676b2a2114

But current behavior is that the typechecker blocks it.

rok-cesnovar:
Example of a reduce_sum model that would run and work with shadowing allowed:

```stan
functions {
  real fun(array[] real y_slice, int start, int end, real m) {
    return sum(y_slice) * m;
  }

  real fun(array[] real y_slice, int start, int end) {
    return sum(y_slice);
  }
}
transformed data {
  int N = 100;
  array[N] real data_y = ones_array(N);

  real sum_1 = reduce_sum(fun, data_y, 1);
  print(sum_1);
  real sum_2 = reduce_sum(fun, data_y, 1, 5);
  print(sum_2);
}
parameters {
   real y;
}
transformed parameters {
   array[N] real param_y = ones_array(N);

   real p_sum_1 = reduce_sum(fun, param_y, 1);
   print(y, " - ", p_sum_1);
   real p_sum_2 = reduce_sum(fun, param_y, 1, y);
   print(y, " -- ", p_sum_2);
}
model {
   y ~ std_normal();
}
```

WardBrian:
Here's what remains:

1. Write a test for variadic odes with overloaded functions
2. Write a test using map_rect with overloading functions
2. Support map_rect having an overloaded function in typechecking
3. I'd love to clean up how the typechecking is done once it all works, especially if we ever plan on having user-defined higher order functions, we'd need a more general approach for that than we currently have. That might be left for later PRs though.

WardBrian:
This needs a few more tests, especially for higher-order functions, but otherwise I'm happy with where it is right now.

Leaving a note for posterity: 

Variadic functions in the typechecker will search for the appropriate function. This logic could be extended to 'regular' functions, but currently the only other higher order function is `map_rect`, which only has one (and exactly one) signature it can accept, so the old logic works fine.

This will need to change if/when we allow user-defined higher order variadic functions. We're a long way off from such a day, so I didn't go through the effort here, but the code here would be able to extend to that case when the time comes (I hope).

rok-cesnovar:
For ODEs making actually runnable examples is a bit more of a challenge. I would just use https://github.com/stan-dev/example-models/blob/master/knitr/convert-odes/sir.stan and https://github.com/stan-dev/example-models/blob/master/knitr/convert-odes/sir.data.R

and then make another version of the function with an unused variable:

```stan
 vector simple_SIR(real t, vector y, real beta,
                    // water contact rate
                    real kappa, // C_{50}
                    real gamma, // recovery rate
                    real xi, // bacteria production rate
                    real delta,
                    int unused) {
```

WardBrian:
@rok-cesnovar that works, and I can confirm it works!

WardBrian:
@SteveBronder if you have a chance this should be ready for review.

We may want to do #1092 first, after which I will just make the same changes to that as the variadic ODEs required here.

WardBrian:
Marked as draft because I realized we need to specifically handle promotions for cases like 

```stan
functions {
   real foo(int x, real y){
     return y;
   }

   real foo(real x, int y){
     return y;
   }
}

model {
   print(foo(1,2));
}
```

WardBrian:
I think this is ready for review, I'm going to pivot to writing doc. 

The PR has two main 'sections'. It really only touches a few code files, the rest of the changes are all in tests/.

1. Typechecking needed to change a lot. I'd appreciate @nhuurre and @SteveBronder both looking at this part. Overloading had the following requirements:
    - A FunDef needs to define either 1) a signature for a previously empty forward declaration or 2) a unique signature (unique by more than just the return type)
    - A function call needs to check multiple signatures. We were doing 99% of this before this PR, because there are a lot of overloads in the math library. 
     The change here was primarily that there needs to be **one**, *unique* minimum for type promotions. For example, you can't call a function with two only overloads `(int,real)` and `(real,int)` with types `(int,int)`.  
    - `reduce_sum` and the variadic ODEs need to _search_ for a valid signature, not just assume they're passed one by the variable in the arguments. Notably, `map_rect` does not need to do this, since there's only ever one signature it would expect.

2. Codegen. This is conceptually pretty straightforward: Rather than printing functors immediately, they're collated into a table and printed all at once after the entire functions block is processed. I'd appreciate if @SteveBronder could take a look at this

test/integration/bad/function-signatures/overloading/stanc.expected
```diff
@@ -0,0 +1,159 @@
...
+(real, int) => real
+  No unique minimum promotion found. Overloaded functions must not have
+  multiple equally valid promotion paths, but this function call has at least
+  two:
+  [int_to_real, None]
+  and
+  [None, int_to_real]
+  Consider defining a new signature for the exact types needed or re-thinking
+  existing definitions.
+(Additional signatures omitted)
```
nhuurre, line 23: These errors just look weird. It should be something more like
```
Arguments with ambiguous types supplied to function 'foo':
(int, int)
No unique minimum promotion found. Overloaded functions must not have
multiple equally valid promotion paths but this function has signatures:
(real, int) => real
(int, real) => real
Consider defining a new signature for the exact types needed or re-thinking
existing definitions.
```

src/frontend/SignatureMismatch.ml
```diff
@@ -70,9 +70,16 @@ and details =
...
+and promotions =
+  | None
+  | IntToRealPromotion
+  | IntToComplexPromotion
+  | RealToComplexPromotion
+
 and function_mismatch =
   | ArgError of int * type_mismatch
   | ArgNumMismatch of int * int
+  | PromotionConflict of promotions list * promotions list
```
nhuurre, line 82: I don't think `PromotionConflict` belongs here. It's not the same kind of error as the others.

src/frontend/SignatureMismatch.ml
```diff
@@ -170,34 +179,56 @@ let promote es promotions =
...
+           UnsizedType.compare_returntype x y ) in
+  let matches, errors =
+    List.partition_map function_types
+      ~f:(fun (rt, tys, funkind_constructor, _) ->
+        match check_compatible_arguments 0 tys args with
+        | Ok p -> Either.First ((rt, tys), funkind_constructor, p)
+        | Error e -> Second ((rt, tys), e) ) in
+  match unique_minimum_promotion matches with
+  | Ok ((rt, _), funkind_constructor, p) -> Ok (rt, funkind_constructor, p)
+  | Error (Some e) -> Error ([e], true)
```
nhuurre, line 225: Returning a single signature here is kind of wrong; this particular signature is acceptable but the problem is that there are many such signatures available.
Let's change the return type of `matching_function` so that it can return multiple _matching_ signatures and then handle that as a new error in the typechecker.
```ocaml
type match_result =
  | UniqueMatch of
      UnsizedType.returntype
      * (bool Middle.Fun_kind.suffix -> Ast.fun_kind)
      * promotions list
  | AmbiguousMatch of (UnsizedType.t list * UnsizedType.returntype) list
  | Errors of signature_error list * bool
```

src/frontend/SignatureMismatch.ml
```diff
@@ -170,34 +179,56 @@ let promote es promotions =
...
-         Error (errors, not (List.is_empty omitted)) )
+  let unique_minimum_promotion ps =
+    let size (_, _, p) =
+      List.fold ~init:0 ~f:(fun acc p -> acc + promotion_cost p) p in
+    let ps =
+      List.sort ~compare:(fun p1 p2 -> Int.compare (size p1) (size p2)) ps in
+    match ps with
+    | (((rt, tys), _, p1) as ans1) :: ((_, _, p2) as ans2) :: _
+      when size ans1 = size ans2 ->
+        Error (Some ((rt, tys), PromotionConflict (p1, p2)))
```
nhuurre, line 209: Instead of sorting and checking the first two items let's compute the minimum and then select all the items equal to the minimum.

nhuurre (changes requested):

test/integration/bad/function-signatures/overloading/stanc.expected
WardBrian: Done, I went with a very similar style:

```
Semantic error in 'no_minimum_promotion3.stan', line 21, column 9 to column 20:
    19:  
    20:  model {
    21:     print(foo3(1,2,3));
                  ^
    22:  }

No unique minimum promotion found for function "foo3".
Overloaded functions must not have multiple equally valid promotion paths.
For args (int, int, int), this function has several:
(int, int, real) => real
(real, int, int) => real
(int, real, int) => real
Consider defining a new signature for the exact types needed or
re-thinking existing definitions.
```

WardBrian:
Thanks for the suggestions @nhuurre - I think that it's much better now

src/frontend/Semantic_error.ml
```diff
@@ -134,6 +140,22 @@ module TypeError = struct
                   , args )
                 , error ) ]
             , false ) )
+    | AmbiguousFunctionPromotion (s, arg_tys, signatures) ->
+        let pp_sig ppf (rt, args) =
+          Fmt.pf ppf "@[<hov>(@[<hov>%a@]) => %a@]"
+            Fmt.(list ~sep:comma UnsizedType.pp_fun_arg)
+            args UnsizedType.pp_returntype rt in
```
nhuurre, line 147: This is a bit simplistic. `Signature_mismatch.pp_signature_mismatch` takes extra effort to make higher-order function signatures readable. But of course we don't have user-defined higher-order functions so it probably doesn't matter that much.

nhuurre (changes requested):
Something's wrong with variadic argument checking. This model crashes:
```stan
functions {
  real foo(real[] x, int s, int end, real z, int k) {
    return 0.0;
  }
  real foo(real[] x, int s, int end, int z, real k) {
    return 0.0;
  }
}
transformed data {
  real x = reduce_sum(foo, {1,2,3,4,5,6}, 1, 2, 3);
}
```

src/frontend/Semantic_error.ml
WardBrian: Yeah, one disadvantage of moving this out of SignatureMismatch was losing that functionality. I suppose we could expose that as a more general function and use it here, but for now I think it’s fine?

That said, if the solution for your failing example is what I suspect, I may have to do it now to get good errors for that. Every time I try to make a simplifying assumption in this PR it’s been wrong, so I should probably stop doing so

test/integration/bad/function-signatures/overloading/no_minimum_reduce_sum.stan
```diff
@@ -0,0 +1,21 @@
+functions {
+
+  real foo(array[] real x, int s, int end, real z, int k) {
+
+    return 0.0;
+
+  }
```
nhuurre, line 7: Some excess spacing here.

WardBrian:
I think I've sucessfully merged the changes from #1091, so this should be mergeable as soon as it's approved without breaking other PRs

src/stan_math_backend/Expression_gen.ml
```diff
@@ -161,6 +161,8 @@ let functor_suffix_select hof =
       variadic_ode_functor_suffix
   | x when Stan_math_signatures.is_variadic_dae_fn x ->
       variadic_dae_functor_suffix
+  | x when Stan_math_signatures.is_variadic_dae_fn x ->
+      variadic_dae_functor_suffix
```
nhuurre, line 165: Duplicate?

WardBrian: Good catch, snuck in the merge

src/frontend/Typechecker.ml
```diff
@@ -116,21 +116,26 @@ let verify_name_fresh_var loc tenv name =
...
   if
-    Stan_math_signatures.is_stan_math_function_name name
-    (* variadic functions are currently not in math sigs *)
-    || Stan_math_signatures.is_reduce_sum_fn name
+    (* variadic functions are currently not in math sigs and aren't
+       overloadable due to their separate typechecking *)
+    Stan_math_signatures.is_reduce_sum_fn name
     || Stan_math_signatures.is_variadic_ode_fn name
     || Stan_math_signatures.is_variadic_dae_fn name
+    || Stan_math_signatures.is_variadic_dae_fn name
```
nhuurre, line 127: More duplication.

nhuurre:
Almost forgot, one more question: this PR doesn't change anything in `Analysis_and_optimization.Optimize` so I assume this completely breaks function inlining?

WardBrian:
> Almost forgot, one more question: this PR doesn't change anything in `Analysis_and_optimization.Optimize` so I assume this completely breaks function inlining?

A good point I hadn't considered. Based on the comments in Optimize.ml, I'm pretty sure function inlining is broken anyway - it says

```ocaml
  (* We only add the first definition for each function to the inline map.
     This will make sure we do not inline recursive functions.
     We also don't want to add any function declaration (as opposed to
     definitions), because that would replace the function call with a Skip.
  *)
```

But recursive functions are possible without forward declarations, which I think will sneak by this and break it if the comment is to be believed.

At any rate, I can make it _no worse than present_ by changing this so it stops trying to inline a certain function if it encounters a second definition of the same function.

WardBrian:
@nhuurre I made that small change just to not cause any further issues, and opened #1096 after confirming the existing code doesn't work

nhuurre (approved):
> _no worse than present_

The standard we all should aspire to.

src/frontend/SignatureMismatch.ml
```diff
@@ -170,34 +189,62 @@ let promote es promotions =
...
       | _ -> exp )
 
-let returntype env name args =
+let promotion_cost p =
+  match p with
+  | None -> 0
+  | RealToComplexPromotion | IntToRealPromotion -> 1
+  | IntToComplexPromotion -> 2
+
+let unique_minimum_promotion ps =
```
SteveBronder, line 208: Why is the name of this input `ps`?

src/frontend/SignatureMismatch.ml
```diff
@@ -170,34 +189,62 @@ let promote es promotions =
...
+  let size (_, p) =
+    List.fold ~init:0 ~f:(fun acc p -> acc + promotion_cost p) p in
+  let sizes = List.map ~f:size ps in
+  let min_promotions = List.min_elt ~compare:Int.compare sizes in
+  let ps = List.zip_exn sizes ps in
```
SteveBronder, line 213: Can we not reuse the name of the input inside of functions? Just hard to keep track of in my mental model

src/frontend/SignatureMismatch.ml
nhuurre: `ps` is the plural of `p` which is short for `promotion`. That's my guess at least.

WardBrian: Niko is correct on the logic. You run into fun plural-of-a-plural issues with trying to describe a list of promotion*s*, but `ps` worked.

I tend to abbreviate non-globals or variables whose meaning should be clear from their type. I suppose it's the same reason most people don't write `for(int index = 0; ...` . I can do a pass on the really short names to try to clear some of them up

src/frontend/SignatureMismatch.ml
WardBrian: Fair point here, I try not to shadow if the type is also changing

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -25,6 +25,10 @@ open Statement_gen
 
 let standalone_functions = ref false
 
+type found_functor = {template: string; signature: string; defn: string}
+
+let (functors : (string, found_functor list) Hashtbl.t) = String.Table.create ()
```
SteveBronder, line 30: I'm not totally following why this has to be module level object. Can you just make it in `pp_prog` and pass it as an argument? Would we at some point want to use `Stan_math_code_gen.functors` outside of `pp_prog`?

test/integration/bad/function-signatures/overloading/stanc.expected
```diff
@@ -0,0 +1,219 @@
...
+                     ^
+    39:    }
+    40:  
+   -------------------------------------------------
+
+No unique minimum promotion found for function "chem_dae".
+Overloaded functions must not have multiple equally valid promotion paths.
+This function has several:
+(real, vector, vector, array[] real, array[] real, real, int) => vector
+(real, vector, vector, array[] real, array[] real, int, real) => vector
```
SteveBronder, line 16: Could we have the error also output what the user passed in? Like it might be nice for the error to be 
```suggestion
No unique minimum promotion found for function "chem_dae" with argument signature:
(real, vector, vector, array[] real, array[] real, int, int) => vector
Overloaded functions must not have multiple equally valid promotion paths.
This function has several:
(real, vector, vector, array[] real, array[] real, real, int) => vector
(real, vector, vector, array[] real, array[] real, int, real) => vector
```

Extremely bonus points / not necessary, but would it be possible to show in bold which arguments we had trouble detecting promotion for?

test/integration/bad/function-signatures/overloading/stanc.expected
```diff
@@ -0,0 +1,219 @@
...
+    10:  
+    11:  model {
+    12:     print(foo(1,2));
+                  ^
+    13:  }
+   -------------------------------------------------
+
+No unique minimum promotion found for function "foo".
+Overloaded functions must not have multiple equally valid promotion paths.
+For args (int, int), this function has several:
```
SteveBronder, line 31: Oh I see you do what I asked about above here, is there a reason this does not happen for the DAE solver?

test/integration/bad/function-signatures/overloading/stanc.expected
```diff
@@ -0,0 +1,219 @@
...
+Semantic error in 'redefine-cdf-same.stan', line 2, column 2 to line 4, column 3:
+   -------------------------------------------------
+     1:  functions {
+     2:    real poisson_lcdf(int n, real x) {
+           ^
+     3:      return -x^2;
+     4:    }
+   -------------------------------------------------
+
+Function 'poisson_lcdf' has already been declared to for signature (int, real) => real
```
SteveBronder, line 173: Should we tell them it is defined in the math library?

test/integration/bad/reduce_sum/stanc.expected
```diff
@@ -440,4 +457,4 @@ Semantic error in 'bad_userdefined_reduce_sum_static.stan', line 2, column 7 to
      4:    }
    -------------------------------------------------
 
-Identifier 'reduce_sum_static' clashes with Stan Math library function.
+Identifier 'reduce_sum_static' clashes with a non-overloadable Stan Math library function.
```
SteveBronder, line 460: Are there _any_ overloadable stan math library functions? idt so? If not then I'd remove the non-overloadable part of this

test/integration/good/code-gen/cpp.expected
```diff
@@ -1215,311 +1217,329 @@ static constexpr std::array<const char*, 427> locations_array__ =
...
+  operator()(const std::complex<T0__>& z, std::ostream* pstream__) const;
+};
+struct foo4_functor__ {
+std::vector<std::complex<double>>
+  operator()(std::ostream* pstream__) const;
+};
+struct foo6_functor__ {
+template <typename
+  T0__>
+  std::vector<std::complex<stan::promote_args_t<T0__>>>
```
SteveBronder: Can this be pretty printed to remove the newline?
```suggestion
template <typename T0__>
  std::vector<std::complex<stan::promote_args_t<T0__>>>
```

test/integration/good/code-gen/cpp.expected
```diff
@@ -1215,311 +1217,329 @@ static constexpr std::array<const char*, 427> locations_array__ =
...
+  operator()(const std::vector<std::complex<T0__>>& z, std::ostream*
+  pstream__) const;
+};
+struct foo5_functor__ {
+template <typename T0__> stan::promote_args_t<T0__>
+  operator()(const std::vector<std::complex<T0__>>& z, std::ostream*
+  pstream__) const;
```
SteveBronder: I'd prefer if it was pretty printed like the following

```suggestion
template <typename T0__> 
inline stan::promote_args_t<T0__> operator()(
  const std::vector<std::complex<T0__>>& z, std::ostream* pstream__) const;
```

SteveBronder: ( I think these could actually be `static` and `inline` though for static you need to remove the const on the rhs of the signature)

SteveBronder (commented):
Mostly a few comments about style but overall I think the code is good!

src/frontend/SignatureMismatch.ml
SteveBronder: > I tend to abbreviate non-globals or variables whose meaning should be clear from their type. I suppose it's the same reason most people don't write for(int index = 0; ... . I can do a pass on the really short names to try to clear some of them up

ftr it took me a minute to long to realize what `ps` is and what it's purpose was. idt we should rely heavily on the ocaml compilers type deduction in order for us to understand code. (I've seen code @nhuurre has fixed for me that for some reason needed type hinting to compile!). If we used type hinting everywhere I'd be cool with smaller names but we don't so I'd prefer more meaningful names for anything that is not trivial like an iterator.

> I tend to abbreviate non-globals or variables whose meaning should be clear from their type. I suppose it's the same reason most people don't write for(int index = 0;

I have a lot of thoughts about this sentence but I would rather keep my comment short and say idt the meaning or intent is clear from the type given by the Ocaml Compiler™ and we should either have meaningful names, type hints, or both. imo I'd also prefer at least one liner docs here briefly describing what functions we make do.

WardBrian: Are you trying to suggest ` ('a * promotion list) list -> ('a * promotion list, 'a list option) result` _isn't_ a crystal clear type?
(Obviously joking here)

I think if you spend all day reading OCaml that can start to look sensible, in the same way C++ templating magic comes naturally to people who do it a lot, and I'm probably guilty of just being too familiar with my own code.  I've changed the name to `promotion_candidates`. 

To elaborate a bit more on the original name, I think it's common for lists to be given some very short name ending in s. This is hardly a perfect convention, but I tend to read that as meaning that the fact that it _is_ a list is more important than what it is a list _of_ (whether or not this applies to this function is highly debatable). 

Anyway, when I wrote the first version of the function (which was originally only used one place, so it wasn't even part of the module at the top level) I just picked something short. When I needed to do the same thing somewhere else, I promoted the function without thinking about renaming it's argument.

src/frontend/SignatureMismatch.ml
```diff
@@ -170,34 +189,62 @@ let promote es promotions =
...
+  let matches, errors =
+    List.partition_map function_types
+      ~f:(fun (rt, tys, funkind_constructor, _) ->
+        match check_compatible_arguments 0 tys args with
+        | Ok p -> Either.First (((rt, tys), funkind_constructor), p)
+        | Error e -> Second ((rt, tys), e) ) in
+  match unique_minimum_promotion matches with
+  | Ok (((rt, _), funkind_constructor), p) ->
+      UniqueMatch (rt, funkind_constructor, p)
+  | Error (Some e) -> AmbiguousMatch (List.map ~f:fst e)
```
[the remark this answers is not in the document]
WardBrian: unique_minimum_promotion expects a `('a * promotions list ) list`. What we fill in for `'a` here is a tuple of ((return type, argument types), function constructor)

This function historically returned the return type and function constructor when it succeeded. The argument types are new, and needed for the new error message. AmbiguousMatch expects (return type, argument types), so it's the first element of the `'a` above. I'll add a short comment describing what fst does here.

src/stan_math_backend/Expression_gen.ml
```diff
@@ -78,6 +78,8 @@ let%expect_test "promote_unsized" =
     [%sexp (tests |> List.map ~f:promote_unsizedtype : UnsizedType.t list)] ;
   [%expect {| (UReal UReal (UArray UReal)) |}]
 
+let to_var s = Expr.{Fixed.pattern= Var s; meta= Typed.Meta.empty}
```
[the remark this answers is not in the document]
WardBrian: That was originally a helper function inside the huge gen_fun_app function: 

https://github.com/stan-dev/stanc3/blob/6c2bc5e1ef08512e104bc7f67ddc895909ff3e67/src/stan_math_backend/Expression_gen.ml#L314

When I split this function up it was needed in a few places so I moved it. It's mainly used to create the `pstream__` argument and extra arguments in the functors.

src/stan_math_backend/Expression_gen.ml
```diff
@@ -295,6 +297,128 @@ and gen_misc_special_math_app (f : string)
...
+and gen_functionals fname suffix es mem_pattern =
+  let contains_hof_vars = function
+    | {Expr.Fixed.pattern= Var _; meta= {Expr.Typed.Meta.type_= UFun _; _}} ->
+        true
+    | _ -> false in
+  let is_hof_call = List.exists ~f:contains_hof_vars es in
+  if not is_hof_call then None
+  else
+    Some
+      (fun ppf es ->
```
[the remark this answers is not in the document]
WardBrian: Yeah, this big chunk of code is probably going to bite us one day. There's an old comment in it asking to refactor it and for C++ devs to stop introducing functions which require special-cased argument orders, but that doesn't seem to have worked so far

test/integration/bad/function-signatures/overloading/stanc.expected
WardBrian: I don't think there is a way to do bold which works both in the console and in raw text output (e.g. in stancjs)

test/integration/bad/function-signatures/overloading/stanc.expected
WardBrian: Yeah, if the user directly calls a function it was easy to print what they supplied. For all the variadic functions however, the function being passed is called internally and which arguments it receives is variable. I could probably do some special casing for each kind of variadic function to figure it out, but it was very complicated when I tried (stuff like the optional tolerances don't get passed to the user function, for example)

test/integration/bad/function-signatures/overloading/stanc.expected
WardBrian: Yes. Easy and good for users

test/integration/bad/reduce_sum/stanc.expected
WardBrian: All non-variadic functions can be overloaded in this PR

src/stan_math_backend/Stan_math_code_gen.ml
WardBrian: Fair point, I'll move it inside

test/integration/good/code-gen/cpp.expected
WardBrian: I spent a lot of time wrestling with the pretty printer. Let me take another pass at it.


A long-term project (likely after tuples) I'm thinking about is defining a proper LIR which maps to C++ so we don't have to do all this string generation, just call `code_gen: MIR -> LIR` followed py `pp_lir: LIR -> str`

WardBrian: I've played around with the C++ output, how does it look in the latest commit?

src/stan_math_backend/Expression_gen.ml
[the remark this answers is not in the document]
SteveBronder: Aight!

test/integration/good/code-gen/cpp.expected
SteveBronder: Good!

PR #1028, opened 2021-11-09
Standardize internal compiler error messages

### LINKED ISSUE #76: Standardize errors that are the programs fault
We would like a standard utility function or a consistent policy for raising errors that we expect to never be hit. They should ideally serialize a relevant data structure in the error message. A simple version of this could be something like:

```
let fatal_error (data_structure : 'a) =
  prerr_endline ("Fatal error: this should never happen. Please file a bug on https://github.com/stan-dev/stanc3/issues/new. Offending data structure:") ;
  raise_s [%sexp (data_structure : 'a)]
```

WardBrian:
This closes #76 

It adds `Common.FatalError.fatal_error_msg`, which is a drop-in replacement for `raise_s` that prepends the following message:
`Fatal error: this should never happen. Please file a bug on https://github.com/stan-dev/stanc3/issues/new.`

and `Common.FatalError.fatal_error` which is the same as the above but doesn't take in any additional message/sexp.

We previously had a Frontend.Errors.fatal_error, but it wasn't used, and it felt like it should live in Common (rather than importing the frontend into the code generator, for example).

I also updated all existing calls to either `raise_s` or `failwith` in the compiler.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

All internal compiler errors now include a link to the stanc3 issues page.

## Copyright and Licensing

### DIFF
--- a/src/analysis_and_optimization/Mir_utils.ml
+++ b/src/analysis_and_optimization/Mir_utils.ml
@@ -241,7 +241,9 @@ let fwd_traverse_statement stmt ~init ~f =
 let vexpr_of_expr_exn Expr.Fixed.({pattern; _}) =
   match pattern with
   | Var s -> VVar s
-  | _ -> raise (Failure "Non-var expression found, but var expected")
+  | _ ->
+      Common.FatalError.fatal_error_msg
+        [%message "FNon-var expression found, but var expected"]
 
 (** See interface file *)
 let rec expr_var_set Expr.Fixed.({pattern; meta}) =
@@ -291,7 +293,9 @@ let expr_assigned_var Expr.Fixed.({pattern; _}) =
   match pattern with
   | Var s -> VVar s
   | Indexed ({pattern= Var s; _}, _) -> VVar s
-  | _ -> raise (Failure "Unimplemented: analysis of assigning to non-var")
+  | _ ->
+      Common.FatalError.fatal_error_msg
+        [%message "Unimplemented: analysis of assigning to non-var"]
 
 (** See interface file *)
 let rec summation_terms (Expr.Fixed.({pattern; _}) as rhs) =
--- a/src/analysis_and_optimization/Mir_utils.ml
+++ b/src/analysis_and_optimization/Mir_utils.ml
@@ -243,7 +243,7 @@ let vexpr_of_expr_exn Expr.Fixed.({pattern; _}) =
   | Var s -> VVar s
   | _ ->
       Common.FatalError.fatal_error_msg
-        [%message "FNon-var expression found, but var expected"]
+        [%message "Non-var expression found, but var expected"]
 
 (** See interface file *)
 let rec expr_var_set Expr.Fixed.({pattern; meta}) =
--- a/src/analysis_and_optimization/Partial_evaluator.ml
+++ b/src/analysis_and_optimization/Partial_evaluator.ml
@@ -22,7 +22,7 @@ let apply_prefix_operator_int (op : string) i =
         | "PPlus__" -> i
         | "PMinus__" -> -i
         | "PNot__" -> if i = 0 then 1 else 0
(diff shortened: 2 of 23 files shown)


src/analysis_and_optimization/Mir_utils.ml
```diff
@@ -241,7 +241,9 @@ let fwd_traverse_statement stmt ~init ~f =
 let vexpr_of_expr_exn Expr.Fixed.({pattern; _}) =
   match pattern with
   | Var s -> VVar s
-  | _ -> raise (Failure "Non-var expression found, but var expected")
+  | _ ->
+      Common.FatalError.fatal_error_msg
+        [%message "FNon-var expression found, but var expected"]
```
mcol, line 246: Typo `FNon-var`?

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -294,9 +294,12 @@ let pp_fun_def ppf Program.({fdrt; fdname; fdsuffix; fdargs; fdbody; _})
...
               ([slice; start ^ " + 1"; end_ ^ " + 1"], rest, `ReduceSum)
         | _ ->
-            raise_s
+            Common.FatalError.fatal_error_msg
               [%message
-                "Ill-formed reduce_sum call! This is bug in the compiler."]
+                "Ill-formed reduce_sum call!"
+                  ( fdargs
+                    : (UnsizedType.autodifftype * string * UnsizedType.t) list
+                    )]
```
nhuurre, line 302:
```suggestion
                "Ill-formed reduce_sum call!" (fdargs : Program.fun_arg_decl)]
```

src/frontend/dune
```diff
@@ -1,47 +1,54 @@
...
+     parser.messages))))
 
 (menhir
  (modules parser)
  (flags :standard --table --strict --unused-tokens --fixed-exception))
 
 (rule
- (with-stdout-to parser_updated.messages
+ (with-stdout-to
+  parser_updated.messages
```
nhuurre, line 27: What's the reason for these spacing changes? `run menhir` above collects the options to the same line but here the line is split into two.

src/analysis_and_optimization/Partial_evaluator.ml
```diff
@@ -22,7 +22,7 @@ let apply_prefix_operator_int (op : string) i =
         | "PPlus__" -> i
         | "PMinus__" -> -i
         | "PNot__" -> if i = 0 then 1 else 0
-        | s -> raise_s [%sexp (s : string)] ) )
+        | s -> Common.FatalError.fatal_error_msg [%message s] ) )
```
nhuurre, line 25: A stupid question from someone who knows nothing about Sexplib: what's the difference between `[%sexp ...]` and `[%message ...]`? `Optimize.ml` has one fatal error that keeps `[%sexp ...]` but here in `Partial_evaluator.ml` they're all changed to `[%message ...]`.

src/frontend/dune
WardBrian: I think I ran ocamlformat on this file by accident, will revert

src/analysis_and_optimization/Partial_evaluator.ml
WardBrian: The difference is subtle and honestly seems kind of arbitrary. `[%message` comes from https://github.com/janestreet/ppx_sexp_message, and `[%sexp` comes from https://github.com/janestreet/ppx_sexp_value

They have slightly different behavior for how they label their values, which is relevant for when the payload is only a string. `[%sexp (s:string)]` will output something like `(s ("The contents of the string"))`, while `[%message s]` should just output `The contents of the string`, but they accomplish the same goal in the end

nhuurre (approved):
Looks good to me.

PR #1030, opened 2021-11-09
Note that function arguments must be both unsized *and* unconstrained

WardBrian:
This came up on the forums [here](https://discourse.mc-stan.org/t/within-chain-parallelization-not-working-with-cmdstanr-on-linux-server/25203/7?u=wardbrian)

We currently generate a warning that just says the type must be unsized, which is confusing when you declare something like `corr_matrix` without sizes. I've added a test that does just this as well.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Improve error messages when a user tries to declare a function argument as a constrained type

## Copyright and Licensing

### DIFF
--- a/src/frontend/parser.messages
+++ b/src/frontend/parser.messages
@@ -3004,6 +3004,7 @@ program: FUNCTIONBLOCK LBRACE VOID TRUNCATE LPAREN RPAREN VOID
 
 Either "{" statement "}" is expected for a function definition or ";" for a function forward declaration.
 
+program: FUNCTIONBLOCK LBRACE VOID TRUNCATE LPAREN WHILE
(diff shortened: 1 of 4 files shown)


WardBrian:
@jgabry - pr from the forums. This also caught a few other unhelpful error messages, like the fact that the error was different depending on if it was the first argument or a latter argument which caused the issue.

jgabry:
Thanks @WardBrian!

rok-cesnovar (changes requested):
Thanks!

rok-cesnovar (approved):
Fat fingers… sorry. Obviously approving.

Issue #1031, opened 2021-11-10
Expose bernoulli_logit_glm_rng


rok-cesnovar:
As reported here: https://discourse.mc-stan.org/t/generated-quantities-from-bernoulli-logit-glm-model/25124/2 the `bernoulli_logit_glm_rng` function is not exposed even though it has a Stan Math implementation. 

We should expose it given that we have the implementation ready.

PR #1034, opened 2021-11-15
Expose bernoulli_logit_glm_rng

adamhaber:
Addresses #1031 .

Several questions before I open the docs PR:
1. There are 2 signatures for `bernoulli_logit_glm_lpmf` for which `y` is an `int` and not an `int` array. For both of them, `x` is a matrix, so AFAIU their `_rng` versions should return an `int` array, right?
2. Similarly, shouldn't the case for which `x` is a row vector, `alpha` is real and `beta` is a vector support an `int` y?
3. Tests - I've added the tests to `bernoulli_logit_glm.stan`, but saw there's also `rngs.stan`. Any preference?
4. Tests - should `bernoulli_logit_glm.stan` contain tests for `bernoulli_logit_glm_lupmf`? Currently it does not.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Expose bernoulli_logit_glm_rng functions.

## Copyright and Licensing


WardBrian:
On 3: I think _rngs is probably better

4: I think that test file is supposed to be named `bernoulli_logit_glm_lpdf` to match the others in the folder, and I think it's probably fine as it was

rok-cesnovar:
> Similarly, shouldn't the case for which x is a row vector, alpha is real and beta is a vector support an int y?

There are quite a few signatures that are not exposed for all GLM functions, even though almost all of the theoretically available GLMs signatures are supported in the Math library. The non-exposed ones are not tested, however, so we can just expose them before we add tests. I think we should handle that separately though.

WardBrian:
@rok-cesnovar - any clue on the C++ failures here?

rok-cesnovar:
Yeah, it seems that only two of the four signatures are supported in Math:

The two with the scalar alpha will have to be added separately:
```stan
  N = bernoulli_logit_glm_rng(d_x_m, d_alpha, d_beta_v);
  N = bernoulli_logit_glm_rng(d_x_rv, d_alpha, d_beta_v);
```

rok-cesnovar (approved):
Good to go, assuming tests pass. Thanks Adam!

WardBrian (changes requested):
LGTM as well, but needs a docs PR first! (should be a simple one though)

rok-cesnovar:
Ah, good catch!

WardBrian (approved):
Thanks again!

Issue #953, opened 2021-08-25
Incorrect parser errors on variable declarations


WardBrian:
The way the parser handles the new array[] syntax leads to a _lot_ of errors all ending in the same state.

In parser.messages, we provide the error:

"Expected non-array type after array in return type."

for the sentence 
program: DATABLOCK LBRACE ARRAY LBRACK IDENTIFIER RBRACK REAL UPPER WHILE

the problem is, this sentence is equivalent to program: DATABLOCK LBRACE UPPER REAL WHILE WHILE in the parser's states.

This means the following program, which a user might reasonably create by misinterpretting multiple declarations:

```stan
data {
real x, int y;
}
```
gives the following unhelpful error:

```
Syntax error in 'bad_multiple_decl.stan', line 2, column 16 to column 17, parsing error:
   -------------------------------------------------
     1:  data {
     2:      real x, int y;
                         ^
     3:  }
   -------------------------------------------------

Expected non-array type after array in return type.
```

-----

This is tricky to fix. The "bit of a hack" we use to parse the new array syntax without making it a reserved word means that we basically can't write a good error message for this rule, because the parser ends in the same state (670) for too many different errors.

Additionally, the way we allow 'offset', 'multiplier' et al as identifiers, and actually parse most of the reserved words to then throw an error in the semantic check, makes the parser messages file very difficult to maintain: an input sentence like `data { real real x; }`
is parsed equivalently to a sentence like 
`data { real x y; }` and so the error has no idea that the type is wrong. It's recommended fix is to put a semicolon before the x, leaving you with `real real;`, which the semantic check would later complain about. The sentence Menhir generates as an example might have the token 'WHILE' or 'REAL' where one would reasonably expect an Identifier, and that's because we do ''allow'' those things as identifiers.

The most straightforward fix for this second issue is to not try to be as cute in the parser - even if we continue to allow offset/multiplier as identifiers, the reserved-words-caught-in-semantic-check method is not ideal. As for the first problem, I think it requires array as a reserved word to simplify the parsing logic sufficiently. I think this is important to address before the parser becomes very difficult to maintain

Issue #475, opened 2015-01-28
reverse deprecation on abs() and fabs()


bob-carpenter:
I think we should deprecate
-  `real fabs(real)`

and go with just
- `real abs(real)`
- `int abs(int)`

The only drawback is that we won't have a function that applies to an integer and returns a real value, but I don't think we need that.  Returning an integer won't lose information, and Stan will just promote it to a real value if necessary to use it elsewhere.  And we'll still get integer absolute values for integers.

I propose getting rid of everything we're deprecating in Stan 3.
- [ ] undeprecate `abs(real)`
- [ ] deprecate `fabs()`
- [ ] update manual explaing what's going on

bob-carpenter:
I think we need to make a decision on this once and for all.  It used to cause us huge headaches before we got better at controlling C++ signatures.

I like the idea of keeping `fabs`,`fmax` and `fmin` because they follow the C++ spec.  But I also like the idea of extending `abs` so that it works the way everyone coming from other languages would expect.

knkumar:
I second this sentiment, decisions are always a good thing. I favor extending abs to have an uniform specification. 

Having an uniform specification is much better than adhering to the C++ constructs. There are pros to adhering to C++, for example if someone wants to tweak the functions it provides a consistent framework. However, Stan is sophisticated as a higher level domain specific language and the use case is perhaps a little different.

bob-carpenter:
Right.  We want the C++ compatibility in our C++ library, but we don't have
to expose that to Stan.  We'll eventually get around to fixing all this now
that we're better at getting things to compile.

WardBrian:
To resurrect this issue, do we want to make abs have all the same signatures as fabs? Should we also add int container signatures `abs(array[] int): array[] int` etc?

bob-carpenter:
> Do we want to make abs have all the same signatures as fabs? 

No.  The fundamental problem is the conflict between:

```
real fabs(int)
```

and

```
int abs(int)
```

So the question is whether we can get away without the `fabs` version?  If so, we can just go back to `abs` everywhere. 

We now also have

```
complex abs(complex)
```

WardBrian:
I believe our type promotion is strong enough for that to be valid everywhere. If it is not, that would itself be a bug. 

I guess that means we are not really talking about deprecating fabs, just un-deprecating abs?

bob-carpenter:
[...]

PR #1044, opened 2021-11-18
Give existing deprecations an expiration date

WardBrian:
This is the first of several PRs to begin the implementation of https://github.com/stan-dev/design-docs/pull/43

This PR simply adds notice to all existing minor deprecations that they will be removed in Stan 2.32.0. This does not apply to the ODE integrators, only the syntax-level changes which can all be replaced with the canonicalizer.

Follow on PRs will:

- Add warnings for 'future keywords': at the very least `array` (#953), but also `offset`, `multiplier`, etc (basically re-doing https://github.com/stan-dev/stan/issues/2712)
- Mark old array syntax as deprecated. This is already stated in the docs but currently produces no warnings. 
  As part of this, I think we should run _at least_ the auto-formatter on our test models, if not also the canonicalizer, to prevent the test output from ballooning due to all the old models.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: There is one docs PR for all 3 of these stanc3 PRs: https://github.com/stan-dev/docs/pull/439
 
## Release notes

Marked existing syntax deprecations as expiring in Stan 2.32.0 in accordance with https://github.com/stan-dev/design-docs/pull/43. Users should update their models (manually or with `stanc --print-canonical`) _before_ that version of Stan.

## Copyright and Licensing


src/frontend/Deprecation_analysis.ml
```diff
@@ -85,50 +88,52 @@ let rec collect_deprecated_expr deprecated_userdefined
...
+             2.32.0. Use the conditional operator (x ? y : z) instead; this \
+             can be automatically changed using stanc --print-canonical" ) ]
       @ List.concat
           (List.map l ~f:(fun e ->
                collect_deprecated_expr deprecated_userdefined [] e ) )
   | FunApp (StanLib _, {name; _}, l) ->
       let w =
         if Option.is_some (String.Map.find deprecated_distributions name) then
+          let rename, version =
+            String.Map.find_exn deprecated_distributions name in
```
nhuurre, line 112: Instead of doing `Map.find` twice you should just `match .. with` the first result. Also I think `Map.find` is preferred over `String.Map.find`.

src/frontend/Deprecation_analysis.ml
```diff
@@ -141,15 +146,18 @@ let rec collect_deprecated_expr deprecated_userdefined
         let type_ = String.Map.find deprecated_userdefined name in
         if Option.is_some type_ then
           [ ( emeta.loc
-            , "Use of the _log suffix in user defined function " ^ name
-              ^ " is deprecated, use "
+            , "Use of the _log suffix in user defined probability function "
+              ^ name
+              ^ " is deprecated and will be removed in Stan 2.32.0, use name '"
               ^ update_suffix name (Option.value_exn type_)
```
nhuurre, line 152: Same here: don't do `if Option.is_some type_ then Option.value_exn type_` but instead do `match type_ with Some ..`

src/frontend/Deprecation_analysis.ml
```diff
@@ -85,50 +88,52 @@ let rec collect_deprecated_expr deprecated_userdefined
...
-          , "The no-argument function `get_lp()` is deprecated. Use the \
-             no-argument function `target()` instead." ) ]
   | FunApp (StanLib FnPlain, {name= "abs"; _}, [e])
     when Middle.UnsizedType.is_real_type e.emeta.type_ ->
       collect_deprecated_expr deprecated_userdefined
         ( acc
         @ [ ( emeta.loc
             , "Use of the `abs` function with real-valued arguments is \
-               deprecated; use functions `fabs` instead." ) ] )
+               deprecated; use function `fabs` instead." ) ] )
```
nhuurre, line 97: Isn't `fabs()` going to be deprecated soon? We won't need this `abs()` branch at all.

src/frontend/Deprecation_analysis.ml
```diff
@@ -176,6 +184,20 @@ let rec collect_deprecated_stmt deprecated_userdefined
       acc
       @ collect_deprecated_lval deprecated_userdefined [] l
       @ collect_deprecated_expr deprecated_userdefined [] e
+  | Tilde {distribution= {name; id_loc}; arg; args; _}
+    when Option.is_some (String.Map.find deprecated_userdefined (name ^ "_log"))
+    ->
```
nhuurre, line 189: One more `Option.(if is_some .. then value_exn ..)`. I also don't understand why this is here. It looks like this is detecting if the tilde statement refers to a user-defined `_log` density function. But the `_log` suffix is not visible here; in my opinion it would make more sense to warn about the function definition, not the places that use the function.

src/frontend/Warnings.ml
```diff
@@ -7,8 +7,8 @@ let pp ?printed_filename ppf (span, message) =
   let loc_str =
     if span = Location_span.empty then ""
     else " in " ^ Location.to_string ?printed_filename span.begin_loc in
-  Fmt.pf ppf "@[<hov>Warning%s: %s@]" loc_str message
+  Fmt.pf ppf "@[<hov 4>Warning%s: %a@]" loc_str Fmt.text message
 
 let pp_warnings ?printed_filename ppf warnings =
   if List.length warnings > 0 then
-    Fmt.(pf ppf "@[<v>%a@]@." (list ~sep:cut (pp ?printed_filename)) warnings)
+    Fmt.(pf ppf "@[<v>%a@]@\n" (list ~sep:cut (pp ?printed_filename)) warnings)
```
nhuurre, line 14: Why this change? We try to avoid `\n` in format strings because the formatter does not recognize it as a newline.

src/frontend/Deprecation_analysis.ml
WardBrian: Yeah, the structure of this entire file is really awful for this reason. I was trying to avoid redoing all of it, especially because it will now mostly be deleted a couple versions from now, but it's probably worth it.

src/frontend/Deprecation_analysis.ml
WardBrian: I'm not sure about the status of #475 so I just left this unchanged for now. It should be handled alongside the changes to the supported signatures

src/frontend/Deprecation_analysis.ml
WardBrian: We currently warn if a user writes 

`target += foo_log(x, ...)`
but not
`x ~ foo(...)` 

even though this are completely equivalent. Really it's the second one which is actually being deprecated/removed - if a user still wants to name a function `bar_log` in the future they can, it's just that it won't have the special meaning it does now. Warning at the definition time would suggest that they won't be able to do such a thing

src/frontend/Warnings.ml
WardBrian: `@\n` and `\n` are not the same, the former works like `@.` but always forces a newline rather than letting the formatter sometimes ignore it. In particular, when ppf is stderr, this guarantees that the formatter always flushes the stream.

src/frontend/Deprecation_analysis.ml
nhuurre: The warning at the definition suggest the meaning of the definition is going to change. Currently `_log` functions can use `_lupdf` and that's going to be deprecated. (which btw means that `target += foo_log()` is not _completely_ equivalent to a tilde)
I also think that use in tilde statement indicates a problem with the function definition which no longer matches the intended use.

src/frontend/Deprecation_analysis.ml
nhuurre: If you decide rewriting all this is not worth it then I'd still change that `Option.is_some  (String.Map.find ..)` to `Map.mem`

src/frontend/Deprecation_analysis.ml
WardBrian: So are y0u suggesting we not warn at any usages, only at the definitions?

nhuurre: I think
* Don't warn at calls. Those warnings are redundant if the definition emits a warning and unnecessary if the definition does not need to change.
* Don't warn at tildes. The tilde does not mention the deprecated suffix so even if the user takes action the statement itself will not change.
* The definition must warn if it either contains `_lupdf` or is used in a tilde, otherwise it's fine as is. That's kind of tricky to check, and the canonicalizer updates the name unconditionally so "always warn" is a reasonable choice here.

WardBrian: I guess the tricky thing is that `_log` isn't technically a reserved suffix as far as I am aware, so the user should be able to write something like

```stan
functions {
   real add_one_log(real x){
     return log1p(x);
   }
}

model {
   real x = add_one_log(3.);
}
```

without it warning anywhere (though it looks like syntax highlighters are too broad on this as well), as this will be the sort of thing _log could totally be used for now and in the future. (as a historical note, it's not clear to me why both the _log suffix and the function multiply_log _both_ needed to be deprecated, but here we are)

If we want, I can try to make it warn only for Tildes and in the body of a definition where lup[dm]f is called? I agree that isn't the ideal location, but it should at least be doable and scoped how we want it to be.

nhuurre: Yes, that's the problem. I'd rather just complain during the transition period and then allow it after `_log` distributions are completely gone. While your proposal is doable for the warnings module I don't see how to make it work in the canonicalizer.

WardBrian: I've settled on warning at the definition, _only_ when:

1. The name ends in "_log"
2. The return type is either UReal or UInt

These are necessary conditions for it to be in the deprecation case, but sadly not sufficient to say it isn't a false alarm. I've updated the warning to account for this:

> ```
> Warning in 'log.stan', line 2, column 8: Use of the _log suffix in user
>     defined probability functions is deprecated and will be removed in Stan
>     2.32.0, use name 'add_one_log_lpdf' instead if you intend on using this
>     function in ~ statements or calling unnormalized probability functions
>     inside of it.
> ```

Canonicalize behavior also updated to make sure the return type is either real or int, otherwise it just leaves the name. Does this sound good?

nhuurre: A function with `_log` suffix must have UReal return type, otherwise it does not pass typecheck
https://github.com/stan-dev/stanc3/blob/9802b07906eaa2014c59a1fdf7e5aa6747f247b4/src/frontend/Typechecker.ml#L1257-L1265

WardBrian: Oh, what's all this business then? It seems like it's just wrong?

https://github.com/stan-dev/stanc3/blob/f1c30d28e143f4fe3fe790eb38c5bda872306c6a/src/frontend/Deprecation_analysis.ml#L75-L77

nhuurre: That should not be the return type but the type of the first argument. Also I'm not sure if `is_real_type` is exactly correct there either, depends on if deeply nested arrays are allowed or not...

WardBrian:
```ocaml
let is_real_type = function
  | UReal | UVector | URowVector | UMatrix
   |UArray UReal
   |UArray UVector
   |UArray URowVector
   |UArray UMatrix ->
      true
  | _ -> false
```

besides my mixup, does that sound like an okay way to proceed?

nhuurre: Oh, wow.
```stan
real foo_log(array[,] real x) {
  return sum(x[1]);
}
real foo_lpdf(array[,] real x) {
  return sum(x[1]);
}
```
The first one is fine, the second is apparently not? This is surely a bug but which one is wrong?

WardBrian: Well that's certainly not ideal. I just checked and the same behavior is in 2.27, so it's not even something introduced by #995, it's just been around a while.

WardBrian: I think it's because _log could be either a lpmf or lpdf, so it would be ambiguous whether it should be a real or int, so nothing is checked at all. I can trivially fix it in a separate PR to check that it's at least an int or a real but nothing else.

nhuurre: I'm not sure if we want to be so restrictive. What's wrong with distribution of an array of arrays? `UnsizedType.is_real_type` and `is_int_type` could be changed to recurse deep into array types, that's how I'd expect them to behave based on their names anyway.

WardBrian: I agree, as does @bob-carpenter. I've opened #1045 about it

WardBrian:
Bugs in the typechecker aside, I think this is good to go @nhuurre

nhuurre (approved):

PR #1046, opened 2021-11-18
User-defined densities allowed over nested and complex values.

### LINKED ISSUE #1045: Inconsistency in allowed arguments to lpdf, lpmf, and _log functions
While reviewing #1044 @nhuurre discovered that these two functions are treated differently:

```stan
real foo_log(array[,] real x) {
  return sum(x[1]);
}
real foo_lpdf(array[,] real x) {
  return sum(x[1]);
}
```

Because `UnsizedType.is_real_type` doesn't check nested arrays, only 1-D ones.

I've asked @bob-carpenter and it seems the original intent was for any nested type to be possible, so the latter is the bug. Changing `UnsizedType.is_real_type` and `UnsizedType.is_int_type` should be the right thing to do here.

Additionally, there is nothing preventing a user from defining `real foo_log(complex z)` at the moment, which isn't possible for `_lp[dm]f`.

WardBrian:
Closes #1045. The is_int_type and is_real_types now check the type recursively to handle nesting, and `lpdf` is defined not by 'takes a real as the first argument' but (properly) as 'does _not_ take an int as the first argument'

#### Submission Checklist

- [x] Run unit tests
- Documentation
      This is consistent with the final paragraph given [here](https://mc-stan.org/docs/2_28/stan-users-guide/user-defined-probability-functions.html) already

## Release notes

Allow user-defined densities over complex arguments and deeply nested arguments.

## Copyright and Licensing

### DIFF
--- a/src/frontend/Deprecation_analysis.ml
+++ b/src/frontend/Deprecation_analysis.ml
@@ -69,9 +69,8 @@ let update_suffix name type_ =
   let open String in
   if is_suffix ~suffix:"_cdf_log" name then drop_suffix name 8 ^ "_lcdf"
   else if is_suffix ~suffix:"_ccdf_log" name then drop_suffix name 9 ^ "_lccdf"
-  else if Middle.UnsizedType.is_real_type type_ then
-    drop_suffix name 4 ^ "_lpdf"
-  else drop_suffix name 4 ^ "_lpmf"
+  else if Middle.UnsizedType.is_int_type type_ then drop_suffix name 4 ^ "_lpmf"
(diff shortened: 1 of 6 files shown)


nhuurre:
Oh cool, I forgot about the complex type.
I'm going to point out that technically we also have _function_ types and I don't think they should be allowed in `_lpdf`s. It's not possible to name them at the moment so it doesn't really matter until Stan gets closures.

WardBrian:
Yeah, we can cross that bridge once a user could actually try to write a function type as an argument declaration

nhuurre (approved):
Oh cool, I forgot about complex type.

Issue #1049, opened 2021-11-19
Move the models that print deprecation warnings into a separate folder


rok-cesnovar:
By separating them in a folder we can run --print-canonical on the rest of the .stan models which reduces the number of warnings printed when compiling the entire set of models.

Also easier to remove them once the deprecated features are thrown out.

You are leading the deprecation stuff @WardBrian, so just wanted to make sure you are good with this?

WardBrian:
This was part of my plan for when I add the warning for the old array syntax as it will produce a ton of errors, so I agree

WardBrian:
I think the natural time to do this will be following #1042, hopefully sometime next week. We can create a `test/integration/good/warnings`, move a few files (like empty.stan and deprecations.stan + a few more) to that folder, and then run `stanc --auto-format --canonicalize=deprecations` on the rest of the tests. 

I might then manually un-format a few tests to make sure we're still properly testing the pretty-printer.

WardBrian:
Going to tackle this now that #1042 is merged. 

My plan is to use this script:

```python
import sys
import subprocess

filename = sys.argv[1]

pretty = subprocess.run(
    ["stanc", "--auto-format", filename], capture_output=True
).stdout
updated = subprocess.run(
    ["stanc", "--auto-format", "--canonicalize=deprecations", filename],
    capture_output=True,
).stdout

if updated != pretty:
    with open(filename, 'wb') as f:
        f.write(updated)
```

To update only the models that need deprecated features removed. I'll then manually scan through to make sure we're not updating, for example, the new warnings folder

PR #1052, opened 2021-11-22
Update Stan_math_signatures.distributions

nhuurre:
#### Submission Checklist

- [X] Run unit tests
- Documentation

Update `Stan_math_signatures.distributions` list to contain all the distribution functions.
Some functions were missing because `mk_declarative_sig` does not support their signatures. However, this list also used by deprecation analysis and typechecker error messages so it must have every distribution name even if some function signatures are also added later.
Also removes duplicated `multinomial` signatures.

## Copyright and Licensing
Copyright holder: Niko Huurre


test/integration/good/function-signatures/distributions/univariate/discrete/hypergeometric/pretty.expected
```diff
@@ -17,3 +17,9 @@ model {
   y_p ~ normal(0, 1);
 }
 
+Warning in 'hypergeometric_log.stan', line 6, column 26: hypergeometric_log
+    is deprecated and will be removed in Stan 2.32.0. Use hypergeometric_lpmf
+    instead. This can be automatically changed using stanc --print-canonical
```
WardBrian, line 22: Just to check, have you verified that the canonicalizer works properly on these?

Otherwise this looks great

nhuurre: No, I didn't test the canonicalizer at all. But the canonicalizer gets the function renames directly from `Deprecation_analysis.ml` so `--print-canonical` will do exactly as the warning says.

Maybe we should have a test that runs the canonicalizer on every warning test file and checks that the warnings are gone.

WardBrian: That wouldn't be a terrible idea, but it would be difficult to instrument (and I agree with Rok in #1049 that we should try to have all the tests which aren't intentionally testing this up to date). For now it might be worth just adding a few of these new signatures to `test/integration/canonicalize/deprecated.stan`

WardBrian (approved):
LGTM

PR #1055, opened 2021-11-22
Properly canonicalize nested if-then-else constructs

WardBrian:
Probably should have caught this during #1008 or #1003: `--print-canonical` currently breaks the way that `--auto-format` wants to do nested `if () ... else if () ...` style statements. 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Ensure the canonicalizer properly prints nested if-else statements.

## Copyright and Licensing


src/frontend/Canonicalize.ml
```diff
@@ -141,9 +141,9 @@ and keep_parens {expr; emeta} =
 
 let parens_lval = map_lval_with no_parens ident
 
-let stmt_to_block ({stmt; smeta} : typed_statement) : typed_statement =
+let stmt_to_block ({stmt; smeta} as s : typed_statement) : typed_statement =
   match stmt with
-  | Block _ -> {stmt; smeta}
+  | Block _ -> s
```
nhuurre, line 146: If you're going to change this then don't bother destructuring `{stmt;smeta}`, just use `s.stmt` and `s.smeta` where appropriate.

nhuurre (approved):
Looks good, just one optional comment.

src/frontend/Canonicalize.ml
WardBrian: Reverted this, initially I was doing more in stmt_to_block

WardBrian:
@nhuurre updated to also recognize instances where an Else contains a block which is only an IfThanElse. Makes it so cases like


```stan
if (...) {
... 
} else {
  if (...) {} else {}
}
```

also get formatted as `else if` when canonicalized

nhuurre:
Kind of nice but also I'm not sure; can the block be used to group comments or something? We don't remove empty blocks or skip statements.

If you remove one block then what about nested blocks?
```stan
if (...) {
... 
} else {
  {
    if (...) {} else {}
    }
  }
}
```

WardBrian:
That code won't have the blocks removed

As for comments, the following two snippets would both get formatted to the third

```stan

model {
    if (N>5) {y ~ std_normal();}
    else // test
      //test 2
      if(N>5) //test 3
       y ~ std_normal();
    else if (N>5) y ~ std_normal();

}
```

```stan
model {
    if (N>5) {y ~ std_normal();}
    else { // test
      //test 2
      if(N>5) //test 3
       y ~ std_normal();
    else if (N>5) y ~ std_normal();
    }
}
```

result:
```stan
model {
  if (N > 5) {
    y ~ std_normal();
  }
  // test
  //test 2
  else if (N > 5) {
    //test 3
    y ~ std_normal();
  } else if (N > 5) {
    y ~ std_normal();
  }
}
```

So I don't believe that local blocks currently work for grouping comments, at least not when there is only an IfThenElse inside them

nhuurre (approved):
Sounds reasonable.

Issue #990, opened 2021-10-04
Improve pretty-printing of strangely placed comments


WardBrian:
The example models contain this model in BPA/Ch.06/M0.stan:

```stan
data {
  int<lower=0> M;               // Size of augumented data set
  int<lower=0> T;               // Number of sampling occasions
  int<lower=0,upper=1> y[M, T]; // Capture-history matrix
}

transformed data {
  int<lower=0> s[M];            // Totals in each row
  int<lower=0> C;               // Size of observed data set

  C = 0;
  for (i in 1:M) {
    s[i] = sum(y[i]);
    if (s[i] > 0)
      C = C + 1;
  }
}

parameters {
  real<lower=0,upper=1> omega;  // Inclusion probability
  real<lower=0,upper=1> p;      // Detection probability
}

model {
  // Priors are imlicitly defined;
  //  omega ~ uniform(0, 1);
  //  p ~ uniform(0, 1);

  // Likelihood
  for (i in 1:M)
    if (s[i] > 0)
      // z[i] == 1
      target += bernoulli_lpmf(1 | omega)
              + binomial_lpmf(s[i] | T, p);
    else // s[i] == 0
      target += log_sum_exp( bernoulli_lpmf(1 | omega)   // z[i] == 1
                             + binomial_lpmf(0 | T, p),
                             bernoulli_lpmf(0 | omega)); // z[i] == 0
}

generated quantities {
  // prob present given never detected
  real omega_nd = (omega * (1 - p)^T) / (omega * (1 - p)^T + (1 - omega));
  int<lower=C, upper=M> N = C + binomial_rng(M - C, omega_nd);
}
```

Our pretty-printer does a pretty poor job on this model block in particular. Output:

```stan
model {
  // Priors are imlicitly defined;
  //  omega ~ uniform(0, 1);
  //  p ~ uniform(0, 1);
  
  // Likelihood
  for (i in 1 : M) 
    if (s[i] > 0) 
    // z[i] == 1
      target += bernoulli_lpmf(1 | omega) + binomial_lpmf(s[i] | T, p);
    // s[i] == 0
    
    else 
    
      target += log_sum_exp(bernoulli_lpmf(1 | omega)
                            + binomial_lpmf(0 | T, p),
                            bernoulli_lpmf(0 | omega)); // z[i] == 0
  
  // ^^^: z[i] == 1
}
```

[...]

PR #1056, opened 2021-11-22
Improve pretty printing comments inside binary ops

nhuurre:
#### Submission Checklist

- [X] Run unit tests
- Documentation

See #990.
I didn't change the AST to record more info; this just extends the "array-expr" hack to binary operators. Probably has some corner cases where comments are misplaced anyway but for now I think it's an improvement over what we had before.
The logic is also not as simple as it probably could be but again, not worse than before, just more of the same.

## Copyright and Licensing
Copyright holder: Niko Huurre

### DIFF
--- a/src/frontend/Ast.ml
+++ b/src/frontend/Ast.ml
@@ -222,7 +222,7 @@ type 's block = {stmts: 's list; xloc: Middle.Location_span.t [@ignore]}
 and comment_type =
   | LineComment of string * Middle.Location_span.t
   | BlockComment of string list * Middle.Location_span.t
-  | Comma of Middle.Location.t
+  | Separator of Middle.Location.t
 
 and 's program =
   { functionblock: 's block option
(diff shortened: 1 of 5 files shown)


src/frontend/Ast.ml
```diff
@@ -222,7 +222,7 @@ type 's block = {stmts: 's list; xloc: Middle.Location_span.t [@ignore]}
 and comment_type =
   | LineComment of string * Middle.Location_span.t
   | BlockComment of string list * Middle.Location_span.t
-  | Comma of Middle.Location.t
+  | Separator of Middle.Location.t
```
WardBrian, line 225: Mind adding a note on this constructor about its purpose? Something like `(** Seperator is used to denote the location of items like commas and keywords which don't have location information stored in the AST, but are useful for placing comments in printing *)`

(assuming my interpretation is correct)

WardBrian (commented):
One remark, otherwise this LGTM and seems like a real nice improvement

WardBrian (approved):

WardBrian:
Actually, this looks weird to me:

``` // y[i] == 0
    else {
```

gets turned into

```
 else // y[i] == 0
    {
```

WardBrian:
Disregard my last comment, I was incorrectly reading the diff output

Issue #1042, opened 2021-11-16
Support pretty-printing without inlining #include-d code


spinkney:
When the '.stanfunctions' file has an '#include some.stanfunctions' in a different directory the auto formatter fails. One typically wouldn't include the path because it would be specified in the call to cmdstanr via include_paths = "./path_to_file".

spinkney:
> So I think the real issue here is "support pretty-printing without inlining `#include`-d code". This should be doable because we keep the fact that it was included around in the location information.

Yes, this would work

rok-cesnovar:
I agree that by default auto-format should not inline `#include`. I would like to still have a way to inline the includes if requested. Not sure on what the API should be. 

The use case is caching models without having to inline includes with some string processing.

WardBrian:
I've been thinking about how the interface should be for formatting, canonicalizing, etc. I think all of these things should be optional and part of the canonicalize command, with `--auto-format` doing the bare minimum.

So I think we'd want something like `--auto-format --canonical=includes,deprecations,parenthesis` with `--print-canonical` becoming an alias for essentially `--auto-format --canonical=[everything we can do]`

PR #1058, opened 2021-11-24
Modular canonicalizer

### LINKED ISSUE #1053: Make Canonicalizer modular
Currently, `--print-canonical` does a lot of different things to your code. Especially with the incoming deprecation changes, I think it would be great to have a way to pick and choose what it does.

I propose something like this, first discussed in #1042:

- All pretty printing is done with `--auto-format`.
- Add an additional argument which allows you to specify certain canonicalizations to use. For example, `--auto-format --canonicalize=deprecations` would only update deprecated features, not add/remove parenthesis etc. This is a comma separated list of options.
  By my count, we would have the following options to start: 
    - `deprecations` - deprecated syntax and function replacement
    - `includes` - see #1042
    -  `parenthesis` - the current canonicalizer removes extra parenthesis
    - `braces` - the current canonicalizer adds `{}` around single-line if/else (...etc) blocks.
- We can then make `--print-canonical` a synonym for `--auto-format --canonicalize=(all options enabled)` 

This also allows us in the future to add possible more opinionated canonicalizers which we don't want enabled for all users, or even mutually exclusive options.

I think this would be good to tackle alongside #24, and a similar design could be used for #549

WardBrian:
Resolves #1053. I think this will be good for #1049 as we can run only the deprecations setting on the test suite so we don't over-regularize.

This also lays the ground work for #1042 but I ran into some issues after fixing a bug in how we handle source positions that I need more time to run down, so I'm submitting this without that work.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/444

## Release notes

The canonicalizer can now have each part enabled seperately. This can be done with commands like `stanc --auto-format --canonicalize=braces,deprecations <model>`. Current options are `deprecations`, `braces`, and `parenthesis`.  The `--print-canonical` flag still works, but is simply synonymous for `--auto-format --canonicalize=<all possible options>`

## Copyright and Licensing


src/frontend/Canonicalize.ml
```diff
@@ -165,30 +166,65 @@ let rec parens_stmt ({stmt; smeta} : typed_statement) : typed_statement =
...
+    | While (e, s) -> While (no_parens e, parens_stmt s)
     | IfThenElse (e, s1, Some ({stmt= IfThenElse _; _} as s2))
      |IfThenElse (e, s1, Some {stmt= Block [({stmt= IfThenElse _; _} as s2)]; _})
       ->
         (* Flatten if ... else if ... constructs *)
-        IfThenElse (no_parens e, parens_block s1, Some (parens_stmt s2))
+        IfThenElse (no_parens e, parens_stmt s1, Some (parens_stmt s2))
     | IfThenElse (e, s1, s2) ->
-        IfThenElse (no_parens e, parens_block s1, Option.map ~f:parens_block s2)
+        IfThenElse (no_parens e, parens_stmt s1, Option.map ~f:parens_stmt s2)
```
nhuurre, line 176: These branches can be removed completely--`While` and `IfThenElse` are handled correctly by the final wildcard `map_statement`.

nhuurre (changes requested):

src/frontend/Canonicalize.ml
WardBrian: Good catch

src/frontend/Canonicalize.ml
```diff
@@ -165,30 +166,57 @@ let rec parens_stmt ({stmt; smeta} : typed_statement) : typed_statement =
...
+        For {f with loop_body= stmt_to_block loop_body}
+    | _ -> map_statement ident blocks_stmt ident ident stmt in
+  {stmt; smeta}
+
+let repair_syntax program settings =
+  if settings.deprecations then
     program
+    |> map_program
+         (repair_syntax_stmt (userdef_distributions program.functionblock))
+  else program
```
nhuurre, line 207: I'm inclined to say that `repair_syntax` should be enabled for all formatting, not just deprecations. It's not about syntax that's _going to be_ deprecated but syntax that _has already been_ deprecated.. A program that "needs" `repair_syntax` isn't even going to typecheck without it.

WardBrian: At the moment, those programs _do_ fail to typecheck without `--print-canonical`. If we don't like that we could run it if and only if the formatter is activated, I suppose.

WardBrian:
As @rok-cesnovar brought up in #577, it would be useful if this was also available in stancjs. Any suggestions on how that should be implemented? I think the js interface can only pass string flags, correct? So each part would need a seperate flag like "canonicalize-deprecations"?

nhuurre:
The changes to the canonicalizer look good. I'm not familiar with cram tests so maybe someone else should review those.

rok-cesnovar:
> I think the js interface can only pass string flags, correct? So each part would need a seperate flag like "canonicalize-deprecations"?

It also handles "name=val" https://github.com/stan-dev/stanc3/blob/master/src/stancjs/stancjs.ml#L114

WardBrian:
Cram tests are described a little in the [dune manual](https://dune.readthedocs.io/en/stable/tests.html#cram-tests). They basically show a command (the line starting with `  $`), its output, and the exit code (in `[``]`). They're great for testing command line arguments, which is all they're doing here

stanc.opam
```diff
@@ -3,7 +3,7 @@ opam-version: "2.0"
 synopsis: "The Stan compiler and utilities"
 depends: [
   "ocaml" {= "4.12.0"}
-  "dune" {>= "2.8"}
+  "dune" {>= "2.7" & >= "2.8"}
```
nhuurre, line 6: What does this do? Looks kind of redundant?

WardBrian: This was generated by dune because we require 2.8 but I had `lang-dune` set to 2.7, and I guess it wasn't smart enough to figure out that only one of those is needed. I'll fix it

nhuurre (approved):

PR #1059, opened 2021-11-25
Add warning for lvalue indexing inconsistency.

### LINKED ISSUE #531: Indexed LValue bug
#### Summary:
Currently, Stan treats LValues (left-hand-side of assignments) as having a flat indexing structure consisting of a single list of indices. This seems in contradiction with the treatment of RValues (right-hand-side of assignments).

#### Description:
For example: `a[:][2]` and `a[:, 2]` are equivalent as an LValue but not as an RValue.

#### Reproducible Steps:
Run 
```
model {
    int a[2, 2] = {{1, 2}, {3, 4}};
    a[:][2] = {17, 18};
    print(a);
}
```
and
```
model {
    int a[2, 2] = {{1, 2}, {3, 4}};
    a[:, 2] = {17, 18};
    print(a);
}
```
They give the same output.

Then run 
```
model {
    int a[2, 2] = {{1, 2}, {3, 4}};
    print(a[:][2]);
}
```
and
```
model {
    int a[2, 2] = {{1, 2}, {3, 4}};
    print(a[:, 2]);
}
```
They give different output.

#### Current Output:
Inconsistent behaviour between indexing in LValues and RValues.

#### Expected Output:
Consistency. I think the behaviour of RValues is the correct one.

#### Current Version:
v2.20.0

nhuurre:
#### Submission Checklist

- [X] Run unit tests
- Documentation
    - [X] If a user-facing facing change was made, the documentation PR is here: stan-dev/docs#464

Fix #531.
Typechecker would check LValues as if they were RValues but codegen just flattens nested indexing. If these interpretations implied different type for the LValue then stanc3 would generate invalid C++ code. This case in now a type error.
For backwards compatibility, if the generated C++ code would have compiled before then the typechecker only emits a warning.

Single indexing may be arbitrarily nested as long as only the outermost index group contains multi-indexes (these cannot cause ambiguity)
```stan
x[1,2][3][4] = ... // ok
x[1, 2][:,4] = ... // ok
x[1, :][3,4] = ... // bad
```

## Release notes
Deprecate nested multi-indexing on lvalues as it is inconsistent with rvalues.

## Copyright and Licensing
Copyright holder: Niko Huurre


src/frontend/Typechecker.ml
```diff
@@ -728,34 +728,69 @@ let verify_assignment_global loc cf block is_global id =
...
+            , flat @ idcs ) in
+      let lval, var, flat = check_inner lval in
+      let idcs = List.map ~f:(check_index cf tenv) idcs in
+      let ad_level = inferred_ad_type_of_indexed lval.lmeta.ad_level idcs in
+      let type_ = inferred_unsizedtype_of_indexed ~loc lval.lmeta.type_ idcs in
+      if List.exists ~f:is_multiindex flat then (
+        add_warning loc
+          "Nested multi-indexing in LValues does not behave the same as nested \
+           indexing in expressions. This is considered a bug and will be \
+           disallowed in the future." ;
```
WardBrian, line 779: I'd love if we either elaborated a bit more on how it is different or linked to some page in the doc which does

src/frontend/Typechecker.ml
```diff
@@ -728,34 +728,69 @@ let verify_assignment_global loc cf block is_global id =
   if (not is_global) || block = cf.current_block then ()
   else Semantic_error.cannot_assign_to_global loc id.name |> error
 
-let mk_assignment_from_indexed_expr assop lhs rhs =
-  Assignment
-    {assign_lhs= Ast.lvalue_of_expr lhs; assign_op= assop; assign_rhs= rhs}
-
 let check_assignment_operator loc assop lhs rhs =
```
WardBrian, line 731: This should now be named `verify` since it doesn't return anything?

test/integration/bad/stanc.expected
```diff
@@ -1379,6 +1379,28 @@ Semantic error in 'lp-error.stan', line 5, column 2 to column 6:
    -------------------------------------------------
 
 Identifier 'lp__' clashes with reserved keyword.
+  $ ../../../../install/default/bin/stanc lvalue_indexes1.stan
```
WardBrian, line 1382: Can you also add tests showing the warning?

WardBrian (changes requested):

src/frontend/Typechecker.ml
nhuurre: I guess a full explanation belongs in the deprecated features section of the user manual. LValues just concatenate the indices, I'm not sure how to describe the difference in a single sentence.

src/frontend/Typechecker.ml
nhuurre: Good catch, I forgot there was a naming convention.

test/integration/bad/stanc.expected
nhuurre: Sure, where do they go? `test/integration/bad/` is for models that don't compile, right? `test/integration/good/` has some models with deprecation warnings so is that the right place? `test/integration/good/lvalue_multiindex.stan`

WardBrian: I think we may want to start thinking about #1049 and make a new folder `test/integration/good/deprecated` or something, with this there. We can then slowly start moving things there and eventually run the canonicalizer on anything we don't want to end up in said folder

src/frontend/Typechecker.ml
WardBrian: Yeah, I think this can go in the deprecations section. If you can give me the longform explanation I could include it in https://github.com/stan-dev/docs/pull/439, or I could merge that now and this could be a separate PR

test/integration/bad/stanc.expected
WardBrian: After thinking about it, `test/integration/good/warning` or `test/integration/good/emits_warning` may be best (we can also put things like empty.stan there)

src/frontend/Typechecker.ml
nhuurre: LValue just concatenates the indexes so that `v[idx1,...][idx2,...] == v[idx1,...,idx2,...]`. RValues are described in https://mc-stan.org/docs/2_28/reference-manual/language-multi-indexing.html#multiple-index-semantics

WardBrian: I think it would be best to clarify this in a docs pr, both on the deprecations page and the one you link there

src/frontend/Typechecker.ml
```diff
@@ -728,34 +728,73 @@ let verify_assignment_global loc cf block is_global id =
...
+            ( {lval= LIndexed (lval, idcs); lmeta= {ad_level; type_; loc}}
+            , var
+            , flat @ idcs ) in
+      let lval, var, flat = check_inner lval in
+      let idcs = List.map ~f:(check_index cf tenv) idcs in
+      let ad_level = inferred_ad_type_of_indexed lval.lmeta.ad_level idcs in
+      let type_ = inferred_unsizedtype_of_indexed ~loc lval.lmeta.type_ idcs in
+      if List.exists ~f:is_multiindex flat then (
+        add_warning loc
+          "Nested multi-indexing in LValues does not behave the same as nested \
```
WardBrian, line 777: Code/tests look good - one wording suggestion and I think this does need a docs PR, but otherwise LGTM

```suggestion
          "Nested multi-indexing on the left hand side of assignment does not behave the same as nested \
```

src/frontend/Semantic_error.ml
```diff
@@ -352,6 +353,9 @@ module StatementError = struct
         Fmt.pf ppf
           "Cannot assign to global variable '%s' declared in previous blocks."
           name
+    | LValueMultiIndexing ->
+        Fmt.pf ppf
+          "Left hand side of an assignment cannot have nested multi-indexing."
```
WardBrian, line 358: More of a question than anything - after removal do you think we should change the parser/AST to enforce this, or keep it in typechecking? 

Is the eventual hope to re-introduce with matching semantics?

src/frontend/Typechecker.ml
```diff
@@ -730,34 +730,73 @@ let verify_assignment_global loc cf block is_global id =
...
+            , var
+            , flat @ idcs ) in
+      let lval, var, flat = check_inner lval in
+      let idcs = List.map ~f:(check_index cf tenv) idcs in
+      let ad_level = inferred_ad_type_of_indexed lval.lmeta.ad_level idcs in
+      let type_ = inferred_unsizedtype_of_indexed ~loc lval.lmeta.type_ idcs in
+      if List.exists ~f:is_multiindex flat then (
+        add_warning loc
+          "Nested multi-indexing on the left hand side of assignment does not \
+           behave the same as nested indexing in expressions. This is \
```
WardBrian, line 780: I think we should indicate what it does mean/what it can be replaced with (All other deprecation messages mention the canonicalize flag, like `This can be automatically changed using the canonicalize flag for stanc`)

src/frontend/Semantic_error.ml
nhuurre: I don't know. Cleaning up the AST seems like a good idea but nested indexing is still allowed when all the indexes are simple integers. If we change the AST then pretty printer will always flatten the indexes. That may be acceptable, it's not like pretty printer preserves all the formatting anyway.
In my opinion nested indexing is just an accidental consequence of the ability to index any expression. It's not necessary to support it in lvalue position, where arbitrary expressions aren't allowed.

WardBrian: Fair enough. Does this change do anything for nested indexing on the left hand side of a `~`, or is that technically fine since it ends up inside a `lpdf`? I'm fuzzy on whether that counts as an l-value or not.

nhuurre: No change there, the "left-hand side" of a tilde statement is not an lvalue.

WardBrian (approved):
Looks good

PR #1065, opened 2021-11-29
Stancjs data-gen

rok-cesnovar:
I re-added the stancjs data-gen test and locally I am seeing no issues, the output is exactly the same each time. Just trying to run on CI for now.

## Copyright and Licensing


Jenkinsfile
```diff
@@ -55,7 +55,7 @@ pipeline {
                     def stanMathSigs = ['test/integration/signatures/stan_math_sigs.expected'].join(" ")
                     skipExpressionTests = utils.verifyChanges(stanMathSigs)
 
-                    def sourceCodePaths = ['src'].join(" ")
+                    def sourceCodePaths = ['src', 'test/integration/good', 'test/stancjs'].join(" ")
```
rok-cesnovar, line 58: We should run all tests also when we just add new tests.

WardBrian: I agree, though ideally we wouldn't rebuild all the binaries if only `test/` changed

rok-cesnovar: Good call. Done.

WardBrian:
For me, running this locally does not yield the same results consistently. Not sure if it's a platform dependent issue (I'm on Ubuntu 20.04.3)

rok-cesnovar:
Weird, I am on the ubuntu 20.04 LTS as well (tried on two different machines). This is not a really crucial thing to waste too much time on it. Lets keep this open and I am going to tag you if I have something to test.

WardBrian:
We can see if it works in CI. The other thing we could do is rather than printing the random numbers, load the json and assert that each member has the correct shape (similar to how our tests end up looking in cmdstan interfaces)

rok-cesnovar:
Ok, CI fails as well, which is good in this case. Will check the shape and be done with this.

WardBrian (approved):

PR #1069, opened 2021-12-02
Make inlining of includes optional in the pretty printer.

### LINKED ISSUE #1042: Support pretty-printing without inlining #include-d code

WardBrian:
Closes #1042. @spinkney I'd appreciate if you could try this on some code, our tests don't feature all too many `#include` statements. I can give you a binary via slack or email.

The basic idea of this PR is very simple: store `#include` directives the same way we store comments, and add the ability to filter during printing based on if a statement has its location as "included" from somewhere. 

Unforunately, there were several bugs with the current way we were creating and processing locations around included files that made this PR larger. In particular, the first token of the new lexbuf was marked as starting in the old file, and fixing this required some finessing mutable fields in the lexer.

Unlike with comments, if the pretty-printer cannot place an `#include` it produces a fatal error. This is to ensure safety - we should never be moving them around. This does mean some technically legal stan programs could not be pretty-printed without inlining includes, like

```stan
if (1) {
} #include <test.stan>
```
test.stan:
```stan
else {

}
```
Will fail and not produce any pretty-printed output.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/449
   
## Release notes

`--auto-format` no longer prints code which originated from an `#include` directive. This can be re-enabled by using `canonicalize=includes`

## Copyright and Licensing


WardBrian:
Built binaries are available here: https://jenkins.mc-stan.org/job/stanc3-test-binaries/175/artifact/bin/

spinkney:
You can try with some or all of the Stan files in the examples folder of this repo https://github.com/spinkney/helpful_stan_functions. 

As a personal note, I may go a bit dark or be slow to respond for the next few weeks as I'm starting a new job on Monday.

rok-cesnovar:
I tried it on Sean's repo. It mostly looks good, I did found one minor thing.

a.stanfunctions:
```stan
real a_fun() {
  return 0;
}
```

b.stanfunctions:
```stan
#include <a.stanfunctions>

real b_fun() {
  return 1;
}
```
The second file gets formatted as:

```stan
#include <a.stanfunctions>
                          
                          real b_fun() {
                            return 1;
                          }
```

rok-cesnovar:
Works great, tried it again on Sean's repo and its 👍🏼 Tests also look good. 

I am not as confident in reviewing the preprocessor code. If Niko has time to have a look at that, that would be better probably.

src/frontend/Pretty_printing.ml
```diff
@@ -47,32 +72,44 @@ let skip_comments loc =
...
     @ List.filter_map (get_comments loc) ~f:(function
+        | `Include, l, loc ->
+            (* This prevents against bad behavior, but also really terrible but technically allowed things fail
+               For example, an if statement where the 'else' is entirely inside the include.
+               This makes the failure noisy rather than ever producing anything invalid for these. *)
+            Common.FatalError.fatal_error_msg
+              [%message
+                "Attempting to move include!"
+                  (l : string list)
+                  (loc : Middle.Location_span.t)]
```
nhuurre, line 83: 1. "Attempting to move include" is not the problem, it is a failed attempt at solving the problem. The error message should say something like "Cannot format #include in this position."
2. `FatalError` is meant for unreachable code, and advises the user to file a bug report. However, that is not appropriate here: tricky includes are allowed by the language and we don't plan to support pretty printing them. We need a new error mechanism for this case.
3. Another bad behavior is when the `if`branch is entirely inside the include but `else` is outside. That is not caught until the validation step at the very end of pretty printing.

src/frontend/Pretty_printing.ml
```diff
@@ -369,63 +411,66 @@ and pp_recursive_ifthenelse ppf (s, loc) =
...
-  | FunDef {returntype= rt; funname= id; arguments= args; body= b} -> (
-      let loc_of (_, _, id) = id.id_loc in
-      pf ppf "%a %a(%a" pp_returntype rt pp_identifier id
-        (box (pp_list_of pp_args loc_of))
-        (args, {loc with end_loc= b.smeta.loc.begin_loc}) ;
-      match b with
-      | {stmt= Skip; _} -> pf ppf ");"
-      | b -> pf ppf ") %a" pp_statement b )
+  if (not !print_included) && Option.is_some loc.end_loc.included_from then ()
+  else
```
nhuurre, line 415:
```suggestion
  if !print_included || Option.is_none loc.end_loc.included_from then
```
The same line is also present in `pp_space`. Would it be possible to move all this logic to `pp_list_of_statements`? I'd rather have the caller decide which statements need to be printed than `pp_statement` and `pp_space` sometimes be no-ops. There are some calls to `pp_statement` outside of `pp_list_of_statements` but I don't think any of those allow #include anyway.

nhuurre (changes requested):

src/middle/Location.ml
```diff
@@ -75,6 +59,24 @@ let rec to_string ?printed_filename ?(print_file = true) ?(print_line = true)
     | None -> "" in
   sprintf "%s%scolumn %d%s" file line loc.col_num incl
 
+let compare loc1 loc2 =
+  let rec unfold = function
+    | {included_from= None; _} as loc -> [loc]
+    | {included_from= Some loc1; _} as loc2 ->
+        (* force tiebreaker to go to the file included from *)
+        loc2 :: unfold {loc1 with line_num= loc1.line_num + 1} in
```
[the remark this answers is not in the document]
WardBrian: This is required due to the fact that comments/includes never call Lexing.token directly, so their position information is sometimes hairy. In particular, if two includes happen in a row the positions can all end up higher than they should from an included file.

The best example I can think of comes from running test/integration/rstanarm/binomial.stan without this line:
https://gist.github.com/WardBrian/d73f77b54339b3af5b621202b041a49c#file-binomial-bad-stan-L342

The functions block here has two includes in a row. Inserting anything in between them, even another comment, seems to fix the issue

It may be possible to avoid doing this if we really sort out 100% the issues with the preprocessor/lexer. The issue in the above is that the immediate next thing after the `#include` is another `#include`, so we never 100% reset the token positions. There's probably a similar hack available as what this currently does in finish_comment, but it's difficult to avoid 'artificially' inserting line breaks. If you're able to spend some time debugging I'd appreciate the help here

src/frontend/Ast.ml
```diff
@@ -347,7 +347,7 @@ let get_first_loc (s : untyped_statement) =
    |IfThenElse (e, _, _)
    |While (e, _) ->
       e.emeta.loc.begin_loc
-  | Profile _ | Block _ -> s.smeta.loc.begin_loc
+  | Assignment _ | Profile _ | Block _ -> s.smeta.loc.begin_loc
```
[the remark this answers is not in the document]
WardBrian: Complicated question! The previous way of getting a location out of an assignment was based around the id location. This led to a lot of headaches with the preprocessor-start-position issues. In particular, assignment is the only statement which 1. began with an identifier (meaning no previous token reset the positions of the parser) and 2. used the position of that identifier for its position. I was not able to make the location of the identifiers correct despite trying for some time, but this reports the same location in all previously correct instances and fixes the issue.

src/frontend/Preprocessor.ml
```diff
@@ -12,6 +13,34 @@ let dup_exists l =
...
+let current_buffer () =
+  let buf = Stack.top_exn include_stack in
+  buf
+
+let pop_buffer () = Stack.pop_exn include_stack
+
+let update_start_pos pos =
+  if size () > 1 then
+    let lexbuf = List.nth_exn (Stack.to_list include_stack) 1 in
+    lexbuf.lex_start_p <- pos
```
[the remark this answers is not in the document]
WardBrian: I think you technically may be correct, if a user had a very long string of includes in a row.

The reason this is necessary at all is because of the way we do preprocessing. Basically, the flow of data is like this:

1. The parser is called and given the top of the preprocessor stack
2. It grabs uses that lexbuf to get one token
3. Goto 1

The issue is that inside bullet point 2 we can change the preprocessor stack, but the parser has already bound a reference to the position information of the buffer it was last called with. 

You're right in that 2 could actually add an arbitrary number of includes to the stack, so we should probably update all the start positions of them.

src/frontend/Ast.ml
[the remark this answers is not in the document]
WardBrian: You're right, I didn't catch that. And yes, it also is wrong:

```stan
  // test tildes

  x ~

  exponential(15);
```

gets turned into the following when included
```stan
  x ~  exponential(15);

      // ^^^:   test tildes
```

I'll see if the same fix works here, or if I need to spend more time trying to actually fix identifiers

src/frontend/Pretty_printing.ml
WardBrian: I don't think we can move it because of the recursive calls inside `pp_statement`, like in the `while` case

src/frontend/Pretty_printing.ml
WardBrian: 1. Agreed
2. I would like to encourage bug reports on this, at least for the first few versions, in case we miss any cases we _should_ support. I suppose its possible to say this outside of a `FatalError` however
3. Yeah, that's a bit tricky. Because we don't keep track of the individual locations of `if` and `else`, it's impossible to tell if one of them was included from or not at the moment.

src/frontend/Pretty_printing.ml
nhuurre: I don't think `while (...) #include` works either way.

WardBrian: Ah yep, it only works if you put braces around it, which means it would be a Block and call the list version instead. 

I'm not able to move the logic out of pp_space still. It's still important to call pp_spac*ing* from inside list_of_statements in order to actually place the `#include`

src/frontend/Preprocessor.ml
```diff
@@ -12,6 +13,34 @@ let dup_exists l =
...
+let current_buffer () =
+  let buf = Stack.top_exn include_stack in
+  buf
+
+let pop_buffer () = Stack.pop_exn include_stack
+
+let update_start_positions pos =
+  List.iter
+    ~f:(fun lexbuf -> lexbuf.lex_start_p <- pos)
+    (Stack.to_list include_stack)
```
nhuurre, line 28:
```suggestion
  Stack.iter
    ~f:(fun lexbuf -> lexbuf.lex_start_p <- pos)
    include_stack
```

WardBrian: Good catch

src/frontend/Pretty_printing.ml
WardBrian: I've been trying for a little bit and it's actually rather difficult to hit this case. Most of the 'bad' inputs I have formulated end up failing inside check_correctness, which is a `FatalError` anyway.

Personally, I'm okay with getting a few false positive issues opened about this. If anything it will help us calibrate how much utility there would be to users if we tried really hard to improve this feature past where it is now

nhuurre (approved):

PR #1074, opened 2021-12-09
Clear lexing stack before parsing starts

WardBrian:
This fixes an issue first spotted by @spinkney in #577. The preprocessing stack is never cleared, so in particular a subsequent parse will inherit anything left on the stack. This is only possible to trigger in the Javascript compiler.

Removing the change in `Preprocessor.ml` makes the new test fail like so:

```
Syntax error in 'string', line 8, column 0 to column 5, parsing error:
Expected "generated quantities {" or end of file after end of model block.

Syntax error in 'string', line 8, column 6 to column 7, parsing error:
Expected "generated quantities {" or end of file after end of model block.
```

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fix an issue with parser errors 'sticking around' on subsequent runs, which primarily affected the Javascript compiler

## Copyright and Licensing

### DIFF
--- a/src/frontend/Preprocessor.ml
+++ b/src/frontend/Preprocessor.ml
@@ -14,7 +14,10 @@ let include_stack = Stack.create ()
 let include_paths : string list ref = ref []
 let included_files : string list ref = ref []
 let size () = Stack.length include_stack
-let init buf = Stack.push include_stack buf
+
+let init buf =
+  Stack.clear include_stack ;
+  Stack.push include_stack buf
 
 let current_buffer () =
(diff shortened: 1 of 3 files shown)


WardBrian:
I believe this behavior can/has affected more recent versions of RStan -- see, for example, [this comment](https://discourse.mc-stan.org/t/function-algebra-solver-newton-not-found/24737/6?u=wardbrian)

spinkney:
was this found by seantalts?

rok-cesnovar (approved):
Nice catch! Thanks!

PR #1075, opened 2021-12-10
Add Jacobian adjustment warning to pedantic mode

### LINKED ISSUE #168: Emit warning whenever lhs of ~-statement or first argument of conditional distribution application is not a variable
Warn about adding a log Jacobian term to target

WardBrian:
Closes #168. This adds a warning to pedantic mode which triggers whenever the item on the left of a `~` or the first argument x in `*_lp[dm]f(x| ...)` is potentially a non-linear transformation. 

I implemented a rough heuristic that removes most of the obvious false alarms when using stan-lib functions or basic arithmetic, but it's definitely not perfect. I tried to make it fairly conservative.

Additionally, this won't detect if you do something in the transformed parameters block, for example. I'm not sure how much of a concern that is, or how feasible it would be to detect.

The warning looks like this:
```
Warning in 'jacobian_warning1.stan', line 7, column 2: Left-hand side of
    sampling statement (~) may contain a non-linear transform of a parameter
    or local variable. If it does, you need to include a target += statement
    with the log absolute determinant of the Jacobian of the transform.
```

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/455

## Release notes

Pedantic mode will now warn you to include a `target +=` Jacobian adjustment if it believes a non-linear transformation is occurring during sampling.

## Copyright and Licensing


rok-cesnovar:
@rybern do you maybe have a moment to review this? Given your expertise on the analysis and optimization codebase, I think it would be  great to get your eyes on this. If you don't have time, that is also fine, obviously :)

WardBrian:
I found the original code for this warning in stanc2, and it was much less sophisticated - any time a non-parameter appeared on the left hand side the warning would be emitted

https://github.com/stan-dev/stan/blob/5642cd6dc78bbe49f75c1e9b0c790caed082e140/src/stan/lang/grammars/semantic_actions_def.cpp#L1043-L1055

bob-carpenter:
Sorry for the meaning drift of the function name.  If you look at the implementation, you can see it tries to cut down on false positives by removing linear functions. .   But it wasn't complete and still generated false positives.  See:

https://github.com/stan-dev/rstan/blob/da2fc9c079534a82d3d26adda51ad17bf22f5e2b/rstan/rstan/src/stan/lang/ast/fun/has_non_param_var_vis_def.hpp

WardBrian:
Thanks @bob-carpenter - I just went through and compared the two and mine was (more or less) equivalent. I've made a few additions (function names I had missed, the conditional operator) and now the 'false_alarm' test here doesn't emit any false alarms.

rybern (approved):
Looks good to me! Didn't deeply check the linearity logic but the code looks good

PR #1082, opened 2022-01-02
Add optimization flags, debug-* flags to stancjs

rok-cesnovar:
This is not completely done yet as I am having two issues:
- I am not sure if/how we can return the "pretty" prints (`debug-mir-pretty`, `debug-transformed-mir-pretty`, etc.) I am not sure how to replace this call to make it return a string: `Program.Typed.pp Format.std_formatter tx_mir`
- The stan math signatures hashtbl seems to be empty when called with stancjs. See comment.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Add optimization level flags, debug-* and dump-math-signatures flags to stancjs.

## Copyright and Licensing

### DIFF
--- a/src/stancjs/stancjs.ml
+++ b/src/stancjs/stancjs.ml
@@ -7,7 +7,6 @@ open Middle
 open Js_of_ocaml
 
 let version = "%%NAME%% %%VERSION%%"
-let opt_lvl = ref Optimize.O0
 
 let warn_uninitialized_msgs (uninit_vars : (Location_span.t * string) Set.Poly.t)
     =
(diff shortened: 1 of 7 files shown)


src/stancjs/stancjs.ml
```diff
@@ -77,16 +80,41 @@ let stan2cpp model_name model_string is_flag_set flag_val =
...
+            , warnings
+            , [] ) ;
+        if is_flag_set "debug-transformed-mir" then
+          r.return
+            ( Result.Ok
+                (Sexp.to_string_hum [%sexp (tx_mir : Middle.Program.Typed.t)])
+            , warnings
+            , [] ) ;
+        if is_flag_set "dump-stan-math-signatures" then (
+          Stan_math_signatures.pretty_print_all_math_sigs Format.str_formatter
```
rok-cesnovar, line 115: This is just for debugging purposes as it would print signatures not return them. However this doesn't print anything. `List.length (Hashtbl.keys stan_math_signatures)` returns 0.

WardBrian: I wonder if there is something about the module loading order in the JS runtime that means the initalizer code doesn't run. You could try refactoring the portion of stan_math_signatures that builds it into

```ocaml

let build_signatures () = (* current contents of the let () = ... block *)

let () = build_signatures ()


```

and then manually call Stan_math_signatures.build_signatures in stancjs before printing?

rok-cesnovar: Ah never mind, it's just the printing that doesn't seem to show up in the test. 

```ocaml
        if is_flag_set "dump-stan-math-signatures" then (
          let num_of_sigs = 
          Hashtbl.to_alist Stan_math_signatures.stan_math_signatures 
          |> List.length 
          |> string_of_int in
          r.return (Result.Ok num_of_sigs, warnings, []) ) ;
```
works just fine.

WardBrian:
Just to note: this is another step toward requiring all developers to have node, as now adding a signature (probably the simplest thing a developer could do) would change the JS tests after this.

We could either just decide this is the case (and probably remove the runjstest alias, merging with runtest) or be careful with how we test it (maybe confirm the existence of a known signature rather than printing all of them in the test)

Another thought: should something which doesn't need the source at all really be part of the `stanc.stanc` function? We could define other functions in the javascript interface right?

rok-cesnovar:
> as now adding a signature (probably the simplest thing a developer could do) would change the JS tests after this. 

Yeah, I turned it into a draft, as I haven't figured that out. The final test will not be a print of all sigs.

> Another thought: should something which doesn't need the source at all really be part of the stanc.stanc function? We could define other functions in the javascript interface right?

Great thought! This should definitely be a separate function, not an arg to stanc3.

WardBrian:
I think something like `print(signatures.results.includes("zeros_array(int) => array[] real"))` should be enough as a simple test?

rok-cesnovar:
Agree, will double-check your commits.

rok-cesnovar (commented):
Looks great. Thanks for the console.assert cleanup, wasn't aware of that existing.

WardBrian:
> Looks great. Thanks for the console.assert cleanup, wasn't aware of that existing.

I was writing the signature test before I looked at your other tests, and google led me to it. If I had seen yours first, I probably would have just done the same thing

WardBrian (approved):

Issue #1096, opened 2022-01-19
Function inlining fails to detect recursive functions, inlines them anyway


WardBrian:
In `Optimize.ml`, the function overlining optimization includes this remark:

```ocaml
  (* We only add the first definition for each function to the inline map.
     This will make sure we do not inline recursive functions.
     We also don't want to add any function declaration (as opposed to
     definitions), because that would replace the function call with a Skip.
  *)
```

But it is possible to declare recursive functions without forward declarations.

The following stan file

```stan
functions {
  real dumb(real x);
   real dumb(real x){
     return dumb(x) + 0.5;
   }
}

model {
  target += dumb(1.0);
}
```

produces 
```c++
    try {
      double lcm_sym4__;
      {
        current_statement__ = 1;
        lp_accum__.add(dumb(1.0, pstream__));
      }
    } catch (const std::exception& e) {
      stan::lang::rethrow_located(e, locations_array__[current_statement__]);
    }
    lp_accum__.add(lp__);
    return lp_accum__.sum();
```

Removing the foward declaration produces

```c++
try {
      double lcm_sym7__;
      {
        double inline_sym1__;
        int inline_sym2__ = std::numeric_limits<int>::min();
        for (int inline_sym3__ = 1; inline_sym3__ <= 1; ++inline_sym3__) {
          inline_sym1__ = (dumb(1.0, pstream__) + 0.5);   /* PROBLEM HERE */
          break;
        }
        lp_accum__.add(inline_sym1__);
      }
    } catch (const std::exception& e) {
      stan::lang::rethrow_located(e, locations_array__[current_statement__]);
    }
    lp_accum__.add(lp__);
    return lp_accum__.sum();
```

This specific example seems to be innocuous, but it's worth noting that the invariant described does not hold. I'm not familiar enough to know if this is a bigger problem or not

SteveBronder:
Where is the for loop coming from in the second example?

WardBrian:
I’m not sure, the Stan code was the same as the first example with only one line commented out. The C++ in both cases was generated with -Oexperimental, so it’s possible it’s from a later optimization which is also buggy?

SteveBronder:
I turned on function inlining with O1 and it also produces the loop? That's very weird.

But is there something else wrong with inlining the first level of the recursive function here? The below code seems correct if we are just inlining one iteration of the inliner

```cpp
inline_sym1__ = (dumb(1.0, pstream__) + 0.5);
```

[...]

PR #1097, opened 2022-01-21
Remove for loop trick for inlining functions w/o multiple returns + optims

SteveBronder:
#### Summary

This PR removes for loops surrounding inlined functions if the inlined function does not have more than one return statement while also allow C++ to perform move semantics from the inlined functions inner return to it's outer return object.

After @WardBrian posted #1096 I thought it was odd that inlined functions are always wrapped in a for loop with breaks at the  bottom. So this PR looks at a function as we are inlining it and if the function only has one return statement then the inlined function's code is placed into a `Block` (curly braces) instead of a for loop. For instance the following code is wrapped inside of a for loop that just runs once.

```cpp
        double inline_sym1__;
        for (int inline_sym3__ = 1; inline_sym3__ <= 1; ++inline_sym3__) {
          inline_sym1__ = (blah(1.0, pstream__) + 0.5);   
          break;
        }
```

But we can see that there are not multiple returns here and so instead of having a for loop we can just have a block

```cpp
        double inline_sym1__;
        {
          inline_sym1__ = (blah(1.0, pstream__) + 0.5);   
        }
```

After cleaning that up, I noticed that the inlining code did a pattern where an  variable is declared outside of the block scope, than another variable is declared at the inlined functions scope, and at the end of the inlined functions scope the inner variable is assigned to the outer variable. The below is pseudocode showing what I mean here.

```cpp
Eigen::Matrix<local_scalar_t__, -1, -1> A_outer;
{
  Eigen::Matrix<local_scalar_t__, -1, -1> A_inner;
  // Do a bunch of stuff with A_inner ...
  assign(A_outer, A_inner);
}
```

This is the exact kind of place where we want to do move semantics, where at the end of the scope instead of copying `A_inner` to `A_outer` we can just hand `A_outer` the memory of `A_inner`. This is safe and good because `A_inner` will be deleted at the end of the inlined functions block scope. To let the C++ know an expression can be moved I made an internal `fun_kind` called `FnMove` where if an expression is wrapped in a `FnMove` the C++ code generator can then call `std::move(expr)` such as the below.

```cpp
Eigen::Matrix<local_scalar_t__, -1, -1> A_outer;
{
  Eigen::Matrix<local_scalar_t__, -1, -1> A_inner;
  // Do a bunch of stuff with A_inner ...
  assign(A_outer, std::move(A_inner));
}
```

While I was reading this PR I was having a hard time following which `inline_sym{number}__` was doing what and where. So now when we generate code for inline functions the local inlined variables have the pattern `inline_{function}_{orig_name}_sym{number}__`. The outer return mentioned above now has the name `inline_{func_name}_return_sym{number}__`. The iterator for the for loop trick when we have multiple returns now has the pattern `inline_{func_name}_{iterator}_sym{number}__`.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

## Copyright and Licensing


src/analysis_and_optimization/Optimize.ml
```diff
@@ -106,6 +112,20 @@ let subst_args_stmt args es =
...
+let rec count_returns (acc : int) Stmt.Fixed.{pattern; _} : int =
+  match pattern with
+  | Return _ -> acc + 1
+  | IfElse (_, true_stmt, false_stmt) ->
+      count_returns acc true_stmt
+      + Option.value_map ~default:0 ~f:(count_returns acc) false_stmt
+  | While (_, body) | For {body; _} -> count_returns acc body
+  | Block lst | SList lst | Profile (_, lst) ->
+      List.fold ~init:acc ~f:count_returns lst
+  | _ -> acc
```
WardBrian, line 127: I think this function could just be a `Stmt.Fixed.fold`?

src/analysis_and_optimization/Optimize.ml
```diff
@@ -114,30 +134,41 @@ let subst_args_stmt args es =
...
-                              Expr.Typed.Meta.
-                                { type_= UInt
-                                ; adlevel= DataOnly
-                                ; loc= Location_span.empty } } )
-                ; meta= Location_span.empty }
-            ; Stmt.Fixed.
-                { pattern= Assignment ((name, Expr.Typed.type_of e, []), e)
-                ; meta= Location_span.empty }
-            ; {pattern= Break; meta= Location_span.empty} ]
+          if num_returns > 1 then
```
WardBrian, line 146: I'd prefer this be a guard clause in the match rather than this big if, so you have two branches:
```ocaml
| Some name, Some e when num_returns > 1 -> ... contents of if
| Some name, Some e -> ... contents of else
```

Similar for the other cases where the body of the match is just an if statement

WardBrian (commented):
A few comments questions. The test output on this looks a lot cleaner than the existing stuff!

src/middle/Internal_fun.ml
```diff
@@ -25,6 +25,7 @@ type 'expr t =
   | FnResizeToMatch
   | FnNaN
   | FnDeepCopy
+  | FnMove
```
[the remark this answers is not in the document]
SteveBronder: Yeah it's very C++ specific in that most other languages do not use RAII (Resource Allocation Is Initialization) with deterministic destructors for classes. 

One way I've been thinking of is kind of a big change. Essentially I'd like the optims to be able to tell the code gen extra information they can choose to use. So for instance below we have an additional `Info` type for `Meta`'s record that for `Var` stores whether the var is "movable" or "lower_triangular" (I'm not saying that is how we should do triangularity fyi it's just an example)

```ocaml
module Typed = struct
  module Meta = struct
    module Info = struct 
      type t =
      | Var of { movable: bool; lower_triangular : bool }
      | Lit 
      | FunApp 
      | TernaryIf 
      | EAnd 
      | EOr 
      | Indexed 
      | PromotionMeta
      [@@deriving sexp, hash, map, fold, compare]
    end
    type t =
      { type_: UnsizedType.t
      ; loc: (Location_span.t[@sexp.opaque] [@compare.ignore])
      ; adlevel: UnsizedType.autodifftype
      ; info : Info.t }
    [@@deriving compare, create, sexp, hash]
...
```

This change would also mean that where we currently have strings for function argument inputs we would need to make them `Var` types. . So the optimizer can go through and say, "This is the last time we use object X so after this call we are free to move the object so mark `movable` as true". Or when doing the transformed MIR we can look for which objects are type `LowerTriangular` and then for code gen mark their associated `Var` types in functions with `lower_triangular = true` so we can print them as `{my_obj}.triangularView<Eigen::Lower>()`. 

The benefit of this is that code gens could always ignore this extra information and still get a compilable program.

But imo I think this is a bigger project we can do at some point. The `FnMove` is pretty similar to internal compiler things like `MakeArray` or even the `Promoter` expression in that it's kind of C++ specific (an R code gen would just done the promotion!). So imo I think it's okie dokie. At worst we can just deprecate it in the future.

src/analysis_and_optimization/Optimize.ml
SteveBronder: Sorry I think I'm just unfamiliar, what would that look like?

src/analysis_and_optimization/Optimize.ml
SteveBronder: Much cleaner!

src/analysis_and_optimization/Optimize.ml
WardBrian: Folding functions will walk over a data structure and call a two argument function on each node. In some languages this is called `reduce` instead of `fold`. The first argument will be the "accumulator", the second will be the node. Since our Stmt actually has two types inside (the expression type and then the statement type itself) you give two functions, one for each.

The result is something like  

```ocaml
let count_returns = 
Stmt.Fixed.fold  (fun acc _ -> acc) 
(fun acc -> function Stmt.Fixed.{pattern=Return _; _} -> acc + 1 | _ -> acc) 
0
```
(I don't have the source in front of me to make sure that compiles, but it should be close)

It basically avoids the need for defining lots of functions that look similar to what you've defined that mostly just walk over all the sub nodes. There are a bunch of usages of it around the optimize code like here:

https://github.com/stan-dev/stanc3/blob/d2358be75753fc13825c5fd45e03c99e67d2011c/src/analysis_and_optimization/Factor_graph.ml#L37-L44

SteveBronder:
Also fyi before we merge this I think I should turn it on for all of the tests that get compiled and run it once through Jenkins to help ensure there are no weird edge cases we are missing

WardBrian:
Do you want this in before the RC? If it can wait we have tons of time to do that

SteveBronder:
It can absolutely wait. Yeah thinking it over that would be good to do. Though I may need some help as that sort of schema is a little over what I'm used to doing

test/integration/good/compiler-optimizations/cppO1.expected
```diff
@@ -11292,32 +12265,594 @@ stan::math::profile_map& get_stan_profile_data() {
...
+      p_multi_ret_vec = in__.template read<
+                          stan::conditional_var_value_t<local_scalar_t__, 
+                            Eigen::Matrix<local_scalar_t__, -1, 1>>>(5);
+      stan::conditional_var_value_t<local_scalar_t__, 
+        Eigen::Matrix<local_scalar_t__, -1, 1>> tp_single_ret_vec =
+         stan::conditional_var_value_t<local_scalar_t__, 
+           Eigen::Matrix<local_scalar_t__, -1, 1>>(
+           Eigen::Matrix<double, -1, 1>::Constant(5,
+             std::numeric_limits<double>::quiet_NaN()));
+      Eigen::Matrix<double, -1, 1> inline_single_ret_fun_return_sym7__;
```
SteveBronder, line 12441: So, I think the DCE scheme is not actually deleting all non-used objects? Else this would be deleted right?

WardBrian:
Is there a reason to update Core? We can, but I’ll need to do another windows port and our CI will need updates

SteveBronder:
No we don't as long as you can tell me how to downgrade the stuff on my laptop after running `opam upgrade` lol. That's the only reason these are updated

WardBrian:
You should be able to do `opam pin core_kernel=0.14.2`

SteveBronder:
`opam pin add core_kernel v0.14.2` did the trick! I think we should add that and `opam pin add fmt 0.8.8` to the setup scripts

WardBrian:
We could, though I think in general there is no reason to run `opam upgrade` in the switch you use for Stan. Effectively everything is pinned

src/analysis_and_optimization/Optimize.ml
```diff
@@ -248,23 +280,60 @@ let rec inline_function_expression propto adt fim (Expr.Fixed.{pattern; _} as e)
...
+                            SComplexRowVector (Expr.Helpers.int 0)
+                        | UFun (_, UnsizedType.ReturnType x, _, _) -> to_sized x
+                        | UFun (_, Void, _, _) | UMathLibraryFunction ->
+                            Common.FatalError.fatal_error_msg
+                              [%message
+                                ( "return type of a function was a void user \
+                                   defined function or math library function."
+                                  : string )] in
+                      Type.Sized (to_sized aa) in
+                let blah = Option.map ~f:unsized_to_sized rt in
```
WardBrian: Rename to something other than blah

src/analysis_and_optimization/Optimize.ml
```diff
@@ -248,23 +280,60 @@ let rec inline_function_expression propto adt fim (Expr.Fixed.{pattern; _} as e)
...
-              let handle = handle_early_returns (Some x) in
+          | Some (rt, args, body) ->
+              let inline_return_name = gen_inline_var fname "return" in
+              let handle =
+                handle_early_returns fname (Some inline_return_name) in
               let d_list2, s_list2, (e : Expr.Typed.t) =
+                let unsized_to_sized rt =
+                  match rt with
+                  | Type.Sized _ as a -> a
+                  | Unsized aa ->
```
WardBrian, line 281: This needs a lot of explanation in the code I think. Both what is doing, why it is needed, why size 0 is safe, etc

WardBrian: Possibly put it in SizedType.ml (or unsizedtype) with a name like `unsafe_unsized_to_sized`

WardBrian (commented):
A few style/organization comments. I think that the logic here is sound in general, but I'd appreciate some clarifying comments in the code, especially around the sized 0 sizedtypes created.

src/analysis_and_optimization/Optimize.ml
SteveBronder: I moved this over to `Mir_utils` as I think this is a thing we are only going to use in the optimizations. Also added a bunch of docs describing the behavior since it is an odd function

SteveBronder:
Ty for looking this over! lmk if you have more Qs or suggestions

WardBrian (approved):
This looks good to me. I propose squashing and merging rather than just a normal merge, but I leave that up to you @SteveBronder

SteveBronder:
Sounds good! I'll merge on Monday

PR #1099, opened 2022-01-25
Remove hacky array syntax parsing, improve error messages

### LINKED ISSUE #953: Incorrect parser errors on variable declarations

WardBrian:
While working on updating the Tuples PR, I realized that it's possible to resolve the shift/reduce conflict while still allowing `array` as an identifier, with a slightly different approach. This makes that PR easier, and greatly improves the ability to make error messages around declarations. See the test output for some examples/improvements. 

This is similar to how @rybern originally implemented it in https://github.com/stan-dev/stanc3/pull/708, with they key difference being that this uses precedence to resolve the conflict rather than reserving `array`. I believe it keeps/improves existing behavior, and until we can reserve array in 2.32 should be fine.

Closes #953. 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Improve error messages for incorrect variable declarations.

## Copyright and Licensing


rybern:
Great solution!

rok-cesnovar (approved):
Nice! Thanks!

WardBrian:
@rok-cesnovar if we merge once tests pass this will make it into the 2.29 release candidate right?

rok-cesnovar:
Yeah, we are still waiting for https://github.com/stan-dev/stan/pull/3095 for the release candidate anyways.

WardBrian:
Great! I think its pretty unlikely this has any incompatibilities (the existing tests all either pass or fail like before, it's just message changes), but it's good it will be part of the RC right away

rok-cesnovar:
Yes of course. We can never be too sure of anything :)

PR #1102, opened 2022-01-27
Move assignment logic into promotion functionality

WardBrian:
Follow on to #1004. This tries to incorporate one of @nhuurre's suggestions there to handle array assignment and literals using the `Promote` expression. I'm thinking Tuples will need something like this as well, if we want to assign (int,int) to (real, int) for example.

I'm not 100% confident this will work as-is (mainly due to my lack of familiarity with the C++ at hand), so I am making this draft to run the tests.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Refactor how arrays have their types promoted internally.

## Copyright and Licensing

### DIFF
--- a/src/frontend/SignatureMismatch.ml
+++ b/src/frontend/SignatureMismatch.ml
@@ -173,6 +173,7 @@ and check_compatible_arguments depth typs args2 :
               else Error (ArgError (i + 1, DataOnlyError)) )
       |> Result.all
 
+let check_of_same_type_mod_conv = check_same_type 0
 let check_compatible_arguments_mod_conv = check_compatible_arguments 0
 let max_n_errors = 5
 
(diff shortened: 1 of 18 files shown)


WardBrian:
@rok-cesnovar this requires more work than I thought, ignore the review and we can revisit after 2.29

WardBrian:
This should be ready for review. A brief summary of changes:

1. Both array expressions and assignment now use the Promotion expression during typechecking.
2. The code formerly in SignatureMismatch for promotion logic has been moved to its own Promotion module
3. In code generation, assignment assumes that C++ will do one level of promotion so it strips away one layer to avoid doing duplicate work (e.g., `double x = 1;` is valid C++, so `double x = promote_scalar<double>(1);` is just silly)
4. Complex scalars were changed to use `=` instead of `assign` (I think this was an oversight before? If there's a reason for it it was not clear to me, as the changed code still compiles)
5. The special-cased code for handling array expressions in codegen was removed, it is all done with Promotion now.
6. I tried my best to remove promotion expressions from things like the partial evaluator -- if we can evaluate it in OCaml directly, we don't need to keep the Promotion information around.

WardBrian:
@rok-cesnovar Would you mind taking a look at this now that the release is out? Thanks!

src/frontend/Promotion.ml
```diff
@@ -0,0 +1,98 @@
...
+  expressions for use in [Ast.Promotion]*)
+type t =
+  | NoPromotion
+  | IntToReal
+  | ToVar (* used in arrays, not functions *)
+  | ToComplexVar (* used in arrays, not functions *)
+  | IntToComplex
+  | RealToComplex
+(* One day:
+   | TuplePromotion of t list *)
```
rok-cesnovar, line 16:
```suggestion
```

src/frontend/Promotion.ml
```diff
@@ -0,0 +1,98 @@
...
+module UnsizedType = Middle.UnsizedType
+
+(** Type to represent promotions in the typechecker.
+  This can be used to return information about how to promote
+  expressions for use in [Ast.Promotion]*)
+type t =
+  | NoPromotion
+  | IntToReal
+  | ToVar (* used in arrays, not functions *)
+  | ToComplexVar (* used in arrays, not functions *)
```
rok-cesnovar, line 12: I am not sure I understand what the comment `used in arrays, not functions` means? Can you give me an example?

rok-cesnovar (approved):

src/frontend/Promotion.ml
WardBrian: We don't (at the moment at least) have any functions which require Var arguments, so these will never be created inside `SignatureMismatch.ml`. But, when building this PR, you do run into situations like this

```stan
parameters {
complex z;
array[2] complex x = {z; 3};  
}
```

Where promoting the 3 to `complex<var>` is necessary (complex<double> would fail).
There are similar cases where we need to promote a double to a var - these don't exist in function calls, ever.

This is why the `promotion_cost` of these is 0, since that promotion cost is for function resolution

rok-cesnovar: Got it. Thanks!

PR #1106, opened 2022-02-01
Add require_* templates to generated functions

### LINKED ISSUE #1105: [BUG] Function overloading when the number of arguments match is a bit brittle
Example of a model that fails to compile:

```stan
functions {
   real foo(row_vector p) {
   	return sum(p) + 1;
   }
   real foo(vector p) {
   	return sum(p) + 1;
   }
}
transformed data{
   vector[5] a = [0.1, 0.1, 0.1, 0.1, 0.1]';
}
parameters {
    real y;
}
model {
    y ~ normal(foo(a), 1);
}
```
In this example model, both functions are generated as the exact same code. See https://bit.ly/3rgYpcY 

The issue is that the UDF arguments are fully templated. The only exceptions are integers and when arrays are involved, then the type is `std::vector<T>`.

Changing the functions to the following fixes that the model compile: 
```stan
functions {
   real foo(row_vector p, real a) {
   	return sum(p) + 1;
   }
   real foo(vector p) {
   	return sum(p) + 1;
   }
}
```

What we need is to use the require template metaprograms. Not fun...

WardBrian:
#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Closes #1105 by adding `require_*` templates to the code generated by user-defined functions.

## Copyright and Licensing

### DIFF
--- a/src/stan_math_backend/Stan_math_code_gen.ml
+++ b/src/stan_math_backend/Stan_math_code_gen.ml
@@ -304,7 +304,7 @@ let pp_fun_def ppf
     pf ppf "%s(@[<hov>%a@]) " name (list ~sep:comma string) arg_strs ;
     Format.close_box () in
   let templates, templated_args = get_templates true `None in
-  pp_templates ~defaults:true ppf templates ;
+  pp_templates ~defaults:(Option.is_some fdbody) ppf templates ;
   pp_sig ppf (fdname, templated_args, `None) ;
   match fdbody with
   | None -> pf ppf ";@ "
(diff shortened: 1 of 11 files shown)


rok-cesnovar (approved):
Looks good to me. Merge assuming tests pass. If @SteveBronder has time to have a look in the next day or two, I wouldn't mind him taking a quick peek at this for a sanity check. 

I think an RC2 is in line with this as well.

Thanks for jumping on this so quick!

WardBrian:
@SteveBronder messaged me to say he can take a look later, which I'd also appreciate (never can be too careful around these templates)

SteveBronder (approved):
Looks good! There's a few cleanups I'd like to do in the codegen but nothing that's a blocker so those can wait till after the release

rok-cesnovar:
One thing that I thought of now. What about if some args are varmat? Or do we not handle varmats with udfs?

SteveBronder:
^not supported yet so all good

PR #1107, opened 2022-02-02
[WIP] Generalize overloading templates

SteveBronder:
#### Summary

This tries to generalize some of the `requires` logic for stan UDFs. 

1. All arguments are templated instead of allowing some to get `int` or `std::vector<int>`. I need to check how this is going to effect multiple declarations that have int and double inputs.

2. Instead of using `std::vector<T0__>` in the signature we use `T{arg_name}__`

3. `promote_scalar_t<>` has been bumped to use `return_type_t<>` which is what we tend to use in the math library

4. `inline` keyword is added to functions and functor `operator()`

5. All `requires` are placed into one template parameter. This is to help cut down on the compile time cost of using requires since instead of having an extra template parameter for each argument we will always only have one additional template parameter

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

## Copyright and Licensing

### DIFF
--- a/src/stan_math_backend/Stan_math_code_gen.ml
+++ b/src/stan_math_backend/Stan_math_code_gen.ml
@@ -224,12 +224,8 @@ let pp_arg ppf (custom_scalar_opt, (_, name, ut)) =
     | Some scalar -> scalar
     | None -> stantype_prim_str ut in
   (* we add the _arg suffix for any Eigen types *)
-  if UnsizedType.is_eigen_type ut then
-    pf ppf "const %a& %s_arg__" pp_unsizedtype_custom_scalar_eigen_exprs
-      (scalar, ut) name
-  else
-    pf ppf "const %a& %s" pp_unsizedtype_custom_scalar_eigen_exprs (scalar, ut)
-      name
+  pf ppf "const %a& %s" pp_unsizedtype_custom_scalar_eigen_exprs (scalar, ut)
+    name
 
 let pp_arg_eigen_suffix ppf (custom_scalar_opt, (_, name, ut)) =
   let scalar =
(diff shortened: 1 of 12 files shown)


WardBrian:
This looks like it will be a lot cleaner. The ease of messing with the template arguments here is really convincing me having more structure to our code gen (i.e. fewer strings being produced directly) is a good direction. I'll try to revisit that during the freeze/while Tuples is out of my hands

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -23,7 +23,72 @@ open Fmt
...
+
+let pp_template_requires_defaults ppf (template : template list) =
+  pf ppf "%a = nullptr" pp_template_requires template
+
+let pp_templates ~defaults ppf templates =
+  let find_requires default_val template =
+    match template with Require _ -> default_val | _ -> not default_val in
+  let templates' = List.filter ~f:(find_requires false) templates in
+  let require_templates =
+    List.filter ~f:(fun x -> find_requires true x) templates in
```
WardBrian, line 74: Obviously too early for a proper review but I just wanted to point out the existence of the super useful [`List.partition_tf`](https://ocaml.janestreet.com/ocaml-core/v0.14/doc/core_kernel/Core_kernel/List/index.html#val-partition_tf) function which does exactly this kind of thing

So you could write something like:

```ocaml
let templates', require_templates = List.partition_tf templates 
       ~f:(function Require -> false | _ -> true)
```

SteveBronder: Thanks! Lol yeah I knew there was some way to easily do this and forgot the name of it

WardBrian:
I believe #1157 handled this, so I'm closing this PR

PR #1110, opened 2022-02-02
[FIX] Partial evaluator only uses fma for elementwise multiplication

SteveBronder:
This PR does a check during the partial evaluator when it sees expressions such as `x * y + z` and only does `fma` if `x` and `y` are not both matrix types. It also adds another pattern match for `Add [z; EltTimes__ [x; y]]` since those are ones we do want to transform to use `fma(x, y, z)`. 

On a side note, I wonder if we should have like a `MatrixTimes__` to separate out the difference between scalar and matrix multiplication?

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fixes https://github.com/stan-dev/cmdstan/issues/1077 so partial evaluator only uses fma for elementwise multiplication

## Copyright and Licensing


src/analysis_and_optimization/Partial_evaluator.ml
```diff
@@ -787,11 +787,41 @@ let rec eval_expr ?(preserve_stability = false) (e : Expr.Typed.t) =
...
                   , [ z
                     ; { pattern=
                           FunApp (StanLib ("Times__", FnPlain, mem), [x; y])
                       ; _ } ] )
+                  when (not preserve_stability)
+                       && not
+                            ( UnsizedType.is_eigen_type x.meta.type_
+                            && UnsizedType.is_eigen_type y.meta.type_ ) ->
+                    let lub_mem = lub_mem_pat [mem] in
+                    FunApp (StanLib ("fma", suffix, lub_mem), [x; y; z])
```
WardBrian, line 806: These first two cases can stay together, right?

SteveBronder: They actually have to be broken up, at least when I did this I get a 

```
Error (warning 57): Ambiguous or-pattern variables under guard;
variables x,y may match different arguments. (See manual section 9.5) ocamllsp
```

WardBrian: Huh, today I learned. There's a great explanation of why this is true in the manual: https://ocaml.org/releases/4.12/htmlman/comp.html#ss%3Awarn57

WardBrian (approved):
LGTM, thanks for hopping on this so quick

SteveBronder:
Np!

PR #1113, opened 2022-02-08
Clean up pretty printing of array[]

WardBrian:
This pulls out some changes made in #1100 which are also useful for things like complex matricies and just generally an attempt to clean up how we're printing types in the pretty printer. The output is identical to the existing code.

@rybern made a similar comment during work on #675 
> This seems like a mess to me. Why are we "discarding" arrays instead of just using unwind_sized_array_type to group them when they're printed? It seems like unwind_sized_array_type is called multiple times on the same input. Should sized types, unsized types, and transformations really be handled in the same function? Why split off pp_transformed if we're already matching on the transformation here?

I've tried to give functions better names and avoid doing this extra work he describes.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Cleaned up internal pretty-printing code

## Copyright and Licensing


test/integration/good/pretty.expected
```diff
@@ -1352,6 +1352,13 @@ generated quantities {
 data {
   //nothing here either
 }
+transformed data {
+  int N;
+  array[N /*test1*/, //test1.5
+        N] real arr;
+  
+  /* ^^^:test2*/
```
WardBrian, line 1360: Note: This is also the behavior on `master` (currently no tests expose this)

PR #1130, opened 2022-02-22
Only show assignment operators which match the lhs in error messages

WardBrian:
Currently our type errors produced from assignment operators like `*=` are very different than something like `operator*`. In particular, we don't truncate the list at all. This PR introduces two changes:

1. Truncate the shown signatures at 5 (same as any other signature error)
2. Only show signatures which match the LHS provided. For example,
    ```stan
    row_vector[N] x;
    x *= x;
    ```
    Does not need to show the `int*int` signature, but should show `row_vector*real`, etc.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Improve error messages when an assignment operator like `+=` is given bad types.

## Copyright and Licensing

### DIFF
--- a/src/frontend/SignatureMismatch.ml
+++ b/src/frontend/SignatureMismatch.ml
@@ -368,21 +368,17 @@ let pp_signature_mismatch ppf (name, arg_tys, (sigs, omitted)) =
 
 let pp_math_lib_assignmentoperator_sigs ppf (lt, op) =
   let signatures =
-    match op with
-    | Operator.Plus | Minus | Times | Divide | EltTimes | EltDivide -> (
-        let errors =
-          Stan_math_signatures.make_assignmentoperator_stan_math_signatures op
-        in
-        let errors =
-          List.filter
-            ~f:(fun (_, args, _) ->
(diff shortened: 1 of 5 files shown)


src/frontend/SignatureMismatch.ml
```diff
@@ -365,3 +365,30 @@ let pp_signature_mismatch ppf (name, arg_tys, (sigs, omitted)) =
     name pp_args arg_tys
     (list ~sep:cut pp_signature)
     sigs pp_omitted ()
+
+let pp_math_lib_assignmentoperator_sigs ppf (lt, op) =
+  let signatures =
+    match op with
+    | Operator.Plus | Minus | Times | Divide | EltTimes | EltDivide -> (
```
nhuurre, line 372: These are all the assignment operators we have, right? I don't see what this match is supposed to do. If it's just guarding against bad inputs, I think it would make more sense to put it inside `make_assigmentoperator_stan_math_signatures`.

WardBrian: I think we can actually remove it entirely. It was a holdover from when the function which did this was returning a string option rather than printing

PR #1132, opened 2022-02-22
Improve tilde statement type errors

WardBrian:
Currently, no matter what goes wrong when typechecking a statement like `x ~ foo(3)`, the same error is output if `foo` is invalid:
`Ill-typed arguments to '~' statement. No distribution 'foo' was found with the correct signature.`

This is much weaker than for other function errors, which will output valid signatures or ambiguous match results. This PR simply tries to make them look more like function signature errors. It's not perfect (a 100% ideal error might try to re-construct the `~` notation rather than formatting them like the function calls they are under the hood) but it's certainly much more informative than before.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Improve type errors generated in `~` statements.

## Copyright and Licensing

### DIFF
--- a/src/frontend/Typechecker.ml
+++ b/src/frontend/Typechecker.ml
@@ -1082,8 +1082,7 @@ let verify_sampling_distribution loc tenv id arguments =
   let name_w_suffix_sampling_dist suffix =
     SignatureMismatch.matching_function tenv (name ^ suffix) argumenttypes in
   let sampling_dists =
-    List.map ~f:name_w_suffix_sampling_dist
-      (Utils.distribution_suffices @ Utils.unnormalized_suffices) in
+    List.map ~f:name_w_suffix_sampling_dist Utils.distribution_suffices in
   let is_sampling_dist_defined =
     List.exists
       ~f:(function UniqueMatch (ReturnType UReal, _, _) -> true | _ -> false)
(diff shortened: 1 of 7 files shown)


src/frontend/Typechecker.ml
```diff
@@ -1079,54 +1079,49 @@ let verify_valid_sampling_pos loc cf =
...
-    | UniqueMatch (ReturnType UReal, _, _) -> true
-    | _ -> false in
-  let is_sampling_dist_in_math =
-    List.exists ~f:is_name_w_suffix_sampling_dist
-      (Utils.distribution_suffices @ Utils.unnormalized_suffices)
+  let name_w_suffix_sampling_dist suffix =
+    SignatureMismatch.matching_function tenv (name ^ suffix) argumenttypes in
+  let sampling_dists =
+    List.map ~f:name_w_suffix_sampling_dist
+      (Utils.distribution_suffices @ Utils.unnormalized_suffices) in
```
nhuurre, line 1086: As far as I know `matching_function` standardizes the suffix before name lookup so `distribution_suffices` already covers all functions here and `unnormalized_suffices` just adds duplicates to the list.

nhuurre (changes requested):
Thanks, this is a much-needed feature. Are you going to handle bad CDFs in truncation too?

src/frontend/Typechecker.ml
WardBrian: Huh, you're right. I was just following existing behavior here, which means it's been doing extra work for who-knows how long

WardBrian:
Truncation checks on the CDF are a bit trickier due to the fact that the requirements AND over each other rather than OR. I think it's a less common use case, so while it's still worth fixing it might be better to do a separate PR

nhuurre (approved):

PR #1140, opened 2022-03-04
expose norm1 & norm2

yizhang-yiz:
#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/501

## Release notes

L1 and L2 norms.

## Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to  Metrum Research Gruop.

### DIFF
--- a/src/middle/Stan_math_signatures.ml
+++ b/src/middle/Stan_math_signatures.ml
@@ -1625,6 +1625,12 @@ let () =
     , SoA ) ;
   add_nullary "negative_infinity" ;
   add_unqualified ("norm", ReturnType UReal, [UComplex], AoS) ;
+  add_unqualified ("norm1", ReturnType UReal, [UArray UReal], SoA) ;
+  add_unqualified ("norm1", ReturnType UReal, [UVector], SoA) ;
+  add_unqualified ("norm1", ReturnType UReal, [URowVector], SoA) ;
+  add_unqualified ("norm2", ReturnType UReal, [UArray UReal], SoA) ;
+  add_unqualified ("norm2", ReturnType UReal, [UVector], SoA) ;
+  add_unqualified ("norm2", ReturnType UReal, [URowVector], SoA) ;
   add_unqualified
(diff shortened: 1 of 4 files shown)


WardBrian:
@rok-cesnovar this seems to be failing due to the math library, any thoughts?

rok-cesnovar:
Yeah, the problem is that the reverse mode implementation for arrays is missing. It's only implemented for data types.

We can go ahead and fix that or just expose vector/row_vector signatures first.

yizhang-yiz:
We don't want to give user surprises. I can fix the math repo later.

andrjohns:
@yizhang-yiz for the Math changes, the easy fix here is to restrict the `rev`/`fwd`/`prim` specialisations to only take eigen inputs, and then add an `apply_vector_unary` overload to the `prim` header.

For example, the `prim` header would then look like:
```cpp
template <typename T,
          require_eigen_vector_vt<std::is_arithmetic, T>* = nullptr>
inline double norm1(const T& v) {
  ref_type_t<T> v_ref = v;
  return v_ref.template lpNorm<1>();
}

template <typename Container, require_std_vector_t<Container>* = nullptr>
inline auto norm1(const Container& x) {
  return apply_vector_unary<Container>::reduce(
      x, [](const auto& v) { return norm1(v); });
}
```

And the templates for `fwd` and `rev`:
```cpp
template <typename Container,
          require_eigen_vector_vt<is_fvar, Container>* = nullptr>
```
```cpp
template <typename T, require_eigen_vector_vt<is_var, T>* = nullptr>
template <typename T, require_var_matrix_t<T>* = nullptr>
```

The `apply_vector_unary` overload will automatically `Eigen::Map` the `std::vector` input before calling the appropriate specialisation (it also allows the function to be called with nested inputs, like `array[] vector`).

WardBrian (approved):

PR #1144, opened 2022-03-10
Turn off unenforce-initialize for assignment of doubles to ad types

### LINKED ISSUE #1141: [BUG] Model crash when compiled with "--O1"
### Current Behavior:
Compiling certain models using stanc with "--O1" generates a model that crashes with a segmentation fault. The same model appears to work fine when compiled with "--O0" or with Stan versions prior to 2.29.0. It looks like it might be related to the update of a multi-dimensional array in a loop.

### Expected Behavior:
The model should run with "--O1".

### Model Code
Here's what I was able to come up with for a minimal example.
```stan
transformed data {
  vector[5] inputs[2] = rep_array(zeros_vector(5), 2);
}
parameters {
  real param;
}
model {
  vector[5] local[2] = inputs;
  for(i in 1:2) {
    for(j in 1:5) {
      local[i, j] += 1;
    }
  }
  param ~ std_normal();
}
```

### Environment:
- OS: macOS 12.2.1 (also happens on Linux) 
- stanc: 2.29.1
- cmdstan: 2.29.1

SteveBronder:
## Summary 

Fixes #1141. The issue is that in c++ `std::vector<var>` and `std::vector<double>` are not assignable to one another, so we need to call the assignment that loops over each element of the assigned from vector and does assignment directly to the sub element. If `unenforce-initialize` is turned on then that vector will not have been filled yet and so we'll be accessing bad memory and getting a seg. This fixes that by checking in `find_assignment_idx` whether the rhs is array of data types and if so then failing the optimization since the decleration does need to be initialized with NA.

The mre from #1141 was added in the optimization test suite (Much appreciated @joewing!!)

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Bug fix for `--O1` when arrays declared in `transformed parameters` or `model` were assigned data 

## Copyright and Licensing


rok-cesnovar (approved):
Looks good to me! Thanks Steve. 

Do we want to do a .2 here or should we wait if something else comes up? Its fairly easy for users to switch to the nightly stanc3 release if needed, so maybe we hold off?

SteveBronder:
Ty! How long until our next release? If it's more than two months than a .2 might be good but otherwise I think it's fine to just put this in the next release.

PR #1146, opened 2022-03-15
Fix compilation error with forward declarations

WardBrian:
#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fixes a bug reported here: https://discourse.mc-stan.org/t/cmdstan-2-29-1-compile-error/26793/10

This updates the test `test/integration/good/code-gen/overloading_templating.stan` such that it fails on `master`

## Copyright and Licensing

### DIFF
--- a/src/stan_math_backend/Stan_math_code_gen.ml
+++ b/src/stan_math_backend/Stan_math_code_gen.ml
@@ -249,10 +249,11 @@ let mk_extra_args templates args =
 *)
 let pp_fun_def ppf
     ( Program.{fdrt; fdname; fdsuffix; fdargs; fdbody; _}
-    , functors
-    , funs_used_in_reduce_sum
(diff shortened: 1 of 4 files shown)


WardBrian:
I don't love how this function is implemented. Doing it in once pass is the fastest way, but in all likelihood we should either do it in multiple passes or use a lower level intermediate representation. I don't think that kind of refactor is worthwhile during a bugfix though

WardBrian:
I'm gonna let @rok-cesnovar possibly do some more testing on this before hitting merge. He can do that and then we can release this as 2.29.2

rok-cesnovar:
Looks good, no complaints on my side.

As for the 2.29.2 patch release, we need to see if we can easily fix https://github.com/stan-dev/stanc3/issues/1150 as well.

PR #1149, opened 2022-03-16
Disable AD levels optimization in `--O1`

rok-cesnovar:


SteveBronder:
fyi running a jenkins job with a math PR to check if I have the math fixes right

https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FStanc3/detail/PR-1149/7/pipeline/249

SteveBronder:
Okay! Now I want to turn on the partial evaluator to see if my math fixes also fix that

SteveBronder:
Mkay! So it looks like with those two math PRs we are good for the partial evaluator

rok-cesnovar:
@SteveBronder does that mean that we merge the 2 math PRs and we should be good to go for 2.29.2?

rok-cesnovar:
Oh and disable AD levels for O1?

SteveBronder:
Yep!

WardBrian:
Is the AD levels optimization a low impact one?

SteveBronder:
Its an optimization where imo the user could also just move the data only operation to `transform data` themselves. imo not a huge one. I think part of the issue is just that we shouldn't run it over functions. The main place it is beneficial is in log_prob

rok-cesnovar:
Both Math PRs were just merged in, I cleaned up this PR which will now just disable AD levels optimizations in O1. Running it once more with `--O1` to make sure all models compile and will then remove changes to Jenkinsfile and tag for review.

WardBrian (approved):
LGTM. Before releasing the patch I'd like to see #1151 merged, hopefully also with the testing from performance-tests-cmdstan

PR #1154, opened 2022-03-18
Adds documentation for exposing a distribution function signature.

valentin12:
See [Issue (stan-dev/stan): Wiki section outdated: Contributing New Functions to Stan 4) Exposing Functions to Stan #3109 ](https://github.com/stan-dev/stan/issues/3109)
The process of adding a new function to Stan involves changes
to src/middle/Stan_math_signatures.ml, but is currently not documented.
This commit adds very basic documentation for exposing a distribution
function to stanc3. It is incomplete, as only few examples and little
explanation is given.
@WardBrian: There was some guesswork involved, could you review the
text so no wrong information is supplied? Some further information on the
format would probably be helpful, but I lack the knowledge to add more.

#### Submission Checklist

- [ ] Run unit tests
- Documentation

## Release notes
Documentation: Process of exposing Stan Math library functions is explained

## Copyright and Licensing

### DIFF
--- a/docs/exposing_new_functions.mld
+++ b/docs/exposing_new_functions.mld
@@ -1,0 +1,31 @@
+{0 Exposing New Functions to stanc3}
+
+
+
+{1 Background}
+
+For a function to be built into Stan, it has to be included in the Stan Math library and its signature has to be
+exposed to the parser. To do the latter, we have to add a corresponding line in [src/middle/Stan_math_signatures.ml].
+The parser uses the signatures defined there to do the type checking. 
+
+
+{1 Adding a distribution function }
+To add a distribution, we have to find the line containing [let distributions =]. The existing distributions can be
+used for reference. The first argument defines the type of function we want to add.
+The second argument is the function name without the [_...] part. The third argument specifies
+the argument types of the function. The last argument describes whether the return type can be represented as an
+Array of Structs (AoS) or as a Struct of Arrays (SoA). The following line exposes a function [foo_lpdf] which takes in
+two real numbers:
+{[
+  ; ([Lpdf], "foo", [DVReal; DVReal], SoA)
+]}
+
+For overloaded functions, we need to add a signature for each combination of variable types. If [foo_lpdf] admits either
+an integer or a real as its first argument, the proper way to expose the function would be:
+{[
+  ; ([Lpdf], "foo", [DVReal; DVReal], SoA)
+  ; ([Lpdf], "foo", [DVInt; DVReal], SoA)
+]}
+
+Once we modified [Stan_math_signatures.ml], we need to compile stanc3.
+
--- a/docs/index.mld
+++ b/docs/index.mld
@@ -19,6 +19,8 @@ to a new developer.
   Wiki on using [Format]/[Fmt] module}
 - {{!page-parser_messages}
   Page on modifying the parser's error messages}
+- {{!page-exposing_new_functions}
+  Page on exposing a new function of the Stan Math library}
 - {{:https://github.com/stan-dev/stanc3/wiki/Software-Engineering-best-practices}
   Wiki on generic software best practices}
 


docs/exposing_new_functions.mld
```diff
@@ -0,0 +1,31 @@
...
+For a function to be built into Stan, it has to be included in the Stan Math library and its signature has to be
+exposed to the parser. To do the latter, we have to add a corresponding line in [src/middle/Stan_math_signatures.ml].
+The parser uses the signatures defined there to do the type checking. 
+
+
+{1 Adding a distribution function }
+To add a distribution, we have to find the line containing [let distributions =]. The existing distributions can be
+used for reference. The first argument defines the type of function we want to add.
+The second argument is the function name without the [_...] part. The third argument specifies
+the argument types of the function. The last argument describes whether the return type can be represented as an
```
WardBrian, line 16: It's worth describing what these types are a bit more. For example, the thing you use in "foo" (`DVReal`) can be a real, array[] real, row_vector, or vector (`V` stands for "vectorized")

docs/exposing_new_functions.mld
```diff
@@ -0,0 +1,31 @@
...
+  ; ([Lpdf], "foo", [DVReal; DVReal], SoA)
+]}
+
+For overloaded functions, we need to add a signature for each combination of variable types. If [foo_lpdf] admits either
+an integer or a real as its first argument, the proper way to expose the function would be:
+{[
+  ; ([Lpdf], "foo", [DVReal; DVReal], SoA)
+  ; ([Lpdf], "foo", [DVInt; DVReal], SoA)
+]}
+
```
WardBrian, line 29: I think we should also add another segment on adding functions which aren't distributions. This uses a bunch of helper functions which may be harder to understand, so I can write it if you'd like.

It begins on the line 

```ocaml
(* -- Start populating stan_math_signaturess -- *)
let () =
```

WardBrian (commented):
Thanks so much for getting this started. I think everything you wrote is correct, but there is a bit more we should say on this. If you don't feel like writing the rest let me know and I would be happy to

docs/exposing_new_functions.mld
```diff
@@ -0,0 +1,31 @@
...
+{[
+  ; ([Lpdf], "foo", [DVReal; DVReal], SoA)
+]}
+
+For overloaded functions, we need to add a signature for each combination of variable types. If [foo_lpdf] admits either
+an integer or a real as its first argument, the proper way to expose the function would be:
+{[
+  ; ([Lpdf], "foo", [DVReal; DVReal], SoA)
+  ; ([Lpdf], "foo", [DVInt; DVReal], SoA)
+]}
```
WardBrian, line 28: This is the correct way to do this but I don't believe we have any distributions overloaded in this way. We do use it in the normal functions section, like:

```ocaml
 add_unqualified ("multiply", ReturnType UComplex, [UComplex; UComplex], AoS) ;
  add_unqualified ("multiply", ReturnType UInt, [UInt; UInt], SoA) ;
  add_unqualified ("multiply", ReturnType UReal, [UReal; UReal], SoA) ;
  add_unqualified ("multiply", ReturnType UVector, [UVector; UReal], SoA) ;
...
```

docs/exposing_new_functions.mld
valentin12: I'd be happy to help adding some more details, I'll see if I can understand what's going on and expand this a bit.

docs/exposing_new_functions.mld
valentin12: Yes, I'd be happy if you could write this part.

docs/exposing_new_functions.mld
valentin12: I'm not sure whether I understand correctly, could you elaborate on this? I wrote this in analogy to the way the `wiener` function is [declared](https://github.com/stan-dev/stanc3/blob/c0a37eff2e19b10163e04c4267dff5031a399761/src/middle/Stan_math_signatures.ml#L305), so these two lines would be inside `distributions`.

WardBrian: My point was that this list of distributions uses the types like `DVReal` to automatically generate overrides, there are not any distributions which explictly define an override as you do in this example.

So, for something like the `wiener` distribution, all 1024 available signatures are generated only by one line. This is different from the 'normal' functions like `multiply(x,y)`, which need to have each signature written out for each overload

WardBrian:
I have added sections on non-distribution functions, as well as brief descriptions of the testing and documentation required

valentin12:
I have added a description of the parameter types and the distribution function types. Could you have another look and check whether the terminology is correct? Is there anything missing that should be in there?

docs/exposing_new_functions.mld
valentin12: Ok, I understand, thanks for the clarification.

WardBrian (approved):
This looks great!

PR #1163, opened 2022-04-04
ensure write_array writes all parameters even if there's an exception

nhuurre:
Fix stan-dev/stan#3114.
I'm pretty sure constraints on generated quantities never had any effect other than suppressing output so with this they're even more useless.

#### Submission Checklist

- [X] Run unit tests
- Documentation

(are bugfixes "user-facing"?)

## Release notes

Exceptions in the generated quantities no longer hide all generated quantities; the ones assigned before the exception are now available in the output.

## Copyright and Licensing


fergusjchadwick:
Thanks for jumping on this @nhuurre. Not sure whether to add this to the issue or here (this is my first issue submission!).

> I'm pretty sure constraints on generated quantities never had any effect other than suppressing output so with this they're even more useless.

I do find them useful in generated quantities (or how I imagine they work/could work), for example, making sure something is a simplex following transformation is very handy.

bob-carpenter:
@nhuurre :  I just saw the issue and feel very strongly that this should *not* be implemented for reasons I describe on the issue:  https://github.com/stan-dev/stan/issues/3114#issuecomment-1087640525

I am not sure about our process at this point.  I would like to close this so it's never merged.  But then you can get a vote of devs to overrule me.  I'm going to add a blocking review right now.

nhuurre:
This PR now depends on stan-dev/stan#3116 and with that, the behavior is as Rok described on the issue https://github.com/stan-dev/stan/issues/3114#issuecomment-1087655016

SteveBronder:
@bob-carpenter @nhuurre how do you feel about the alt approach in #1165? If any part of write_array fails then we just return NaN values for everything that is not a parameter

bob-carpenter:
> If any part of write_array fails then we just return NaN values for everything that is not a parameter

This is my preference.

WardBrian:
We decided to go with #1165 so I'm going to close this

PR #1169, opened 2022-04-08
Add vectorized signatures for get_real and get_imag

WardBrian:
#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/515

## Release notes

Added vectorized signatures for `get_real` and `get_imag`, e.g. `get_real(complex_matrix) => matrix`

## Copyright and Licensing


src/middle/Stan_math_signatures.ml
```diff
@@ -103,6 +103,14 @@ let rec ints_to_real unsized =
   | UArray t -> UArray (ints_to_real t)
   | x -> x
 
+let rec complex_to_real = function
```
SteveBronder, line 106: Can we put this in UnsizedType?

SteveBronder (changes requested):
One small change then this is good

src/middle/Stan_math_signatures.ml
WardBrian: We could but I don't think we should. Like the function above it (`ints_to_real`) I think the only safe uses for it in the compiler are in writing signatures. If you can think of other places we would want it I can move it over

SteveBronder (approved):
Good!

PR #1161, opened 2022-03-30
Use Yojson for --info output

WardBrian:
This is similar to what we do in code generation, and will make writing the types of tuples much easier (e.g #820). The contents of the output are identical.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [ ] If a user-facing facing change was made, the documentation PR is here: <LINK>
    - [x] OR, no user-facing changes were made

## Release notes

The `--info` command now uses the Yojson library for it's output.

## Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause)

PR #1175, opened 2022-04-20
Pretty-print debug data using yojson

WardBrian:
#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Similar to #1161, `--debug-generate-data` uses Yojson rather than defining a bespoke json pretty-printer

## Copyright and Licensing


src/analysis_and_optimization/Debug_data_generation.ml
```diff
@@ -274,27 +274,34 @@ and generate_value m st t =
...
+      let value = generate_value m sizedtype transformation in
+      ((name, value) :: l, Map.set m ~key:name ~data:value) )
+  |> fst |> List.rev
+
+open Yojson
+
+let generate_json_entries (name, expr) : string * t =
+  let rec expr_to_json e : t =
+    match e.Expr.Fixed.pattern with
+    | Lit (Real, s) when String.is_suffix s ~suffix:"." -> `Floatlit (s ^ "0")
```
nhuurre, line 289: This suffix fudging is kind of ugly. Changing this to `Float (Float.of_string s)` seems to lose one decimal, not sure if that matters (and not sure if it's any less of a hack...)

nhuurre (approved):
Great. Just one optional comment.

src/analysis_and_optimization/Debug_data_generation.ml
WardBrian: I don't think JSON makes any distinction between `0`, `0.`, and `0.0`, does it?

The best way of doing this would be to avoid turning the float we generate higher up in the file into a string in the first place, but that's not really easy to do with the way Expressions are set up. 

We could also change `Float.to_string` to something like `Float.to_string_hum ~decimals:16 ~strip_zero:false`, which will generate `0.0000000000000000` instead, though obviously it would be great if there was a 'strip all but one' zeros option

WardBrian: Alternatively, we could move this match logic into `Expr.Helpers.float`

nhuurre: I think `0` and `0.0` are the same but `0.` is a syntax error. I don't like `0.0000000000000000`. This isn't really worth overengineering so let's leave it as a suffix check for now.

Issue #1054, opened 2021-11-22
Should trailing commas be allowed in functions, etc?


WardBrian:
Currently, the following produces a parse error:

```stan
functions {
  void foo(real x, real y,){
    return;
  }
}
```

Similar for function calls `print("Hello", "World",)`

In Python a trailing comma is often used for when a function signature may change, since it allows the diff to be smaller

e.g, changing 

```python
def foo(a,
    b,
    ):
```
to 

```python
def foo(a,
    b,
    c,
    ):
```

only has a 1-line diff, but without the trailing comma both the previous final argument and the new final argument would need to change.

I am not an R programmer, but I believe R does not allow this.

rok-cesnovar:
R does not allow this. But this does seem like a useful feature to me, the git diff is really annoying when you add an argument. If this doesnt complicate anything, I am fine with this.

bob-carpenter:
I didn't see this the first time around, but I don't like changing the language to make git diffs cleaner.  I think it's going to be confusing optionally having trailing commas.  I also don't like these one-line per argument formats---our auto formatter will get rid of them anyway, right?

that is, we'll never see something like this

```stan
foo(a,
    b,
...);
```

in code.

WardBrian:
Right now the autoformatter would get rid of them, yes. It's a design choice whether we would want to keep them - `black` (a popular python formatter) uses a trailing comma as a sign to split over multiple lines:

```shell
$ echo "def foo(x,y): pass" | black -
def foo(x, y):
    pass


$ echo "def foo(x,y,): pass" | black -
def foo(
    x,
    y,
):
    pass
```

I think the amount of additional work to do this in our formatter would not be worth it for such a minor thing, but in principle it would be possible to do the same.

adamhaber:
@WardBrian what would be the best way to proceed with this issue (if at all)?

bob-carpenter:
Hi, @admahber.  This is just my opinion.  I'm not going to stand in the way of an issue if you and @WardBrian think it's the right thing to do.

adamhaber:
I don't have a strong opinion on this; I can see the practical benefit in making diffs clearer, but also the advantages of  keeping the language more "strict"/"terse" in this sense. I'm fine with whatever @rok-cesnovar and @WardBrian would decide on this.

WardBrian:
[...]

PR #1176, opened 2022-04-21
Feature/allow trailing comma

adamhaber:
Address #1054 .

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Allow trailing commas in functions to make git diffs less cluttered.

## Copyright and Licensing


src/frontend/parser.mly
```diff
@@ -810,3 +810,13 @@ top_vardecl_or_statement:
     { grammar_logger "top_vardecl_or_statement_statement" ; [s] }
   | v=top_var_decl
     { grammar_logger "top_vardecl_or_statement_top_vardecl" ; v }
+
+arg_list_trailing_comma(sep, X):
```
WardBrian, line 814: Can we make this %inline?

Also, if we want to name it `arg_list_trailing_comma` I don't think we need to take in sep, X as arguments, we can hard-code them as COMMA and arg_decl.

If we want to use this anywhere else (I can't think of any such place at the moment) we should keep it as a higher-order rule and rename it to something like `seperated_list_trailing`

test/integration/good/parser-generator/fun-args-trailing-comma.stan
```diff
@@ -0,0 +1,5 @@
+functions {
+  void foo(real x, real y,){
```
WardBrian, line 2: Could you also add a test to /bad which shows that something like `void foo(real x, real y ,,)`  [note the multiple , s] fails?

test/integration/good/parser-generator/pretty.expected
```diff
@@ -107,6 +107,13 @@ transformed data {
   print("expect: [3,7,7,8,5]  v: ", v);
 }
 
+  $ ../../../../../install/default/bin/stanc --auto-format fun-args-trailing-comma.stan
+functions {
+  void foo(real x, real y) {
```
WardBrian, line 112: I wonder how difficult it would be to preserve these commas through pretty-printing. If they idea is that this better supports a git workflow, we probably don't want it to be counter to using the auto-formatter.

Also, can you add some comments between/after the commas here to see what happens when pretty-printing?

WardBrian:
Also, the original issue also included trailing commas in _calls_ - this is a bit more of an invasive change, since it would need to edit the parsing of function calls, sampling, and print/reject

WardBrian:
Thanks @adamhaber - I'm going to close this per the discussion in #1054. We may want to look at it again one day.

PR #1177, opened 2022-04-21
Show source code context with errors in Javascript

WardBrian:
This refactors some of the error printing code to isolated the file-reading from the logic which creates an error.

As a bonus this allows us to pass in the code string from the JS compiler to also format an error.

So, something which currently looks like
```
Semantic error in 'string', line 3, column 8 to column 11: 
Function is declared without specifying a definition.
```
Would now look like 
```
Semantic error in 'string', line 3, column 8 to column 11:
     1:  
     2:  functions {
     3:      int foo(real a);
                 ^
     4:  }
     5:  transformed data {

Function is declared without specifying a definition.
```

This example is taken from the test output of the stancjs integration tests.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

The javascript interface (stancjs) now generates errors which look more like stanc3's by pointing out where in the supplied code the error occurred. 

## Copyright and Licensing

### DIFF
--- a/src/frontend/Errors.ml
+++ b/src/frontend/Errors.ml
@@ -22,46 +22,49 @@ type t =
   | Syntax_error of syntax_error
   | Semantic_error of Semantic_error.t
 
-let pp_context_with_message ppf (msg, loc) =
-  Fmt.pf ppf "%a@,%s" (Fmt.option Fmt.string)
-    (Middle.Location.context_to_string loc)
-    msg
+let pp_context_with_message ?code ppf (msg, loc) =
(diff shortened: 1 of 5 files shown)


hsbadr:
> I want to check with the RStan developers that this won't cause any headaches for them

@WardBrian Looks good to me. The error recovery of `stanc3` JS in RStan is handled by `try` function. This will actually provide more informative errors, as mentioned in the OP.

rok-cesnovar (approved):

PR #1179, opened 2022-04-26
Clean up module interfaces

### LINKED ISSUE #358: .mli file usage
We've been going back and forth for a while on where we should be using .mli files are where they are unnecessary. I've now asked a couple of questions related to this on OCaml discuss [here](https://discuss.ocaml.org/t/jane-streets-intf-ml-usage/3848) and [here](https://discuss.ocaml.org/t/guidance-on-mli-files/4575) and wanted to summarize my understanding and get some consensus on this.

The first question I asked was about why Jane Street use three files `[X].ml`, `[X].mli` and `[X]_intf.mli` in their libraries. The answer is basically that it saves repeating, and therefore having to update, signatures in two places. The approach is only needed when a module exposes a functor.

My second question was to clarify why you would use an interface file in three situations. The easy cases are:

- When you want to hide the implementation you clearly need to use one.
- When a file only exposes a module signature you do not need one.

The third case is where we have an implementation file but the interface doesn't hide anything. The response to my questions suggested that this is indeed discretionary but [recommended avoiding using `module type of`](https://discuss.ocaml.org/t/guidance-on-mli-files/4575/4) which we do in a number of places at the moment. Another respondent suggested that .mli files are usually added when a module is matured and the [Jane Street style guide recommends putting your docs in the .mli](https://opensource.janestreet.com/standards/).

The final feedback was that there is no need to include 'index modules' like [`Common.ml`](https://github.com/stan-dev/stanc3/blob/master/src/common/Common.ml) or [`Middle.ml`](https://github.com/stan-dev/stanc3/blob/master/src/middle/Middle.ml) since dune generates this for you. This is true even in cases where you don't want to expose all the modules in the library since this can be handled using dunes `private_modules` stanza.

I asked about the impact of .mli on compilation times but this wasn't mentioned a pro/con by any respondents.

Based on this, my instinct is to:

1. Move to the Jane Street style of `[X].ml`, `[X].mli` and `[X]_intf.mli` for modules exposing functors
2. Remove usage of `module type of`
3. Remove 'index modules' (which should mostly take care of point 2).
4. Start to create .mli where the implementation is more settled and being to add docs there.

This would save us a few files (with point 3) but cost us a few with point 1. However, adopting 1 will help with maintenance.

@rybern @seantalts @VMatthijs I'd like to get your thoughts on this before making any changes.

WardBrian:
This does a few things:

1. Closes #358 by deleting the 'index' modules which dune creates for us, like Middle.ml
2. Adds interfaces and documentation for a couple more existing modules, namely `Canonicalize` and `Stan_math_signatures`
3. Removes `Frontend_utils` by finding better homes for what lived there (basically all of it's functions were only called in one place and not generalizable)

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Cleaned up the internal structuring of modules in the compiler

## Copyright and Licensing


rok-cesnovar (approved):
🧹

PR #1183, opened 2022-05-03
Fix tests for optimizations

rok-cesnovar:
#### Submission Checklist

- [ ] Run unit tests
- Documentation

@WardBrian can we add a "simple" optimization bug to stress test this? Something that will produce the wrong results.

## Copyright and Licensing

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -447,7 +447,7 @@ pipeline {
                                 echo "CXXFLAGS+=-march=core2" > cmdstan/make/local
                                 echo "PRECOMPILED_HEADERS=false" >> cmdstan/make/local
                                 cd cmdstan; make clean-all; git show HEAD --stat; cd ..
-                                CXX="${CXX}" ./compare-optimizer.sh "--tests-file all.tests --num-samples=10" "--O1"
+                                CXX="${CXX}" ./compare-optimizer.sh "--tests-file all.tests --num-samples=10" "--O1" "\$(readlink -f ../bin/stanc)"
                             """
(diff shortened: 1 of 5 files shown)


WardBrian:
I've introduced a bug which swaps the order of arguments to `fma` in the partial evaluator. If it still passes tests that is probably because none of the tests we're running feature a float multiply add. We next need to address https://github.com/stan-dev/performance-tests-cmdstan/issues/38, e.g. we need more tests in this set that get optimized.

rok-cesnovar:
> We next need to address https://github.com/stan-dev/performance-tests-cmdstan/issues/38, e.g. we need more tests in this set that get optimized.

Agreed. Where do we want to put these? In the perf-cmdstan repo or stanc3?

We can then do the same for https://github.com/stan-dev/performance-tests-cmdstan/issues/30

WardBrian:
It may be arbitrary, but to me I think anything which is testing that the model compiles (e.g., what you're proposing in https://github.com/stan-dev/performance-tests-cmdstan/issues/30) can live here in stanc3, but anything which we want to actually run should be in performance-tests-cmdstan. Sound reasonable?

rok-cesnovar:
Yeah, good call.

WardBrian:
We already have https://github.com/stan-dev/stanc3/tree/master/test/integration/downstream, which is a good place to potentially put more of those models we want to make sure compile. I don't believe we're currently compiling that folder all the way down to a binary

WardBrian (approved):

PR #1195, opened 2022-05-10
Vectorize abs, deprecate fabs

### LINKED ISSUE #475: reverse deprecation on abs() and fabs()

WardBrian:
This closes #475. It won't be ready to merge until @bob-carpenter figures out some things on the C++ side (see https://github.com/stan-dev/math/issues/2729), but I'm opening it now so we can run tests against it.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/531

## Release notes

`abs` is now vectorized for container types. `fabs` is now deprecated, as all usages can be replaced with `abs` thanks to array promotion rules.

## Copyright and Licensing


WardBrian:
I'm now running this with https://github.com/stan-dev/math/pull/2734

rok-cesnovar (approved):

Issue #839, opened 2021-03-05
./setup_dev_env.sh: line 15: t: command not found 


LutzWeischerFujitsu:
[jw@cn05 scripts]$ sh ./setup_dev_env.sh
+ sh /dev/fd/63
++ curl -sL https://raw.githubusercontent.com/ocaml/opam/master/shell/install.sh
## Where should it be installed ? [/usr/local/bin]
Write access to /usr/local/bin required, using 'sudo'.
Command: mv /usr/local/bin/opam /usr/local/bin/opam.2.0.8
## /usr/local/bin/opam backed up as opam.2.0.8
Write access to /usr/local/bin required, using 'sudo'.
Command: install -m 755 /tmp/opam-2.0.8-arm64-linux /usr/local/bin/opam
## Run this script again with '--restore 2.0.8' to revert.
+ opam init
[NOTE] Will configure from built-in defaults.
Checking for available remotes: rsync and local, git, mercurial, darcs. Perfect!

<><> Fetching repository information ><><><><><><><><><><><><><><><><><><><><><>
[default] Initialised

<><> Required setup - please read <><><><><><><><><><><><><><><><><><><><><><><>

  In normal operation, opam only alters files within ~/.opam.

  However, to best integrate with your system, some environment variables
  should be set. If you allow it to, this initialisation step will update
  your bash configuration by adding the following line to ~/.bash_profile:

    test -r /home/jw/.opam/opam-init/init.sh && . /home/jw/.opam/opam-init/init.sh > /dev/null 2> /dev/null || true

  Otherwise, every time you want to access your opam installation, you will
  need to run:

    eval $(opam env)

  You can always re-run this setup with 'opam init' later.

Do you want opam to modify ~/.bash_profile? [N/y/f]
(default is 'no', use 'f' to choose a different file)
A hook can be added to opam's init scripts to ensure that the shell remains in sync with the opam environment when they
are loaded. Set that up? [y/N] n

<><> Creating initial switch (ocaml-base-compiler) ><><><><><><><><><><><><><><>

<><> Gathering sources ><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[ocaml-base-compiler.4.12.0] downloaded from cache at https://opam.ocaml.org/cache

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
-> installed base-bigarray.base
-> installed base-threads.base
-> installed base-unix.base
-> installed ocaml-options-vanilla.1
[...]

PR #1200, opened 2022-05-17
Scripts cleanup

### LINKED ISSUE #797: Reorganize scripts/
I think we can probably merge a few of the scripts and slim things down a bit. I think a fine flow might be:

1. Install opam - this isn't idempotent and will just reinstall over whatever you've got. It also has interactive prompts that need answering. This is different for windows vs. linux
2. Install 4.07.1 OCaml compiler switch (errors if already installed), pin build, js, and dev deps (idempotent). this is the same on Linux and Windows.
3. (optional) a script with Windows cross-compile dependencies
4. (optional) .bat script for Windows WSL builds (do we use this?)

This will essentially throw everything most developers care about into the 2nd script and let devs re-run that in case we've messed up our environment (:D) at the expense of making some of the Jenkins builds for those steps slightly longer. I think it will be pretty minimal since these will all be built in parallel and there aren't that many additional dev or JS dependencies. I think at the moment it's a bit tricky to figure out what you need to run to get up to speed.

Any thoughts on this? I'm happy to take a crack at it if people think this is a good idea. We can see how much longer Jenkins builds take on the PR. If it adds <1m I think I'm happy.

WardBrian:
Closes #797 #839. I've tested this in a fresh environment, it correctly handles commands which are not idempotent and also asks the user about optional things like the windows cross compiler packages and the javascript compiler

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Made installing the developer tools for the Stanc repository an interactive script with error handling. 

## Copyright and Licensing


rok-cesnovar (approved):
🧹 🧹

PR #1203, opened 2022-05-25
Clean up AST by specifying vardecl as sized

WardBrian:
We currently have the type of VarDecl set up so that it could hold an unsized type. This is never done in the AST, and leads to a lot of unecessary hand-wringing, like this unreachable code in the typechecker:

```ocaml
  | VarDecl {decl_type= Unsized _; _} ->
      (* currently unallowed by parser *)
      Common.FatalError.fatal_error_msg
        [%message "Don't support unsized declarations yet."]
```

We cannot do the same in the MIR, since certain optimizations actually can produce unsized declarations, namely function inlining 
#1097 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Internally clean up how variable declarations are represented.

## Copyright and Licensing

### DIFF
--- a/src/frontend/Ast.ml
+++ b/src/frontend/Ast.ml
@@ -156,13 +156,13 @@ type ('e, 's, 'l, 'f) statement =
   | Profile of string * 's list
   | Block of 's list
   | VarDecl of
-      { decl_type: 'e Middle.Type.t
+      { decl_type: 'e SizedType.t
       ; transformation: 'e Transformation.t
       ; identifier: identifier
       ; initial_value: 'e option
       ; is_global: bool }
   | FunDef of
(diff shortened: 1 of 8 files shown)


nhuurre:
This cleans up the code nicely but I thought the plan has been to add unsized local variables to the language? Have you discussed this with Bob?

WardBrian:
@bob-carpenter and I both seem to think now that something like `auto` or `var` from Java would be better than unsized declarations, for what it is worth.

If we do decide to add unsized declarations then reverting this change is just as easy as making it, but leaving this code there "just in case" seems unnecessary to me

nhuurre (approved):
Alright.

PR #1204, opened 2022-05-26
Remove unused MIR specializations

WardBrian:
We've had two MIR specializations (`Labelled` and `NoMeta`) which have been there since the refactor in #346 but are never instantiated. Removing them requires no additional code changes to the compiler, and it is unclear what they were intended for.

I believe removing these also clears up any circular dependency issues which would arise if we wanted to move things like `SizedType` and `UnsizedType` from `Middle` to `Common` in the future.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Simplified internal representations in the compiler.

## Copyright and Licensing


nhuurre (approved):

rybern:
FYI, IIRC Labeled was meant as a possible replacement for the MIR data
structure in analysis/optimization, but the switch was never made

WardBrian:
That would explain why the `Non_recursive` specialization is labeled as "a temporary replacement". Unless someone is dedicated to finishing that change, I don't think it will end up happening (much like how the AST never got the same treatment as the MIR).

Perhaps it is just different sensibilities, but my instinct is to go directly in the opposite direction and define `untyped_ast` and `typed_ast` patterns which are distinct from each other. The current structure means we have a fair number of illegal states which the type system is unaware of, such as `Promotion` AST nodes being in the untyped AST, which logically can never happen.

PR #1208, opened 2022-06-03
Vectorize to_complex

WardBrian:
This also refactors some of the binary function helpers we have to re-use code between them. 

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/545

## Release notes

`to_complex` is now vectorized. This means signatures like `to_complex(matrix, matrix) => complex_matrix` are now available. 

## Copyright and Licensing


src/middle/Stan_math_signatures.ml
```diff
@@ -565,25 +573,25 @@ let add_binary name supports_soa =
...
       List.iter
         ~f:(fun j ->
-          add_unqualified
-            (name, ReturnType (ints_to_real i), [i; j], supports_soa) )
-        [UnsizedType.UInt; UReal] )
-    [UnsizedType.UInt; UReal] ;
+          add_unqualified (name, ReturnType (return_fn i), [i; j], supports_soa)
+          )
+        scalars )
+    scalars ;
```
nhuurre, line 584: This is optional, and kind of conflicts with the general style in the file but am I the only who would rather see
```ocaml
  List.iter (List.cartesian_product scalars scalars) ~f:(fun (i, j) ->
      add_unqualified (name, ReturnType (return_fn i), [i; j], supports_soa) ) ;
```

WardBrian: I don't have a strong preference either way in terms of style. I think that `List.cartesian_product` would use more memory, but I doubt it is too much more?

I don't think we have an equivalent of a 3-dimensional cartesian product, so for me I would prefer to use the nested iters everywhere rather than have a mix of things (further down in this function we have a triple-nested version to do sizes of arrays)

nhuurre: It probably does allocate a temporary lists (or maybe OCaml can optimize it out? I dunno) but `add_unqualified` going to store every item anyway.

Three-list product is not a problem if you're willing to do
```ocaml
let ( * ) = List.cartesian_product in
List.iter (list1 * list2 * list3) ~f:(fun ((x,y),z) ->
```
although maybe that's a bit too convoluted.

What throws me off most about these iters is that the list is placed after the function so it's
```ocaml
List.iter
  ~f:(fun i ->
    (* loop body here *)
  )
  (List.range 0 8)
```
instead of
```ocaml
List.iter (List.range 0 8) ~f:(fun i ->
    (* loop body here *)
  )
```
The named arguments like `~f` don't care about order.

WardBrian: I guess the nice thing about having them at the end is that the lists sort of 'stack up' down there, so it's relatively clear what you're iterating over. If you move them to the start but don't do a cartesian product call, they end up not being particularly near each other

nhuurre: Really? I thought having them at the top makes it clearer. Most languages have
```
for (x in list) {
  do stuff
}
```
instead of
```
for {
  do stuff
} where (x in list);
```

WardBrian: I think the specific syntax here is why I prefer them at the end - they end up being all grouped very locally, seperated only by the next closing `)`

```ocaml
List.iter ~f:( fun x -> 
  List.iter ~f:( fun y ->
    List.iter ~f:( fun z ->
(* some function of x,y,z *)
    ) zlist
  ) ylist
) xlist
```

The alternative mixes the `iter` calls, `fun` keywords, and lists all together

```ocaml
List.iter xlist ~f:( fun x -> 
  List.iter ylist ~f:( fun y ->
    List.iter zlist ~f:( fun z ->
(* some function of x,y,z *)
    ) 
  ) 
) 
```

This doesn't really matter too much IMO, and I can see the benefits of either. For this PR I just preserved what was already there

nhuurre (approved):

PR #1217, opened 2022-06-20
Add a proper suffix to auxiliary variables

### LINKED ISSUE #1216: "context" not listed as a reserved name, but should probably be
#### Summary:
Naming a variable "context" in the data block leads to obscure errors when compiling, so "context" should probably listed as a reserved name.

#### Description:
Naming a variable "context" in the data block leads to obscure errors when compiling. Naming a variable "context" in the parameters block doesn't cause any issue however. However, "context" is not listed as a reserved name in the manual (see [here](https://mc-stan.org/docs/2_29/reference-manual/variables.html)), and the errors returned by rstan and cmdstanr are not particularly helpful.

#### Reproducible Steps:

Model doesn't compile:
```
data {
  int<lower=0> N;
  vector[N] context;
}

parameters {
  real mu;
  real<lower=0> sigma;
}

model {
  context ~ normal(mu,sigma);
}
```

Model compiles without any problem:
```
data {
  int<lower=0> N;
  vector[N] y;
}

parameters {
  real context;
  real<lower=0> sigma;
}

model {
  y ~ normal(context,sigma);
}
```

#### Current Output:
I tried both rstan (v2.29.0.900) and cmdstanr (with cmdstan v2.29.2), both return different errors, neither of which is particularly helpful (although cmdstan at least mentioned `context__` and `var_context`, which helped me figure out what was wrong).

nhuurre:
#### Submission Checklist

- [X] Run unit tests
- Documentation

Fix #1216 

## Copyright and Licensing
Copyright holder: Niko Huurre


WardBrian:
Ah, I assumed the bug was because of the class named context, not that we were renaming something to context__ too. 

Thanks for the fix, do you mind adding a model which previously would have failed to compile as a test?

WardBrian (approved):
Thanks!

Issue #823, opened 2021-02-22
Chained comparison


nhuurre:
What does this Stan code do?
```stan
if (x < y < z) {
```
The natural (naïve?) reading is that the condition asks if x < y and y < z and that is indeed how Python or Julia would interpret it. (R considers it a syntax error.)
But Stan follows C/C++ convention where it reads `(x < y) < z` which is either `1 < z` or `0 < z` depending on whether or not x is less than y.

I think this should at least raise warning but preferably Stan could just implement comparison operators the same way Python and Julia do.

PR #1220, opened 2022-07-07
some deprecations

### LINKED ISSUE #494: Deprecate implicit double-to-bool conversion
Consider the program fragment
```stan
real x = 1.0;
if (x) {
  target += (x) ? 1.0 : 2.0;
}
```
Third line is an error because ternary operator expects an integer condition but `x` is type real.
The second line, however, compiles and is equivalent to 
```stan
if (x != 0.0) {
```
The inconsistency is annoying but more importantly floating-point equality comparisons are fragile and should be avoided. Let's at least force the users to make them explicit.

nhuurre:
#### Submission Checklist

- [X] Run unit tests
- Documentation
    - [X] If a user-facing facing change was made, the documentation PR is here: stan-dev/docs#563

Closes #494 and touches on #823 . Those issues have not received any feedback, not sure if it's because everyone agrees or because no one reads the issue tracker.

## Release notes

* Deprecate using real scalar as a boolean
* Warn about ambiguous comparison operator chains

## Copyright and Licensing

Copyright holder: Niko Huurre

### DIFF
--- a/src/frontend/Canonicalize.ml
+++ b/src/frontend/Canonicalize.ml
@@ -85,7 +85,27 @@ let rec replace_deprecated_expr
           , replace_boolean_real ~parens:true deprecated_userdefined e2 )
     | BinOp (({expr= BinOp (_, op1, _); _} as e1), op2, e2)
       when Middle.Operator.(is_cmp op1 && is_cmp op2) ->
-        BinOp ({e1 with expr= Paren e1}, op2, e2)
+        BinOp
+          ( { e1 with
+              expr= Paren (replace_deprecated_expr deprecated_userdefined e1) }
+          , op2
+          , replace_deprecated_expr deprecated_userdefined e2 )
+    | BinOp
+        ( ({emeta= {type_= UInt; _}; _} as e1)
+        , Divide
+        , ({emeta= {type_= UInt; _}; _} as e2) ) ->
+        BinOp
+          ( replace_deprecated_expr deprecated_userdefined e1
+          , IntDivide
+          , replace_deprecated_expr deprecated_userdefined e2 )
+    | BinOp
+        ( ({emeta= {type_= UArray UMatrix | UMatrix; _}; _} as e1)
+        , Pow
+        , ({emeta= {type_= UInt | UReal; _}; _} as e2) ) ->
+        BinOp
+          ( replace_deprecated_expr deprecated_userdefined e1
+          , EltPow
+          , replace_deprecated_expr deprecated_userdefined e2 )
     | _ ->
         map_expression
           (replace_deprecated_expr deprecated_userdefined)
--- a/src/frontend/Canonicalize.ml
+++ b/src/frontend/Canonicalize.ml
@@ -75,11 +75,41 @@ let rec replace_deprecated_expr
               , {name; id_loc}
               , List.map ~f:(replace_deprecated_expr deprecated_userdefined) e
               ) )
+    | PrefixOp (PNot, e) ->
+        PrefixOp
+          (PNot, replace_boolean_real ~parens:true deprecated_userdefined e)
+    | BinOp (e1, ((And | Or) as op), e2) ->
+        BinOp
+          ( replace_boolean_real ~parens:true deprecated_userdefined e1
+          , op
+          , replace_boolean_real ~parens:true deprecated_userdefined e2 )
+    | BinOp (({expr= BinOp (_, op1, _); _} as e1), op2, e2)
+      when Middle.Operator.(is_cmp op1 && is_cmp op2) ->
+        BinOp ({e1 with expr= Paren e1}, op2, e2)
     | _ ->
         map_expression
           (replace_deprecated_expr deprecated_userdefined)
           ident expr in
   {expr; emeta}
+
+and replace_boolean_real ?(parens = false) deprecated_userdefined e =
+  match e with
+  | {emeta= {type_= UReal; _}; _} when parens ->
+      { emeta= {e.emeta with type_= UInt}
+      ; expr=
+          Paren (replace_boolean_real ~parens:false deprecated_userdefined e) }
+  | {emeta= {type_= UReal; _}; _} ->
+      { emeta= {e.emeta with type_= UInt}
+      ; expr=
+          BinOp
+            ( replace_deprecated_expr deprecated_userdefined e
+            , NEquals
+            , { expr= RealNumeral "0.0"
+              ; emeta=
+                  { type_= UInt
+                  ; loc= Middle.Location_span.empty
+                  ; ad_level= DataOnly } } ) }
+  | _ -> replace_deprecated_expr deprecated_userdefined e
 
 let replace_deprecated_lval deprecated_userdefined {lval; lmeta} =
   let is_multiindex = function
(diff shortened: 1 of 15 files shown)


src/frontend/Deprecation_analysis.ml
```diff
@@ -137,6 +137,40 @@ let rec collect_deprecated_expr (acc : (Location_span.t * string) list)
...
+  | BinOp ({expr= BinOp (e1, op1, e2); emeta= {loc; _}}, op2, e3)
+    when Operator.(is_cmp op1 && is_cmp op2) ->
+      let pp = Operator.pp in
+      let acc =
+        acc
+        @ [ ( loc
+            , Fmt.str
+                "Found x %a y %a z. This is interpreted as (x %a y) %a z but \
+                 that will change in Stan 2.34. Consider if the intended \
+                 meaning was x %a y && y %a z"
```
WardBrian, line 168: I don't think the current deprecation policy would allow for this to be changed before 3.0, since it changes the meaning of the same program before and after the version, rather than just being a syntax update.

src/frontend/Deprecation_analysis.ml
```diff
@@ -137,6 +137,40 @@ let rec collect_deprecated_expr (acc : (Location_span.t * string) list)
                 ) ]
           | _ -> [] ) in
       acc @ w @ List.concat_map l ~f:(fun e -> collect_deprecated_expr [] e)
+  | PrefixOp (PNot, ({emeta= {type_= UReal; loc; _}; _} as e)) ->
+      let acc =
+        acc
+        @ [ ( loc
+            , "Using a real as a boolean value is deprecated and will be \
```
WardBrian, line 144: I feel like it would be good to potentially note that exact comparison of a float to zero is also a bad idea, and they should consider `abs(x) < epsilon` or something else. I'm not sure of the exact wording, but it feels weird to me to deprecate this but recommend something with the same poor behavior

test/integration/cli-args/canonicalize/deprecated.stan
```diff
@@ -57,6 +57,11 @@ model {
     x ~ exponential_log(1);
     increment_log_prob(foo_cdf_log(x, 1));
   }
+  while (0.0) {
+    
+  }
+
+  int bool = !b < 2 && d || x;
```
WardBrian, line 64: Could you add a simple ` 1 < 2 < 3 < 4` type statement here?

src/frontend/Deprecation_analysis.ml
```diff
@@ -137,6 +137,40 @@ let rec collect_deprecated_expr (acc : (Location_span.t * string) list)
...
+               disallowed in Stan 2.34. Use an explicit != 0 comparison \
+               instead." ) ] in
```
WardBrian, line 146: We should also note that the canonicalizer can rewrite this for you, like we do for the other warnings

test/integration/good/warning/unreachable_statement.stan
```diff
@@ -1,6 +1,6 @@
 functions {
    void foo(real x){
-     if(x) ; // bug - misplaced ; makes the next statement unconditional
+     if(x == 0.0) ; // bug - misplaced ; makes the next statement unconditional
```
WardBrian, line 3: I agree it shouldn't be this test necessarily, but we should add something to the deprecated_syntax test in this same `warning` folder which triggers all the new warnings.

src/frontend/Deprecation_analysis.ml
nhuurre: Yes, I wasn't sure if assigning a new meaning would be a good idea. The wording "but that will change in [the future]" is deliberately vague about what the new interpretation will be or if there even is one. In my opinion the current behavior is surprising enough that just making it a syntax error is worth it.

src/frontend/Deprecation_analysis.ml
nhuurre: We don't say that for binary op warnings (int-division and matrix-power) which this was modeled after but I'm fine with adding such a comment here.

WardBrian: Those don’t say so because it (I believe) the canonicalizer doesn’t do anything for them, correct?

src/frontend/Deprecation_analysis.ml
WardBrian: You could either say change in 3.0 or become a syntax error on 2.34, but I think we can’t have a program that compiles in 2.21 compile in 2.34 but mean something different

src/frontend/Deprecation_analysis.ml
nhuurre: Yes, something like that definitely makes sense. I'm not sure about exact wording either and it probably should trigger on all uses of `==` and `!=`, not just the implicit ones.
But I think comparison with _zero_ in particular is more justified, at least for DataOnly, because it might be used as a marker value indicating missing data.

src/frontend/Deprecation_analysis.ml
nhuurre: Oh no, you're right. Canonicalizer definitely should change int division to `%/%`, no?

src/frontend/Canonicalize.ml
```diff
@@ -75,12 +75,62 @@ let rec replace_deprecated_expr
...
+    | BinOp (e1, ((And | Or) as op), e2) ->
+        BinOp
+          ( replace_boolean_real ~parens:true deprecated_userdefined e1
+          , op
+          , replace_boolean_real ~parens:true deprecated_userdefined e2 )
+    | BinOp (({expr= BinOp (_, op1, _); _} as e1), op2, e2)
+      when Middle.Operator.(is_cmp op1 && is_cmp op2) ->
+        BinOp
+          ( { e1 with
+              expr= Paren (replace_deprecated_expr deprecated_userdefined e1) }
```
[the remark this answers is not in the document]
nhuurre: The code adds parens only when there's nothing between the binops so it's not going to add it more than once.
The new parens are sandwiched between binops and parens canonicalization never changes those (it doesn't know the ops' precedence rules).

src/frontend/Deprecation_analysis.ml
```diff
@@ -137,6 +137,52 @@ let rec collect_deprecated_expr (acc : (Location_span.t * string) list)
...
+      let pp_e = Pretty_printing.pp_typed_expression in
+      let pp = Operator.pp in
+      let acc =
+        acc
+        @ [ ( loc
+            , Fmt.str
+                "Found %a. This is interpreted as %a. Consider if the intended \
+                 meaning was %a but if it wasn't Stan 2.34 will require \
+                 explicit parenthesis. This can be automatically changed using \
+                 the canonicalize flag for stanc"
```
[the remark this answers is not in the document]
WardBrian: > leave it a perpetual warning but never remove the syntax.

I think for now we should plan on doing this. We can always turn it into a removal warning later, but for now I personally wouldn't want it removed

I sort of see it as the typechecker warnings are useful for things like matrix-pow since they're not really deprecations, but rather type-specific warnings about unexpected or error-prone behavior. We could move them all elsewhere (which would also simplify the return type of the typechecker), but we'd want to either put them somewhere besides the module called `Deprecation_analysis` or rename that module to something more generic.

src/frontend/Canonicalize.ml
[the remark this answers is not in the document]
WardBrian: Should this canonicalization actually be done as _part_ of the parenthesis regularization, if we aren't labeling it explicitly as deprecated?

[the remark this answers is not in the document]
nhuurre: How do you feel about canonicalizing `%/%`? Is it close enough to deprecated or should we make a new canonicalize flag?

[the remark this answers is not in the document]
WardBrian: Yeah, `%/%` and `.^` don't really fit in any of the existing categories. Deprecations also feels like the wrong place to put them, so I'd be happy with a new category. That would also require a (small) docs PR

As always, the hardest part is naming that category, maybe "operators", or more generally "confusing-syntax" or something?

WardBrian:
I think this is good to go. I was going to ask why you weren't warning inside of the TernaryIf expression, but it turns out that already requires ints for the conditionals. The more you know.

Would you mind opening a docs PR to add the deprecation of reals as bools to [this page](https://mc-stan.org/docs/reference-manual/deprecated-features.html)? We might also want to update the [warnings guide](https://mc-stan.org/docs/stan-users-guide/understanding-stanc3-errors-and-warnings.html) but that is optional, IMO

WardBrian (approved):

Issue #1224, opened 2022-07-15
C++ compilation error when using slicing in recursive function call 


spinkney:
The following recursive function fails to compile. I believe this happens because if D - 1 == 1 then the index is 1:1 and the container becomes a real. Wrapping `gamma[1:D - 1]` in `to_vector()` allows the model to compile. 

```stan
functions {
vector test2(vector gamma);
 vector test2(vector gamma) {
   int D = num_elements(gamma);

   if (D == 1)
      return rep_vector(D, 0);
    else
      return test2(gamma[1:D - 1]);
  }

}
data {
  int N;
  int times;
}
parameters {
  vector[times] gamma;
}
transformed parameters {
  vector[N] z_hat = test2(gamma);
}
```
@WardBrian @rok-cesnovar

WardBrian:
The error specifically mentions 

```
stan/lib/stan_math/lib/eigen_3.3.9/Eigen/src/Core/MatrixBase.h:48:34: fatal error: template instantiation depth exceeds maximum of 900 (use ‘-ftemplate-depth=’ to increase the maximum)
   48 | template<typename Derived> class MatrixBase
```

and a ton of nested Eigen::Block functions. Wrapping this in to_vector seems to stop the recursive template instantiation.

I'm not sure what we can do to resolve this to be honest

andrjohns:
A quick fix could be to add `-ftemplate-depth=1500` (or whichever limit) to the Math library default compiler flags

WardBrian:
I think the problem is the code generated by this will genuinely recurse infinitely. Each new template expansion adds another wrapping Eigen::Block<…> which then leads to a new template expansion

andrjohns:
Ahhh of course. Would it be tricky on the Stanc3 side of things to always eval recursive functions? Or having a `plain_type_t<>` return type maybe?

WardBrian:
I think it would be tricky, or at least quiet messy. By the time we're doing code generation we don't want to be trying to do function resolution to figure out if a call is recursive (as opposed to an overloaded call, or just a normal function call). 

If there was something we could do in terms of the return types or declarations to resolve this it would be great.

For some reason this doesn't have the same problem when the code is `test2(gamma + gamma)`, even though I would expect that to create an infinite chain of `CWiseUnaryOps`. Any idea what the difference is @andrjohns? Could we try to update the templating of `rvalue` to avoid this?

PR #1228, opened 2022-07-25
Call eval on UDF arguments which are indices in UDF bodies

WardBrian:
Closes https://github.com/stan-dev/stanc3/issues/1224

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fix an issue where recursive functions which called themselves with a slice of their arguments would lead to infinite template expansion in C++ compilation.

## Copyright and Licensing


WardBrian:
I went with the safe but possibly over-eager choice to evaluate whenever a function call in a function body has the same name as the function itself. As noted in the test and in the discussion in #1224, this is not necessarily a recursive call in the context of overloading. 

So, this may produce a few extra calls to eval than it needs to.

nhuurre:
> So, this may produce a few extra calls to eval than it needs to.

Unfortunately this misses mutually recursive functions.
```stan
functions {
vector test3(vector gamma);
vector test2(vector gamma);
 vector test2(vector gamma) {
   int D = num_elements(gamma);

   if (D == 1)
      return rep_vector(D, 0);
    else
      return test3(gamma[1:D - 1]);
  }
vector test3(vector gamma) {
  return test2(gamma);
}
}
data {
  int N;
  int times;
}
parameters {
  vector[times] gamma;
}
transformed parameters {
  vector[N] z_hat = test2(gamma);
}
```

WardBrian:
Yes, it does. I don't think there's any good way around that short of re-implementing most of the C++ compiler's template resolution pass? Or we could eval all indexed arguments to UDF calls within UDFs.

Even if we don't fix the mutually recursive case, I think this change is still an improvement relative to the current behavior

WardBrian:
Updated to 

1) handle void functions

2) Call `eval` on every indexed argument to a UDF when inside a UDF, to handle possible mutual recursion. My guess is that this is probably not a super common edge case so I don't mind going for the safe option here.

nhuurre:
Fixing it fully is indeed difficult and even a partial solution is an improvement, yes. `eval()`ing _every_ UDF call at least avoids the terrible C++ errors. How much does `eval()` impact performance?

I think the ideal solution would be to do a topological sort and build a DAG whose vertices are functions and edges are calls, and if there are cycles then break them by inserting `eval()`s.

WardBrian:
I think the expense of eval will vary depending on the usage. In the simplest of cases, say where the function only cares about the length of the vector but not the values, it will be doing an extra copy each time as a result of the `eval`. But (I believe, anyway) if the function is doing essentially anything nontrivial it will eventually evaluate the argument anyway

@rok-cesnovar or @SteveBronder may be able to say more about this

bob-carpenter:
> But (I believe, anyway) if the function is doing essentially anything nontrivial it will eventually evaluate the argument anyway

Eigen will only implicitly evaluate an argument if it decides it's more efficient that way.  Typical uses that only access values once, like a dot product, won't force an eval.

WardBrian:
In the short term I think the extra evals are a worthwhile tradeoff for safety. I expect this is probably not super commonly written code by users where UDFs are calling other UDFs with slices of eigen types. 

We can later try a more sophisticated analysis as to when they are necessary. I think the proposal @nhuurre makes for that is the correct approach, but implementing that would be nontrivial.

WardBrian:
I think my latest commit will be the happy-middle of the current options. I've restricted the extra `eval`s to only be applied to functions who have forward declarations. This will catch all functions which could end up being recursive or mutually recursive, but won't affect the vast majority of UDFs.

WardBrian:
@nhuurre thoughts on the current state of this?

I discussed with @rok-cesnovar and we are going to do a minor release with just the recent stanc bug fixes (hopefully including this), Cherry-picking around the to_int and new warnings

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -30,6 +30,24 @@ let rec change_kwrds_stmts s =
     | x -> map Fn.id change_kwrds_stmts x in
   {s with pattern}
 
+let eval_eigen_indexed e =
+  let open Expr.Fixed in
+  let f ({pattern; meta} as expr) =
+    match (pattern, UnsizedType.is_eigen_type (Expr.Typed.type_of expr)) with
+    | Indexed _, true ->
+        {meta; pattern= FunApp (StanLib ("eval", FnPlain, AoS), [expr])}
+    | _ -> expr in
```
nhuurre, line 39: `Indexed` is not enough. Some math lib functions also fail.
```stan
 vector test2(vector gamma) {
   int D = num_elements(gamma);

   if (D == 1)
      return rep_vector(D, 0);
    else
      return test2(head(gamma,D - 1));
  }
```

WardBrian: It's easy enough to also wrap arbitrary math functions, but figuring out an exhaustive list of which ones will trigger this behavior seems daunting. It's probably reasonable to start with the functions listed [here](https://mc-stan.org/docs/functions-reference/slicing-and-blocking-functions.html)? If we get reports or find more in the future it is easy enough to add them

nhuurre (approved):
LGTM

PR #1230, opened 2022-07-25
[Fix] Function inliner name collision with local scope

### LINKED ISSUE #1229: [Bug] Function inliner fails when a variable name is the same as one in the enclosing scope
MRE:

```stan
functions {
  vector foo(vector oR) {
    int ot = num_elements(oR);
    
    vector[ot] R = oR;
    
    return (R);
  }
}
data {
  int N;
}
transformed parameters {
  vector[N] R;
  vector[N] out;
  out = foo(R);
}

```

Renaming `R` in either the `transformed parameters` block or in the function resolves the issue. I believe the issue stems from [here](https://github.com/stan-dev/stanc3/blob/fa5fc06a7b667b47b04c8689c094bef52c256f9c/src/analysis_and_optimization/Optimize.ml#L289-L291)

Pinging @SteveBronder

_Originally posted by @WardBrian in https://github.com/stan-dev/stanc3/issues/1227#issuecomment-1194319053_

WardBrian:
Closes #1229 . @SteveBronder this is just a simple re-ordering of how we do the name substitution, and I believe this is the correct thing to do. It didn't change any other files but fixes this bug and makes the generated code look like it would if the variables had different names. 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fix an issue with the function inliner when a function local variable had the same name as a variable in the calling scope.

## Copyright and Licensing


WardBrian:
I am able to compile the model from EpiNow2 with this PR @hsbadr

hsbadr (commented):
LGTM. It does fix the compilation of `EpiNow2` when `stanc3` optimizations are enabled.

hsbadr (approved):

SteveBronder (approved):

WardBrian:
Thanks for trying these models out @hsbadr. Super useful to have rstan be able to try the newer features on all the existing models in R packages

PR #1231, opened 2022-07-25
Allow driver code to set printed argument list in generated hpp

WardBrian:
#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

The generated `model_compile_info` will now feature arguments  passed in from stancjs

## Copyright and Licensing

### DIFF
--- a/src/stan_math_backend/Stan_math_code_gen.ml
+++ b/src/stan_math_backend/Stan_math_code_gen.ml
@@ -26,17 +26,8 @@ open Function_gen
 
 let standalone_functions = ref false
 
-let stanc_args_to_print =
-  let sans_model_and_hpp_paths x =
-    not
-      String.(
-        is_suffix ~suffix:".stan" x
-        && not (is_prefix ~prefix:"--filename-in-msg" x)
-        || is_prefix ~prefix:"--o" x) in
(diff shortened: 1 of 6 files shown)


WardBrian:
I’d like to write a good test for this before merging, I eyeballed that it looks correct

hsbadr (commented):
@WardBrian This does print the flags in the compile info:
```r
  inline std::vector<std::string> model_compile_info() const noexcept {
    return std::vector<std::string>{"stanc_version = stanc3 v2.30.0-37-g09a0377", "stancflags = O1"};
  }
  ```
  
It might be useful to add them with the version comment too:
```c++
// Code generated by stanc v2.30.0-37-g09a0377
```

hsbadr (commented):
@WardBrian Would it make sense to unify how the flags are printed? For example, `stanc` binary gets `stancflags = --O1` while the JS generates `stancflags = O1`. Alternatively, you may mention that it's command line or JS in the model compile info.

WardBrian:
@rok-cesnovar should be ready now. I would also like to include this in the patch release

rok-cesnovar (approved):
Great, thanks!

Issue #354, opened 2019-10-23
Don't print cuts between empty blocks in pretty printer


seantalts:
https://github.com/stan-dev/stanc3/blob/master/src/middle/Program.ml#L117-L129

PR #1233, opened 2022-07-29
[Refactor] Add a lower-level IR for C++ code generation

### LINKED ISSUE #850: Collate pretty printers for C++ and structure the code gen files
Right now we have a mix-match of things inside of `Expression_gen.ml` and `Stan_math_code_gen.ml` that actually prints out and deduces information from the MIR about what those pretty printers should print out. I think it would be good to separate these things into two or four separate files. We could have something like 

- `Expression_gen_mir.ml`: Helpers for querying the MIR data structure for information
- Expression_gen_pp.ml`: All the pretty printers for general C++ expression gen
- `Stan_math_code_gen`: Just holds printers for printing out the model class.

Or something like that? Of course the printers can have queries about the MIR as well so I'm not sure if this is a nice structure. Though it might make sense to have the first two just so it's easier to find things.

I'd also like to change the names of the pretty printers for the c++ from `pp_*` to `cpp_*`. When I first started on the project I was confused because there are also `pp_*` functions for pretty printing out things like the cleaned up stan code and mir. It would be nice if the names were a bit more obvious. In the same vein we could have `tfpp_*` for printing tensorflow probability code.

WardBrian:
Closes #850, #354.

This PR introduces a relatively simple[^1] C++ representation for the code generation backend. 

Rather than code generating by creating strings directly, we first lower the MIR to this representation and then print it out.

Benefits:
1. Code generation is now a mostly-pure computation. With the exception of a few things like functor generation and tracking map-rect calls, there are no side effects in the code gen. Each function takes in some chunk of input and returns some portion of the C++ datatype. 
2. Separation of the code-generation logic and the pretty-printing. This means formatting is more consistent, and makes it impossible for some part of the code gen to miss a semicolon where other parts wouldn't. As an added plus, we don't need to pass printing callbacks around to get reasonable formatting of things like method bodies.
3. Ability to re-use code and write expressively. For example, we can write OCaml code like
    ```ocaml
    let open Expression_syntax in
    let open Types in
    let vector = Constructor (row_vector Double, [Literal "3"]) in
    let values = [Literal "1"; Var "a"; Literal "3"] in
    (vector << values).@!("finished")
   ```
   which gets turned into the C++
   ```C++
   (Eigen::Matrix<double,1,-1>(3) << 1, a, 3).finished()
   ```
   I tried not to go over board on making syntax for this, it is more or less limited to doing method calls (the alternative would be `(MethodCall (Parens (StreamInsertion (Constructor (row_vector Double, [Literal "3"]),  [Literal "1"; Var "a"; Literal "3"]), "finished", [], [])`, which moves all the least-important information to the front of the code).

There is a lot of potential to add further processing of this generated C++ before it is stringified and output, but that is not part of this PR.

The code generated by this PR is intended to be semantically identical; whitespace, the ordering of some functions, and other superficial changes are common however. I recommend viewing the diffs with whitespace changes hidden. I have eyeballed all of them, but this PR is the first time some of it will actually be compiled to executable.

[^1]: "Relatively simple" in that it 1) does not try to capture all of C++, just the portion of C++ we currently generate and 2) is a basic recursive data type, no two-level-type reasoning is needed at this level

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Refactored how C++ is code-generated in the compiler.

## Copyright and Licensing


src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,664 @@
...
+  let std_vector t = Vector t
+  let bool = Type_literal "bool"
+  let complex s = Complex s
+  let vector s = Matrix (s, -1, 1)
+  let row_vector s = Matrix (s, 1, -1)
+  let matrix s = Matrix (s, -1, -1)
+  let string = Type_literal "std::string"
+  let size_t = Type_literal "size_t"
+  let const_ref t = Const (Ref t)
+  let str_array i = Array (Const (Pointer (Type_literal "char")), i)
```
SteveBronder, line 39: String is a very different type in C++ from a `const char*`

```suggestion
  let const_char_array i = Array (Const (Pointer (Type_literal "char")), i)
```

WardBrian: Good point, I've done the rename you suggest.

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,732 @@
...
+
+  (** Helper for [std::numeric_limits<int>::min()] *)
+  let int_min = fun_call "std::numeric_limits<int>::min" []
+
+  (** Fold a list of expressions together with a C++ operator,
+      useful for e.g. adding a list of numbers. *)
+  let binop_list es ~f ~default : expr =
+    match es with
+    | [] -> default
+    | head :: rest -> List.fold ~init:head ~f:(fun a b -> Parens (f a b)) rest
```
SteveBronder, line 124: Should this be `binop_fold`?

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,732 @@
...
+  include Exprs
+
+  (** A pun for the C++ [operator<<] *)
+  let ( << ) a b = StreamInsertion (a, b)
+
+  (** Method call: Call a no-argument method
+
+      E.g. [foo.bar()]
+  *)
+  let ( .@!() ) obj name = method_call obj name [] []
```
SteveBronder, line 140: [optional] Would be nice to give result types here if possible

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,732 @@
...
+  { static: bool [@default false]
+  ; constexpr: bool [@default false]
+  ; type_: type_
+  ; name: identifier
+  ; init: init [@default Uninitialized] }
+[@@deriving make, sexp]
+
+type stmt =
+  | Expression of expr
+  | VarDef of var_defn
```
SteveBronder, line 201: I'd rather have this be `obj_decl`. My thinking is that this is just the info that's used for the declaration of an object

SteveBronder: var is kind of overused because we have variables and the literal `var` autodiff type

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,732 @@
...
+type fun_defn =
+  { templates_init: template_parameter list list * bool [@default [[]], false]
+        (** Double list since some functions (mainly reduce_sum functors) need two sets of templates *)
+  ; name: identifier
+  ; inline: bool [@default false]
+  ; return_type: type_
+  ; args: (type_ * string) list
+  ; cv_qualifiers: cv_qualifiers list [@default []]
+  ; body: stmt list option }
+[@@deriving make, sexp]
```
SteveBronder, line 311: Do you mind shifting these around to follow the standard syntax scheme of a C++ function?
```suggestion
type fun_defn =
  { templates_init: template_parameter list list * bool [@default [[]], false]
        (** Double list since some functions (mainly reduce_sum functors) need two sets of templates *)
  ; inline: bool [@default false]
  ; return_type: type_
  ; name: identifier
  ; args: (type_ * string) list
  ; cv_qualifiers: cv_qualifiers list [@default []]
  ; body: stmt list option }
[@@deriving make, sexp]
```

I would be nice to use the same verbage as the standard for each of these parts of the function decleration

https://en.cppreference.com/w/cpp/language/function

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,732 @@
...
+and struct_defn =
+  {param: template_parameter option; struct_name: identifier; body: defn list}
+
+and defn =
+  | FunDef of fun_defn
+  | Class of class_defn
+  | Struct of struct_defn
+  | TopVarDef of var_defn
+  | TopComment of string
+  | TopUsing of string * type_ option
```
SteveBronder, line 345: Does top always just mean "At the top of the generated file?" Should we just call them global?

SteveBronder (changes requested):
I only got about halfway through Cpp.ml today, but so far I like what I see! Mostly seeing a few places it would be nice to have more verbose names / names that more closely align with C++ verbage. But the overall scheme seems very nice so far

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,732 @@
...
+  | Complex of type_
+  | TemplateType of identifier
+  | Vector of type_
+  | Array of type_ * int
+  | Type_literal of identifier  (** Used for things like Eigen::Index *)
+  | Matrix of type_ * int * int
+  | Ref of type_
+  | Const of type_
+  | Pointer of type_
+  | TypeTrait of string * type_ list
```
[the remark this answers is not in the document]
WardBrian: I tried to stick pretty hard to the "You Aren't Gonna Need It" philosophy for the C++ representation, and as it turns out we actually never generate those explicit types. Everything is `local_scalar_t__` rather than a specific autodiff type. 

It would be really trivial to add later, but for now I only did the types we use (OCaml emits warnings if you have type variants which are never instantiated, too). 

Similar reasoning behind not having `long double` etc as types. Easy to add later, not needed now.

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,732 @@
...
+      (** e.g. stan::promote_scalar, stan:base_type *)
+[@@deriving sexp]
+
+module Types = struct
+  (** Helpers for constructing types *)
+
+  let local_scalar = Type_literal "local_scalar_t__"
+
+  (** A [std::vector<t>] *)
+  let std_vector t = Vector t
```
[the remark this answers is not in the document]
WardBrian: All Eigen types are represented by the `Matrix of type_ * int * int` variant, to keep it similar to how they are defined in C++. That's why this helper function exists a bit down from the function you're highlighting:

```ocaml
  let vector s = Matrix (s, -1, 1)
```

I can rename the std::vector variant to `StdVector` rather than `Vector` if you'd like. For now I'll just add a doc comment clarifying

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,732 @@
...
+  | MethodCall of expr * identifier * type_ list * expr list
+  | StaticMethodCall of type_ * identifier * type_ list * expr list
+  | Constructor of type_ * expr list  (** printed as [type(expr1, expr2, ...)]*)
+  | InitializerExpr of type_ * expr list
+      (** printed as [type{expr1, expr2, ...}]*)
+  | ArrayLiteral of expr list
+  | TernaryIf of expr * expr * expr
+  | Cast of type_ * expr
+  | Index of expr * expr
+  | New of string option * type_ * expr list
```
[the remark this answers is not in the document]
WardBrian: Renamed (I believe we only ever generate the later). Good distinction to make

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,732 @@
...
+  let std_vector_expr t es = InitializerExpr (Vector t, es)
+
+  let fun_call s es = FunCall (s, [], es)
+  let templated_fun_call s ts es = FunCall (s, ts, es)
+
+  (** Helper for [std::numeric_limits<double>::quiet_NaN()] *)
+  let quiet_NaN = fun_call "std::numeric_limits<double>::quiet_NaN" []
+
+  (** Helper for [std::numeric_limits<int>::min()] *)
+  let int_min = fun_call "std::numeric_limits<int>::min" []
```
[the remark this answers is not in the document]
WardBrian: Names which begin with a capital are reserved for constructors in OCaml so this unfortunately cannot be done

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,732 @@
...
+module Exprs = struct
+  (** Some helper values and functions *)
+
+  (** Call a method on object, wrapping it in parentheses if
+      it is not a variable
+  *)
+  let method_call obj name templates args =
+    match obj with
+    | Var _ -> MethodCall (obj, name, templates, args)
+    | _ -> MethodCall (Parens obj, name, templates, args)
```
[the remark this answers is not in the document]
WardBrian: This helper method is needed to make sure that the precendence you specify in OCaml is actually what gets printed.

Consider the code example:

```ocaml
let open Expression_syntax in 
let a = Var "a" in
let b = Var "b" in
(a + b).@!("eval")
```

If _not_ for the above logic, this would be a `MethodCall(BinOp(...), ...)`, but that would print as
`a + b.eval()`, which is clearly not the correct order. What we want is `(a + b).eval()`.

The reason we match on a Var is so avoid printing the () in redundant cases like `c.eval()` becoming `(c).eval()`

src/stan_math_backend/Cpp.ml
WardBrian: Yes, but even better, it's unused, so I can delete it.

src/stan_math_backend/Cpp.ml
WardBrian: The return type of everything in Expression_syntax is `Cpp.expr`, so I'm not sure if it needs to be specified?

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,732 @@
...
+  let to_var s = Var s
+
+  (** Turn an OCaml string into a quoted and escaped C++ string*)
+  let literal_string s = Literal ("\"" ^ Cpp_str.escaped s ^ "\"")
+
+  (** Equivalent to [std::vector<t>{e1,...,en}] *)
+  let std_vector_expr t es = InitializerExpr (Vector t, es)
+
+  let fun_call s es = FunCall (s, [], es)
+  let templated_fun_call s ts es = FunCall (s, ts, es)
```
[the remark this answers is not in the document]
WardBrian: Updated

src/stan_math_backend/Cpp.ml
WardBrian: It means "Top-level". I'm not married to the name Top, but I also think Global isn't the right one either (what is a "Global" comment, really?)

src/stan_math_backend/Cpp.ml
WardBrian: Do you mind `variable_defn`? It's nitpicky but there are plenty of types which aren't objects

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,732 @@
...
+  | Eq
+  | LEq
+  | GEq
+  | Lthn
+  | Gthn
+  | And
+  | Or
+[@@deriving sexp]
+
+type unary_op = PMinus | Incr [@@deriving sexp]
```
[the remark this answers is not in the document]
WardBrian: This seems counter to your above suggestion that the StreamInsertion operator be subsumed in to `Nary`, but I agree with this more of the two suggestions (again for "You Aren't Gonna Need It" reasons - I don't forsee us using other N-ary operators, and if we do, it's easy enough to do them later)

src/stan_math_backend/Cpp.ml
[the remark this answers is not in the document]
SteveBronder: Yeah imo I'd change it to `StdVector` so it doesn't get more confusing down the road

src/stan_math_backend/Cpp.ml
SteveBronder: Let me read over this a bit more to see how it's used and we can circle back

src/stan_math_backend/Cpp.ml
SteveBronder: I think a global comment is a thing, it's just a comment placed in global scope

WardBrian:
@nhuurre - I talked to @SteveBronder and he doesn't have the bandwidth to review this soon. Do you feel comfortable reviewing a PR in this area, and would you mind?

src/stan_math_backend/Lower_stmt.ml
```diff
@@ -0,0 +1,328 @@
...
+      [While (lower_bool_expr cond, Stmts.block (lower_statement body))]
+  | For
+      { body=
+          { pattern=
+              Assignment
+                (_, {pattern= FunApp (CompilerInternal (FnReadParam _), _); _})
+          ; _ } as body
+      ; _ } ->
+      lower_statement body
+      (* Skip For loop part, just emit body due to the way FnReadParam emits *)
```
nhuurre, line 288: This case is never reached in our tests. (Thank goodness, I don't want to think about that comment.)

src/stan_math_backend/Locations.ml
```diff
@@ -31,20 +31,27 @@ let prepare_prog (mir : Program.Typed.t) : Program.Numbered.t * state_t =
...
+         ~name:"profiles__" () )
+  ; GlobalVariableDefn
+      (make_variable_defn ~static:true ~constexpr:true ~type_:arr_type
+         ~name:"locations_array__"
+         ~init:(Assignment (ArrayLiteral location_list)) () ) ]
 
-let pp_smeta ppf location_num =
-  if location_num = no_span_num then ()
-  else Fmt.pf ppf "current_statement__ = %d;@;" location_num
+let create_loc_assignment location_num =
```
nhuurre, line 51: Wouldn't `gen_loc_assignment` be more in line with the other function names? Although I might prefer `assign_loc` or `set_loc`.

nhuurre (changes requested):
You say that this closes #354 but that issue is about MIR dump format, something this PR does not touch at all.

> With the exception of a few things like functor generation and tracking map-rect calls, there are no side effects in the code gen.

I didn't see functor generation side effects; it's forward declarations that have side effects but those are fully contained in `Lower_functions.ml` so I wouldn't worry too much.

More annoying is the `map_rect` tracking and registration.
Do you think it would be reasonable to separate `map_rect` id assignment from code lowering?
For example, `Locations.prepare_prog` walks the whole program and collects and relabels all the location metadata. Maybe it could also transform `map_rect` calls
```
map_rect(fn, args...) -> map_rect(id, fn, args...)
```
and return a list of the `id`s.

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,731 @@
...
+  | StdVector of type_
+      (** A std::vector. For Eigen Vectors, use [Matrix] with a row or column size of 1 *)
+  | Array of type_ * int
+  | Type_literal of identifier  (** Used for things like Eigen::Index *)
+  | Matrix of type_ * int * int
+  | Ref of type_
+  | Const of type_
+  | Pointer of type_
+  | TypeTrait of string * type_ list
+      (** e.g. stan::promote_scalar, stan:base_type *)
```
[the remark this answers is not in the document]
WardBrian: Agreed (and renamed `Type_literal`)

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,731 @@
...
+    | PMinus e -> pf ppf "-%a" pp_expr e
+    | Increment e -> pf ppf "++%a" pp_expr e
+    | BinOp (e1, op, e2) ->
+        pf ppf "%a@ %a@ %a" pp_expr e1 pp_operator op pp_expr e2
+
+  let pp_variable_defn ppf {static; constexpr; type_; name; init} =
+    let pp_init ppf init =
+      match init with
+      | Uninitialized -> ()
+      | Assignment e -> pf ppf "@ =@ %a" pp_expr e
```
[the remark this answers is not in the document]
WardBrian: `@, ` will add an extra space after the newline if a break is printed, but I agree we don't need a breaking space before the `=`

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,731 @@
...
+      ; base
+      ; private_members
+      ; constructor
+      ; destructor_body
+      ; public_members } =
+    pf ppf
+      "@[<v 2>class %s%s : %a{@,\
+       @[<v 2>private:@,\
+       %a@]@,\
+       @[<v 2>public:@,\
```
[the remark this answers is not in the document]
WardBrian: Huh, this was actually an intentional choice I made, but it turns out the Google Style Guide recommends 1 space.

src/stan_math_backend/Lower_expr.ml
```diff
@@ -0,0 +1,615 @@
...
+        [%message (x : UnsizedType.t) "not implemented yet"]
+
+let rec lower_logical_op op e1 e2 =
+  let prim e = Exprs.fun_call "stan::math::primitive_value" [lower_expr e] in
+  Parens (BinOp (prim e1, op, prim e2))
+
+and lower_binop op es =
+  Parens (BinOp (lower_expr (first es), op, lower_expr (second es)))
+
+and lower_binary_fun f es = Exprs.fun_call f (lower_exprs (List.take es 2))
```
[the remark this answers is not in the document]
WardBrian: Correct, not sure why I wrote it that way

src/stan_math_backend/Lower_stmt.ml
```diff
@@ -0,0 +1,328 @@
...
+      (* When the current block or this transformation doesn't require unconstraining,
+         use vanilla write *)
+      | None, _ | _, None -> out.@?(("write", [lower_expr var])) |> wrap_e
+      (* Otherwise, use stan::io::serializer's write_free functions *)
+      | Some trans, Some unconstrain_string ->
+          let unconstrain_args = transform_args trans in
+          let write_fn = "write_free_" ^ unconstrain_string in
+          out.@?((write_fn, lower_exprs (unconstrain_args @ [var]))) |> wrap_e )
+  | NRFunApp (CompilerInternal f, args) ->
+      let fname, extra_args = trans_math_fn f in
```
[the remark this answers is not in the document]
WardBrian: Huh, I was just maintaining this from the existing source, but you're right, it is never used. Glancing through the blame, at no point in time did that function return anything other than `[]` for the second value, so I'm guessing it was unnecessary future proofing

src/stan_math_backend/Lower_stmt.ml
WardBrian: I was too fearful of that comment to try touching it, but you're correct in that deleting it does not change any output. Phew.

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,731 @@
...
+        if String.contains s '\n' then pf ppf "/@[<v>*@[@ %a@]@,@]*/" text s
+        else pf ppf "//@[<h> %s@]" s
+    | TryCatch (Block trys, (exn_ty, exn_name), Block thn) ->
+        (* When we know this contains blocks, we can do better pretty-printing*)
+        pf ppf "@[<v 2>try {@ %a@]@,@[<v 2>} catch(%a %a) {@ %a@]@,}"
+          (list ~sep:cut pp_stmt) trys pp_type_ exn_ty pp_identifier exn_name
+          (list ~sep:cut pp_stmt) thn
+    | TryCatch (trys, (exn_ty, exn_name), thn) ->
+        pf ppf "@[<v>@[<hov>try@ %a@]@,@[<hov>catch(%a %a)@ %a@]@]" pp_stmt trys
+          pp_type_ exn_ty pp_identifier exn_name pp_stmt thn
```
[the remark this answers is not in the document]
WardBrian: Huh, I assumed `try`/`catch` followed the `if`/`else` rule where a single-statement body was permitted without the enclosing braces. I'll clean this up.

WardBrian:
> You say that this closes https://github.com/stan-dev/stanc3/issues/354 but that issue is about MIR dump format, something this PR does not touch at all.

My bad, I thought that was about the (nearly identical) issue we currently have in the C++ code gen for printing a lot of empty cuts.

> map_rect

Giving that suggestion a try now

WardBrian:
@nhuurre I quite like the result of moving map_rect into a pre-processing phase. The end result is that we shouldn't have any global state left in the code gen, it should all be within the scope of one function/module

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,724 @@
...
+    | Semicolon -> string ppf ";"
+    | VariableDefn vd -> pf ppf "%a;" pp_variable_defn vd
+    | For (init, cond, incr, Block stmts) ->
+        (* When we know a block is here, we can do better pretty-printing*)
+        pf ppf "@[<v 2>for(@[<hov>%a; %a; %a@]) {@ @[<v>%a@]@]@,}"
+          pp_variable_defn init pp_expr cond pp_expr incr
+          (list ~sep:cut pp_stmt) stmts
+    | For (init, cond, incr, s) ->
+        pf ppf "@[<v 2>for(@[<hov>%a; %a; %a@])@ @[%a@]@]" pp_variable_defn init
+          pp_expr cond pp_expr incr pp_stmt s
```
nhuurre, line 509: That better pretty-printing for blocks could be refactored into a helper function (not sure what to call it)
```ocaml
let pp_with_block pp ppf = function
  | Block stmts ->
      pf ppf "@[<v 2>%a {@ @[<v>%a@]@]@,}" pp () (list pp_stmt) stmts
  | stmt -> pf ppf "@[<v 2>%a@ @[<v>%a@]@]" pp () pp_stmt stmt
```
and then you can simplify this
```suggestion
    | For (init, cond, incr, s) ->
        let pp ppf () =
          pf ppf "for (@[<hov>%a; %a; %a@])" pp_variable_defn init pp_expr cond
            pp_expr incr in
        pp_with_block pp ppf s
```
(I added a space to `for(` as that looks a bit cleaner.)
The same helper can be used for `ForEach`, `While`, and `IfElse` (and those also need a space between the keyword and paren.)

src/stan_math_backend/Cpp.ml
```diff
@@ -0,0 +1,724 @@
...
+    | Using (s, init) ->
+        pf ppf "using %s%a;" s
+          (option (fun ppf defn -> pf ppf " = %a" pp_type_ defn))
+          init
+    | Comment s ->
+        if String.contains s '\n' then pf ppf "/@[<v>*@[@ %a@]@,@]*/" text s
+        else pf ppf "//@[<h> %s@]" s
+    | TryCatch (trys, (exn_ty, exn_name), thn) ->
+        (* When we know this contains blocks, we can do better pretty-printing*)
+        pf ppf "@[<v 2>try {@ %a@]@,@[<v 2>} catch(%a %a) {@ %a@]@,}"
```
nhuurre, line 535:
```suggestion
        pf ppf "@[<v 2>try {@ %a@]@,@[<v 2>} catch (%a %a) {@ %a@]@,}"
```

nhuurre (changes requested):

src/stan_math_backend/Lower_expr.ml
```diff
@@ -0,0 +1,609 @@
...
+      vector_literal ~column:true st es
+  | FunApp (CompilerInternal f, es) ->
+      lower_compiler_internal meta.adlevel meta.type_ f es
+  | FunApp (StanLib (f, suffix, mem_pattern), es) ->
+      let ret_type = Some (UnsizedType.ReturnType meta.type_) in
+      lower_fun_app suffix f es mem_pattern ret_type
+  | FunApp (UserDefined (f, suffix), es) -> lower_user_defined_fun f suffix es
+  | Indexed (e, idx) -> (
+    match e.pattern with
+    | FunApp (CompilerInternal (FnReadParam _), _) -> lower_expr e
```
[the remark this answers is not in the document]
WardBrian: Deleting it also does nothing, nice catch

src/stan_math_backend/Lower_program.ml
```diff
@@ -0,0 +1,871 @@
...
+  | _ -> None
+
+let lower_num_param (dims : Expr.Typed.t list) =
+  match dims with
+  | [] -> Literal "0"
+  | [d] -> lower_expr d
+  | d1 :: dims ->
+      Parens
+        (List.fold ~init:(lower_expr d1) ~f:Expression_syntax.( * )
+           (lower_exprs dims) )
```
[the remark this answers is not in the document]
WardBrian: Much cleaner!

src/stan_math_backend/Numbering.ml
```diff
@@ -0,0 +1,97 @@
...
+    Program.Numbered.t * state_t * map_rect_registration_t =
+  let label_to_location = Int.Table.create () in
+  let map_rect_calls = Int.Table.create () in
+  let location_to_label = Hashtbl.create (module Location_span) in
+  Hashtbl.set label_to_location ~key:no_span_num ~data:Location_span.empty ;
+  Hashtbl.set location_to_label ~key:Location_span.empty ~data:no_span_num ;
+  (* turn locations into numbers for array printing *)
+  let rec number_locations_stmt ({pattern; meta} : Stmt.Located.t) :
+      Stmt.Numbered.t =
+    let pattern = Stmt.Fixed.Pattern.map Fn.id number_locations_stmt pattern in
```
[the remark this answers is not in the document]
WardBrian: Great!

src/stan_math_backend/Cpp.ml
WardBrian: This took some tweaking to get right for `if/else` but looks really nice in the end!

nhuurre (approved):
A couple of nitpicky comments but those are optional so I'll approve now.

WardBrian:
Thanks for the thorough review @nhuurre!

@SteveBronder - I believe merging will still be blocked so long as you have 'requested changes'. Do you mind clearing that?

SteveBronder (approved):
lgtm!

Issue #1240, opened 2022-08-24
[FR] get_param_names should be able to only return object names in parameters block


SteveBronder:
### Current Behavior:

`get_param_names()` defined in the model returns all `Program.outvars`, but it should only return the stan programs parameters declared in the parameters block.

This is important for a bug I found recently in initialization where trying to set hard coded used [-2, 2] random sampling for initialization. At the line [here](https://github.com/stan-dev/stan/blob/develop/src/stan/services/util/initialize.hpp#L80) if the user has any values in generated quantities, unless the user knew about this bug and created values for the variables declared in generated quantities then that check fails and it tries over and over again with the user specified values.

To fix this I think we need a separate `Program` section holding the names of the parameters 

### Model Code

All models

WardBrian:
I was able to confirm that changing the signature in model base and changing the generated signature in stanc did not require any changes to the call sites, so I think we can do that if we want to

roualdes:
Thanks, y'all.  I do think this will help BridgeStan.  Cheers.

PR #1241, opened 2022-08-25
Add get_param_names overload with flag arguments

WardBrian:
This would be the first part of #1240. After this is merged, we can safely update Stan's model_base to require these arguments, then delete the original overload, _then_ fix the bug @SteveBronder noticed by setting the flags to false during initialization. 

While we are at it, would we want to make the same change to `get_dims`?

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Added a new version of the model method `get_param_names` which allows specifiying if the transformed parameters or generated quantities should be emitted. 

## Copyright and Licensing

### DIFF
--- a/src/stan_math_backend/Stan_math_code_gen.ml
+++ b/src/stan_math_backend/Stan_math_code_gen.ml
@@ -181,6 +181,7 @@ let pp_method ppf rt name params intro ?(outro = nop)
  *)
 let pp_get_param_names ppf {Program.output_vars; _} =
   let add_param = fmt "%S" in
+  (* old -- kept as overload for compat until model_base changes *)
   let extract_name var = Mangle.remove_prefix (fst var) in
   pp_method ppf "void" "get_param_names"
     ["std::vector<std::string>& names__"]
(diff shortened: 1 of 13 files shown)


WardBrian:
Discussed in the weekly Stan meeting, @bob-carpenter confirmed it was fine to add the new arguments with defaults in model_base for backwards compatibility purposes. 

After this is merged we can update the model_base signatures to be

```c++
  virtual void get_param_names(std::vector<std::string>& names,
                                bool include_tparams = true,
                                bool include_gqs = true) const = 0;
```

```c++
  virtual void get_dims(std::vector<std::vector<size_t> >& dimss,
                        bool include_tparams = true,
                        bool include_gqs = true) const = 0;
```

WardBrian:
@SteveBronder this is ready for review and should be relatively quick, if you have a moment

SteveBronder (approved):
lgtm!

src/stan_math_backend/Stan_math_code_gen.ml
```diff
@@ -176,11 +176,21 @@ let pp_method ppf rt name params intro ?(outro = nop)
...
+       %s.reserve(%s.size() + temp.size());@,\
+       %s.insert(%s.end(), temp.begin(), temp.end());@,"
+      type_ (list ~sep:comma pp_elt) elts name name name name
+
 (** Print the [get_param_names] method of the model class
   @param ppf A pretty printer.
  *)
 let pp_get_param_names ppf {Program.output_vars; _} =
   let add_param = fmt "%S" in
+  (* old -- kept as overload for compat until model_base changes *)
```
SteveBronder, line 193: del this and the other new comment tho'

WardBrian: After the change to model_base I plan on deleting the old code, and those comments

SteveBronder: Lets not delete the old code since we want to keep backwards compatibility (C++ sees these as two different virtual function signatures)

WardBrian: We cannot expose both 

`  virtual void get_param_names(std::vector<std::string>& names) const = 0;` 

and 

`  virtual void get_param_names(std::vector<std::string>& names,  const bool emit_transformed_parameters = true, const bool emit_generated_quantities = true) const = 0;`

since it makes calls with only the vector argument ambiguous. Because we're providing default values for the arguments which match previous behavior, we should just keep the second

PR #1256, opened 2022-10-11
Improve errors for function re-declaration or missing definition

WardBrian:
#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Improved the error generated when a function is re-declared, or when a function is not given a definition. 

## Copyright and Licensing

### DIFF
--- a/src/frontend/Typechecker.ml
+++ b/src/frontend/Typechecker.ml
@@ -437,7 +437,7 @@ let verify_fn_target_plus_equals cf loc id =
     && not
          ( cf.in_lp_fun_def || cf.current_block = Model
          || cf.current_block = TParam )
-  then Semantic_error.target_plusequals_outisde_model_or_logprob loc |> error
+  then Semantic_error.target_plusequals_outside_model_or_logprob loc |> error
 
 (** Rng functions cannot be used in Tp or Model and only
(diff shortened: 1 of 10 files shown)


src/frontend/Typechecker.ml
```diff
@@ -1575,7 +1575,7 @@ and get_fn_decl_or_defn loc tenv id arg_tys rt body =
   match body with
   | {stmt= Skip; _} ->
       if exists_matching_fn_declared tenv id arg_tys rt then
-        Semantic_error.fn_decl_without_def loc |> error
+        Semantic_error.fn_decl_without_def loc id.name true |> error
```
nhuurre, line 1578: Is "re-declared without specifying a definition" really the best description of this error? Looks to me this happens when there are two identical declarations, regardless of whether there is also a definition later. I think "Function has already been declared. Expected a definition" would be more appropriate. (aka FunDeclExists, an error that does not seem to be in use currently?)

WardBrian: Good catch. It seemed odd to me that these used the same error, but I didn't spot the one it should be using

nhuurre (approved):
Looks good to me.

Issue #576, opened 2020-06-16
Add algebra_solver_powell and deprecate algebra_solver


rok-cesnovar:
`algebra_solver_powell()` should replace what is currently `algebra_solver()`, which should be deprecated. 

Before proceeding with this issue, the function names in Math should be changed.

charlesm93:
This issue needs some love. I'll dive into it.

charlesm93:
Going through the `stan-math` library, it looks like we already have the `algebra_solver_powell` signature in place. So we should be ready to update the stanc3. We also need to expose the variadic argument signature.

I realize this is not super kosher, but I'll fix both issues at once.

PR #1259, opened 2022-10-13
Refactor: encapsulate variadic functions in typechecking

WardBrian:
In anticipation of #576 and potential future variadic higher-order functions, I decided it was worth taking a good look at cleaning up the shared code between existing variadic functions. 

The DAE and ODE functions, and the future `solve_newton` etc are completely described by a few rules:

1. What type does the higher-order function itself return?
2. What arguments are for this function, rather than passed to the first argument (I named these the `control_args`)?
3. What return type do we require for the function passed in?
4. What, if any, arguments do we require for the function passed in?

This PR reorganizes the existing code around this to remove current duplication and prevent more going foward.

Note that the above are **not** sufficient to capture `reduce_sum`, since item 4 is actually dynamic depending on the first argument. So, for now at least, this is left as a special case. 

This would also let us try to print these as part of `--dump-stan-math-signatures` eventually, if we wanted to. 

Finally, I'd like to note that it would eventually be nice to also clean up the code gen for these higher-order functions, which is currently also very special cased due to the differing positions of arguments, different requirements for functors being generated, etc. 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

## Copyright and Licensing

### DIFF
--- a/src/frontend/Typechecker.ml
+++ b/src/frontend/Typechecker.ml
@@ -122,8 +122,7 @@ let verify_name_fresh_udf loc tenv name =
     (* variadic functions are currently not in math sigs and aren't
        overloadable due to their separate typechecking *)
     Stan_math_signatures.is_reduce_sum_fn name
-    || Stan_math_signatures.is_variadic_ode_fn name
-    || Stan_math_signatures.is_variadic_dae_fn name
+    || Stan_math_signatures.is_stan_math_variadic_function_name name
   then Semantic_error.ident_is_stanmath_name loc name |> error
   else if Utils.is_unnormalized_distribution name then
     Semantic_error.udf_is_unnormalized_fn loc name |> error
@@ -194,10 +193,12 @@ let stan_math_return_type name arg_tys =
   match name with
   | x when Stan_math_signatures.is_reduce_sum_fn x ->
       Some (UnsizedType.ReturnType UReal)
-  | x when Stan_math_signatures.is_variadic_ode_fn x ->
-      Some (UnsizedType.ReturnType (UArray UVector))
-  | x when Stan_math_signatures.is_variadic_dae_fn x ->
-      Some (UnsizedType.ReturnType (UArray UVector))
+  | x when Stan_math_signatures.is_stan_math_variadic_function_name x ->
+      Some
+        (UnsizedType.ReturnType
+           (Hashtbl.find_exn Stan_math_signatures.stan_math_variadic_signatures
+              x )
+             .return_type )
   | _ ->
       SignatureMismatch.matching_stanlib_function name arg_tys
       |> match_to_rt_option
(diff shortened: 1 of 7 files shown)


src/frontend/Typechecker.ml
```diff
@@ -194,10 +193,12 @@ let stan_math_return_type name arg_tys =
...
-      Some (UnsizedType.ReturnType (UArray UVector))
+  | x when Stan_math_signatures.is_stan_math_variadic_function_name x ->
+      Some
+        (UnsizedType.ReturnType
+           (Hashtbl.find_exn Stan_math_signatures.stan_math_variadic_signatures
+              x )
+             .return_type )
   | _ ->
       SignatureMismatch.matching_stanlib_function name arg_tys
       |> match_to_rt_option
```
nhuurre, line 204: How about
```ocaml
let stan_math_return_type name arg_tys =
  match (Hashtbl.find Stan_math_signatures.stan_math_variadic_signatures name) with
  | _ when Stan_math_signatures.is_reduce_sum_fn name ->
      Some (UnsizedType.ReturnType UReal)
  | Some { return_type ; _ } -> Some(UnsizedType.ReturnType return_type)
  | None ->
      SignatureMismatch.matching_stanlib_function name arg_tys
      |> match_to_rt_option
```
and maybe move the whole function to SignatureMismatch since it's not needed here?

src/frontend/Typechecker.ml
```diff
@@ -635,79 +638,30 @@ and check_reduce_sum ~is_cond_dist loc cf tenv id tes =
...
+and check_variadic ~is_cond_dist loc cf tenv id tes =
+  let Stan_math_signatures.
+        { control_args
+        ; required_fn_args
+        ; required_fn_rt
+        ; allow_fn_lpdf
+        ; return_type } =
+    Hashtbl.find_exn Stan_math_signatures.stan_math_variadic_signatures id.name
+  in
   let fail () =
```
nhuurre, line 650: This `fail ()` is only used once so it doesn't need to be a separate function. Just move the code to where it is invoked.

src/middle/Stan_math_signatures.ml
```diff
@@ -90,6 +90,14 @@ type fkind =
 type fun_arg = UnsizedType.autodifftype * UnsizedType.t
 type signature = UnsizedType.returntype * fun_arg list * Mem_pattern.t
 
+type variadic_signature =
+  { return_type: UnsizedType.t
+  ; control_args: fun_arg list
+  ; allow_fn_lpdf: bool [@default false]
```
nhuurre, line 96: I don't think this `allow_lpdf_fn` field serves any purpose. Allowing lpdfs really only works for reduce_sum which was special-cased anyway.
That's because the user can wrap a normal lpdf in a plain function so `allow_lpdf` is really more like `allow_lupdf` and _that_ is very hard to get right. Even something like `reduce_log_sum_exp`, when applied to `binomial_lupdf` would drop the binomial coefficients and often return nonsense.

nhuurre (changes requested):

src/middle/Stan_math_signatures.ml
WardBrian: You’re right, I was originally hopeful that I could fit reduce_sum into this paradigm but this flag ended up not being useful.

src/frontend/Typechecker.ml
WardBrian: Agreed on the refactor, I prefer to keep it in Typechecker (in part because I am still hopeful for something like #1182, where it makes the most sense to encapsulate it inside of Typechecking)

nhuurre (approved):

PR #1263, opened 2022-10-31
Allow truncation syntax in vectorized sampling statements

### LINKED ISSUE #788: Vectorized truncation
Is it possible to vectorize the truncation syntax (`T[U, L]`)?

This came up on the forums recently: https://discourse.mc-stan.org/t/truncation-with-vectorized-sampling-statements/20084

The underlying distributions function should already be vectorized enough.

Edit: fixed syntax

WardBrian:
Closes #788. This allows statements like 

```stan
  y ~ normal(mu1, sigma1) T[L, U];
```

In place of 

```stan
  for (n in 1 : N) {
    y[n] ~ normal(mu2, sigma2) T[L, U];
  }
```

We previously translated the loop into (MIR syntax):

```
for(n in 1:N) {
    if((y[n] < L)) 
        target += FnNegInf__(); 
    else if((y[n] > U)) 
        target += FnNegInf__(); 
    else 
        target += PMinus__(log_diff_exp(normal_cdf_log(U, mu2, sigma2), normal_cdf_log(L, mu2, sigma2)));
    target += normal_lupdf(y[n], mu2, sigma2);
}
```

We now turn the vectorized version into 
```
if((min(y) < L)) 
    target += FnNegInf__(); 
else if((max(y) > U)) 
    target += FnNegInf__(); 
else 
    target += PMinus__((log_diff_exp(normal_cdf_log(U, mu1, sigma1), normal_cdf_log(L, mu1, sigma1)) * size(y)));

target += normal_lupdf(y, mu1, sigma1);
```

NB the calls to `min`, `max`, and the multiplication by `size(y)`. This formulation is based on #788 and the linked thread, but also to me seems more or less necessarily equivalent to the for loop.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Vectorized sampling statements can now be used with the truncation syntax `T[ , ]`

## Copyright and Licensing


nhuurre (changes requested):
Thanks, this looks very useful!
We'll need to sort out some trickier cases though. Remember that Stan allows both y _and_ mu to be vectorized in `y ~ normal(mu, 1)`.
If mu (or sigma) is a vector (necessarily of the same size as y) you shouldn't multiply by the size of y; the lcdf function sums the terms for each mu.
Vector-like mu also requires that `T[L,U]` (with both upper and lower truncation) creates a loop, unless new Stan Math functions are added, such as vectorized stan-dev/stan#1154

bob-carpenter:
I agree this is going to be super useful.  And thanks for catching the alternative vectorization cases, @nhuurre.

src/frontend/Typechecker.ml
```diff
@@ -1110,12 +1110,6 @@ let is_cumulative_density_defined tenv id arguments =
   && ( valid_arg_types_for_suffix "_lccdf"
      || valid_arg_types_for_suffix "_ccdf_log" )
 
-let verify_can_truncate_distribution loc (arg : typed_expression) = function
-  | NoTruncate -> ()
-  | _ ->
-      if UnsizedType.is_scalar_type arg.emeta.type_ then ()
-      else Semantic_error.multivariate_truncation loc |> error
-
```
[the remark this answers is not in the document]
WardBrian: It was originally called on line 1146, also removed in this PR

WardBrian:
@nhuurre great catches! 

My latest two commits:

1. Make it so the truncation is only multiplied by `size(y)` if none of the later args are vectors
2. Generate a for loop in the `T[U, L]` vector case. 
3. move the `lupdf` statement above this logic. I believe this is sound (they're both just accumulation statements), my reasoning is that item 2 here could generate a weird error if `size(mu) != size(sigma)`, but the `normal_lupdf` call will necessarily check this and produce a nice error if it is not true.

src/frontend/Ast_to_Mir.ml
```diff
@@ -97,60 +97,120 @@ let neg_inf =
...
+            let bound =
+              Expr.Helpers.internal_funapp FnLength [e]
+                {e.meta with type_= UnsizedType.UInt} in
+            let bodyfn (idx : Expr.Typed.t) =
+              let ast_args = trans_exprs ast_args in
+              let args =
+                List.map
+                  ~f:(fun (e : Expr.Typed.t) ->
+                    if UnsizedType.is_container e.meta.type_ then
+                      Expr.Helpers.add_int_index e (Index.Single idx)
```
nhuurre, line 204: `add_int_index` expects a `Var` or `Indexed` so code like `y ~ normal(a*x + b, sigma)T[0,1];` blows up. You need temporaries to store the arguments because you certainly don't want to recompute the full vector `a*x + b` for each iteration of the loop.

WardBrian: Yep, good catch again.

In theory we could store temporaries for this entire operation, since even without the loop it would still be computing that expression ~twice (once for the lupdf, once for the cdf)?

nhuurre: Yes, that makes sense.

src/middle/Stmt.ml
```diff
@@ -115,26 +115,40 @@ module Numbered = struct
...
 module Helpers = struct
+  let temp_vars exprs : Located.t list * Expr.Typed.t list * (unit -> unit) =
+    let sym, reset = Gensym.enter () in
+    let rec loop es sym inits vars =
+      match es with
+      | [] -> (inits, vars)
+      | e :: es -> (
+        match e with
+        | Expr.{Fixed.pattern= Var _; _} -> loop es sym inits (e :: vars)
+        | _ ->
```
nhuurre, line 126: can you combine these `match`es? Like
```ocaml
match es with
| [] -> (inits, vars)
| Expr{Fixed.pattern=Var _;_} as e :: es -> loop es sym inits (e :: vars)
| e :: es -> ...
```
Also, I'm not sure if that should be `pattern=Var _` or `pattern=Var _ | Indexed _`? Does `Indexed` as an rvalue create copies or references?

src/frontend/Ast_to_Mir.ml
```diff
@@ -97,60 +97,124 @@ let neg_inf =
...
+            ast_args
+        with
+        (* If any of the arguments (besides the data) are vectors, need to generate a loop
+           This can go away if https://github.com/stan-dev/stan/issues/1154 is implemented
+        *)
+        | Some e ->
+            let e = trans_expr e in
+            let bound =
+              Expr.Helpers.internal_funapp FnLength [e]
+                {e.meta with type_= UnsizedType.UInt} in
```
nhuurre, line 197: You should use the temporary for the loop size. I'm seeing generated code like
```cpp
for (int sym3__ = 1;
     sym3__ <= stan::math::size(
                 stan::math::add(stan::math::multiply(a, x), b));
     ++sym3__) {
```

src/middle/Stmt.ml
WardBrian: I believe this would create copies. The rvalue in C++ ends up being something like `Eigen::Block`, but when that is assigned to an `Eigen::Matrix` it would get evaluated.

That said, and `Indexed` could very well be something crazy like `(a +  b * X)[:]`, so I think we want anything other than a Var here to have a temporary created.

src/frontend/Ast_to_Mir.ml
WardBrian: Fixed

nhuurre (approved):

WardBrian:
Thanks @nhuurre. I just circled back to improve the error message which will show if a vectorized lcdf is not available and add a test for that. 

I'll merge after the tests pass.

PR #1265, opened 2022-11-02
Treat int returns from functions as DataOnly

### LINKED ISSUE #803: Unexpected error compiling (something to do with `data` arguments)
Model is:

```stan
parameters {
  vector[5] x;
}
model {
  x ~ normal(0, linspaced_array(size(x), 1.0, 10.0));
}
```

Error is:

```
Semantic error in 'normals.stan', line 6, column 16 to column 51:
     4:  
     5:  model {
     6:    x ~ normal(0, linspaced_array(size(x), 1.0, 10.0));
                         ^
     7:  }
     8:  
Ill-typed arguments supplied to function 'linspaced_array'. Available signatures: 
(data int, data real, data real) => real[]
Instead supplied arguments of incompatible type: int, real, real.
```

I think this should work.

WardBrian:
Fixes #803. This marks any function call which results in an integer as `data` rather than autodiffable. This will include things like `size`, `cols`, etc. 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Functions which return integer types are considered as "data" for purposes of data-only arguments. 

## Copyright and Licensing

### DIFF
--- a/src/frontend/Typechecker.ml
+++ b/src/frontend/Typechecker.ml
@@ -458,8 +458,14 @@ let verify_unnormalized cf loc id =
     && not ((cf.in_fun_def && cf.in_udf_dist_def) || cf.current_block = Model)
   then Semantic_error.invalid_unnormalized_fn loc |> error
 
-let mk_fun_app ~is_cond_dist (x, y, z) =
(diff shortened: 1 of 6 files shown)


rok-cesnovar (approved):
Looks great!

PR #1271, opened 2022-11-09
Mark dae time parameters as DataOnly

WardBrian:
Marked the `initial_time` and `times` parameters of the `dae` functions as DataOnly. If you look in the C++, they're `double`s:  https://github.com/stan-dev/math/blob/f4c68170c1f1e9bc1284520560204ea9e76c62cd/stan/math/rev/functor/dae.hpp#L167-L168

This causes a C++ compilation error which we should be catching in stanc, see https://discourse.mc-stan.org/t/compilation-error-on-using-the-differential-algebraic-equation-dae-solver-on-time-that-depends-on-the-parameters/29404

I'm going to ping @yizhang-yiz on this as the original author to confirm this is correct.

@rok-cesnovar IMO we can merge this during the freeze, but it doesn't warrant an RC2 all on its own. 

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/594
    
## Release notes
Marked the `initial_time` and `times` parameters of the `dae` functions as `data`. 

## Copyright and Licensing

### DIFF
--- a/src/middle/Stan_math_signatures.ml
+++ b/src/middle/Stan_math_signatures.ml
@@ -2608,8 +2608,8 @@ let variadic_dae_tol_arg_types =
 
 let variadic_dae_mandatory_arg_types =
   [ (UnsizedType.AutoDiffable, UnsizedType.UVector); (* yy *)
-    (UnsizedType.AutoDiffable, UnsizedType.UVector); (* yp *)
-    (AutoDiffable, UReal); (AutoDiffable, UArray UReal) ]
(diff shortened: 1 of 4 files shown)


rok-cesnovar (approved):
Based on the C++ implementation this is definitely correct. If those could or should be parameters is a different story, that is more of a question for @yizhang-yiz and also irrelevant for this PR.

> IMO we can merge this during the freeze, but it doesn't warrant an RC2 all on its own.

Agree in both cases.

WardBrian:
@yizhang-yiz confirmed on the forum thread that this is the intended behavior

PR #1274, opened 2022-12-09
More consistently print requested file name in errors 

WardBrian:
#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Improved `--filename-in-msg` for the command line version of `stanc`. Errors now use the requested name, not just warnings. 

## Copyright and Licensing


nhuurre (approved):
Great! And stancjs already handled this.

Not important for now but I noticed the docstring for `Location.to_string` says
> If printed_filename is passed, it will replace the filename printed for this Location.t and all recursively included ones.

and I think that the **recursive** part is a bad idea; overriding the names of `#include`d files has potential for some confusing messages.

WardBrian:
@nhuurre good catch - that behavior also isn't tested! I've added a test and changed it so that the argument only affects the top-level name (e.g. the file passed on the command line), rather than all of them.

For context, I'm particularly interested in this for things like https://github.com/ivan-bocharov/stan-vscode/pull/11, which sometimes needs to run the compiler on a temporary file. It's nice to print the original name in the error if so.

PR #1277, opened 2023-01-09
Deprecate forward decls

### LINKED ISSUE #976: Allow mutual recursion without forward declaration
Relevant to @bob-carpenter 's comment here: https://github.com/stan-dev/docs/issues/408#issuecomment-928228449. 

We currently allow functions to call themselves without a forward declaration, but mutual recursion would still require at least some subset of the required functions to be declared. We can instead do a first pass through the function block which only adds each function's signature to the symbol table, and then our second pass as normal which checks the bodies etc. This would mean all functions are available to all other functions in an order-independent way.

After this change, function declarations are only really useful for defining functions in other (stan|c++) files. In the long term we may want to move toward this being the only valid use of them for simplicity.

nhuurre:
#### Submission Checklist

- [X] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: stan-dev/docs#612

Closes #976 

The backend simply generates a forward declaration for every function (even nonrecursive ones) which also simplifies functor struct generation a bit.
Also needed to do something smarter for potentially cyclic eigen templates (#1228 heuristic has been broken for a while now because simple recursive functions don't actually need forward decls)

And also I've allowed variables to shadow user defined functions and not just stan math functions. The reason is shown here:
```stan
functions {
  void foo() {
    xyz(); // function visible even without a forward declaration - yay
  }
  void bar() {
    real xyz; // name conflicts with the function `xyz` unless shadowing allowed - oh no
  }
  void xyz() {
  }
}
```

## Release notes

Recursive user-defined functions no longer need forward declarations.

## Copyright and Licensing

Copyright holder: Niko Huurre


WardBrian:
How would this interact with the external C++ functions feature?

I think a lot of this is good - generating forward decls for everything and removing any language from the documentation about the need for them is good. We can also make function declarations completely order-independent with this, if we’re willing to do 2 passes over the function block in typechecking, which I would like to see as part of a change like this. 

That said, I think we probably want to pair this with something like an `@external` annotation (see https://github.com/stan-dev/design-docs/pull/45) for their use with external C++, before any deprecation warning was emitted.

More specific review on the code to follow, those are just my thoughts based on the PR description

WardBrian:
`extern "C++" real foo(real);` is an interesting suggestion. For what it's worth, I think "being similar to C++" isn't a strong reason for doing something in Stan, but historically that has been the case.

The thing I like about `@extern` is the potential to do something like this:

```stan
@extern("foo.hpp")
real foo(real);
```

This annotation does 3 things:
1. Disables the "declared but not defined" error for _just this function_
2. Disables all code-gen for this function (maybe functors could stay around, but IMO we don't want to generate _anything_ since that means the user must write code which uses our templates if e.g. we generate a declaration)
3. Place the _contents_ of foo.hpp where we would have generated any function code. This solves the "namespace issue" that currently makes external C++ a pain outside RStan.

I think an extern keyword could easily do 1 and 2, but I'm not sure how to do 3 with it.

src/stan_math_backend/Lower_functions.ml
```diff
@@ -204,141 +189,98 @@ let lower_fun_def (functors : (string, struct_defn) Hashtbl.t)
...
+           else
+             let functors =
+               Map.find_multi functor_required d.fdname |> List.stable_dedup
+             in
+             let fn, st = lower_fun_def functors d in
+             let decl, defn = Cpp.split_fun_decl_defn fn in
+             Some
+               (FunDef decl :: List.map ~f:(fun s -> Struct s) st, FunDef defn) )
+    |> List.unzip in
+  List.concat fun_decls @ fun_defns
```
WardBrian, line 283: I think this concat can happen in the above logic (such that fun_decls and fun_defns always have the same type and can be trivially concatenated)

src/stan_math_backend/Lower_functions.ml
```diff
@@ -204,141 +189,98 @@ let lower_fun_def (functors : (string, struct_defn) Hashtbl.t)
...
-      match
-        Hashtbl.find Stan_math_signatures.stan_math_variadic_signatures x
-      with
-      | Some {required_fn_args; _} ->
-          Map.add_multi accum
-            ~key:(Utils.stdlib_distribution_name f)
-            ~data:(List.length required_fn_args)
-      | _ -> Expr.Fixed.Pattern.fold find_functors_expr accum pattern )
-    | _ -> Expr.Fixed.Pattern.fold find_functors_expr accum pattern in
+let get_functor_requirements (p : Program.Numbered.t) =
```
WardBrian, line 244: I really appreciate how much more general this is than the prior code. Am I correct that as a side effect, we no longer generate any functor if the function is never used in a higher order function?

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -488,30 +467,59 @@ let trans_prog (p : Program.Typed.t) =
...
-        ; meta } in
-  let eval_udf_indexed_stmts (s : 'a Program.fun_def) =
-    if Set.mem possibly_recursive_fns (s.fdname, s.fdargs) then
-      {s with fdbody= Option.map ~f:map_stmt s.fdbody}
-    else s in
-  let p =
-    { p with
-      functions_block= List.map ~f:eval_udf_indexed_stmts p.functions_block }
-  in
+  let callmap = Hashtbl.create (module String) in
```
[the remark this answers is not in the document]
WardBrian: Is callmap a map from name -> direct calls or name -> all calls (including through other functions)?

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -488,30 +467,59 @@ let trans_prog (p : Program.Typed.t) =
...
+      let calls = Hash_set.create (module String) in
+      let fndef = eval_eigen_cycles fun_args calls fd in
+      if not (Hash_set.is_empty calls) then (
+        let calls = Hash_set.to_list calls |> String.Set.of_list in
+        Hashtbl.map_inplace callmap ~f:(fun x ->
+            if Set.mem x fdname then Set.union calls x else x ) ;
+        Hashtbl.update callmap fdname
+          ~f:(Option.value_map ~f:(Set.union calls) ~default:calls) ) ;
+      fndef in
+  let p = {p with functions_block= List.map ~f:break_cycles p.functions_block} in
```
WardBrian, line 522: I think I can follow the above logic well enough to see it is correct, but I'd appreciate some comments on the major steps/flow of information.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -488,30 +467,59 @@ let trans_prog (p : Program.Typed.t) =
...
+  let break_cycles (Program.{fdname; fdargs; _} as fd) =
+    let fun_args =
+      List.filter_map
+        ~f:(fun (_, n, t) ->
+          if UnsizedType.is_eigen_type t then Some n else None )
+        fdargs
+      |> String.Set.of_list in
+    if Set.is_empty fun_args then fd
+    else
+      let calls = Hash_set.create (module String) in
```
WardBrian, line 513:
```suggestion
      let calls = String.Hash_set.create () in
```

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -488,30 +467,59 @@ let trans_prog (p : Program.Typed.t) =
...
+      let fndef = eval_eigen_cycles fun_args calls fd in
+      if not (Hash_set.is_empty calls) then (
+        let calls = Hash_set.to_list calls |> String.Set.of_list in
```
WardBrian, line 516:
```suggestion

```

I think you should either make the type of callmap at the top a `String.Hash_set.t String.Table.t` or make `calls` a `String.Set.t` from the start, rather than doing this intermediate thing.

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -488,30 +467,59 @@ let trans_prog (p : Program.Typed.t) =
...
+          else if is_rec then
+            {e with pattern= FunApp (StanLib ("eval", FnPlain, AoS), [e])}
+          else (Hash_set.add calls name ; e) )
+    and rewrite_expr = function
+      | {pattern= FunApp ((UserDefined (name, _) as kind), args); _} as e ->
+          {e with pattern= FunApp (kind, map_args name args)}
+      | e -> {e with pattern= Pattern.map rewrite_expr e.pattern} in
+    let open Stmt.Fixed in
+    let rec rewrite_stmt = function
+      | {pattern= Pattern.NRFunApp ((UserDefined (name, _) as kind), args); _}
```
WardBrian, line 498: You can make the scope of this `open` tighter:

```suggestion
    let rec rewrite_stmt = 
      let open Stmt.Fixed in
      function
      | {pattern= Pattern.NRFunApp ((UserDefined (name, _) as kind), args); _}
```

WardBrian (changes requested):
Overall I think this is very nice. Most of my comments are stylistic or asking for more clarification

src/stan_math_backend/Lower_functions.ml
```diff
@@ -204,141 +189,98 @@ let lower_fun_def (functors : (string, struct_defn) Hashtbl.t)
...
+    |> List.filter_map ~f:(fun d ->
+           if Set.mem fun_has_def d.fdname && Option.is_none d.fdbody then None
+           else
+             let functors =
+               Map.find_multi functor_required d.fdname |> List.stable_dedup
+             in
+             let fn, st = lower_fun_def functors d in
+             let decl, defn = Cpp.split_fun_decl_defn fn in
+             Some
+               (FunDef decl :: List.map ~f:(fun s -> Struct s) st, FunDef defn) )
```
WardBrian, line 281: I believe the current test failures (`../../ml/stanc3/test/integration/good/overloading/sir_overloaded_odes.hpp:112:8: error: redefinition of ‘struct sir_overloaded_odes_model_namespace::simple_SIR_variadic2_functor__’`) are because of this not properly "collating" functor structs into one struct per name

src/stan_math_backend/Lower_functions.ml
```diff
@@ -162,39 +161,25 @@ let mk_extra_args templates args =
...
   arg_strs
 
-let add_functor_decl functors (name : string)
-    (param : template_parameter option) (f : fun_defn) =
-  let f existing =
-    match existing with
-    | None -> make_struct_defn ~param ~name ~body:[FunDef f] ()
-    | Some sd -> {sd with body= sd.body @ [FunDef f]} in
-  Hashtbl.update functors name ~f
-
```
WardBrian, line 182: Actually it's the removal of this function that I think broke the overloading tests. We probably need something like it still (I think any version which doesn't have this sort of global variable ends up being O(N^2) because you need to search for overloads for each struct you generate)

src/stan_math_backend/Cpp.ml
```diff
@@ -309,6 +309,10 @@ type fun_defn =
   ; body: stmt list option }
 [@@deriving make, sexp]
 
+let split_fun_decl_defn (fn : fun_defn) =
+  ( {fn with body= None}
+  , {fn with templates_init= (fst fn.templates_init, false)} )
```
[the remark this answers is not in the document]
nhuurre: Functor structs' `operator()` methods have a body and `_init=true`. And in the other direction, external function forward declarations turn into a forward declaration and a "function definition" that's secretly just a second declaration with `_init=false`. (we should fix that at some point but I don't think it _breaks_ anything for now...)

src/stan_math_backend/Lower_functions.ml
```diff
@@ -162,39 +161,25 @@ let mk_extra_args templates args =
...
-(** This function produces the function and any functor definitions.
-    Functor {b declarations} need to be collated, and are therefore stored in the
-    functors hashtable *)
-let lower_fun_def (functors : (string, struct_defn) Hashtbl.t)
-    (forward_decls :
-      (string * (UnsizedType.autodifftype * string * UnsizedType.t) list)
-      Hash_set.t ) (funs_used_in_reduce_sum : String.Set.t)
-    (variadic_fns : int list String.Map.t)
-    Program.{fdrt; fdname; fdsuffix; fdargs; fdbody; _} : fun_defn list =
+let lower_fun_def (functors : Lower_expr.variadic list)
```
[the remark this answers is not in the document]
nhuurre: The previous docstring called attention to the side effects. This version has no side effects and the fact that it takes a MIR `fun_def` and produces a LIR `fun_def` and structs for HOFs is pretty obvious from the signature. Is there anything particular you want the docstring to say?

src/stan_math_backend/Lower_functions.ml
nhuurre: That's correct.

src/stan_math_backend/Transform_Mir.ml
nhuurre: I added the type change to signal that we are done collecting calls from this function and the value becomes immutable. But of course that's not correct--the code _on the next line_ mutates the sets inside `callmap`! I'll change it to `Hash_set` everywhere.

src/stan_math_backend/Lower_functions.ml
```diff
@@ -267,20 +268,34 @@ let collect_functors_functions (p : Program.Numbered.t) : defn list =
       | {fdname; fdbody= Some _; _} -> Some fdname
       | _ -> None )
     |> String.Set.of_list in
+  (* overloaded functions generate only one functor struct per name *)
+  let structs = String.Table.create () in
+  let is_overload (d : _ Program.fun_def) arg_types =
+    List.equal UnsizedType.equal
```
WardBrian, line 274: If this is testing for equality of the argument lists then isn't it testing that the function is actually the same, rather than testing that it is an overload?

src/stan_math_backend/Lower_functions.ml
```diff
@@ -204,141 +189,113 @@ let lower_fun_def (functors : (string, struct_defn) Hashtbl.t)
...
+  let fun_decls, fun_defns =
     p.functions_block
-    |> List.concat_map
-         ~f:(lower_fun_def functors forward_decls reduce_sum_fns variadic_fns)
-    |> List.map ~f:(fun f -> FunDef f) in
-  let functor_struct_decls =
-    functors |> Hashtbl.data |> List.map ~f:(fun s -> Struct s) in
-  functor_struct_decls @ fun_and_functor_defs
+    |> List.filter_map ~f:(fun d ->
+           if Set.mem fun_has_def d.fdname && Option.is_none d.fdbody then None
```
WardBrian, line 294: This can be falsely triggered by an overload, I think?

```stan
functions {
   void foo();

   void foo(int a) {
    print(a);
   }
}
```

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -487,31 +466,70 @@ let trans_prog (p : Program.Typed.t) =
...
+    if Set.is_empty fun_args then fd
+    else
+      let calls =
+        Hashtbl.find_or_add callgraph fdname ~default:String.Hash_set.create
+      in
+      let fndef = eval_eigen_cycles fun_args calls fd in
+      (* update `callgraph` with the call paths going through the current function *)
+      Hashtbl.map_inplace callgraph ~f:(fun x ->
+          if Hash_set.mem x fdname then Hash_set.union calls x else x ) ;
+      fndef in
```
WardBrian, line 531: Perhaps I'm missing something, but does this code still not have an order-dependence in the callgraph population? I don't see how it would have information "flow upward" if the final function in the functions block implicated prior ones, for example.

Also, not sure if this is relevant, but might as well ask: can you imagine a situation in which a function is recursive through the use of a higher-order function? Could that exhibit this poor eigen behavior?

src/frontend/Deprecation_analysis.ml
```diff
@@ -156,9 +171,15 @@ let rec collect_deprecated_expr (acc : (Location_span.t * string) list)
...
-let rec collect_deprecated_stmt (acc : (Location_span.t * string) list) {stmt; _}
-    : (Location_span.t * string) list =
+let rec collect_deprecated_stmt fundefs (acc : (Location_span.t * string) list)
+    {stmt; _} : (Location_span.t * string) list =
   match stmt with
+  | FunDef {body= {stmt= Skip; _}; funname; arguments; _}
+    when is_redundant_forwarddecl fundefs funname arguments ->
+      acc
+      @ [ ( funname.id_loc
+          , "Forward declarations are deprecated and not needed for recursion."
```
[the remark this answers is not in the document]
WardBrian: We also generally mention if the canonicalize flag can fix this, and if there is an expiration on the deprecation.

The latter requires a bit more discussion of intent. I don't think we need to settle what it looks like, but if the goal is that somewhere down the line there is a special syntax for external functions and the current forward decls are completely gone, that's worth discussing

src/stan_math_backend/Lower_functions.ml
nhuurre: Yes. This is definitely a case where the function name is just a random word that was on my mind while writing the code. `is_correct_overload` would have been more accurate.

src/stan_math_backend/Lower_functions.ml
nhuurre: Oops, didn't think of that.
I think it's going to be easier to filter out the extra decls during AST -> MIR transformation and then not worry about it in the backend.

src/stan_math_backend/Transform_Mir.ml
nhuurre: If this function could implicate a prior one then that prior function must already implicate this function. We eval()'ed all cycles that would go through this function so at the point where `callgraph` updates it's still cycle-free.

Good point about higher-order functions, this fails on master
```stan
functions {
  vector foo(real x, vector s, matrix y) {
    return ode_rk45(foo, [1]', 0.0, {1.0}, y[2:])[1];
  }
}
parameters {
  real x;
  vector[3] s;
  matrix[3,3] y;
}
transformed parameters {
   vector[3] z = foo(x,s,y);
}
```

WardBrian: Assuming you can fix that in this framework, it would be good to add that as a test.

We should also probably add the small program you have in the original PR text as a regression test

WardBrian: After staring at it a bit longer, I think I understand: we only need to insert an eval _somewhere_ in the chain to prevent the issue. So if we're looking at functions A, B, C, it doesn't matter if A doesn't know to eval calls to C, since when we get to C it will know it should.

This is actually even better than a solution which does what I thought it was, since it won't insert more than are strictly needed.

test/integration/good/code-gen/cpp.expected
```diff
@@ -25219,7 +25305,8 @@ s(const std::vector<T0__>& y_slice, const int& start, const int& end,
   try {
     current_statement__ = 221;
     return stan::math::reduce_sum<s_rsfunctor__>(y_slice, 1, pstream__, a, b,
-             c, d, e, f, g, h, i, j, k, l, m, n, o, p, q);
+             stan::math::eval(c), stan::math::eval(d), stan::math::eval(e),
+             f, g, h, i, j, k, l, m, n, o, p, q);
```
nhuurre, line 25309: This is overly cautious because the code did compile before but note that if the Stan source line were changed from
```stan
    return reduce_sum(s, y_slice, 1, a, b, c, d, e, f, g, h, i, j, k, l, m,
                      n, o, p, q);
```
to
```stan
    return reduce_sum(s, y_slice, 1, a, b, c[2:], d, e, f, g, h, i, j, k, l, m,
                      n, o, p, q);
```
it blows up without the `eval()`s.

src/analysis_and_optimization/Optimize.ml
```diff
@@ -448,47 +448,42 @@ let rec inline_function_statement propto adt fim Stmt.Fixed.{pattern; meta} =
...
+            | `Duplicate -> accum )
+        | Some fdname' ->
+            let data = create_data false in
+            let data' = create_data true in
+            let m = Map.Poly.of_alist_exn [(fdname, data); (fdname', data')] in
+            Map.merge_skewed accum m ~combine:(fun ~key:_ f _ -> f) ) in
+  List.fold l ~init:Map.Poly.empty ~f
 
 let function_inlining (mir : Program.Typed.t) =
+  (* We add only the functions with a single definition to the inline map.
```
WardBrian, line 473: [Comment] Does the fact that this file changed but none of the generated C++ in the tests did suggest to you that everything is good, or that we may be under-testing this particular optimization?

src/stan_math_backend/Lower_functions.ml
```diff
@@ -291,8 +286,9 @@ let collect_functors_functions (p : Program.Numbered.t) : defn list =
   let fun_decls, fun_defns =
     p.functions_block
     |> List.filter_map ~f:(fun d ->
-           if Set.mem fun_has_def d.fdname && Option.is_none d.fdbody then None
-           else Some (declare_and_define d) )
+           (* FIXME: external functions don't need decls
```
WardBrian, line 289: Is this FIXME intended to stay in for the merge (I'm mostly fine if the answer is yes, just checking)

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -467,8 +467,16 @@ let trans_prog (p : Program.Typed.t) =
...
 
+     Infinite expansion can happen only when the call graph is cyclic.
+     The strategy here is to build the call graph one edge at the time
+     and check if adding that edge creates a cycle. If it does, insert
+     an `eval()` to stop template expansion.
+
      All relevant function calls are recorded transitively in `callgraph`,
      meaning if `A` calls `B` and `B` calls `C` then `callgraph[A] = {B,C}`.
+     In the worst case every function calls every other, `callgraph` has
+     size O(n²) and this algorithm is O(n³) so it's important to include
```
WardBrian, line 478: Thanks for these comments, I find them very helpful.

A nitpick:
```suggestion
     size O(n^2) and this algorithm is O(n^3) so it's important to include
```

Just to keep our file encoding simpler I think we should stick to ASCII

WardBrian (commented):
I suspect that this + the unresolved comments above are probably the last substantial review this needs before approval.

Would you mind adding a test which shows function definitions being order-free (the snippet in the original PR text would suffice)? Besides that, I would also like for the PR to stan-dev/docs be written before merging this, since this requires some rather noticeable changes to the user-facing docs.

src/stan_math_backend/Lower_functions.ml
nhuurre: Yes. It only affects external functions that are used directly in higher-order functions. You can't use math library functions like that either so I think this can wait until we have better support for external functions. It's not too difficult to fix once external functions are marked in some way in the MIR.

src/analysis_and_optimization/Optimize.ml
WardBrian: Mind adding that example into `test/integration/good/compiler-optimizations`?

test/integration/bad/functions-bad18.stan
```diff
@@ -4,9 +4,9 @@ functions {
   }
 }
 parameters {
-  real foo;  // should be name conflict here
+  real foo;
 }
 model {
   foo ~ normal(0.0, 1.0);
-  foo();
+  foo();  // the function has been shadowed, this `foo` is a number
```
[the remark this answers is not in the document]
nhuurre: That math library "contextual name shadowing" doesn't make much sense to me and I couldn't find if it's documented anywhere.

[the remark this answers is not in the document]
WardBrian: I'm not sure why it was done this way, but my understanding is:

1. It was desirable to let users name variables the same as built in functions, because names like `beta` are common
2. The fact that C++ namespaces allow us to differentiate between a function (in `stan::math`) and a variable (in `::`) meant we could also let users call that function too. So, we allow it. 

Essentially, [Stan is a Lisp2](https://www.nhplace.com/kent/Papers/Technical-Issues.html), at least when it comes to the standard library. Emulating this same behavior in user functions is possible, but would require extra work due to the fact that we don't namespace-qualify UDF calls.

test/integration/good/compiler-optimizations/dune
```diff
@@ -74,6 +75,8 @@
    fails-test.stan
    off-small.stan
    off-dce.stan
+   overloaded-fn.stan
+   overloaded-fn2.stan
```
nhuurre, line 79: Apparently I don't how these tests work. Why isn't this list just `(glob_files *.stan)`? And was missing `overloaded-fn.stan` a bug in the test?

WardBrian: I think it was a historical thing, where some of the optimizations could cause the compiler to crash so we didn’t run all of them? Not completely sure, it predates me on the project. 

I wouldn’t be opposed to just making them a glob like the other tests are

WardBrian: Looked through the blame, and it seems that at first there was only one file in this directory and the dune file was written without the glob, and that seems to have just persisted without much interrogation.

WardBrian:
I'm going to hold off on a final review/approval until after the [language meeting](https://discourse.mc-stan.org/t/monthly-language-dev-meeting/24161/55) tomorrow in case anyone has some novel thoughts on `extern` functions which are worth considering before this change

WardBrian:
During the language meeting we discussed the `extern` syntax. I'll be opening another issue here to discuss it more fully but long story short is it isn't a blocker for this. @bob-carpenter is very excited this change is finally being made

WardBrian (approved):
LGTM!

PR #1296, opened 2023-03-20
Explicitly promote integers to reals for UDF calls

WardBrian:
Closes https://github.com/stan-dev/stan/issues/2738.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fixes an issue where a user-defined-function expecting real arguments could be passed an integer and exhibit integer behavior like truncating division. 

## Copyright and Licensing


src/stan_math_backend/Lower_expr.ml
```diff
@@ -483,15 +483,21 @@ and lower_indexed_simple e idcs =
...
   | Lit (Str, s) -> literal_string s
   | Lit (Imaginary, s) ->
       fun_call "stan::math::to_complex" [Literal "0"; Literal s]
   | Lit ((Real | Int), s) -> Literal s
-  | Promotion (expr, UReal, _) when is_scalar expr -> lower_expr expr
+  | Promotion (expr, UReal, _) when is_scalar expr ->
+      if promote_reals then
+        (* this can be important for e.g. templated function calls
+           where we might generate an incorrect specification for int *)
+        static_cast Cpp.Double ((lower_expr ~promote_reals:true) expr)
```
nhuurre, line 499: The inner expression does not need `~promote_reals:true` unless you want to promote the value twice (normal code should not have nested promotions but it can happen after optimization, e.g. if you add `~promote_reals:true` to the `FnMakeArray` code on line 401 some tests in `test/integration/good/compiler-optimizations` produce
`static_cast<double>(static_cast<double>(1000000))`

WardBrian: Fair enough (though I assume that the C++ compiler is smart enough to delete the outer one)

nhuurre: If we were to rely on the C++ compiler to ignore superfluous casts (which we certainly can do) then it would be simpler to ignore `promote_reals` completely and generate a `static_cast` for every `Promotion`. I commented because this `~promote_reals:true` looks very much intentional but cannot really accomplish anything.

nhuurre (approved):

src/stan_math_backend/Lower_expr.ml
WardBrian: Yeah, I went for the change which would fix the bug and result in minimal changes in the generated source. `static_cast`ing everything is certainly an option, I confirmed looking at godbolt that it generates identical assembly as the "automatic" upcasting C++ normally does.

PR #1305, opened 2023-04-05
Update generated `transform_inits_impl` to directly accept a var_context.

### LINKED ISSUE #1304: [BUG] `transform_inits` code for complex numbers is incorrect.
`transform_inits_impl` generates `z = in__.read<local_scalar_t__>();` to read in a complex number `z`, but this should be `z = in__.read<std::complex<local_scalar_t__>>();`

The result is that z has a `0` imaginary component and subsequent parameters are read in from the wrong part of the buffer.

I have a partial fix (for scalars) in a branch https://github.com/stan-dev/stanc3/tree/fix/deserializer-read-containers. Containers requires some additional logic in one of the overloads of `transform_inits`

### Model Code

<details><summary>Model which exhibits the issue</summary>

```stan
parameters {
   complex_vector[2] z ;//= [1+2i,3+4i]';
}
model {
   print(z);
}
```

```json
{"z":[[1,2], [3,4]]}
```

```shell
./test_model sample num_warmup=1 num_samples=1 init=params.json
```
prints  `[(1,0),(3,0)]`

</details>

### Environment:

Stan 2.31.0

WardBrian:
#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Closes #1304.

This PR does a couple things:

1. Re-using the code gen we already have for the constructor, `transform_inits_impl` now takes in a var_context directly. This is what fixes #1304, and also will help out with #1100. The previous behavior of flattening it down to a flat vector was hacky, and for complex numbers or tuples, wrong.

2. Adds a new function called `unconstrain_array` which looks much more like the former `transform_inits`. This is the twin of `write_array` and can be eventually exposed/used in place of a lot of the places we're creating a `array_var_context`. The idea for a signature like this dates back to when the deserializer was added: https://github.com/stan-dev/stan/pull/3013#issuecomment-795906563. I added this here to keep the existing code for `transform_inits_impl` around rather than deleting it and re-coding it later, but it will currently be unused.

The second item was actually my original goal, but while working on it I uncovered #1304. The first nicely kills two birds with one stone, fixing that issue and helping out with something I had been stuck on in the tuples PR.

This was able to be done mostly by re-using code already in the compiler. Unfortunately, since the reading code is all generated in `Transform_mir` rather than lower down, the simplest solution was to add another (empty until the backend) item to the MIR.

## Copyright and Licensing

### DIFF
--- a/src/middle/Internal_fun.ml
+++ b/src/middle/Internal_fun.ml
@@ -9,7 +9,7 @@ type 'expr t =
   | FnNegInf
   (* In AST_to_MIR being used as StanLib *)
   | FnReadData
-  | FnReadSerializer
+  | FnReadDeserializer
   (* XXX move these to a backend specific file?*)
   | FnReadParam of
       { constrain: 'expr Transformation.t
@@ -45,13 +45,13 @@ let pp (pp_expr : 'a Fmt.t) ppf internal =
 
 (* Does this function call change state? Can we call it twice with the same results?
 
-   E.g., FnReadSerializer moves the serializer forward, so calling it again has
+   E.g., FnReadDeserializer moves the serializer forward, so calling it again has
    different results
 
    Useful for optimizations
 *)
 let can_side_effect = function
-  | FnReadParam _ | FnReadData | FnReadSerializer | FnWriteParam _
+  | FnReadParam _ | FnReadData | FnReadDeserializer | FnWriteParam _
    |FnValidateSize | FnValidateSizeSimplex | FnValidateSizeUnitVector
    |FnReadWriteEventsOpenCL _ ->
       true
--- a/src/middle/Internal_fun.ml
+++ b/src/middle/Internal_fun.ml
@@ -45,10 +45,10 @@ let pp (pp_expr : 'a Fmt.t) ppf internal =
 
 (* Does this function call change state? Can we call it twice with the same results?
 
-   E.g., FnReadDeserializer moves the serializer forward, so calling it again has
-   different results
+   E.g., FnReadDeserializer moves the deserializer forward, so calling it again has
+    different results
 
-   Useful for optimizations
+    Useful for optimizations
 *)
 let can_side_effect = function
   | FnReadParam _ | FnReadData | FnReadDeserializer | FnWriteParam _
--- a/src/middle/Internal_fun.ml
+++ b/src/middle/Internal_fun.ml
@@ -9,7 +9,7 @@ type 'expr t =
   | FnNegInf
   (* In AST_to_MIR being used as StanLib *)
   | FnReadData
-  | FnReadDataSerializer
+  | FnReadSerializer
   (* XXX move these to a backend specific file?*)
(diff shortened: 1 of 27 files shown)


src/stan_math_backend/Transform_Mir.ml
```diff
@@ -295,15 +295,15 @@ let read_constrain_dims constrain_transform st =
...
-let data_serializer_read loc out_constrained_st =
+let plain_deserializer_read loc out_constrained_st =
   let ut = SizedType.to_unsized out_constrained_st in
-  let dims = SizedType.get_dims_io out_constrained_st in
+  let dims = SizedType.get_dims out_constrained_st in
   let emeta = Expr.Typed.Meta.create ~loc ~type_:ut ~adlevel:AutoDiffable () in
   Expr.(
     Helpers.(
-      internal_funapp FnReadDataSerializer dims Typed.Meta.{emeta with type_= ut}))
+      internal_funapp FnReadSerializer dims Typed.Meta.{emeta with type_= ut}))
```
nhuurre, line 304: Stupid question: why does a function called `deserializer_read` create a call to `ReadSerializer`?

WardBrian: Not at all a stupid question, though the answer might be. It's probably for the same reason the code gen for this function uses a OCaml variable called `serializer_in` which corresponds to the C++ declaration `stan::io::deserializer in__`...
We haven't been the tidiest with our naming conventions. I've been trying to fix them as I noticed, but these are ones I missed and will fix up now

src/middle/Internal_fun.ml
```diff
@@ -45,13 +45,13 @@ let pp (pp_expr : 'a Fmt.t) ppf internal =
 
 (* Does this function call change state? Can we call it twice with the same results?
 
-   E.g., FnReadDataSerializer moves the serializer forward, so calling it again has
+   E.g., FnReadDeserializer moves the serializer forward, so calling it again has
```
nhuurre, line 48:
```suggestion
   E.g., FnReadDeserializer moves the deserializer forward, so calling it again has
```

src/middle/Program.ml
```diff
@@ -31,20 +31,16 @@ type ('a, 'b) t =
   ; input_vars: (string * 'a SizedType.t) list
   ; prepare_data: 'b list (* data & transformed data decls and statements *)
   ; log_prob: 'b list (*assumes data & params are in scope and ready*)
-  ; generate_quantities: 'b list (* assumes data & params ready & in scope*)
+  ; generate_quantities: 'b list
+        (* assumes data & params ready & in scope*)
+        (* TODO the following two items are really backend-specific *)
```
SteveBronder, line 36: Agree these should be moved to something on the backend, but this might be better just as an issue somewhere or as a top level comment

SteveBronder (approved):
Nice, much cleaner! Code and gen'd c++ look good

PR #1311, opened 2023-04-13
Add c bytecode target for stanc

andrjohns:
This draft PR is for testing and getting feedback on an approach to bundling the C bytecode for stanc, and the minimum necessary sources from ocaml, so that users can compile stanc locally (no more precompiled binaries!).

From my (limited) local testing, this is working across Linux, Mac, and Windows


andrjohns:
@serban-nicusor-toptal it looks like the docker image is missing the `file` utility:
```
./configure: line 6771: /usr/bin/file: No such file or directory
```
 Would you mind updating the image to install it when you get the time? Thanks!

stanc/Makefile
WardBrian, line 1: I don't think this folder should be called `stanc`, to me that makes it seem like it might be where you look to find the actual compiler code rather than in `src/stanc`.

Maybe `bytecode_build` or something similar? With a folder-specific README.

andrjohns:
Alright, this is ready for review. 

I've added a new makefile `Makefile.package-stanc` which contains the recipes for obtaining the stanc C bytecode and extracting the ocaml sources and headers that are needed to compile it.

These are placed in a folder (`stanc`) which has the Makefile and `main.c` needed for compiling a standalone binary. Currently this process requires the Ocaml `configure` script to generate machine-specific headers (`caml/m.h` and `caml/s.h`) but this is probably something that could be streamlined/simplified.

I've included a github actions which shows the bytecode and C files being generated under ubuntu, and then being built under ubuntu, mac, and windows and running the cmdstan tests

Makefile.package-stanc
```diff
@@ -0,0 +1,116 @@
...
+stanc-bytecode:
+	# Temporarily remove the bigarray requirement from sexplib - otherwise unix is required
+	sed -i'.bak' -e 's/requires bigarray/requires/g' $(SEXLIB_DUNEPACKAGE)
+	# Temporarily mask the unix library so it can't be included
+	mv $(OCAML_LIBDIR)/unix.cma $(OCAML_LIBDIR)/unix.cma.bak
+
+	dune build src/stanc/stanc.bc.c
+	mv $(OCAML_LIBDIR)/unix.cma.bak $(OCAML_LIBDIR)/unix.cma
+	$(RM) $(SEXLIB_DUNEPACKAGE)
+	mv $(SEXLIB_DUNEPACKAGE).bak $(SEXLIB_DUNEPACKAGE)
```
WardBrian, line 46: I believe if `dune build` fails this will leave the sexplib in a bad state.

There are probably ways to make this safer, but in general I don't really want anything in our makefiles to touch anything inside the opam folder

Makefile
```diff
@@ -1,3 +1,5 @@
+-include Makefile.package-stanc
```
WardBrian, line 1: Is there a reason why the packaging makefile couldn't live entirely inside the bytecode build directory? It seems unnecessary to include any of this in the top-level makefile since it won't be the default way of building stanc

.github/workflows/binaries-downstream.yml
```diff
@@ -0,0 +1,103 @@
...
+        run: |
+          ${{ matrix.config.make }} -j2 build
+          ${{ matrix.config.make }} install-tbb
+          rm bin/${{ matrix.config.stanc }}
+          cp stanc/${{ matrix.config.stanc }} bin/${{ matrix.config.stanc }}
+        shell: bash
+
+      - name: Run CmdStan tests
+        run: |
+          python ./runCmdStanTests.py -j2 src/test/interface
```
WardBrian, line 102: I guess the cmdstan interface tests rely on building a few models, but does running cmdstan with this executable provide us any confidence that simply building and running the bytecode version would not?

I think the most we'd need to do is 
1. Run `stanc --help`, check that the exit code is nonzero
2. Possibly compile an example file and assert that the diff between the code generated by the bytecode stanc and a native build is the same (It is almost impossible for me to imagine why this second test would ever fail, so it probably isn't even necessary)

In particular, running any of the cmdstan tests as part of the stanc CI can lead to a deadlock if we need to make a breaking change. I guess doing it as a github action rather than in our main jenkins pipeline is fine, since it would not prevent the release of the `nightly` tag, but I'm still weary of it

WardBrian (changes requested):
This is all really impressive. I had tried to do similar things before but I was under the impression you needed to have compiled versions of the ocaml libs around and figured that was too big a hurdle to make local compilation any easier than just installing `ocamlc` would be.

A more broad question is if you think this really needs to be in the `stanc3` repo? Besides the change to the one `dune` file, it seems like this could be done out-of-tree, and it wouldn't ever break unless/until we update OCaml or our dependencies (right?), so it seems like we could do it somewhere else with a submodule that gets pulled in weekly or something rather than generating noise each PR
If there _could_ be changes on normal PRs which break this build, then it would be nice to know that, too

Makefile.package-stanc
WardBrian: (Note: if we did want to move this process to another repo, it would be reasonable IMO to use something like https://github.com/tarides/opam-monorepo and then edit the local version of the dependency. If it's going to be inside the `stanc3` repo, I don't want it to step on people's environments even by accident)

andrjohns:
Thanks! It's been one of those things that's been stuck in my "should be possible" list for a while!

Absolutely agree that it doesn't have to (or should) live in `stanc3`. I could setup another repo (`stanc3-bytecode`?) with CI to regularly pull in `stanc3` and update the bytecode & sources (and test). Then if nothing breaks after a release cycle or two of normal development, we could open a discussion about including that in `cmdstan` as a submodule to replace the `stanc3` binary?

WardBrian:
Do you have sufficient permissions that you can make a new repo under the `stan-dev` organization's name?

> we could open a discussion about including that in cmdstan as a submodule to replace the stanc3 binary?

I'd be curious if there's a noticeable performance difference between the bytecode executables this produces and the native binaries we currently ship. 

I'd probably argue in favor of including the `linux-stanc`, `mac-stanc`, and `windows-stanc` binaries in releases, and probably keep our CI downloading `nightly` releases rather than building the bytecode version from source, but it definitely seems like a reasonable option for platforms other than the "big three" or if there are users who don't want to use something they didn't compile (not that the bytecode file is particularly readable, but that's another story).

 Am I correct in guessing you'll want to move RStan to use this rather than stanc3js?

andrjohns:
> Do you have sufficient permissions that you can make a new repo under the stan-dev organization's name?

Good shout, no I don't. But thinking it through more, it might be too early to have the repo under `stan-dev` since that might provide some kind of implicit endorsement  by Stan - while this is probably still a bit too experimental. I'll open the repo under my account instead and we can discuss moving it under the stan-dev umbrella later if all appears stable.

> I'd be curious if there's a noticeable performance difference between the bytecode executables this produces and the native binaries we currently ship.

Me too, I could add some timing tests/comparisons as a future goal for the separate repo CI

> I'd probably argue in favor of including the linux-stanc, mac-stanc, and windows-stanc binaries in releases

Interesting, I was thinking it would be easier to have this as the main format for `stanc` - with the native binaries for those that have some need for speed (e.g., a workflow involving parsing many many models). Since this would allow for us to distribute a single `cmdstan` release source which could be used for every platform. But I don't have any hard plans or strong opinions on that.

One of the other places where I see this being useful is that it makes it much easier to package `stanc` for package distribution platforms (`homebrew`, `apt`, etc.) as a standalone program. But I'm not sure how much of a usecase there is for that just yet.

> Am I correct in guessing you'll want to move RStan to use this rather than stanc3js?

Still a bit early to say. CRAN has strict requirements on the C/C++ in packages (has to compile without warnings under `-Wpedantic`) which can be a bit of pain if you're working with external code (currently dealing with this for [replacement V8 engine](https://cran.r-project.org/web/checks/check_results_QuickJSR.html)). I haven't tried building under `-Wpedantic` yet to see how much work would be needed

WardBrian:
Sounds good. I’m very interested in how it all shakes out!

The wpedantic restriction would probably be an issue is my guess. Shame, because using an executable of stanc that could access the file system would fix a lot of issues with things like how RStan handles `#include` directives

andrjohns:
I've changed this PR to just be adding the `(byte c)` target to the stanc `dune` file. The `(best exe)` is the current default behaviour, so there's no change to the current executables

WardBrian (approved):

WardBrian:
One thing to flag @andrjohns is that the OCaml source code and runtime are distributed under the LGPL. This would mean that the bytecode source (which must actually distribute the source, rather than just link to a library) would fall under the LGPL as well.

I don't believe this is necessarily a huge issue, since none of the code which would be affected is incorporated into the resulting Stan programs, so they would be unaffected (like how GCC is GPL, but the produced executables are not), but wanted to point it out

andrjohns:
Ahh good point. I'll read up a bit on the details so I don't run into any surprises

PR #1313, opened 2023-04-19
Use full return type in --standalone-functions definitions

### LINKED ISSUE #1312: [FR] `--standalone-functions` change auto return to plain type?
The `--standalone-functions` flag currently produces functions with an `auto` return type, but this needs to be replaced with a plain return type for compatibility with exposing functions through `Rcpp` (`Rcpp` generates intermediate definitions/functions which don't always play nicely with `auto`). 

We're currently parsing the rest of the c++ to extract the plain type and replace `auto`. But this causes failures if the specification of plain types in the rest of the generated c++ changes in stanc3 versions (this is the cause of one of the current `cmdstanr` failures with 2.32). 

Would it be possible to change this behaviour in `stanc3` so that the plain type is returned instead of `auto`? 

If there's a preference to keep the `auto` return as an option for propagating expressions, there could be an additional flag requesting plain types instead (e.g., `--standalone-functions --plain-return`, or similar)?

As concrete examples, the function:
```stan
functions {
  matrix ret_matrix(matrix x) { return x; }
}
```

Currently becomes:
```cpp
auto ret_matrix(const Eigen::Matrix<double, -1, -1>& x, 
                std::ostream* pstream__ = nullptr)  {
 return dummy_model_namespace::ret_matrix(x, pstream__);
}
```

With the desired c++ instead being:
```cpp
Eigen::Matrix<double, -1, -1> ret_matrix(const Eigen::Matrix<double, -1, -1>& x, 
                                         std::ostream* pstream__ = nullptr)  {
 return dummy_model_namespace::ret_matrix(x, pstream__);
}
```

WardBrian:
Closes #1312 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

`--standalone-functions` now generates signatures which have the return type written out in full, rather than using `auto`.

## Copyright and Licensing

### DIFF
--- a/src/stan_math_backend/Lower_functions.ml
+++ b/src/stan_math_backend/Lower_functions.ml
@@ -88,9 +88,7 @@ promotion.*)
 let lower_returntype arg_types rt =
   let scalar = lower_promoted_scalar arg_types in
   match rt with
-  | UnsizedType.ReturnType ut when UnsizedType.is_int_type ut ->
-      lower_type ut Int
-  | ReturnType ut -> lower_type ut scalar
+  | UnsizedType.ReturnType ut -> lower_type ut scalar
   | Void -> Void
 
 let lower_eigen_args_to_ref arg_types =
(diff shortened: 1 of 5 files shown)


WardBrian:
@andrjohns does this look like it would solve the issue you ran into?

src/stan_math_backend/Lower_functions.ml
```diff
@@ -316,7 +316,9 @@ let lower_standalone_fun_def namespace_fun
   let return_type, return_stmt =
     match fdrt with
     | Void -> (Void, fun e -> Expression e)
-    | _ -> (Auto, fun e -> Return (Some e)) in
+    | ReturnType ut when UnsizedType.is_int_type ut ->
+        (lower_type ut Int, fun e -> Return (Some e))
+    | ReturnType ut -> (lower_type ut Double, fun e -> Return (Some e)) in
```
nhuurre, line 321: Special-casing on `is_int_type` is not necessary; `lower_type` ignores the `scalar` argument when the inner type is integer.

nhuurre (approved):

PR #1316, opened 2023-04-24
Validate the dimensionality of variables in transform_inits

### LINKED ISSUE #1315: [BUG] `transform_inits` does not check validity of var_context
### Current Behavior:

See https://github.com/stan-dev/stan/issues/3182. 

In short, `transform_inits` will happily read in a var_context object which is the wrong size. For example:

```stan
parameters {
  array[0] real a;
  array[1] real b;
}
```
```json
{ "a":  2.1223971837869353 , "b": [ 5.1902179174507062 ] }
```

After #1305, this will read in a value for `a` even though one isn't expected, causing the `stan::io::serializer` to run out of room when trying to write the value for `b`, yielding an error like

```
Unrecoverable error evaluating the log probability at the initial value.
Exception: In serializer: Storage capacity [1] exceeded while writing value of size [1] from position [1]. This is an internal error, if you see it please report it as an issue on the Stan github repository. (found before start of program)
```

Before #1305, a different kind of error could happen:
```stan
parameters {
  array[1] real a;
}
```
```json
{ "a": [] }
```

Running this with a compiler from before #1305 results in a segmentation fault, after it just doesn't read anything for `a`.

### Expected Behavior:
An error when the transform_inits var context has data of the wrong size (or, possibly, wrong shape? it seems a lot of R code is relying on us not being too picky about the shapes here...)

WardBrian:
#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Closes #1315. The shapes and sizes of parameter initial values are now checked against the specified sizes in the model before being read in, matching the existing behavior for data. 

Note: This change may cause existing code which worked "by accident" to raise new errors about mismatches between the model's declarations and the provided initialization.

Errors in transform_inits should now properly highlight the location in the original Stan program.

## Copyright and Licensing


test/integration/good/code-gen/cl.expected
```diff
@@ -1392,6 +1392,24 @@ class optimize_glm_model final : public model_base_crtp<optimize_glm_model> {
     // suppress unused var warning
     (void) DUMMY_VAR__;
     try {
+      context__.validate_dims("parameter initialization", "alpha_v",
+        "double", std::vector<size_t>{static_cast<size_t>(k)});
```
nhuurre, line 1396: I don't see assignment to `current_statement__` so if this fails the error is going to say something like "(found before start of program)" instead of a line number, right? Would it be possible to include location of the declaration in Program.outvar and add `Numbering.assign_loc` right before the `validate_dims` call?

WardBrian: Yeah, currently transform inits basically lacks any use of the locating mechanism. E.g, try:

```stan
parameters {
  simplex[2] s;
}
```
```json
{ "s": [ 0.3, 0.8 ] }
```

The output says 
```
Exception: stan::math::simplex_free: Simplex variable is not a valid simplex. sum(Simplex variable) = 1.1, but should be 1 (found before start of program)
```

WardBrian: I was able to make this change but it was kind of big/annoying. Starting a separate branch if that works for you?

PR #1318, opened 2023-04-24
Add locations to output variables, adjust code generation

WardBrian:
This includes the changes of #1316 but was big enough I wanted to split it out for review. There might be a "cleaner" way to do this that doesn't end up threading an extra generic type parameter through `Program.t`

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Errors in transform_inits should now properly highlight the location in the original Stan program

## Copyright and Licensing


WardBrian:
This improves the error messages in #1316:

```
Exception: mismatch in number dimensions declared and found in context; processing stage=parameter initialization; variable name=a; dims declared=(0); dims found=() (in '../../ml/stanc3/inits.stan', line 2, column 16 to column 17)
```

It also improves the existing errors for things like when a simplex fails to sum to 1 in an inits file:

```
Exception: stan::math::simplex_free: Simplex variable is not a valid simplex. sum(Simplex variable) = 1.1, but should be 1 (in '../../ml/stanc3/inits.stan', line 2, column 13 to column 14)
```

Both of these previously said `(before start of program)`

nhuurre (approved):
> There might be a "cleaner" way to do this that doesn't end up threading an extra generic type parameter through `Program.t`

The only other way I can see is hard-coding it as
```ocaml
type outvar_loc = Loc of Location_span.t | Number of int
```
The downside is dynamic typing: we don't have a type-level guarantee that `Numbering.prepare_prog` does its job. A generic type is functionally "purer" although dynamic `outvar_loc` may or may not be "cleaner".

WardBrian:
Agreed.

@nhuurre is your intention that this is merged into the branch for #1316 and then that PR is reviewed separately, or is your approval for merging directly into master?

PR #1319, opened 2023-04-25
Add optional `--data-file` for `--debug-generate-inits`

### LINKED ISSUE #1317: Debug flag `--debug-generate-inits` fails if sizes of parameters are based on data
Added in #1306.

Not clear what the correct thing to do is - maybe also generate data at the same time? Just omit them?

nhuurre:
Closes #1317 
There's no real error handling here; if the input data does not conform to the expected type it is ignored silently.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fixed an issue with the `--debug-generate-inits` debugging flag if parameters depended on data for sizes.

## Copyright and Licensing
Copyright holder: Niko Huurre


WardBrian:
I agree that I don't think we should be doing input validation, but it would be nice to improve the error message if you _don't_ supply this file. Even with this PR, the result when the argument is not supplied (or, I suppose, if it is, but the required entries are missing) is still

```
  ("Fatal error: this should never happen. Please file a bug on https://github.com/stan-dev/stanc3/issues/new and include the model that caused this issue."
   "Cannot convert size to number.")

Raised at Base__Error.raise in file "src/error.ml" (inlined), line 8, characters 14-30
Called from Base__Error.raise_s in file "src/error.ml", line 9, characters 19-40
Called from Analysis_and_optimization__Debug_data_generation.unwrap_int_exn in file "src/analysis_and_optimization/Debug_data_generation.ml" (inlined), line 47, characters 38-58
...
```

Since this is all debugging flags, I don't think we need it to be _much_ better, but something like this in place of the current `Fatal_error`

```ocaml
      failwith
        (Fmt.str
           "Unable to evaluate expression '%a' down to a number literal. \
            Consider supplying the --debug-data-file argument to supply the \
            necessary variables "
           Expr.Typed.pp e )
```

Yields the much more informative

```
Uncaught exception:
  
  (Failure
   "Unable to evaluate expression ((N + K)) down to a number literal. Consider supplying the --debug-data-file argument to supply the necessary variables ")

Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from Analysis_and_optimization__Debug_data_generation.unwrap_int_exn in file "src/analysis_and_optimization/Debug_data_generation.ml" (inlined), line 51, characters 38-58
```

while also not requesting a bug report.

I think we also need to catch any errors in the JS frontend and stringify them.

Also (I may have tipped my hand in the earlier example message), how do you feel about renaming this to `--debug-data-file` to keep all of these flags with the same prefix?

Finally, we should add some minimal tests. We don't want the output in the test files since it is RNG based, but some simple cram-style tests which pipe the output to dev/null and simply show the return code would be nice, I think. I can write those later if you don't want to now, they're definitely low priority.

nhuurre:
Good point about fatal errors. I've added some exception-catching. JS frontend should also work now.

src/stanc/stanc.ml
```diff
@@ -82,6 +83,9 @@ let options =
       , Arg.Set generate_inits
       , " For debugging purposes: generate a mock initial value for each \
          parameter" )
+    ; ( "--debug-data-file"
+      , Arg.String (fun s -> data_file := Some s)
+      , " For --debug-generate-data or --debug-generate-inits" )
```
WardBrian, line 88: This doesn't seem to do the "right thing" for `--debug-generate-data` if you give it a partially completed JSON.

Ideally we'd make it so it does, or we'd remove the mention of it from this help string

Easy test:
```stan
data {
  int<lower=0> N;
  int<lower=0> K;
}
```
```json
{ "N":10 }
```

I'd expect the output json to be something like 
```json
{ "N": 10, "K": <any value> }
```
but I got one with `"N" : 4`

Obviously if further values depended on N it would be even more interesting if we got it right here

nhuurre: Whoops, that was silly. Fixed.
I guess this is why it would be good to have tests.

WardBrian: Yeah. With cram tests we could use something like `jq`:

```stan
data {
  int<lower=0> N;
  array[N] real y;
}
```
```json
{ "N": 14 }
```

and the test:
```
 stanc --debug-generate-data debug.stan --debug-data-file partial_data.json | jq -r ".y | length"
14
```

src/stanc/stanc.ml
```diff
@@ -306,13 +310,35 @@ let use_file filename =
...
+      Debug_data_generation.gen_values_json ~filter:true ~data
+        (Ast_to_Mir.gather_declarations typed_ast.parametersblock)
+    with
+    | Ok s -> print_endline s
+    | Error e ->
+        Errors.pp Fmt.stderr ?printed_filename (Errors.DebugDataError e) ;
+        if Option.is_none !data_file then
+          Fmt.pf Fmt.stderr "Supplying a --debug-data-file may help@;" )
+  else if Option.is_some !data_file then
+    Fmt.pf Fmt.stderr "Warning: ignoring --data-file" ;
```
WardBrian, line 341:
```suggestion
    Fmt.pf Fmt.stderr "Warning: ignoring --debug-data-file" ;
```

src/stanc/stanc.ml
```diff
@@ -306,13 +310,35 @@ let use_file filename =
...
+          Debug_data_generation.json_to_mir
+            (Ast_to_Mir.gather_declarations typed_ast.datablock)
+            (Yojson.Basic.from_file file) in
+    match
+      Debug_data_generation.gen_values_json ~filter:true ~data
+        (Ast_to_Mir.gather_declarations typed_ast.parametersblock)
+    with
+    | Ok s -> print_endline s
+    | Error e ->
+        Errors.pp Fmt.stderr ?printed_filename (Errors.DebugDataError e) ;
```
WardBrian, line 337: We should exit with a nonzero code here IMO

WardBrian:
I put together some really minimal tests:
https://github.com/wardbrian/stanc3/tree/tests/debug-data-file/

I decided to use `python` rather than `jq`, since I believe our docker image already has it. If they look good to you, mind merging that branch into yours?

src/stanc/stanc.ml
nhuurre: Should we also exit with zero code in the `Ok` branch? Right now we always generate a `.hpp` file but that seems weird to me...

nhuurre: > similar to how it currently functions with `--auto-format`
Oh, neat. Done.

src/analysis_and_optimization/Debug_data_generation.mli
```diff
@@ -1,5 +1,13 @@
...
 
+val json_to_mir :
+     (Expr.Typed.t SizedType.t * 'a * string) list
+  -> Yojson.Basic.t
+  -> (string, Expr.Typed.t) Map.Poly.t
+
 val gen_values_json :
-     (Expr.Typed.t SizedType.t * Expr.Typed.t Transformation.t * string) list
-  -> string
+     ?filter:bool
```
WardBrian, line 10: Mind adding a brief doc comment for both these functions? Particularly for things like what `filter` does.

Otherwise I'm pretty happy with this PR as it stands. It's a nice fix and improved functionality.

PR #1320, opened 2023-04-26
Handle Lexing.position <-> Location.t conversions in Preprocessor

nhuurre:
Stateful alternative to #1310 .

With this `include_from` locations are cached in Preprocessor and we don't need `Location.of_string_opt` at all.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Copyright and Licensing


src/frontend/Preprocessor.ml
```diff
@@ -5,16 +5,39 @@ open Lexing
...
+let locations_map : (string * Middle.Location.t option) String.Table.t =
+  String.Table.create ()
+
+let reset_locations () = Hashtbl.clear locations_map
+
+let new_file_start_position filename included_from =
+  if Option.is_none included_from then
+    {Lexing.pos_fname= filename; pos_lnum= 1; pos_bol= 0; pos_cnum= 0}
+  else
+    let key = "\u{0}" ^ string_of_int (Hashtbl.length locations_map) in
```
WardBrian, line 22: Why do we need the key to start with NULL?

WardBrian (commented):
This seems much cleaner than my approach, and a bit faster to boot. A few minor questions

src/frontend/Preprocessor.ml
```diff
@@ -5,16 +5,39 @@ open Lexing
...
 let include_paths : string list ref = ref []
 let included_files : string list ref = ref []
 let size () = Stack.length include_stack
 
+let locations_map : (string * Middle.Location.t option) String.Table.t =
+  String.Table.create ()
+
+let reset_locations () = Hashtbl.clear locations_map
+
+let new_file_start_position filename included_from =
```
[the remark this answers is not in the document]
nhuurre: The idea is that `new_file_start_position` is called once per included file; but that logic is fully contained in `Preprocessor`, so yes, the exported version could be combined with `reset_locations`. And I agree `Preprocessor.init` could be called earlier and be part of that as well.

At the start of `Parse.parse` the code initializes `Lexer.comments` and calls `Input_warnings.init`. Do you think those too should be handled by `Preprocessor.init`? (requires moving the relevant mutable state to `Preprocessor` in order to avoid dependency cycles--IMO reasonable for `Lexer.comments` at least)

And then there's the cleanup at the end of `Parse.parse`. The code clears out `Lexer.comments` after it's no longer needed but leaves the global state in `Input_warnings` in whatever state it was at the end of parsing. Neither of these really matter; if `parse` is called again it cleans up the state anyway, and if it's not called then the globals are never touched again so who cares.
Calling `reset_locations` there is pointless too, I now think.

src/frontend/Preprocessor.ml
nhuurre: To distinguish it from  a filename (which never begins with NUL, I hope)

There's a micro-optimization here that I'm not sure is actually worth it. (I didn't profile the code or run benchmarks)
My reasoning is that most Stan code lives in the toplevel file, not the `#include`d ones. In that case there's no `included_from` and it would be better to just store the filename directly in `pos_fname` instead of going through the hash table. So that's what we do! Then we just just need a fast way to recognize which `pos_fname` really are hash table keys and which are plain filenames. The easiest way to do that is to tag all keys with a character that is unlikely to occur in a filename. 

I can add a comment explaining this--or just remove the "toplevel fast path"; while a NUL check is surely faster than a hash table lookup I don't know if the difference is noticeable.

src/frontend/Preprocessor.ml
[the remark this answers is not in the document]
WardBrian: I think the `Input_warnings` state should stay put. I don't have much of an opinion on the comments - it's not really a part of "preprocessing" per se, but the fact that it's so closely tied to the `#include` """comments""" mean I can justify it.

I do think it's worth making the other clean-ups you mention

src/frontend/Preprocessor.ml
WardBrian: I did some minimal profiling of this branch, and then again with this additional logic commented out. 

It does seem to be faster for a model with no `#include`s (168.22K vs 281.29K total cycles spent in the `location_of_position` function) to have this specialized behavior, but not in a huge way (the existing `Location.of_position_opt` function is 175.6K cycles for this model, and the total runtime of the compiler is 93 million cycles)

So far as I can tell, no platform allows a filename to contain NUL, so it's probably fine, but I think a comment would go a long way.

WardBrian: Huh, for some reason `Preprocessor.location_span_of_position` is showing up as expensive in the profiling I'm doing, even though all it does is call `location_of_position` twice and return. 

Odd, but the overall time on `Parse.parse` is essentially the same for a model with no `#include`s and much (up to 10x) faster for a model which is just one big `#include`. I imagine it would be even better if you have `includes` which themselves have further `includes`.

src/frontend/Preprocessor.mli
```diff
@@ -20,6 +25,8 @@ val update_start_positions : Lexing.position -> unit
 val pop_buffer : unit -> Lexing.lexbuf
 (** Pop the buffer at the top of the include stack *)
 
+val comments : Ast.comment_type Queue.t
```
WardBrian, line 28: Maybe this should be two functions `val add_comment: Ast.comment_type -> unit` and `val get_comments: unit -> Ast.comment_type list`?

WardBrian (approved):
This looks good to me. Thanks for the clean up, very nice.

One optional comment, otherwise I think this is good to go. I'm going to wait until after 2.32.1 to merge, since it's not really a bug fix

PR #1326, opened 2023-05-08
Add get_param_sizes function to generated C++

WardBrian:
This PR splits off some small changes from #1100 and adds a new function called `get_param_sizes` with signature:

```C++
std::vector<size_t> get_param_sizes(const bool emit_transformed_parameters__ = true,
                                    const bool emit_generated_quantities__ = true)
```

This function is forward-compatible with all future types we have discussed for Stan (they will always have a size), and makes it possible to re-write things like `random_var_context` to not depend on `get_dims`, which is _not_ forward compatible and assumes rectangular dimensions.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Added code-generation for a new function `get_param_sizes`

## Copyright and Licensing

### DIFF
--- a/src/middle/SizedType.ml
+++ b/src/middle/SizedType.ml
@@ -83,6 +83,21 @@ let rec get_dims st =
       [dim]
   | SArray (t, dim) -> dim :: get_dims t
 
+let rec io_size st =
+  let two = Expr.Helpers.int 2 in
+  match st with
+  | SInt | SReal -> Expr.Helpers.one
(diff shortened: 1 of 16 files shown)


WardBrian:
There is a corresponding C++ branch here: https://github.com/stan-dev/stan/tree/model-base-get-sizes/

This isn't strictly necessary for tuples (this final usage of `get_dims` can be hacked around) so I'm happy to avoid it for the time being.

PR #1327, opened 2023-05-09
Updates the model to have a seperate log prob for reverse mode

SteveBronder:
#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

This adds a `log_prob_impl` specialization specifically for reverse mode. The main reason for making a separate `log_prob_impl` for reverse mode is to make the code gen simpler when dealing with Stuct of Array style matrices and (hopefully) soon matrices that can go on the GPU.

Instead of having to generate a `conditional_var_value_t<...>` type trait for deciding whether we can use SoA inside of `log_prob`, by making a specialization we can just generate a `stan::math::var_value<...>` whenever we see that a matrix is SoA. We will then be able to do a similar thing for matrices tagged with `GPU`

## Copyright and Licensing


SteveBronder:
@WardBrian I'll ping you when this is ready for review

test/integration/good/code-gen/ode/cpp.expected
```diff
@@ -282,7 +282,77 @@ class ode_adjoint_test_model_model final : public model_base_crtp<ode_adjoint_te
   }
   template <bool propto__, bool jacobian__, typename VecR, typename VecI,
             stan::require_vector_like_t<VecR>* = nullptr,
-            stan::require_vector_like_vt<std::is_integral, VecI>* = nullptr>
+            stan::require_vector_like_vt<std::is_integral, VecI>* = nullptr,
+            stan::require_not_st_var<VecR>* = nullptr>
```
SteveBronder, line 286: This is written in a kind of specific way. I'd like this overload to work if we also use a `stan::math::var_value<Eigen::VectorXd>` or `stan::math::var_value<stan::matrix_cl<double>>`. The latter would be nice if the whole model is on the GPU

SteveBronder:
@WardBrian Alright I think this is ready!

Using an optional type for `reverse_mode_log_prob` made for some kind of goofy logic branches down in the code generation. So instead I just start `reverse_mode_log_prob` with an empty list, then in `transform_mir` I assign the transformed `log_prob` to `reverse_mode_log_prob`. Then we just pass it through the optimizations and code gen.

This always produces a reverse mode specialization for log prob, which imo I think is fine and good. Once we have this in there are several smaller optimizations I'd like to look at doing on the lir that I think would be nice. One example is making the parameters typed `auto` instead of hard types as that will let Eigen matrices and vectors of parameters stay as `Eigen::Map<>` types which never need copied by that math library

WardBrian (commented):
A few small questions but this mostly looks good to me.

> Using an optional type for reverse_mode_log_prob made for some kind of goofy logic branches down in the code generation

It's fine if it's not super feasible, but I think there is a (possibly meaningful?) distinction between 'theres nothing here' (`Option.None`) and 'theres something here but its an empty list' so I'd prefer it be an option. By code gen we know it always should be `Some` so that does get an bit annoying

> This always produces a reverse mode specialization for log prob, which imo I think is fine and good

I also think this is fine. It would make it easier to review if it was only for optimizations, but that also makes it more likely for there to be subtle differences we don't notice.

WardBrian:
It's worth grepping around for existing usages of `log_prob` that need to be updated. For example, the `--debug-mem-patterns` flag code:

https://github.com/stan-dev/stanc3/blob/0b1d1cbeb0d4e5cdc1b732bb011340081a3e6a6c/src/analysis_and_optimization/Memory_patterns.ml#L653-L662

WardBrian:
Seems like a copy/paste error led to both overloads now being `require_st_var`

src/analysis_and_optimization/Dependence_analysis.ml
```diff
@@ -231,7 +234,22 @@ let log_prob_build_dep_info_map (mir : Program.Typed.t) :
     Stmt.Fixed.{meta= Location_span.empty; pattern= SList mir.log_prob} in
   build_dep_info_map mir log_prob_stmt
 
+let reverse_mode_log_prob_build_dep_info_map (mir : Program.Typed.t) :
```
WardBrian, line 237: The two new functions here both appear to be dead-code-on-arrival. The version for the normal log prob may also be, since it looks like it is only used in a test.

src/middle/Program.ml
```diff
@@ -120,6 +131,8 @@ let pp pp_e pp_s ppf
   Fmt.cut ppf () ;
   pp_log_prob pp_s ppf log_prob ;
   Fmt.cut ppf () ;
+  pp_log_prob pp_s ppf reverse_mode_log_prob ;
```
WardBrian, line 134:
```suggestion
  pp_reverse_mode_log_prob pp_s ppf reverse_mode_log_prob ;
```

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -718,6 +719,7 @@ let trans_prog (p : Program.Typed.t) =
         p.output_vars in
     { p with
       log_prob= log_prob @ maybe_add_opencl_events_clear
+    ; reverse_mode_log_prob= log_prob @ maybe_add_opencl_events_clear
```
WardBrian, line 722: I think it's worth adding a comment that right here is the first time reverse_mode_log_prob is initialized, hence assigning it to log prob is not a typo/error

WardBrian (commented):
A few small clean ups I missed earlier but then I think this is good to go!

WardBrian (approved):

WardBrian:
Looks good to me. Do you want to squash out any of the debugging commits before a merge? I'll leave it up to you

PR #1330, opened 2023-05-12
Ensure proper cleanup during CI stage

serban-nicusor-toptal:
#### Submission Checklist

- [ ] Run unit tests
- Documentation

## Release notes

Because we do not clean after `VerifyChanges` master is keeping an older state on disk, this PR fixes this behavior.

## Copyright and Licensing

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,4 @@
-@Library('StanUtils')
+@Library('StanUtilsDev')
 import org.stan.Utils
 
 utils = new org.stan.Utils()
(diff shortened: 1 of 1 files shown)


serban-nicusor-toptal:
I think this fixed it @WardBrian the problem was that we never cleaned up after Verify Changes so it was keeping the old files there thus an older version of master.
I will need to apply this fix to Math, Stan, CmdStan too.

WardBrian (approved):
Thanks!

serban-nicusor-toptal:
It's because of trying to git diff with `origin/{hash}` which seems to not work for hashes, I reverted that and we're good!

PR #1332, opened 2023-06-17
Better error for unsupported type in reduce_sum

nhuurre:
#### Submission Checklist

- [x] Run unit tests
- Documentation

`reduce_sum` documentation says the array being sliced can contain any type but that's not really true. If you try to use complex number you get a convoluted error.
```stan
functions {
  real my_func(array[] complex y_slice, int start, int end) {
    return std_normal_lpdf(get_real(y_slice) |) + std_normal_lpdf(get_imag(y_slice) |);
  }
}
parameters {
  array[10] complex a;
}
model {
  target += reduce_sum(my_func, a, 1);
}
```
```
Semantic error in 'example.stan', line 10, column 12 to column 37:
     8:  }
     9:  model {
    10:    target += reduce_sum(my_func, a, 1);
                     ^
    11:  }

Ill-typed arguments supplied to function 'reduce_sum':
(<F1>, array[] complex, int)
where F1 = (array[] complex, int, int) => real
Available signatures:
(<F2>, array[] real, int) => real
where F2 = (array[] real, data int, data int) => real
  The first argument must be
   (array[] real, data int, data int) => real
  but got
   (array[] complex, int, int) => real
  These are not compatible because:
    The types for the first argument are incompatible: one is
     array[] complex
    but the other is
     array[] real
```
Despite the error saying that only `array[] real` is allowed, `reduce_sum` does work with arrays of vectors and matrices as well; it's only the complex number types that are not supported in the math library.

This PR adds a new error case that explains the limitations more precisely.

## Copyright and Licensing
Copyright holder: Niko Huurre

### DIFF
--- a/src/frontend/Typechecker.ml
+++ b/src/frontend/Typechecker.ml
@@ -594,14 +594,8 @@ and check_reduce_sum ~is_cond_dist loc cf tenv id tes =
       mandatory_fun_args UReal (get_arg_types tes) in
   let matching remaining_es fn =
     match fn with
-    | Env.
-        { type_=
-            UnsizedType.UFun
-              (((_, sliced_arg_fun_type) as sliced_arg_fun) :: _, _, _, _) as
-            ftype
-        ; _ }
-      when List.mem Stan_math_signatures.reduce_sum_slice_types
-             sliced_arg_fun_type ~equal:( = ) ->
+    | Env.{type_= UnsizedType.UFun (sliced_arg_fun :: _, _, _, _) as ftype; _}
+      ->
         let mandatory_args = [sliced_arg_fun; (AutoDiffable, UInt)] in
         let mandatory_fun_args =
           [sliced_arg_fun; (DataOnly, UInt); (DataOnly, UInt)] in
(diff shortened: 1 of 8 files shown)


src/frontend/Typechecker.ml
```diff
@@ -612,22 +606,31 @@ and check_reduce_sum ~is_cond_dist loc cf tenv id tes =
...
-        |> error
-    | SignatureErrors (expected_args, err) ->
-        Semantic_error.illtyped_reduce_sum loc id.name
-          (List.map ~f:type_of_expr_typed tes)
-          expected_args err
-        |> error )
+  | {expr= Variable fname; _}
+    :: ({emeta= {type_= slice_type; _}; _} :: _ as remaining_es) -> (
+      let slice_type, n = UnsizedType.unwind_array_type slice_type in
+      if n = 0 then (
```
WardBrian, line 612: Does this implicitly allow larger dimensionalities than before (I assume the C++ doesn't care, just checking)?

WardBrian (approved):
Thanks, this is a nice catch/improvement.


One quick question, and do you mind running `make format` before we merge?

src/frontend/Typechecker.ml
nhuurre: Yes. A real array with 9 dimensions compiled when I tested it. I don't think there's any explicit limits in the C++, it's just templates that can recurse to any depth. Stanc3 limit was probably for `Stan_math_signatures.pretty_print_all_math_sigs`.

PR #1335, opened 2023-07-12
Ensure return statements use local_scalar_t 

### LINKED ISSUE #1334: [BUG] Array literals in `return` statements may generate wrong type in C++
### Current Behavior:

A literal like `{ 0.0 }` is always marked as DataOnly and generated as `std::vector<double>`, even when it is in a return position that requires the type to be autodiff. 

This can lead to the C++ compiler to fail as a function which is meant to return `std::vector<var>` tries to return `std::vector<double>`

### Model Code

```stan
functions {
  array[] real foo(real a) {
    return {0.1};
  }
}
parameters {
  real a;
}
model {
  array[1] real z = foo(a);
}

```

### Environment:

stan 2.32.2

### Anything else:
First noticed on forums: https://discourse.mc-stan.org/t/compilation-error-passes-syntax-check/32059

WardBrian:
Closes #1334.

The solution here was to update the typechecker to promote arrays in the return position to fix this issue. Scalars and Eigen types happily promote from double to var, so it was really only an issue for `std::vector`s.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fixed an issue where returning array literals could produce the wrong types in C++.

## Copyright and Licensing


rok-cesnovar (approved):

SteveBronder (commented):
good!

PR #1339, opened 2023-07-26
--filename-in-msg updates wording of runtime exceptions, too

### LINKED ISSUE #1333: `--filename-in-msg` should also update error strings in generated C++
Many of the downstream interfaces end up using a temporary file for Stan code, which ends up being called something like `model-3414234231232.stan`. It would be nice if we could specify that runtime messages should print `Error in "foo.stan" ...` rather than `Error in "/tmp/zdfs2/model-3414234231232.stan"...`

WardBrian:
Closes #1333 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

`--filename-in-msg` now also affects the name of the file shown in runtime exceptions. Useful for interfaces which use mangled or temporary names during compilation.

## Copyright and Licensing

### DIFF
--- a/src/stan_math_backend/Lower_program.ml
+++ b/src/stan_math_backend/Lower_program.ml
@@ -885,12 +885,13 @@ let new_model_boilerplate prog_name =
       (IfNDef ("USING_R", [GlobalComment "Boilerplate"; new_model; profile_data])
       ) ]
 
-let lower_program (p : Program.Typed.t) : Cpp.program =
+let lower_program ?printed_filename (p : Program.Typed.t) : Cpp.program =
   (* First, do some transformations on the MIR itself before we begin printing it.*)
   let p, s, map_rect_calls = Numbering.prepare_prog p in
   let model_namespace_str = namespace p in
   let model_contents =
-    usings @ Numbering.gen_globals s
+    usings
+    @ Numbering.gen_globals ?printed_filename s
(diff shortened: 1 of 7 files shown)


nhuurre:
Weird corner case:
```stan
// example.stan
model {
  vector[1
  #include part.stan
}
```
```stan
// part.stan
-2] x;
```
generates an error message that displays the original file name
```
(in 'newname.stan', line 3, column 9 to './part.stan', line 2, column 2, included from
'example.stan', line 4, column 2)
```

The fix is easy enough, add `?printed_filename` to line 17 in Location_span.ml, or alternatively (perhaps more consistently) skip the `end_loc_str` entirely if `end_loc.included_from` is not None (equivalent: `end_loc.filename` does not equal `begin_loc.filename`).

WardBrian:
Huh. The compiler definitely doesn't like when `#includes` contain less than a full statement, and can cause weird behavior. Luckily this isn't a common occurence, but I did put together some examples a while back of technically valid programs which crash the pretty printer: https://github.com/WardBrian/stanc3/tree/tests/failing-pretty-printer/test/integration/bad/includes

The example you provide there is another that crashes when you use `--auto-format` without `--canonicalize=includes`. These deserve a better look one day, but I think your proposed solution to the problem in this PR is a good one.

PR #1340, opened 2023-07-28
Consider a return statement in an otherwise infinite loop as 'complete'

WardBrian:
This fixes an issue that @bob-carpenter pointed out in (at least) October 2019 [on the forums](https://discourse.mc-stan.org/t/real-to-integer-conversion/5622/9?u=wardbrian) where a function with an infinite loop which obviously is the only way out of the function still requires a dummy return statement after the loop to typecheck.

The solution here is pretty narrow: loops are considered infinite if and only if they have a nonzero integer literal as their loop expression. If a loop is infinite, we treat what would normally be considered an "incomplete" return as a full one.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Hard-coded infinite loops like `while (1)` which contain a return statement are now considered valid without needing to put a dummy return statement afterwards.

## Copyright and Licensing

### DIFF
--- a/src/frontend/Typechecker.ml
+++ b/src/frontend/Typechecker.ml
@@ -1365,15 +1365,28 @@ and check_while loc cf tenv cond_e loop_body =
   let hardcoded_true e =
     (* heuristic for "will this loop forever" *)
     match e.expr with
-    | Ast.IntNumeral s -> String.count s ~f:(fun c -> c > '0' && c <= '9') >= 1
+    | Ast.IntNumeral s -> String.exists s ~f:(fun c -> c > '0' && c <= '9')
     | _ -> false in
+  let rec could_break s =
+    match s.stmt with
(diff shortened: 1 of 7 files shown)


nhuurre (changes requested):
It's not that simple; breaks statements can exit the loop from the middle so you need to check for those too.
```stan
functions {
  real not_endless() {
    while (1) {
      if (0) return 1.0;
      break;
    }
  }
}
```

WardBrian:
Good catch. That's much more involved to check. Not impossible, but I'm not sure if it's worth it relative to the requirement that you put a dummy return (I think these cases are relatively few and far between, even the original example could have been written in such a way that the loop terminated and _then_ returned)

WardBrian:
I ended up moving the resolution of this into #1341

PR #1341, opened 2023-07-28
Improve error for wrong return type, allow promotion in return position

WardBrian:
While working on #1340, I decided it would be good to separately improve the message for this case. This meant storing the return type when you're inside a function definition, which then also allowed us to clean up the edge case of promotion inside a return. This can be seen as a (much cleaner) alternative to the fix I implemented in #1335 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Improved error message provided when the wrong return type is returned from a function.
Return statements now follow the same type promotion rules as assignment and function argument passing.
The typechecker now allows infinite loops (`while (1)` without an internal `break`) to be the end of a returning function. Previously, an (unreachable) return statement was needed following the loop.

## Copyright and Licensing


src/frontend/Typechecker.ml
```diff
@@ -34,23 +34,28 @@ let attach_warnings x = (x, List.rev !warnings)
...
+  | Returning of UnsizedType.t
+
 (* Record structure holding flags and other markers about context to be
    used for error reporting. *)
 type context_flags_record =
   { current_block: Env.originblock
   ; in_toplevel_decl: bool
-  ; in_fun_def: bool
-  ; in_returning_fun_def: bool
+  ; in_fun: function_indicator
```
nhuurre, line 47: Other `in_*` prefixed fields are of type `bool` (`in_fun_def` reads like "is in a function definition" and expects a yes-or-no answer). Can you name this `fun_return`?

I also wonder if the fields `in_rng_fun_def`, `in_lp_fun_def`, and `in_udf_dist_def` could be replaced with a single field of type `unit Fun_kind.suffix`.

src/frontend/Typechecker.ml
```diff
@@ -1294,7 +1302,9 @@ let returntype_leastupperbound loc rt1 rt2 =
    |ReturnType UInt, ReturnType UReal ->
       UnsizedType.ReturnType UReal
   | _, _ when rt1 = rt2 -> rt2
-  | _ -> Semantic_error.mismatched_return_types loc rt1 rt2 |> error
+  | _ ->
+      (* should no longer be possible with improved return statement type checking *)
+      Semantic_error.mismatched_return_types loc rt1 rt2 |> error
```
nhuurre, line 1307: Now that these checks are redundant, `statement_returntype` does not need to keep track of the type and can be simplified: `AnyReturnType` and `Complete` are equivalent, as are `NoReturnType` and `Incomplete`.

And for #1340 you can define
```ocaml
type statement_returntype =
  | Incomplete of bool
  | Complete
```
where the boolean indicates whether there are any `break`s present.

WardBrian: I went with a 3-part enum rather than adding the bool, 
```ocaml
type statement_returntype =
  | Incomplete
  | UnclearControlFlow (* is any break present *)
  | Complete
```
but otherwise agree this is much nicer!

nhuurre: I wouldn't call it `UnclearControlFlow`.
`Complete` means any possible execution of the statement must exit the function, either through a `return` or `reject`. `Incomplete` means some possible execution passes control flow to the following statement. The third option is more like `NonlocalControlFlow` meaning the control flow can jump out of this statement into some later statement in the same function (i.e. `break` jumps to the end of the loop, wherever that may be. But we don't consider a `continue` to "nonlocal" for the purposes of `while` completeness checking.)

WardBrian: I changed the name and wrote a longer docstring based on your comment

src/frontend/Typechecker.ml
```diff
@@ -1288,57 +1295,20 @@ and list_until_escape xs =
...
-   |Incomplete rt, NoReturnType ->
-      Incomplete rt
-  | Complete rt, AnyReturnType | AnyReturnType, Complete rt -> Complete rt
-  | AnyReturnType, AnyReturnType -> AnyReturnType
+  | Complete, Complete -> Complete
+  | NonlocalControlFlow, _ | _, NonlocalControlFlow -> NonlocalControlFlow
+  | _ -> Incomplete
+
+(* when we exit a loop, the loop's entire return type is either complete or not *)
+let prevent_break_escape = function Complete -> Complete | _ -> Incomplete
```
nhuurre, line 1311:
```suggestion
let compute_loop_statement_returntype = function
  | Complete -> Complete
  | Incomplete | NonlocalControlFlow -> Incomplete
```

src/frontend/Typechecker.ml
```diff
@@ -1357,19 +1327,29 @@ let rec check_if_then_else loc cf tenv pred_e s_true s_false_opt =
...
+    check_statement {cf with loop_depth= cf.loop_depth + 1} tenv loop_body in
+  let te =
     check_expression_of_int_or_real_type cf tenv cond_e
       "Condition in while-loop" in
-  mk_typed_statement
-    ~stmt:(While (te, ts))
-    ~return_type:ts.smeta.return_type ~loc
+  let return_type =
+    match (hardcoded_true te, ts.smeta.return_type) with
+    | true, Incomplete ->
```
nhuurre, line 1347: I think I would prefer
```ocaml
    match ts.smeta.return_type with
    | Incomplete when hardcoded_true te ->
```
or `unconditional_true` (as I mentioned on the other PR).

nhuurre:
> Hard-coded infinite loops like while (1) which contain a return statement are now considered valid without needing to put a dummy return statement afterwards.

Loops don't need return statements in them to be considered endless. More correct statement is that previously a returning function had to end in either `return` or `reject()` but now a `while (1) { .. }` loop without `break`s is also accepted as terminal statement.

src/frontend/Typechecker.ml
WardBrian: That's better, yes.

I prefer `hardcoded_true` to `unconditional_true` since to me it captures more correctly the idea that we're not checking anything clever, e.g. using the partial evaluator or something like that. `(10 %/% 2 - 1)` is also unconditional, but it is not hard coded (IMO)

nhuurre (approved):

PR #1343, opened 2023-08-10
Stop reserved words in parser to prevent worse parse errors later

WardBrian:
See: https://discourse.mc-stan.org/t/ill-formed-phrase-should-be-followed-by-a-statement-variable-declaration-or-expression/32400

This reverts some portions of #771, but rather than using the UNREACHABLE token and updating the messages file, we simply raise the error ourselves directly. 

This unfortunately does make some errors worse, since continuing parsing let us do things like an informative error when a user writes `real foo, int bar;` to try to declare multiple variables of different types on one line. The slightly worse error in that case seems worth it compared to very confusing errors like in the linked thread. 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Improved errors when using a reserved word as a variable name.

## Copyright and Licensing

### DIFF
--- a/src/frontend/parser.messages
+++ b/src/frontend/parser.messages
@@ -2,7 +2,7 @@ program: WHILE
 ##
 ## Concrete syntax: while
 ##
-## Ends in an error in state: 599.
+## Ends in an error in state: 609.
 ##
 ## program' -> . program [ # ]
 ##
@@ -22,7 +22,7 @@ program: DATABLOCK LBRACE CHOLESKYFACTORCORR WHILE
 ##
 ## Concrete syntax: data { cholesky_factor_corr while
 ##
-## Ends in an error in state: 791.
+## Ends in an error in state: 801.
 ##
 ## top_var_type -> CHOLESKYFACTORCORR . LBRACK lhs RBRACK [ WHILE VOID VECTOR UPPER UNITVECTOR TUPLE TRUNCATE TARGET SIMPLEX RPAREN ROWVECTOR RETURN REJECT REAL PROFILE PRINT POSITIVEORDERED PARAMETERSBLOCK ORDERED OFFSET MULTIPLIER MODELBLOCK MATRIX LOWER INT IN IF IDENTIFIER GETLP FUNCTIONBLOCK FOR ELSE DATABLOCK COVMATRIX CORRMATRIX CONTINUE COMPLEXVECTOR COMPLEXROWVECTOR COMPLEXMATRIX COMPLEX COMMA CHOLESKYFACTORCOV CHOLESKYFACTORCORR BREAK ARRAY ]
 ## top_var_type -> CHOLESKYFACTORCORR . LBRACK non_lhs RBRACK [ WHILE VOID VECTOR UPPER UNITVECTOR TUPLE TRUNCATE TARGET SIMPLEX RPAREN ROWVECTOR RETURN REJECT REAL PROFILE PRINT POSITIVEORDERED PARAMETERSBLOCK ORDERED OFFSET MULTIPLIER MODELBLOCK MATRIX LOWER INT IN IF IDENTIFIER GETLP FUNCTIONBLOCK FOR ELSE DATABLOCK COVMATRIX CORRMATRIX CONTINUE COMPLEXVECTOR COMPLEXROWVECTOR COMPLEXMATRIX COMPLEX COMMA CHOLESKYFACTORCOV CHOLESKYFACTORCORR BREAK ARRAY ]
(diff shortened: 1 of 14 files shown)


WardBrian:
Note: as part of preparing this it made me realize that the terms `generated`, `quantities`, and `transformed` are all reserved when they don't in fact need to be. It's perfectly unambiguous to use them as variable names, so we could allow them if we wanted

test/integration/bad/lang/stanc.expected
```diff
@@ -502,16 +502,16 @@ Syntax error in 'bad_cov_exp_quad_vec_rvec_param.stan', line 17, column 0 to col
...
-     3:  }
+     2:    // unfortunately the error here got much worse
+     3:    real x, int y;
+                   ^
+     4:  }
    -------------------------------------------------
 
-Expected a new identifier after comma in declaration.
-All variables declared must be of the same type, and any initializing assignment must follow the identifier before the next comma.
+Identifier 'int' clashes with reserved keyword.
```
nhuurre, line 514: I don't think it got _much_ worse. The old error is pretty bad because it points to the identifier and says "expected a new identifier." At least the new error points to the problematic keyword instead!
I think my above suggested rewording helps here too.

If you want to special-case this error to be even more informative, it might be possible to split the `reserved` error into two cases
```
let reserved id =
  raise (Errors.SyntaxError (Errors.Parsing
          ("Identifier '" ^ id.name ^ "' clashes with reserved keyword.\n",
            id.id_loc)))

let reserved_decl id =
  raise (Errors.SyntaxError (Errors.Parsing
          ("Expected a new identifier after comma in declaration.\nAll variables declared must be of the same type, and any initializing assignment must follow the identifier before the next comma.\n",
            id.id_loc)))
```
and have two rules for `decl_identifier`
```
decl_identifier:
  | id=identifier { id }
  | id=reserved_word { reserved id }

decl_identifier_after_comma:
  | id=identifier { id }
  | id=reserved_word { reserved_decl id }

reserved_word:
  (* Keywords cannot be identifiers but it is nice to
    let them parse as such to provide a better error *)
  | FUNCTIONBLOCK { build_id "functions" $loc }
  | DATABLOCK { build_id "data" $loc }
  | PARAMETERSBLOCK { build_id "parameters" $loc }
  | MODELBLOCK { build_id "model" $loc }
  | RETURN { build_id "return" $loc }
  ...
```

and then use the use the second one here
```
id_and_optional_assignment(rhs):
  | identifier=decl_identifier_after_comma initial_value=optional_assignment(rhs)
    { Ast.{identifier; initial_value} }
```

nhuurre (commented):
> This effectively reverts https://github.com/stan-dev/stanc3/pull/771

No, it doesn't. Postponing keyword/identifier clashes to semantic checking was an incidental change in that PR whose main point to change how error locations are reported in `Parse.ml`

src/frontend/parser.mly
```diff
@@ -16,6 +16,20 @@ let build_id id loc =
...
+let reserved id =
+  raise (Errors.SyntaxError (Errors.Parsing
+          ("Expected a new identifier but found reserved keyword '" ^ id.name ^ "'.\n",
+            id.id_loc)))
+
+let reserved_decl (id, is_type) =
+  if is_type then
+    raise (Errors.SyntaxError (Errors.Parsing
+          ("Found a type ('"^id.name^
+            "') where an identifier was expected.\nAll variables declared in a comma-separated list must be of the same type.\n",
```
nhuurre, line 28: That `^id.name^` needs space and the following line is too long. Since ocamlformat doesn't recognize this `.mly` file I usually copy-paste the modified functions to a random `.ml` file, format that file, and then copy-paste the formatted functions back here.

src/frontend/parser.mly
```diff
@@ -314,10 +332,15 @@ optional_assignment(rhs):
...
+id_and_optional_assignment(rhs, decl):
+  | identifier=decl initial_value=optional_assignment(rhs)
     { Ast.{identifier; initial_value} }
 
+remaining_declarations(rhs):
+  | COMMA decls=separated_nonempty_list(COMMA, id_and_optional_assignment(rhs, decl_identifier_after_comma))
+    { decls }
+
+
 (*
```
nhuurre, line 344: One empty line is enough.

src/frontend/parser.mly
```diff
@@ -193,47 +207,51 @@ future_keyword:
...
-  | COVMATRIX { build_id "cov_matrix" $loc }
-  | PRINT { build_id "print" $loc }
-  | REJECT { build_id "reject" $loc }
-  | TARGET { build_id "target" $loc }
-  | GETLP { build_id "get_lp" $loc }
-  | PROFILE { build_id "profile" $loc }
-  | TUPLE { build_id "tuple" $loc }
+  (* Keywords cannot be identifiers but it is nice to
+    let them parse as such to provide a better error *)
+  | FUNCTIONBLOCK { build_id "functions" $loc, false }
```
nhuurre, line 219: I wonder if `build_id` is worth it when the `Ast.identifier` is destructured immediately. Alternative would just be the tuple `("functions", $loc, false)`.

src/frontend/parser.messages
```diff
@@ -2,7 +2,7 @@ program: WHILE
 ##
 ## Concrete syntax: while
 ##
-## Ends in an error in state: 599.
+## Ends in an error in state: 609.
```
nhuurre, line 5: Definitely not something for this PR but these state number changes make reviewing diffs pretty annoying. And the numbers don't really matter anyway, so maybe we could just write a pre-commit script that deletes these comments.

nhuurre (approved):
Looks good, I left a couple of optional formatting changes.

src/frontend/parser.messages
WardBrian: I find them useful because they’re printed if a parse error occurs when using `—debug-parse` and I can use them to quickly jump to errors (especially if there are duplicate messages or if I’m writing errors for new states in the parser)

nhuurre: I believe some variation of `add_missing_messages.py` script can regenerate them on demand but I forget how all that works. If that would work then I'd like to keep the numbers out of git history.

PR #1345, opened 2023-08-14
Fix: Logic for scalar, data matrix, and autodiff matrices

### LINKED ISSUE #1232: SoA optimization fail?
Using `--O1` the following code fails to use SoA for `beta` and `mu`
```
data {
  int N;
}
parameters {
  real alpha;
  vector[N] beta;
}
model{
  vector[N] mu = alpha + rep_vector(0.0, N) + beta;
  target += sum(mu);
}
```
but the following equivalent code does work (showing only the model block)
```
model{
  vector[N] mu = alpha + (rep_vector(0.0, N) + beta);
  target += sum(mu);
}
```
but it's not just the order of summing as following works, too
```
model{
  vector[N] tmp = rep_vector(0.0, N);
  vector[N] mu = alpha + tmp + beta;
  target += sum(mu);
}
```

@WardBrian commented
>I believe the problem is the formulation `alpha + rep_vector(0.0, N) + beta` leads to the first being a `Plus((AutoDiffable UReal) (DataOnly UVector))`, which does not contain any eigen autodiff variables so the optimization says oh, SoA can not be used

> whereas `alpha + (rep_vector(0.0, N) + beta)` is a `Plus((DataOnly UVector) (AutoDiffable UVector))` which does contain it

> `tmp` is a model-block level variable which means it is always a AutoDiffable

Is there anything to be done to make the first code to use SoA, as figuring out these subtleties can be challenging for users

SteveBronder:
### Summary

Fixes #1232 where some mixes of scalars, data matrices, and autodiff matrices would not be promoted to SoA matrices appropriately. 

For the following code we would not promote to SoA when we should have 

```stan
data {
  int N;
}
parameters {
  real alpha;
  vector[N] beta;
}
model{
  vector[N] mu = alpha + rep_vector(0.0, N) + beta;
  target += sum(mu);
}
```

but for the following model block it does make SoA matrices correctly

```stan
model{
  vector[N] mu = alpha + (rep_vector(0.0, N) + beta);
  target += sum(mu);
}
```

This is because of a bug in the logic for detecting whether we should demote expressions that have a mix of scalars and data matrices. 

The difference between the two model blocks above is that you have, in pseudo mir code for the first has a scalar and data matrix operation inside of the expression

```
assign(mu, add(add(alpha, rep_vector(0.0, N)), beta)
```

while the second, because of the parenthesis, has a scalar and autodiff matrix operation

```
assign(mu, add(alpha, add(rep_vector(0.0, N), beta))
```

The first fails because the optimizer sees a scalar and data matrix operation and fails the whole line. But the second just has a scalar and autodiff matrix operation so it does not fail.

Overall I need to look deeper to understand why I had this particular piece of logic. It was an issue as I was building the optimization, but I forget the exact instance of when this can go badly and need to look into it.

But just checking first in the expression if any side is an autodiff matrix, then checking whether there are scalar / data matrix operations for failing does fix the example model here.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fix logic for deducing whether mixes of scalars, data matrices, and autodiff matrices can be promoted to SoA

## Copyright and Licensing


src/analysis_and_optimization/Memory_patterns.ml
```diff
@@ -268,35 +268,17 @@ and is_any_ad_real_data_matrix_expr_fun (kind : 'a Fun_kind.t)
...
-                 match (x, UnsizedType.contains_eigen_type y) with
-                 | UnsizedType.AutoDiffable, true -> false
-                 | _ -> true )
-               fun_args in
-        match is_args_autodiff_real_data_matrix with
-        | true -> true
-        | false -> List.exists ~f:is_any_ad_real_data_matrix_expr exprs ) )
+        List.for_all
+          ~f:(fun (x, y) ->
+            match (x, y) with
```
WardBrian, line 277:
```suggestion
          ~f:(function
```

it's a bit nicer to use `function` when you immediately match on the arguments

test/integration/good/compiler-optimizations/cppO1.expected
```diff
@@ -1504,15 +1504,14 @@ class ad_levels_deep_model final : public model_base_crtp<ad_levels_deep_model>
     // suppress unused var warning
     (void) function__;
     try {
-      stan::math::var_value<Eigen::Matrix<double,-1,-1>> X_p;
+      Eigen::Matrix<local_scalar_t__,-1,-1> X_p;
```
WardBrian, line 1507: Was it expected that this version is more conservative?

SteveBronder:
Closing this as I figured out a simpler solution

PR #1347, opened 2023-08-16
Fix: Logic for demotion rules around scalar ad types and data matrices

### LINKED ISSUE #1232: SoA optimization fail?

SteveBronder:
### Summary

Fixes #1232 where some mixes of scalars, data matrices, and autodiff matrices would not be promoted to SoA matrices appropriately. 

For the following code we would not promote to SoA when we should have 

```stan
data {
  int N;
}
parameters {
  real alpha;
  vector[N] beta;
}
model{
  vector[N] mu = alpha + rep_vector(0.0, N) + beta;
  target += sum(mu);
}
```

but for the following model block it does make SoA matrices correctly

```stan
model{
  vector[N] mu = alpha + (rep_vector(0.0, N) + beta);
  target += sum(mu);
}
```

This is because of a bug in the logic for detecting whether we should demote expressions when we have a mix of scalars and data matrices. 

The difference between the two model blocks above is that, in pseudo mir code, the first has a scalar and data matrix operation inside of the expression while the second, because of the parenthesis, has a scalar and autodiff matrix operation

```
assign(mu, add(add(alpha, rep_vector(0.0, N)), beta)
```

```
assign(mu, add(alpha, add(rep_vector(0.0, N), beta))
```

The first fails because the optimizer sees a scalar and data matrix operation and fails the whole line. But the second just has a scalar and autodiff matrix operation so it does not fail.

Though we do want the following to fail for trying to promote `p`. In this code, `p` is derived from an autodiff scalar and a data matrix, which we can't promote to SoA.

```stan
data {
  int J;
  array[J] int n;
  vector[J] x;
  array[J] int y;
  real r;
  real R;
}
transformed data {
  vector[J] threshold_angle = asin((R - r) ./ x);
}
parameters {
  real<lower=0> sigma;
}
model {
  vector[J] p = 2 * Phi(threshold_angle / sigma) - 1;
  y ~ binomial(n, p);
}
```

I found a bug in the logic where if the optimizer saw any `op(autodiff scalar, data matrix)` operations before it saw a `autodiff matrix` it would fail the expression and the left hand side would need to be demoted to a AoS.

This PR fixes that by, for each expression with a `autodiff matrix` type, it goes through checking if it is derived only by autodiff scalars and data matrices. If so then we do a second pass to check if there are any autodiff matrices in the expression graph for that statement that are derived outside of the expression. If we find an autodiff matrix that is not derived in the expression then we do not force demotion of the left hand side.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fix logic for deducing whether mixes of scalars, data matrices, and autodiff matrices can be promoted to SoA

## Copyright and Licensing


src/analysis_and_optimization/Memory_patterns.ml
```diff
@@ -195,111 +195,60 @@ and query_initial_demotable_funs (in_loop : bool) (acc : string Set.Poly.t)
...
-     |TupleProjection (expr, _) ->
-        is_any_soa_supported_expr expr
-    | Var (_ : string) | Lit ((_ : Expr.Fixed.Pattern.litType), (_ : string)) ->
-        true
-    | TernaryIf (_, texpr, fexpr) ->
-        is_any_soa_supported_expr texpr && is_any_soa_supported_expr fexpr
-    | EAnd (lhs, rhs) | EOr (lhs, rhs) ->
-        is_any_soa_supported_expr lhs && is_any_soa_supported_expr rhs
+let print_string s = print_endline s
+let print_set s = Set.Poly.iter ~f:print_string s
```
WardBrian, line 199:
```suggestion

```

src/analysis_and_optimization/Memory_patterns.ml
```diff
@@ -195,111 +195,60 @@ and query_initial_demotable_funs (in_loop : bool) (acc : string Set.Poly.t)
...
-  | UserDefined ((_ : string), (_ : bool Fun_kind.suffix)) -> false
-  | CompilerInternal (_ : 'a Internal_fun.t) -> true
-  | Fun_kind.StanLib (name, (_ : bool Fun_kind.suffix), _) -> (
-    match name with
-    | "check_matching_dims" -> true
-    | _ ->
-        is_fun_soa_supported name exprs
-        && List.exists ~f:is_any_soa_supported_expr exprs )
+let print_bool nn bb =
+  Printf.printf "%s: %s \n" nn (if bb then "true" else "false")
```
WardBrian, line 202:
```suggestion

```

src/middle/Stan_math_signatures.ml
```diff
@@ -2166,8 +2166,8 @@ let () =
         (List.range 1 3) )
     bare_types ;
   add_unqualified ("rep_matrix", ReturnType UMatrix, [UReal; UInt; UInt], SoA) ;
-  add_unqualified ("rep_matrix", ReturnType UMatrix, [UVector; UInt], AoS) ;
-  add_unqualified ("rep_matrix", ReturnType UMatrix, [URowVector; UInt], AoS) ;
+  add_unqualified ("rep_matrix", ReturnType UMatrix, [UVector; UInt], SoA) ;
+  add_unqualified ("rep_matrix", ReturnType UMatrix, [URowVector; UInt], SoA) ;
```
WardBrian, line 2170: Mind updating https://github.com/stan-dev/stanc3/blob/master/test/integration/good/function-signatures/math/matrix/rep_matrix.stan?

src/analysis_and_optimization/Memory_patterns.ml
```diff
@@ -195,111 +195,60 @@ and query_initial_demotable_funs (in_loop : bool) (acc : string Set.Poly.t)
...
 
-(**
-  Return true if the rhs expression of an assignment contains only
-   combinations of AutoDiffable Reals and Data Matrices
- *)
-let rec is_any_ad_real_data_matrix_expr
-    Expr.Fixed.{pattern; meta= Expr.Typed.Meta.{adlevel; _}} : bool =
-  if UnsizedType.is_dataonlytype adlevel then false
-  else
+let rec get_lowest_level
```
WardBrian, line 204: I think these functions deserve a docstring (and possibly a better name?)

src/analysis_and_optimization/Memory_patterns.ml
```diff
@@ -336,57 +285,65 @@ let rec query_initial_demotable_stmt (in_loop : bool) (acc : string Set.Poly.t)
...
-              (Set.Poly.union assign_demotes (query_var_eigen_names rhs))
+              (Set.Poly.union rhs_and_idx_demotions (query_var_eigen_names rhs))
+              name
+        | _ -> rhs_and_idx_demotions in
+      let assign_demotions =
+        let is_eigen_stmt = UnsizedType.contains_eigen_type rhs.meta.type_ in
+        if is_eigen_stmt then
+          (* LHS (2)*)
+          let is_rhs_not_promoteable_to_soa =
+            let blah = get_lowest_level rhs in
```
WardBrian, line 317:
```suggestion
            let expression_parts = get_lowest_level rhs in
```

Or any other name! Just not blah

WardBrian (changes requested):
A few (hopefully quick) comments. The overall logic looks good to me

src/analysis_and_optimization/Memory_patterns.ml
SteveBronder: lmao sorry that was from debugging

src/analysis_and_optimization/Memory_patterns.ml
SteveBronder: Just added doc to explain what the recursion is looking for. And gave these better names

SteveBronder:
Think this is ready for review!

WardBrian (approved):
LGTM (assuming tests pass)! Thanks for taking this on

PR #1350, opened 2023-08-21
Use -fsyntax-only in Jenkins

WardBrian:
#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Try to use `-fsyntax-only` argument to Clang to speed up CI during the phases that only check if something actually compiles.

>-fsyntax-only
>
>Run the preprocessor, parser and semantic analysis stages

## Copyright and Licensing


WardBrian:
This shaves ~20-30 minutes off of the CI compared to what we do right now. If we don't actually care about code-generating or linking of the stanc-generated code, it seems like a nice speedup

PR #1351, opened 2023-08-28
Fix: Optimizations crash on mod 0 partial evaluation

WardBrian:
Crash found by fuzzing 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fixed an issue with the partial evaluator crashing on statements containing a mod-by-zero. 

## Copyright and Licensing


nhuurre (approved):
Thanks, looks good.

Speaking of obscure errors in `Partial_evaluator`, I think `apply_operator_int` needs an overflow check. Overflowing arithmetic does not _crash_ stanc3 (Ocaml has 64-bit ints as opposed to Stan's 32-bit ints, and anyway truncates on overflow) but the generated C++ won't compile. (Also not sure if stancjs follows the same rules, checks for overflowing literals caused some trouble #624)

WardBrian:
> 64-bit ints

[63,](https://blog.janestreet.com/what-is-gained-and-lost-with-63-bit-integers/) actually (one of OCaml's weirder defaults).

But yes, sounds like it could be an issue. Can you write down a model that experiences that problem?

nhuurre:
> Can you write down a model that experiences that problem?

```stan
model {
  array[1] int T = {1000000000 * 1000000000};
  print(T);
}
```
Compile with `stanc --O1` and `clang++` stops at
```
error: constant expression evaluates to 1000000000000000000 which cannot be narrowed to type 'int'
```

WardBrian:
That actually seems like it is catching what would be a runtime error at compile time, (with a pretty clear error message, on gcc as well) so I guess it's not the end of the world?

If you compile without optimizations GCC actually warns you anyway:
```
../../ml/stanc3/big.hpp:101:59: warning: integer overflow in expression of type ‘int’ results in ‘-1486618624’ [-Woverflow]
  101 |       stan::model::assign(T, std::vector<int>{(1000000000 * 1000000000)},
```

nhuurre:
Yes, overflows are always runtime errors, just like divide-by-zero. (The partial evaluator does some "runtime work" at compile time so runtime undefined behavior can turn into a compile-time crash.) We should try and catch them in stanc3 and not the C++ compiler.

PR #1353, opened 2023-08-29
[WIP] Expose OpenCL optimization pass

SteveBronder:
#### Submission Checklist

- [ ] Run unit tests
- Documentation

## Summary

Creates a user exposed flag `-fopencl` that will attempt to promote log_prob for reverse mode such that we can run the entire thing on a GPU via OpenCL.

This is about halfway there with a few things to figure out

1. The lower level C++ mir needs to be able to promote vectors and scalars so that they get moved over to the GPU. I think this means that we need to add a tag to arrays and scalars in the mir for a `Mem_pattern.t`. @WardBrian can you think of another way to do this that wouldn't require that? Would just be annoying since we would have to touch a lot of code.

2. The optimization pass is all or none aka we either are able to move the entire `log_prob` over to the GPU or we go back to running only on the CPU. If we are doing this scheme then I think I also need to add a logger so that when a parameter fails we can give the user a reason as to why their model failed to be moved over to the GPU. (UPDATE: done but it's just a writer to stderr)

3. The current scheme right now is 
   a. Do the exact same pass as the SoA optimization using the monotone framework
   b. At the end check the compiler checks whether all the parameters declared in the parameters,  transformed parameters, and model block are able to go on the GPU. If so then we are good and continue, otherwise we stop here and throw an error
   c. If (b) passes then we do another pass over those blocks collecting the names of all of the data used in the model
   d. Take the data names from (c) and in the data section add declarations and assignments of that data over to the GPU using `to_matrix_cl` like in the code below.

    ```c++
    matrix_cl<{TYPE}> {DATA_NAME}_opencl__ = to_matric_cl({DATA_NAME});
    ```

I ripped out all the previous OpenCL code and my goal right now is just to get all of those tests compiling and working correctly. 

4. I think it would be a good idea for now to leave the `target` on the CPU as when we write the scalar from the GPU to CPU that's a stopping point for the async opencl code to know it needs to finish before passing that scalar back.

5. `Mem_pattern.t` now has an `OpenCL` type that indicates a statement or expression can be used on the GPU. For functions, every function in the math library that supports OpenCL also supports the new matrix type, but all functions that support the new matrix type do not support OpenCL. So for the table of available function signatures, if something is tagged `OpenCL` we assume it can support both `SoA` and `OpenCL`. For now I just tagged everything but before we merge and start testing I need to go through the math library and see which functions actually support OpenCL

## Release notes

Allow `-fopencl` that performs a pass on log prob to attempt to promote the model to run on the GPU via OpenCL

## Copyright and Licensing


rok-cesnovar:
Awesome stuff!! Let me know if I can help in any way.

SteveBronder:
Ty! At this point it's mostly just brain storming nice patterns for all this. If you can look at https://github.com/stan-dev/stan/pull/3219 that is a PR we are waiting on before we can merge this

andrjohns:
This is great! Looking forward to this!

src/middle/Mem_pattern.ml
```diff
@@ -11,13 +10,18 @@ open Core_kernel.Poly
...
   | SoA -> Fmt.string ppf "SoA"
+  | OpenCL -> Fmt.string ppf "OpenCL"
+
+let is_soa mem = match mem with SoA -> true | _ -> false
+let is_aos mem = match mem with AoS -> true | _ -> false
+let is_opencl mem = match mem with OpenCL -> true | _ -> false
 
 let lub_mem_pat lst =
-  let find_soa mem_pat = mem_pat = SoA in
+  let find_soa mem_pat = match mem_pat with SoA -> true | _ -> false in
```
WardBrian, line 25: Equivalent to `is_soa` above

src/middle/Mem_pattern.ml
```diff
@@ -11,13 +10,18 @@ open Core_kernel.Poly
...
+type t = AoS | SoA | OpenCL [@@deriving sexp, compare, map, hash, fold, equal]
 
 let pp ppf = function
   | AoS -> Fmt.string ppf "AoS"
   | SoA -> Fmt.string ppf "SoA"
+  | OpenCL -> Fmt.string ppf "OpenCL"
+
+let is_soa mem = match mem with SoA -> true | _ -> false
+let is_aos mem = match mem with AoS -> true | _ -> false
+let is_opencl mem = match mem with OpenCL -> true | _ -> false
```
WardBrian, line 22: There is a relationship between these right? All "OpenCL" memory layouts are automatically SoA, right?

Should `is_soa` take this into account?

src/middle/SizedType.ml
```diff
@@ -229,6 +238,16 @@ let rec has_mem_pattern = function
...
+let is_eigen_type st =
+  match st with
+  | (SVector (mem, _) | SRowVector (mem, _) | SMatrix (mem, _, _))
+    when Mem_pattern.is_opencl mem ->
+      false
+  | SVector _ | SRowVector _ | SMatrix _ | SComplexRowVector _
+   |SComplexVector _ | SComplexMatrix _ ->
+      true
+  | _ -> false
+
```
WardBrian, line 250: I feel a little nervous that `SizedType.is_eigen_type st` and `UnsizedType.is_eigen_type (SizedType.to_unsized st)` would return different things in some circumstances. I suppose it depends on how/where both of them are used, but I'd feel a bit more confident with a `is_eigen_type` which matches Unsized and using `is_eigen st && not (is_opencl st)` where needed.

What do you think?

src/analysis_and_optimization/Memory_patterns.ml
```diff
@@ -103,13 +126,16 @@ let query_stan_math_mem_pattern_support (name : string)
             |> Result.is_ok )
           namematches in
       let is_soa ((_ : UnsizedType.returntype), _, mem) =
-        mem = Mem_pattern.SoA in
+        match requested_mem with
+        | Mem_pattern.SoA -> mem = Mem_pattern.SoA || mem = OpenCL
```
WardBrian, line 130: This seems like one place where the SoA and OpenCL things are considered equivalent, so I wonder if the `is` functions in `mem_pattern.ml` should do the same?

src/middle/Stan_math_signatures.ml
WardBrian, line 1: Are there any functions with are SoA but not OpenCL? It seems like this was a total find-and-replace, but that also seems incorrect? (The math `opencl/` folder is much smaller than the others, no?)

src/stan_math_backend/Cpp.ml
```diff
@@ -385,28 +385,33 @@ module Printing = struct
...
-          pf ppf "stan::math::var_value<Eigen::Matrix<double,%i,%i>>" i j )
-    | Const t -> pf ppf "const %a" pp_type_ t
-    | Ref t -> pf ppf "%a&" pp_type_ t
-    | Pointer t -> pf ppf "%a*" pp_type_ t
+          pf ppf "Eigen::Matrix<%a,%i,%i>" pp_type_ inner_t i j
+      | SoA -> pf ppf "stan::math::var_value<Eigen::Matrix<double,%i,%i>>" i j
+      | OpenCL -> (
+        match inner_t with
+        | Double -> pf ppf "stan::math::matrix_cl<double>"
+        | Int -> pf ppf "stan::math::matrix_cl<int>"
```
WardBrian, line 410: Do we actually generate matrices of ints?

src/stan_math_backend/Transform_Mir.ml
WardBrian, line 1: I'm very glad this can all be removed. There are probably a few more ways to clean this up once the opencl stuff is moved out of here.

One thing worth considering is pass ordering: previously, the opencl stuff was all done _before_ any optimization, now I think it gets done last?

src/stancjs/stancjs.ml
WardBrian, line 1: (Eventually) needs to have the same flag logic as stanc.ml

src/stanc/stanc.ml
```diff
@@ -190,6 +191,9 @@ let options =
     ; ( "-fsoa"
       , Arg.Unit (fun () -> soa_opt := true)
       , "\tTurn on the Struct of Arrays optimization" )
+    ; ( "-fopencl"
+      , Arg.Unit (fun () -> opencl_opt := true)
+      , "\tTurn on OpenCL optimization" )
```
WardBrian, line 196: It's unfortunate that it would disagree with the `-fsoa` flag, but I personally prefer the `--use-opencl` name.

src/stanc/stanc.ml
```diff
@@ -359,7 +360,11 @@ let use_file filename =
         if !no_soa_opt then {base_optims with optimize_soa= false}
         else if !soa_opt then {base_optims with optimize_soa= true}
         else base_optims in
-      Optimize.optimization_suite ~settings:set_optims tx_mir in
+      let set_optims_opencl =
+        if !opencl_opt then
+          {set_optims with optimize_opencl= true; optimize_soa= false}
```
WardBrian, line 365: If `opencl` and `soa` are mutually exclusive, I'd rather enforce this inside the optimization module than requiring the settings to keep it correct (you can just have the soa pass run iff `optimize_soa && not optimize_opencl`)

src/middle/SizedType.ml
```diff
@@ -220,6 +220,15 @@ let modify_sizedtype_mem (mem_pattern : Mem_pattern.t) st =
   match mem_pattern with
   | AoS -> demote_sizedtype_mem st
   | SoA -> promote_sizedtype_mem st
+  | OpenCL -> promote_sizedtype_mem st
```
WardBrian, line 223: OpenCL is "promoted" to SoA by this function?

src/middle/SizedType.ml
```diff
@@ -220,6 +220,15 @@ let modify_sizedtype_mem (mem_pattern : Mem_pattern.t) st =
...
+
+let rec promote_mem (mem_pattern : Mem_pattern.t) st =
+  match st with
+  | SVector (_, dim) -> SVector (mem_pattern, dim)
+  | SRowVector (_, dim) -> SRowVector (mem_pattern, dim)
+  | SMatrix (_, dim1, dim2) -> SMatrix (mem_pattern, dim1, dim2)
+  | SArray (inner_type, dim) -> SArray (promote_mem mem_pattern inner_type, dim)
+  | _ -> st
```
WardBrian, line 231: Does this need to consider things inside tuples?

src/middle/SizedType.ml
```diff
@@ -220,6 +220,15 @@ let modify_sizedtype_mem (mem_pattern : Mem_pattern.t) st =
   match mem_pattern with
   | AoS -> demote_sizedtype_mem st
   | SoA -> promote_sizedtype_mem st
+  | OpenCL -> promote_sizedtype_mem st
+
+let rec promote_mem (mem_pattern : Mem_pattern.t) st =
```
WardBrian, line 225: This seems like it should be called `replace_mem` rather than `promote`

src/middle/Index.ml
```diff
@@ -43,6 +43,13 @@ let apply ~default ~merge op (ind : 'a t) =
   | Between (expr_top, expr_bottom) -> merge (op expr_top) (op expr_bottom)
   | MultiIndex exprs -> op exprs
 
+let map_expr ~f = function
```
WardBrian, line 46: We already derive `map` for the type `t`, so this is equivalent to `Index.map`

src/analysis_and_optimization/Optimize.ml
```diff
@@ -1281,7 +1333,8 @@ let settings_const b =
   ; lazy_code_motion= b
   ; optimize_ad_levels= b
   ; preserve_stability= not b
-  ; optimize_soa= b }
+  ; optimize_soa= b
+  ; optimize_opencl= false }
```
WardBrian, line 1337: I think we'd want this to be enabled as part of `all_optimizations` so this should be `b`

WardBrian (commented):
First round - really more asking questions than anything proper review-y

src/middle/Stan_math_signatures.ml
SteveBronder: Yes for now I just did a find and replace but before we merge I need to go through and strip out what is SoA and what is OpenCL

src/middle/Mem_pattern.ml
SteveBronder: > All "OpenCL" memory layouts are automatically SoA, right?

Yes, but I think we should keep them different. For instance like printing the cpp we probably want to know the difference. I need to think about a right way to describe this

src/middle/SizedType.ml
SteveBronder: Yeah I agree it's very shaky. I think splitting it out it a better idea

src/analysis_and_optimization/Memory_patterns.ml
SteveBronder: For checking the functions `OpenCL` -> `SoA` but `SoA` != `OpenCL`

src/stan_math_backend/Cpp.ml
SteveBronder: Yes for bernoulli for example we will want the array of integers transfered to the GPU. We use OpenCL matrices for that

src/stan_math_backend/Transform_Mir.ml
SteveBronder: Yes not it's the very last thing we do. imo I think that's correct. It's like SOA where we want inlined functions and whatnot

src/stanc/stanc.ml
SteveBronder: That's fine I can switch this back.

It makes sense that it is one or the other since `--use-opencl` should use only the GPU in log_prob. So no chance to use SoA

src/stanc/stanc.ml
SteveBronder: Yes, though do we need to still use this for O1? Maybe not actually

src/middle/SizedType.ml
SteveBronder: If we want to allow OpenCL to use tuples then yes. 

Also the `Decls` having a mem_pattern tag won't work because of tuples either :( I think we do just need to tag all sized types as having a memory pattern

src/middle/Index.ml
SteveBronder: Yeah this needs deleted

src/analysis_and_optimization/Optimize.ml
SteveBronder: idk, personally this is a very specific optimization for a piece of hardware. I'd rather the user explicitly flips it on

src/middle/SizedType.ml
WardBrian: Ah, that’s too bad. You could do something like what we had to do for Autodiff level and have a tuple specific variant in the type, but I wasn’t super happy with that either.

SteveBronder: Yeah eod I think it's going to look a little odd in some places but I think it's fine to just add it to the sized types

PR #1356, opened 2023-09-07
Fix two issues with tuple functions

WardBrian:
1. Accidentally returning a reference
2. Integer arguments failed under certain circumstances, are now templated

These were found because the [Stan tests](https://jenkins.flatironinstitute.org/job/Stan/job/Stan/job/develop/162/pipeline_clang/folder.1170986407/category.1710051484/) were raising a `-Wreturn-stack-address` warning. We should try to catch those types of things in stanc going forward.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fixed a few issues with the C++ generated for functions which accept tuples as arguments. 

## Copyright and Licensing


WardBrian:
IMO this will justify a 2.33.1, but the nice thing is the only thing that would be different between the two versions is the Stanc executables. 

@rok-cesnovar @serban-nicusor-toptal

rok-cesnovar:
Great catch. Do we wait for a few days or just do the patch release?

WardBrian:
@SteveBronder and I still want to take a slightly deeper look at this, so it might be nice to also see if anything else comes in in the next couple days

WardBrian:
Ok I think this is finally ready for review @SteveBronder 

It's a bit tricky but honestly I think I like the code better now than before.

src/frontend/Ast_to_Mir.ml
```diff
@@ -143,10 +143,17 @@ let truncate_dist ud_dists (id : Ast.identifier)
           , Some y ) } in
   let funapp meta kind name args =
     Expr.{Fixed.pattern= FunApp (trans_fn_kind kind name, args); meta} in
+  let ensure_type tp lb : Expr.Typed.t =
```
SteveBronder, line 146: (optional) Should this be like `promote_int_to_double`?

test/integration/good/code-gen/cpp.expected
```diff
@@ -877,20 +901,43 @@ static constexpr std::array<const char*, 11> locations_array__ =
...
           std::vector<
-            std::tuple<std::vector<std::tuple<int, int>>,
-              std::vector<std::vector<int>>>>>& x2, std::ostream* pstream__);
+            std::tuple<std::vector<std::tuple<T1__0__0__, T1__0__1__>>,
+              T1__1__>>>& x2, std::ostream* pstream__);
+/* tuple(int, array[] tuple(int, real)) f(int,
+                                         array[,] tuple(array[] tuple(int,
+  int),
+                                                        array[,] int))
+ */
```
SteveBronder, line 922: (optional) if we can figure out a way to not make this so funky that would be nice

SteveBronder (approved):
Two minor comments but imo looks good!

SteveBronder (approved):
Lgtm!

WardBrian:
@rok-cesnovar @serban-nicusor-toptal I think we should try to do 2.33.1 in the middle of next week, giving a few more days for new issues to come in.

At the moment it would just be the 2.33.0 branches from all the repos except for stanc3

PR #1358, opened 2023-09-11
Add a warning when a variable is reassigned to itself

### LINKED ISSUE #1357: Issue a warning when a variable is assigned to itself
Didn't take long to get a package removal threat from CRAN / Ripley...  this is for ctsem, I guess the only approach is to wait for a new Rstan?

> using R Under development (unstable) (2023-09-08 r85108)
>     using platform: x86_64-pc-linux-gnu
>     R was compiled by
>         clang version 16.0.6
>         GNU Fortran (GCC) 13.2.0
>     running under: Fedora Linux 36 (Workstation Edition)
>     using session charset: UTF-8
>     using option ‘--no-stop-on-test-error’
>     checking for file ‘ctsem/DESCRIPTION’ ... OK
>     checking extension type ... Package
>     this is package ‘ctsem’ version ‘3.8.1’
>     package encoding: UTF-8
>     checking package namespace information ... OK
>     checking package dependencies ... OK
>     checking if this is a source package ... OK
>     checking if there is a namespace ... OK
>     checking for executable files ... OK
>     checking for hidden files and directories ... OK
>     checking for portable file names ... OK
>     checking for sufficient/correct file permissions ... OK
>     checking whether package ‘ctsem’ can be installed ... [41m/100m] WARNING
>     Found the following significant warnings:
>       ./stanExports_ctsm.h:2052:13: warning: explicitly assigning value of variable of type 'local_scalar_t__' (aka 'typename boost::math::tools::promote_args<type-parameter-0-0>::type') to itself [-Wself-assign-overloaded]
>       ./stanExports_ctsmgen.h:2071:13: warning: explicitly assigning value of variable of type 'local_scalar_t__' (aka 'typename boost::math::tools::promote_args<type-parameter-0-0>::type') to itself [-Wself-assign-overloaded]
>     See the [install log](https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-clang/ctsem-00install.html) for details.
>     used C++ compiler: ‘clang version 16.0.6’
>     checking C++ specification ... OK
>       Not all R platforms support C++17
>     checking installed package size ... NOTE
>       installed size is 15.8Mb
>       sub-directories of 1Mb or more:
>         R 1.2Mb
>         data 1.6Mb
>         libs 11.8Mb
>     checking package directory ... OK
>     checking ‘build’ directory ... OK
>     checking DESCRIPTION meta-information ... OK
>     checking top-level files ... OK
>     checking for left-over files ... OK
>     checking index information ... OK
>     checking package subdirectories ... OK
>     checking R files for non-ASCII characters ... OK
>     checking R files for syntax errors ... OK
>     checking whether the package can be loaded ... [15s/21s] OK
>     checking whether the package can be loaded with stated dependencies ... [14s/18s] OK
>     checking whether the package can be unloaded cleanly ... [15s/18s] OK
>     checking whether the namespace can be loaded with stated dependencies ... [13s/15s] OK
>     checking whether the namespace can be unloaded cleanly ... [14s/16s] OK
>     checking loading without being on the library search path ... [14s/17s] OK
>     checking whether startup messages can be suppressed ... [14s/16s] OK
>     checking use of S3 registration ... OK
>     checking dependencies in R code ... NOTE
>     Namespace in Imports field not imported from: ‘rstantools’
>       All declared Imports should be used.
>     checking S3 generic/method consistency ... OK
>     checking replacement functions ... OK
>     checking foreign function calls ... OK
>     checking R code for possible problems ... [124s/134s] OK
>     checking Rd files ... OK
>     checking Rd metadata ... OK
>     checking Rd line widths ... OK
>     checking Rd cross-references ... OK
>     checking for missing documentation entries ... OK
>     checking for code/documentation mismatches ... OK
>     checking Rd \usage sections ... OK
>     checking Rd contents ... OK
>     checking for unstated dependencies in examples ... OK
>     checking contents of ‘data’ directory ... OK
>     checking data for non-ASCII characters ... [14s/14s] OK
>     checking LazyData ... OK
>     checking data for ASCII and uncompressed saves ... OK
>     checking line endings in shell scripts ... OK
>     checking line endings in C/C++/Fortran sources/headers ... OK
>     checking line endings in Makefiles ... OK
>     checking for GNU extensions in Makefiles ... NOTE
>     GNU make is a SystemRequirements.
>     checking for portable use of $(BLAS_LIBS) and $(LAPACK_LIBS) ... OK
>     checking use of PKG_*FLAGS in Makefiles ... OK
>     checking use of SHLIB_OPENMP_*FLAGS in Makefiles ... OK
>     checking pragmas in C/C++ headers and code ... OK
>     checking compilation flags used ... OK
>     checking compiled code ... OK
>     checking sizes of PDF files under ‘inst/doc’ ... OK
>     checking installed files from ‘inst/doc’ ... OK
>     checking files in ‘vignettes’ ... OK
>     checking examples ... [42s/43s] OK
>     checking for unstated dependencies in ‘tests’ ... OK
>     checking tests ... [154s/147s] OK
>       Running ‘testthat.R’ [154s/147s]
>     checking for unstated dependencies in vignettes ... OK
>     checking package vignettes in ‘inst/doc’ ... OK
>     checking re-building of vignette outputs ... [67s/70s] OK
>     checking PDF version of manual ... [10s/11s] OK
>     checking HTML version of manual ... OK
>     checking for non-standard things in the check directory ... OK
>     checking for detritus in the temp directory ... NOTE
>     Found the following files/directories:
>       ‘par-7269706c6579’
>     DONE
> 
> Status: 1 WARNING, 4 NOTEs

WardBrian:
Closes #1357.

This is marked as a draft just because it should not be merged before the 2.33.1 patch release.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

The Stan compiler will now warn you when you reassign a variable to itself.

## Copyright and Licensing


nhuurre:
Some additional test cases.
```stan
model {
  real x = 1.0;
  x = (x); // should warn
  x = -x;
  x = (x > 0) ? x : 0.0;
  tuple(real, real) tup = (1.0, 2.0);
  tup.1 = tup.1; // what about this?
}
```

WardBrian:
@nhuurre I just made some changes that ended up adding a warning for this line in `deprecated.stan`:

```
  idxs[1][:] = idxs[1][:];
```

To be perfectly honest I've never quite wrapped my head around the semantic problem which lead to this being deprecated, but am I correct that this is a case where the lvalue and rvalue are actually _not_ the same?

WardBrian:
On the other hand, nested multi-indexing like that is disallowed as of the most recent release, so maybe it's not a big deal if that warning is slightly wrong?

nhuurre:
The specific case
```stan
  idxs[1][:] = idxs[1][:];
```
is _not_ deprecated, and is an example where the lhs and rhs _are_ equal.
The deprecated construct is
```stan
  idxs[:][1] = idxs[:,1];
```
and the broken variant where lhs and rhs are secretly different is
```stan
  idxs[:][1] = idxs[:][1];
```

WardBrian:
@nhuurre mind giving this a review now that 2.33.1 was released?

test/integration/good/warning/pretty.expected
```diff
@@ -563,6 +563,53 @@ Warning in 'overloaded_forward_decl.stan', line 2, column 6: Functions do not
...
+  foo = (foo);
+  // no warnings
+  foo = -foo;
+  foo = (foo > 0) ? foo : 0.0;
+  
+  tuple(real, real) tup = (1.0, 2.0);
+  // line 22: should warn
+  tup.1 = tup.1;
+  // line 24: should warn (decl)
+  tuple(real, real) tup2 = (tup2.2, tup2.2);
```
nhuurre, line 590:
```suggestion
  // line 24: should warn (decl)
  tuple(real, real) tup2 = (tup2.1, tup2.2);
```
Also let's add a test for this "tuple reconstruction" case outside of declaration. It's similar to line 27 (could warn, currently does not)

WardBrian: I think we should warn if any element of a tuple is accessed during declaration, so even something like this:

`tuple(real, real) x = (0.5, x.1);`

should warn, even though there is no memory address being directly assigned to itself in the end. 

Not during a declaration, I think we’d only want to warn if the same element is being assigned to itself (which this PR as currently written wouldn’t do, but I’m happy to write up the tests)

nhuurre: > we’d only want to warn if the same element is being assigned to itself (which this PR as currently written wouldn’t do

just like the "line 27" test.

> we should warn if any element of a tuple is accessed during declaration

yes, but the test as written confusingly _looks_ like a typo.
In all honesty, I'd rather have `--warn-uninitialized` enabled by default in stead of this special-cased "used while declared" warning.

nhuurre:
I note that the original bug report in #1357 was a C++ warning. As far as I understand, this PR makes that warning redundant by adding a Stan warning but does not actually silence the C++ warning. I think we should also change the C++ backend to
> go one step further and just remove such a statement, since it is semantically meaningless in Stan.

before closing the issue.

test/integration/good/warning/pretty.expected
WardBrian: `--warn-uninitialized` is run in the MIR layers, so in particular it does not produce any output when using something like `--auto-format` which exits before lowering to MIR.

It also looks like it warns when using a higher-order function that the functor argument is uninitialized, which is a bug.

There's some redundancy between the two but I think that's fine if we know this subset is correct

nhuurre: I know `--warn-uninitialized` is broken. I just wish less effort was spent on reimplementing it in AST and more on fixing it.

nhuurre:
The user's code in the original issue would have worked correctly but had a no-op statement that C++ devs have decided is probably an error. In my opinion the warning had negative value because all it did is tell user to fix what isn't broken. Regardless, whether no-op assigns are considered errors in Stan shouldn't be for C++ devs to decide. `stanc` should never emit C++ code that causes further warnings or errors.

WardBrian:
I think the warning has independent value. The fact that this statement is a no-op means it is very likely a programmer error, e.g. you meant to do `some_variable_a = some_variable_b` but accidentally wrote `some_variable_a = some_variable_a`. This will always trivially typecheck, so spotting that you make this typo is not easy!

Even if we make it so the C++ compiler never produces this warning, I think it is still worth alerting to it on the Stan level.

And for decls, I think I would actually be in favor of having any use of the variable being declared on the right hand side be a hard error, but that's obviously a pretty major breaking change (even if the code we were breaking is almost certainly wrong)

nhuurre:
> The fact that this statement is a no-op means it is very likely a programmer error

You're probably right about that, even though it was a false positive in the reported case.

> Even if we make it so the C++ compiler never produces this warning, I think it is still worth alerting to it on the Stan level.

And I think that even if we alert to it on the Stan level, it is still worth making it so the C++ compiler never produces this warning.

I believe adding
```ocaml
  | Assignment ((LVariable x, []), _, {pattern= Var y;_}) when String.equal x y -> [] (* self-assign is a no-op *)
```
to the big match in `Stan_math_backend.Lower_stmt.lower_statement` should be enough to silence the C++ compiler.

test/integration/good/warning/pretty.expected
WardBrian: I tried enabling `--warn-uninitialized` by default and it caused the tests to take _significantly_ longer to run. There are a few obvious improvements we could make to it (like fixing the higher-order function issue) but it also seems to have some more underlying performance issue we'd need to sort out

src/stan_math_backend/Lower_stmt.ml
```diff
@@ -191,6 +191,10 @@ let rec lower_statement Stmt.Fixed.{pattern; meta} : stmt list =
   location
   @
   match pattern with
+  | Assignment (l, _, e)
+    when Expr.Typed.compare (Stmt.Helpers.expr_of_lvalue l ~meta:e.meta) e = 0
```
nhuurre, line 195: [Optional] Technically, it's possible for an index expression to have side effects so you should check with `Analysis_and_optimization.Optimize.cannot_remove_expr` but that's a really far-fetched edge case.

nhuurre (approved):

src/stan_math_backend/Lower_stmt.ml
WardBrian: Oh, it's pretty easy to write a case someone _could_ want to do:

```
generated quantities {
  array[20] real gq;
  for (i in 1:20)
    gq[to_int(uniform_rng(1, 20))] = gq[to_int(uniform_rng(1, 20))];
}
```

I think I'll change it to be more conservative and let the dead code elimination pass in the optimization suite worry about indexed expressions

PR #1360, opened 2023-09-14
Feature: Tuple unpacking

### LINKED ISSUE #1349: [FR] Tuple unpacking
As described in the [design doc](https://github.com/stan-dev/design-docs/blob/master/designs/0018-tuples.md#destructuring-assignment-optional), it would be nice to support tuple unpacking or "destructuring assignment".
Example:
```stan
int x;
int y;
(x, y) = (1, 2)
```
This would be great to support for tuple-returning functions like #1346 

Some things to keep in mind:
```stan
(x,y) = (y,x); // should swap variables gracefully
(a,b) = something_expensive(); // should only call function _once_
```

This can be implemented entirely in the frontend, since it in general desugars to (using the example above)
```
temp_tuple = something_expensive()
a = temp_tuple.1
b = temp_tuple.2
```

WardBrian:
Closes #1349.

Two notes on implementation:
1. For the moment, this is desugared before the MIR layer. This is nice because it keeps the implementation simpler, but a bit annoying because it means we can't use `std::tie` in cases where that would be the fastest option. This can of course be changed later.
2.  Parsing these kinds of values on the left hand side of an `=` proved very difficult. I updated the grammar such that we no longer distinguish (in the BNF grammar of the language) between expressions that can be on the left hand side of an assignment and other expressions. We still make this distinction in the AST by checking it as part of the semantic action for the assignment parse rule. 
  This actually dramatically improved the state of `parser.messages`, since we had duplicate errors for states containing a `lhs` and a (non-lhs) `expression`.
  I think this is probably fine (we should probably go even further in this direction -- see #614). Something very similar was done in [ANSI C](https://www.lysator.liu.se/c/ANSI-C-grammar-y.html#assignment-expression), which is why `++foo = 4` is a type error in C, not a parse error.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Added the ability to unpack a tuple during an assignment. For example, the following is now valid:
```stan
parameters {
  matrix[N, M] A;
}
model {
  matrix[N, M] Q;
  matrix[M, M] R;

  (Q, R) = qr_thin(A); // qr_thin returns a tuple(matrix, matrix)
}
```

## Copyright and Licensing


WardBrian:
Note: there is a further optional part of the design doc pertaining to having `_` as the equivalent of something like C++'s `std::ignore`. I chose not to implement that (for now) since it requires complicating the typechecking (`_` would either need to have a new anything-goes type, or we'd need type information to flow more bidirectionally)

src/frontend/Typechecker.ml
```diff
@@ -976,10 +976,10 @@ let verify_assignment_global loc cf block is_global id =
 (* Until function types are added to the user language, we
    disallow assignments to function values
 *)
-let verify_assignment_non_function loc ut id =
+let verify_assignment_non_function loc ut =
   match ut with
   | UnsizedType.UFun _ | UMathLibraryFunction ->
-      Semantic_error.cannot_assign_function loc ut id.name |> error
+      Semantic_error.cannot_assign_function loc ut |> error
   | _ -> ()
```
nhuurre, line 983: Not good enough.
```stan
functions {
  int foo() {
    int x;
    (foo, x) = (foo, 1);
    return x;
  }
}
```
We might want to make function names always readonly and omit this type-based check.

src/frontend/Typechecker.ml
```diff
@@ -1028,13 +1028,60 @@ let check_assignment_operator loc assop lhs rhs =
...
+    is being unpacked), we want to ensure that the same memory is not being
+    written to multiple times. This lets us preserve the illusion of
+    simultaneous assignment and generally avoid confusion. *)
+let verify_lvalue_unique lv =
+  let rec flatten_lvalues lv =
+    match (lv.lval, lv.lmeta.type_) with
+    | LTuplePacking lvs, _ -> List.concat_map ~f:flatten_lvalues lvs
+    | LIndexed (l, _), _ ->
+        (* Prevent assigning to multiple indices in the same object at once.
+           In principal it is safe to assign to disjoint indices, but statically
```
nhuurre, line 1041:
```suggestion
           In principle it is safe to assign to disjoint indices, but statically
```

src/frontend/Typechecker.ml
```diff
@@ -1028,13 +1028,60 @@ let check_assignment_operator loc assop lhs rhs =
...
+           assigning to any slot in this statement. *)
+        lv
+        :: List.concat_mapi ts ~f:(fun i ty ->
+               flatten_lvalues
+                 { lval= LTupleProjection (lv, i + 1)
+                 ; lmeta= {lv.lmeta with type_= ty} } )
+    | _ -> [lv] in
+  let all_lvals =
+    flatten_lvalues lv |> List.map ~f:Ast.untyped_lvalue_of_typed_lvalue in
+  match List.find_all_dups ~compare:Ast.compare_untyped_lval all_lvals with
```
nhuurre, line 1055: `compare_untyped_lval` is not correct here, e.g.
```stan
int i = 1, j = 1;
array[1] tuple(int, int) tup;
(tup[i].1, tup[j].1) = (1, 2);
(tup, tup[1].1) = ({(1,2)}, 3);
```
I'd erase all array-indexing when flattening the lvalues
```ocaml
  let rec flatten idx lv = function
    match lv.lval with
    | LTuplePacking lvs -> List.concat_map ~f:(flatten []) lvs
    | LIndexed (l,_) -> flatten idx lv
    | LTupleProjection (l, ix) -> flatten (ix :: idx) l
    | LVariable id -> [(id.name, idx)]
```

src/frontend/Typechecker.ml
```diff
@@ -1063,9 +1110,14 @@ let rec check_lvalue cf tenv = function
...
             , flat @ idcs )
+        | {lval= LTuplePacking _; _} ->
+            Common.FatalError.fatal_error_msg
+              [%message
+                "Found a tuple LHS where it should never be"
+                  (lval : Ast.untyped_lval)]
         | {lval= LVariable _ | LTupleProjection _; _} as lval ->
             (*  I think the right thing to do here is treat tuples like variables *)
-            let tval = check_lvalue cf tenv lval in
+            let tval = (check_lvalue cf tenv) lval in
```
nhuurre, line 1120: Seems unnecessary?

test/integration/bad/tuples/bad_unpack6.stan
```diff
@@ -0,0 +1,10 @@
+data {
+  tuple(array[3] int, array[4] int) x;
+}
+model {
+  array[7] real x_flat;
+
+  # we disallow this because it's not obvious if unique
+  # locations are being assigned to
```
nhuurre, line 8:
```suggestion
  // we disallow this because it's not obvious if unique
  // locations are being assigned to
```

nhuurre (changes requested):

src/frontend/Typechecker.ml
WardBrian: Making them readonly would work but doesn't produce as nice an error message. The fix here is easy enough that I think we can leave it as the type-based check

src/frontend/Typechecker.ml
WardBrian: These were tricky but I think it is working now.

One unfortunate result of stripping away this information is we can't then print it, so we have to resort to just printing `tup` rather than `tup.1` (misleading) or `tup[].1` (ideal, but that information is lost) in the errors

WardBrian: Ok, I found a reasonable way of having the error mention `tup[...].1`

nhuurre:
I refactored the code a bit to "make illegal states unrepresentable" (and also makes `_` easier), see https://github.com/nhuurre/stanc3/commit/850392fc26402b0341bb8cd23ca090d8f0e973d7

How do you feel about a more precise uniqueness check that would allow
```stan
vector[5] x = ...;
(x[1], x[2:]) = (x[5], x[:4]);
```
Implemented on https://github.com/nhuurre/stanc3/commit/523c931ab5b20edd61543a40a504443575c023b7

Also, we need to decide what to do about
```stan
int x = 1;
vector[2] y = [1,1]';
(x, y[x]) = (2,2);
```

WardBrian:
I definitely like the illegal states refactor, I'll pull that in. 

The extra precision on the uniqueness check seems nice but also a lot of code for a relative edge case. IMO, since it would be backwards compatible to allow more cases later, I think we should stick to the simpler cases now and see if there is a demand.

That's a nasty test case. I think we'd either want to disallow it or give it the semantics that the values on the left hand side are all what they are _before_ any assignments (such that it would be equivalent to x=2; y[1] = 2, **not** y[2] = 2), but that's the opposite of what currently happens. Disallowing it is probably easier?

nhuurre:
I'm also in favor of disallowing it but note that currently this compiles:
```stan
array[2] int x = {1,2};
x[x[1]] = 2;
```
It's rather silly though, so maybe disallowing that too is fine.

WardBrian:
I think that has the "reasonable" semantics, which is less obvious in the tuples case where the assignments are supposed to be simultaneous

src/frontend/Typechecker.ml
```diff
@@ -976,10 +976,11 @@ let verify_assignment_global loc cf block is_global id =
...
    disallow assignments to function values
 *)
-let verify_assignment_non_function loc ut id =
+let rec verify_assignment_non_function loc ut =
   match ut with
   | UnsizedType.UFun _ | UMathLibraryFunction ->
-      Semantic_error.cannot_assign_function loc ut id.name |> error
+      Semantic_error.cannot_assign_function loc ut |> error
+  | UTuple ts -> List.iter ~f:(verify_assignment_non_function loc) ts
   | _ -> ()
```
nhuurre, line 984: This function does not need to change if it's called from the same place as `verify_assignable_id` (ie. on line 1173). I think the error is a lot nicer with the function name.

src/frontend/Typechecker.ml
```diff
@@ -1001,40 +1002,178 @@ let warn_self_assignment loc lhs rhs =
...
+  |> List.map ~f:Ast.untyped_lvalue_of_typed_lvalue
+
+let variables_accessed_in lv =
+  (* We only care about values being read inside the lvalue here,
+     which means only the expressions inside of LIndexed *)
+  let rec extract_indices lv =
+    match lv.lval with
+    | LVariable _ -> []
+    | LTupleProjection (lv, _) -> extract_indices lv
+    | LIndexed (lv, es) -> extract_indices lv @ es in
```
nhuurre, line 1097: If you're going to call `List.dedup_and_sort` at the end then you could use a `Set` instead of `List` in the first place.
```suggestion
  let rec extract_indices lv =
    match lv.lval with
    | LVariable _ -> String.Set.empty
    | LTupleProjection (lv, _) -> extract_indices lv
    | LIndexed (lv, es) ->
        List.concat_map ~f:exprs_in_index es
        |> List.concat_map ~f:extract_ids
        |> List.map ~f:(fun {name; _} -> name)
        |> String.Set.of_list
        |> Set.union (extract_indices lv) in
```

src/frontend/Typechecker.ml
```diff
@@ -1001,40 +1002,178 @@ let warn_self_assignment loc lhs rhs =
...
+    | LVariable id -> id
+    | LTupleProjection (lv, _) -> extract_id lv
+    | LIndexed (lv, _) -> extract_id lv in
+  let all_variables =
+    List.map ~f:extract_id all_lvals
+    |> List.dedup_and_sort ~compare:Ast.compare_identifier in
+  let accessed_lvals = variables_accessed_in lv in
+  match
+    List.find_all_dups ~compare:Ast.compare_identifier
+      (all_variables @ accessed_lvals)
```
nhuurre, line 1148: If `variables_accessed_in` produces a set like I suggested in the above comment, this simplifies
```suggestion
    List.filter ~f:(Set.mem accessed_lvals) all_variables
```

src/frontend/Typechecker.ml
```diff
@@ -1001,40 +1002,178 @@ let warn_self_assignment loc lhs rhs =
...
+        [(UnsizedType.AutoDiffable, type_of_lvalue lhs); arg_type rhs] in
       let return_type = assignmentoperator_stan_math_return_type op args in
-      match return_type with Some Void -> rhs | _ -> err op |> error )
+      match return_type with
+      | Some Void -> rhs
+      | _ -> err lhs op rhs.emeta.type_ )
+
+let overlapping_lvalues lvals =
+  (* Prevent assigning to multiple indices in the same object at once.
+     In principal it is safe to assign to disjoint indices, but statically
```
nhuurre, line 1055:
```suggestion
     In principle it is safe to assign to disjoint indices, but statically
```

WardBrian: I swear I fixed this already. I must have copy/pasted an older version

nhuurre: The funny thing is _I_ saw it when doing the `lvalue_pack` refactor and then forgot to fix it.

src/frontend/Typechecker.ml
WardBrian: that location (check_lvalue) does not have access to the rhs or its type, but I was able to move it to `check_assignment_operator` with the other typechecks

nhuurre (approved):

WardBrian:
@nhuurre I'm not entirely happy with having a `loc` directly inside the LTuplePack variant, since we don't really have that metadata as a direct part of the type anywhere else. 

I was able to remove it in https://github.com/stan-dev/stanc3/commit/145cc345184a254ab87bdb7f1adb40a1884769f8 by re-constructing the tuple packing location as merging all of the individuals, but that misses the opening and closing `( )`. I was worried this would negatively impact the comment printing (if they were inside those), and it does, but it turns out that the current state of this pr is worse (compiler fails):

```stan
transformed data {
  int x = 3;
  int y = 4;
  print(x, y);
  (/* test */x, y) = (5, 6);
```

Can you think of both a way to fix this error and maybe also remove the `loc` part?

WardBrian:
Current failures are actually due to that loc not being marked as ignore for comparison. Fixing that, the result of formatting 

```
transformed data {
  int x = 3;
  int y = 4;
  print(x, y);
  (x, y) = (5, 6);
  print(x, y);
  (x, y) = (y, x);
  print(x, y);
  (/* comment 1*/x/*comment 2*/,/*comment 3*/y/* comment 4 */)/*comment 5*/ = (y, x);
  (//comment 1
  x //comment 2
  , //comment 3
  y //comment 4
   ) // comment 5
    = (y, x);
}
```

is now

```
transformed data {
  int x = 3;
  int y = 4;
  print(x, y);
  (x, y) = (5, 6);
  print(x, y);
  (x, y) = (y, x);
  print(x, y);
  (/* comment 1*/ x /*comment 2*/, /*comment 3*/ y /* comment 4 */) = (y, x);
  
  /* ^^^:comment 5*/
  (//comment 1
   x //comment 2
   , //comment 3
   y //comment 4
   ) = (y, x);
  
  // ^^^: comment 5
}

```

nhuurre:
I'm not a fan of that `loc` either but it's not possible to reliably reconstruct the location information, especially if we later add a `_` without `loc`. Every expression and statement does have metadata `loc` anyway even if it's usually more indirect.
And I totally forgot it needs a `[@compare.ignore]`.

The only problem with that auto-format result is misplaced comment 5, right? I think adding `=` as a `Separator` in the lexer and printing `=` as if it were a binary operator should fix that; but also, auto-formatting is known to misplace comments in unexpected positions so IMO we could just say that specific placement is not supported.

WardBrian:
Yes, 5 is the only bad one currently. If we remove the loc others get worse (1 and 4)

Fair enough, I think it's fine to leave now that it has the `ignore` set

WardBrian:
I changed LTuplePacking to hold a record `{lvals; loc}` rather than a tuple, so it looks/feels more like the other types in the AST. 

I think this is good for a first implementation, I'm going to look more into `_` soon

nhuurre (approved):

PR #1366, opened 2023-10-03
Update to OCaml 4.14

WardBrian:
This updates to OCaml 4.14, the [last update of the 4.x release line](https://ocaml.org/releases/4.14.0).
It is source-compatible with OCaml 5.0 if someone wishes to compile with that instead.

It also updates the Jane Street libraries to v0.16, which included [migrating from Core_kernel to Core](https://blog.janestreet.com/goodbye-Core_kernel/).

Finally, it also bumped some other dependencies:

- fmt: 0.8.8 -> 0.9.0
- menhir: 20210929 -> 20230608
- yojson: 1.7.0 -> 2.1.0
- ocamlformat: 0.19.0 -> 0.26.1

This final change required some re-formatting, but with proper configuration it was not too bad.

**Existing developers should install and use a new opam switch after this is merged: `cd scripts/; ./setup_dev_env.sh`**

The required updates for building on Windows were done in: https://github.com/ocaml-cross/opam-cross-windows/pull/277 https://github.com/ocaml-cross/opam-cross-windows/pull/278 https://github.com/ocaml-cross/opam-cross-windows/pull/279 https://github.com/ocaml-cross/opam-cross-windows/pull/280 https://github.com/ocaml-cross/opam-cross-windows/pull/286

@serban-nicusor-toptal: this will need new versions of all of our docker images, and a new opam environment on the Jenkins mac. Let me know if you need any help!

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Updated the compiler to use OCaml 4.14 and newer versions of its dependencies.

## Copyright and Licensing


serban-nicusor-toptal:
Hey Brian, find the docker images at:
- stanorg/stanc3:debian-ocaml-4.14
- stanorg/stanc3:debian-windows-ocaml-4.14
- stanorg/stanc3:static-ocaml-4.14

For the multiarch image, one of the deps seems to not be compatible with arm32 & 32bit

```
 > [linux/arm/v7 12/13] RUN curl https://raw.githubusercontent.com/stan-dev/stanc3/update/4.14/scripts/install_build_deps.sh | bash:
100   217  100   217    0     0    837      0 --:--:-- --:--:-- --:--:--   951
#75 1.167 [WARNING] Running as root is not recommended
#75 1.431 [WARNING] Running as root is not recommended
#75 3.667 core is now pinned to version v0.16.1
#75 3.800 [WARNING] Running as root is not recommended
#75 11.46 [ERROR] core = v0.16.1 unmet availability conditions: arch != "arm32" & arch != "x86_32"
```

Our target platforms for multi arch build are: `--platform linux/arm/v6,linux/arm/v7,linux/arm64,linux/ppc64le,linux/mips64le,linux/s390x`
You can find the ci-scripts changes here: https://github.com/stan-dev/ci-scripts/pull/29

Is there a way we can fix that or do we ditch arm support ?

WardBrian:
Can you try core 0.16.0 instead of 0.16.1? It doesn’t look like it has the same restriction in the metadata, but that doesn’t mean it will build necessarily.

WardBrian:
See: https://github.com/ocaml/opam-repository/pull/23918

0.16.1 had this restriction added back, but it should be fine if those platforms use a different minor version

serban-nicusor-toptal:
Nice! I'll rebuild, it might take a while ...

serban-nicusor-toptal:
Looks like I got timeout for downloading so many packages haha, I'll try again in the morning.
Do we update `core v0.16.0` in the stanc3 PR too or leave this image with core v0.16.0 and the others core v0.16.1 ?

WardBrian:
I think the main Linux image and the developers should still be 16.1

serban-nicusor-toptal:
Hmmh, looks like it might not be in the repositories v0.16.0 ?
I'm running into:
```
 > [linux/mips64le 13/19] RUN eval $(opam env) && opam install -y core.v0.16.0:
#81 0.469 [WARNING] Running as root is not recommended
#81 0.564 [WARNING] Running as root is not recommended
#81 194.8 Sorry, no solution found: there seems to be a problem with your request.
#81 194.8
#81 194.8 No solution found, exiting
------
```

Initially I thought it's a timeout but looks like it it's a missing package in the repo I think.

WardBrian:
Two things: was there an `opam update` before that, and is that colon at the end part of the command or just part of the display? (Hopefully the latter)

serban-nicusor-toptal:
I've already tried with `RUN eval $(opam env) && opam update && opam upgrade` before the install package command.
The colon is just a printing artifact, not in the command.
If it matters, I managed to build all arch EXCEPT `linux/mips64le` which fails with the above.

WardBrian:
While I'm debugging the `mips64le` build I decided to check, according to https://tooomm.github.io/github-release-stats/?username=stan-dev&repository=cmdstan, that architecture has only been downloaded 45 times total (across all versions) since we started offering it in 2.28

(for the record, the version we build for google collab has been downloaded 44 times _for just 2.33.1_, and the base cmdstan 2.33.1 ~15 thousand times)

If we need to disable builds on that platform for a little while I think that is OK

WardBrian:
@andrjohns - you originally added the different arch builds for stanc. Was there a specific reason for the selection of archs? mips64el and s390x both seem relatively obscure

andrjohns:
> @andrjohns - you originally added the different arch builds for stanc. Was there a specific reason for the selection of archs? mips64el and s390x both seem relatively obscure

Not particularly, it was just all archs supported by Debian. I figured that would cover the majority of use-cases

WardBrian:
This is now building with mips64el disabled. @serban-nicusor-toptal is there a good way to test build the other binaries we normally only build on `master`?

serban-nicusor-toptal:
One fast way would be to comment out the test stages, the whens for each binary build and run the pipeline. ( this can also be done in the Jenkins UI after hitting Replay on a pipeline without the need to commit here ).
I can trigger it after this one finishes just so we test the binary builds.

WardBrian:
Thanks to some help from some Opam maintainers we determined that the issue with the mips64el architecture comes from the [`num`](https://github.com/ocaml/num) library. The fix has already been merged into `num`, so it's just a matter of re-enabling after it gets released.

SteveBronder (approved):
> The 'mips64el' port is a 64-bit little endian port using the N64 ABI, hardware floating point and the MIPS64R2 ISA

I'm not sure the math library is compatible with the nintendo 64 so I think we can remove this jenkins path (unless??)

WardBrian:
We actually just fixed mips64el! I just have to uncomment the Jenkins and test to make sure, but it should be the same as before now

WardBrian:
Existing developers should install and use a new opam switch after this merge: `git pull; cd scripts/; ./setup_dev_env.sh`

Tagging those that come to mind: @SteveBronder  @rok-cesnovar  @andrjohns  @spinkney

PR #1372, opened 2023-10-27
jenkins: remove old build logs

dylex:
Add jenkins configuration to remove old build logs and artifacts after some time to avoid running out of disk space. Feel free to tweak the specific numbers, as long as we don't keep them forever.


WardBrian (approved):
Thanks

PR #1374, opened 2023-11-03
jenkins: pass appropriate parallel arg to runPerformanceTests.py

dylex:
Currently when jenkins runs `compare-compilers.sh` and `compare-optimizer.sh`, they end up running `runPerformanceTest.py` with no `-j` argument, so the resulting builds end up running `-j80` or something (from number of CPUs), while in fact they only have access to 4 cores (as set with `PARALLEL` in jenkins env).  As such the builds are running quite slowly. This should hopefully help.

## Copyright and Licensing


WardBrian (approved):
Thanks

PR #1400, opened 2024-02-02
Fix ci reply

serban-nicusor-toptal:
#### Submission Checklist

- [ ] Run unit tests
- Documentation

## Release notes

Fix CI reply functionality by removing the use of unstash for binaries builds. Made steps optional by adding additional params.

## Copyright and Licensing

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -79,9 +79,6 @@ pipeline {
         booleanParam(name:"skip_compile", defaultValue: false, description:"Skip compile tests")
         booleanParam(name:"skip_math_func_expr", defaultValue: false, description:"Skip math functions expressions test")
         booleanParam(name:"skip_ocaml_tests", defaultValue: false, description:"Skip ocaml tests")
-        booleanParam(name:"skip_code_formatting", defaultValue: false, description:"Skip code formatting")
-        booleanParam(name:"skip_build", defaultValue: false, description:"Skip code formatting")
-        booleanParam(name:"skip_verify_changes", defaultValue: false, description:"Skip verify changes")
         string(defaultValue: 'develop', name: 'cmdstan_pr',
                description: "CmdStan PR to test against. Will check out this PR in the downstream Stan repo.")
         string(defaultValue: 'develop', name: 'stan_pr',
@@ -106,12 +103,6 @@ pipeline {
     }
     stages {
         stage('Verify changes') {
-            when {
-                beforeAgent true
-                expression {
-                    !params.skip_verify_changes
-                }
-            }
             agent {
                 dockerfile {
                     filename 'scripts/docker/debian/Dockerfile'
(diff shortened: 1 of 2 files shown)


serban-nicusor-toptal:
Question @WardBrian should I replace the use of stash/unstash everywhere or leave it only for the binaries builds so they work on re-runs/custom runs?

WardBrian:
I think you should avoid running `dune subst` in the helper function, because on our mac builder we have a different switch name (`eval $(opam env --switch=stanc-4.14.1 --set-switch)`)

WardBrian:
@serban-nicusor-toptal can you add a `dune subst` step to each builder?

serban-nicusor-toptal:
Oh my bad, I thought we didn't want it. Adding it back now per builder.

Jenkinsfile
```diff
@@ -58,11 +58,30 @@ def runPerformanceTests(String testsPath, String stancFlags = ""){
...
     agent none
     parameters {
         booleanParam(name:"skip_end_to_end", defaultValue: false, description:"Skip end-to-end tests ")
         booleanParam(name:"skip_compile_O1", defaultValue: false, description:"Skip compile tests that run with STANCFLAGS = --O1")
+        booleanParam(name:"skip_compile", defaultValue: false, description:"Skip compile tests")
+        booleanParam(name:"skip_math_func_expr", defaultValue: false, description:"Skip math functions expressions test")
+        booleanParam(name:"skip_ocaml_tests", defaultValue: false, description:"Skip ocaml tests")
+        booleanParam(name:"skip_code_formatting", defaultValue: false, description:"Skip code formatting")
+        booleanParam(name:"skip_build", defaultValue: false, description:"Skip code formatting")
+        booleanParam(name:"skip_verify_changes", defaultValue: false, description:"Skip verify changes")
```
WardBrian, line 84: These 3 are fast and shouldn't be skippable, IMO. Verify changes also seems maybe redundant with the others?

Jenkinsfile
```diff
@@ -827,8 +881,12 @@ pipeline {
                     }
                     steps {
                         dir("${env.WORKSPACE}/linux-armel"){
-                            unstash "Stanc3Setup"
-                            sh "bash -x scripts/build_multiarch_stanc3.sh armel"
+                            cleanCheckout()
+                            sh """
+                                eval \$(opam env)
+                                dune subst
```
WardBrian, line 887: I don't think this will work, because it's outside the building docker instance. Could we do it in `scripts/build_multiarch_stanc3.sh`? Just another thing in the `&&` chain in that file.

serban-nicusor-toptal:
Removed the redundant ones! All binaries seem to build just fine https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FStanc3/detail/PR-1400/5/pipeline

WardBrian:
I realize I said stage, but it can also just be a step of the already existing build stage for linux

WardBrian (approved):
Thanks!

PR #1401, opened 2024-02-12
Include function suffixes when looking for "Did you mean" suggestions

nhuurre:
#### Submission Checklist

- [x] Run unit tests
- Documentation

Since #1024, unknown identifier related errors give a similar identifier as a suggestion but the algorithm only considered typos involving at most three letters. Another likely error is omitting the distribution suffix from a function name:
https://discourse.mc-stan.org/t/how-to-interpret-a-returning-function-was-expected-but-an-undeclared-identifier-was-supplied/34078

## Release notes

When encountering an unknown identifier that matches a known suffixed function without its suffix, add the known function name to the error message.

## Copyright and Licensing

Copyright holder: Niko Huurre

### DIFF
--- a/src/frontend/Environment.ml
+++ b/src/frontend/Environment.ml
@@ -109,5 +109,10 @@ let max_distance = 3
 let nearest_ident env name =
   try
     (* catch any errors in distance and just ignore them, no big deal *)
-    Distance.find_min ~max:max_distance (Map.keys env) name
+    Option.first_some
+      (Distance.find_min ~max:max_distance (Map.keys env) name)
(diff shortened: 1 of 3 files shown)


WardBrian:
This is going to end up being based on the order of the lists of suffixes in Utils, right? E.g. if a lpdf is available, it will never suggest lccdf

That’s already an improvement so I’m happy to merge as is, but how would you feel about giving a list of suggestions?

nhuurre:
If any of them matches then almost all of them are going to match, and I think that's excessive. One option would be to omit `_*cdf`s and choose either `_lpdf/_lpmf` or `_rng` based on the argument types but that adds some complexity.

WardBrian (approved):

PR #1402, opened 2024-02-13
Add `fatal_error()` to the language

### LINKED ISSUE #1393: Add exit() to the language
#### Summary:

Add an `exit()` function to the language. 

We could also call it `error()`, `fatal()`, or `fatal_error()`.

#### Description:

Currently, we have `reject()` in the Stan language. Using `reject()` rejects the current iteration. 

We should have matching `exit()` calls in the Stan language. Using `exit()` should stop algorithms from continuing. This will indicate that there is something wrong and shouldn't be ignored. As opposed to `reject()` which will continue onto the next iteration.
- [ ] Decide on what to throw in the C++ when `exit()` is used
- [ ] If necessary, add a function to call to the math library.
- [ ] Add `exit(...)` to the language to match the function signatures of `reject()`.
- [ ] Update manual with description of behavior of `reject()` and `exit()`.
- [ ] Update algorithms to exit with `exit()` call.
#### Reproducible Steps:

Here's an example:

``` stan
parameters {
  real<lower = 0, upper = 1> theta;
}
model {
  exit();
}
```
#### Current Output:

Currently, this will show that the 100 tries for initialization failed.
#### Expected Output:

Exit the algorithm.
#### Current Version:

v2.11.0

WardBrian:
Closes #1393 .

Note that there is no way to parse this and also allow variables named `exit`, so this would immediately reserve the keyword. Based on a github search, I found only one package which uses `exit` as a variable name. If this is a serious concern we could consider different names like `stop` or something like `fatal_error` which is almost certainly unused.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Added `fatal_error()`, which is similar to `reject()` but rather than signaling an error which is recoverable, signals an error which should terminate execution immediately. `fatal_error` is now a reserved word. 

## Copyright and Licensing


nhuurre:
I prefer `fatal_error` over `exit`. Note that in C++, `std::exit(0)` indicates a successful program completion.

bob-carpenter:
Good point.  `fatal_error` sounds more descriptive and will match the logging statement that comes out as an error.

src/middle/Expr.ml
```diff
@@ -221,7 +222,7 @@ module Helpers = struct
       | Var _ | TupleProjection _ -> Fixed.Pattern.Indexed (e, [i])
       | Indexed (e, indices) -> Indexed (e, indices @ [i])
       | _ ->
-          Common.FatalError.fatal_error_msg
+          Common.ICE.internal_compiler_error
```
nhuurre, line 225:
```suggestion
          ICE.internal_compiler_error
```
There's `open Common` at the start of this file and the previous `ICE.internal_compiler_error` omits `Common.` prefix so I'd like to be consistent.

src/middle/Expr.ml
```diff
@@ -234,7 +235,7 @@ module Helpers = struct
       match Typed.(type_of e) with
       | UTuple ts -> List.nth_exn ts (i - 1)
       | t ->
-          Common.FatalError.fatal_error_msg
+          Common.ICE.internal_compiler_error
```
nhuurre, line 238: Same as above.

src/frontend/Typechecker.ml
```diff
@@ -1372,19 +1372,23 @@ let check_reject loc cf tenv ps =
   let tps = List.map ~f:(check_printable cf tenv) ps in
   mk_typed_statement ~stmt:(Reject tps) ~return_type:Complete ~loc
 
+let check_exit loc cf tenv ps =
```
nhuurre, line 1375:
```suggestion
let check_fatal loc cf tenv ps =
```

nhuurre (approved):
Looks good, a couple of optional suggestions.

PR #1407, opened 2024-02-15
Experimental support for unicode identifiers.

WardBrian:
I know for a fact that this requires a few changes in `stan-dev/stan`'s json data handler to recognize unicode names, which is just one of several reasons this is a draft.

The basic overview:
OCaml strings should be treated mostly like arrays of bytes, and `ocamllex` handles inputs as sets of bytes. We can define rules that recognize UTF-8-compatible bytes, and then do validation on them after the fact based on the the [Unicode Annex 31: Unicode Identifiers](https://www.unicode.org/reports/tr31/) standard. 

We then pretend for most of the compiler like it's just bytes, which is fine, because we never do things like subslice variable names.

Finally, at output time, we already had string escaping (since #952), so most of the code-gen works fine. Recent C++ standards require that compilers support UTF-8 names based on the same UAX31 rules linked above, but older ones may not. For now I've got it generating "Universal character names" which seem like the legacy version of this, which hopefully means older compilers will be happy with it.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    
## Release notes

stanc3 can now accept a flag `--allow-unicode` which enables the use of non-ascii characters in Stan files. All files are expected to be encoded in UTF-8. 
This is experimental and may not work with older C++ compilers.

## Copyright and Licensing


bob-carpenter:
Isn't there sub slicing when we peel off `_lpdf` suffixes in sampling statements?

WardBrian:
> Isn't there sub slicing when we peel off `_lpdf` suffixes in sampling statements?

1. Not quite, since we generally go _from_ something without a suffix (`Y ~ foo(...)`) to the thing with the suffix (`target += foo_lpdf(Y | ...)`), so even there it is a concatenation problem more than a subslicing.
2. Even in situations where we do need to do it, I think that should be fine, since the cut point is happening somewhere we _require_ ASCII characters to appear. So even a `θ_lpdf`, the cut point in the string of bytes would still just be 5 bytes from the end

But, that is definitely one area of this that would need much more testing before it could be merged.

WardBrian:
There is some prior art now, in the OCaml compiler itself:
https://github.com/ocaml/ocaml/pull/12664

They define an explicit set of characters they allow, which sidesteps a lot of the issues here (no need for things like UUSeg, etc): https://github.com/ocaml/ocaml/blob/6c298db0e356d0e04dd45acf6684f693f8baa7db/utils/misc.ml#L265-L272

PR #1409, opened 2024-02-27
[CI] Fix multiarch builds

serban-nicusor-toptal:
#### Submission Checklist

- [ ] Run unit tests
- Documentation

## Release notes

Add an optional stage to build the multiarch with a custom docker tag.
Add a variable to hold the docker tag used for multiarch builds, so developers can change it when making changes to the docker image in a PR.
Eliminate the usage of `--privileged` for the multiarch builds themselves.
Lock multiarch build & execution on `linux && triqs` agent to ensure permissions are correct across runs.

## Copyright and Licensing


scripts/docker/multiarch/Dockerfile
```diff
@@ -1,18 +1,31 @@
 # Base image
 FROM debian:buster-20220622-slim
 
-ARG STANC3_BRANCH=update/4.14
+ARG STANC3_BRANCH=master
```
WardBrian, line 4: I think pulling from master (and pushing to `latest`) makes this not super useful because then we cant use it to test something new on a branch without pushing to a place that everywhere now pulls from.

Could we have it pull from the PR branch, and push to a user-defined name or something like that?

serban-nicusor-toptal: I see what you mean, in case you change the build deps during your PR.
``` 
RUN curl https://raw.githubusercontent.com/${STANC3_ORG_ENV}/stanc3/${STANC3_BRANCH_ENV}/scripts/install_build_deps.sh | bash
```

We can have this script parametrized to receive a stanc3 branch (won't work with forks?) and pull the script from that branch then build this image specifically for that branch.
One downside of this is that it will take ~45 minutes to build. (only needed when during a PR we want to build the multiarch).
And I think for master we can have it build on every merge to master and publish to :latest and then master will pull that for the release builds.

What do you think ?

WardBrian: I think not working on forks is probably a good thing for this kind of functionality. Users without perms won't be able to request to build it, either. 

I don't think we need to build on every merge to master. If a PR publishes to a specific place (e.g, I publish to a tag `with-unicode` for #1407), it will have to update the usage sites to be that specific place. When that PR is merged, `master` will keep pulling from there until it changes the next time.

serban-nicusor-toptal: I'm not sure I got the last part right.
So #1407 in this case will publish docker image `:pr-1407` to dockerhub so the PR steps can pull that and do the builds.
When this is merged in master, we want to use `:pr-1407` rather than build a new one ?

WardBrian: Basically, though I'd prefer if we could give it a name like `:update-menhir` or whatever the actual change in the dependencies was. 

This is basically what we've done before, like how at the moment the image is called `:ocaml-4.14'

serban-nicusor-toptal: Before I built the image manually so I gave it that tag.
Now I want to automate this as part of this PR so we need a way to automatically deduct the tag, the best would be from the pr branch name, in this case `:feature/unicode-identifiers` (which might not work because of the `/`).
The PR will build the image and when merged into master, I will read the source branch of the merge commit and deduct the tag master needs to pull.

WardBrian: Why does the tag need to be automatic? Instead of a check box asking if we should build it or not, could it be a text box for the name?

serban-nicusor-toptal: That could work for the first part with the PR but then I'm not sure how on master I would know what's the chosen tag so I could pull that image.

serban-nicusor-toptal: I could try to get the merge message
image
extract the PR-XXX and then call Jenkins api to get the parameter from the last successful pipeline there (or iterate through all in case it's not the latest build)

serban-nicusor-toptal: This is so we can avoid master to rebuild the same image again on merge, to reuse the one build in that pr. We also need to remember somehow this tag for the next merges that do not come with a change in the multiarch.

WardBrian: > I'm not sure how on master I would know what's the chosen tag so I could pull that image.

The PR author is responsible for updating the jenkinsfile so it does

serban-nicusor-toptal: Okay then that makes it much more simple.
I'll add a parameter for the tag, and separate the build of the image from the tag itself, and make the tag static in a variable.
When working on a PR like this, we'll build the image, change the reference tag manually, test and merge to master.
Master will reference the correct tag and everything should work smoothly.
If that sounds good, I'll make the changes accordingly.

serban-nicusor-toptal:
All changes are here I think, I'm now moving into testing this to make sure everything still works as expected.

serban-nicusor-toptal:
Hey @WardBrian this is ready for a review.
I ran 34 and 35 to make sure it build and the second run won't get permission errors.
https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FStanc3/detail/PR-1409/34/pipeline
https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FStanc3/detail/PR-1409/35/pipeline

Feel free to give it a few test runs. I've updated the description with the changes too.

WardBrian (commented):
Thanks! Are there links to runs of each of these steps working?

serban-nicusor-toptal:
@WardBrian this is ready now, see my comment above with the two jenkins urls, those are successful builds, one after another to ensure we do not run into the same error as before.

WardBrian:
Those links both seem to go to builds where the `Build and push multiarch docker…` is grayed out

serban-nicusor-toptal:
Oh my bad, I thought you were looking for the actual multiarch builds based on this image.
See: https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FStanc3/detail/PR-1409/29/pipeline

scripts/docker/builder/Dockerfile
WardBrian, line 1: Is the main difference between this and `docker/static` the lack of `Skopeo`? Could we re-use the same image there?

As an aside, I don't think `static` needs opam or anything like that

serban-nicusor-toptal: We use static for binaries https://github.com/stan-dev/stanc3/blob/master/Jenkinsfile#L661
This builder Dockerfile is just to do docker-in-docker and make sure permissions align.

WardBrian (approved):

PR #1413, opened 2024-03-25
Add a special-case error message for old array syntax

### LINKED ISSUE #1412: Iinformative error message for expired array syntax
We're still seeing some confusion from users when they try to run Stan code with the old array syntax, since the error message references a missing semicolon instead of the syntax usage ([recent example](https://discourse.mc-stan.org/t/expected-after-variable-declaration/34571/4)).

Can the error message be updated to specify that the array syntax is the problem (and possibly include a link to the doc so they know what to change)?

WardBrian:
Closes #1412. Surprisingly, this doesn't introduce any conflicts in the parser. If it ever does, we could remove it again, but for now it seems like a good quality of life feature.

One downside is printing a nice error message introduced a circular dependency between pretty printing and parsing. I resolved this by splitting the pretty printing module in two, but I'm not super happy with that.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Added a better error message when a program attempts to use the removed array syntax.

## Copyright and Licensing


WardBrian:
Examples of the new message:

```
Syntax error in 'declarations.stan', line 3, column 18 to column 18, parsing error:
     1:  data {
     2:    int a0;
     3:    int<lower=0> a1[1];
                           ^
     4:    int a2[2,3];
     5:    array[1] int a3;

";" expected after variable declaration.
It looks like you are trying to use the old array syntax.
Please use the new syntax:
array[...] int<lower=0> a1;
```

```
Syntax error in 'unsized-old-array.stan', line 2, column 18 to column 18, parsing error:
     1:  functions{
     2:    real foo(  real[,] v){
                           ^
     3:      return sum(v);
     4:    }

An identifier is expected after the type as a function argument name.
It looks like you are trying to use the old array syntax.
Please use the new syntax: 
array[...] real
```

src/frontend/parser.mly
```diff
@@ -375,6 +389,18 @@ remaining_declarations(rhs):
...
 decl(type_rule, rhs):
+  (* This is just a helper for the fact that we don't support the old array syntax any more *)
+  | ty=type_rule id=decl_identifier LBRACK {
+    raise
+    (Errors.SyntaxError
+       (Errors.Parsing
+          ( Fmt.str
+              "\";\" expected after variable declaration.@ It looks like you \
+               are trying to use the old array syntax.@ Please use the new \
+               syntax:@ @[<h>array[...] %a %s;@]@\n"
```
nhuurre, line 400: I don't think this needs the generic "expected semicolon" part since the specific error is identified quite precisely.
Also, it wouldn't be too difficult to print the exact array dimensions instead of `[...]`:
```suggestion
  | ty=type_rule id=decl_identifier LBRACK dims=separated_nonempty_list(COMMA, expression) RBRACK {
    let (ty, trans) = ty in
    let ty = List.fold_right ~f:(fun e ty -> SizedType.SArray (ty, e)) ~init:ty dims in
    let ty = (ty, trans) in
    raise
    (Errors.SyntaxError
       (Errors.Parsing
          ( Fmt.str
              "Deprecated array syntax.@ Please use the new \
               syntax:@ @[<h>%a %s;@]@\n"
```
And actually, since this doesn't cause parsing ambiguities, maybe the canonicalizer could just accept the old array syntax.

WardBrian: > I don't think this needs the generic "expected semicolon" part since the specific error is identified quite precisely

I left it in as a guard for there being a genuine mistake where the user didn't want an array at all but had a stray character which happened to be a `[`. It seemed odd that any other special character there would mention a `;`, but this wouldn't.

> maybe the canonicalizer could just accept the old array syntax.

It would need to be a bit more complicated; if you wanted to really allow that you'd also need to support the optional assignment (essentially go back to the parsing rule pre-#1388). I don't think this would cause any conflicts at the moment either.

I don't want to do that, mainly because as long as the canonicalizer would support it, some interfaces would just update their pipeline to `canonicalize -> stanc -> c++ compile` and therefore never actually update the Stan code. That would "work" unless/until we actually wanted to do something syntactically which was incompatible, at which point we're either back to where we are now, or our hands remain tied and the deprecation policy was not able to achieve the main goal of allowing language evolution.

nhuurre: > a genuine mistake where the user didn't want an array at all but had a stray character

That seems unlikely, and I think that as long as the caret points to the errant bracket, the user can spot the mistake easily even if the error message is nonsense.
Also, this problem disappears if you go with my suggestion of parsing the whole dimension expression.

> some interfaces would just update their pipeline to canonicalize -> stanc -> c++ compile and therefore never actually update the Stan code. 

Is this a hypothetical, or do you know some interface that opposes our deprecation policy?

WardBrian: I know BRMS was using that set up until 2.34 forced them not to. I do not know of any others concretely, but I would be entirely unsurprised if some other downstream users of RStan have done this

I made the change to give a complete suggestion. It introduced a few spurious error cases in the messages file for situations where you have `int foo[` followed by an invalid expression - I gave them all a generic message in that case.

test/stancjs/stancjs.expected
```diff
@@ -246,29 +246,35 @@ $ node pedantic.js
...
 $ node removed.js
-Syntax error in 'string', line 3, column 7 to column 8, parsing error:
+Syntax error in 'string', line 3, column 8 to column 10, parsing error:
    -------------------------------------------------
      1:  
      2:  parameters {
      3:  	real y[10];
-                ^
+                 ^
      4:  }
```
nhuurre, line 255: note: these look off because a tab counts as one space when positioning the caret. Presumably this has always been an issue. Tabs are rarely used so maybe not worth fixing.

nhuurre (approved):

PR #1416, opened 2024-03-25
Improve internal error handling and message

### LINKED ISSUE #1408: Improve message and handling of internal compiler errors
We may want to install a top-level exception handler which catches all exceptions, manually prints them and the backtrace, and then outputs the message `Fatal error: this should never happen. Please file a bug...`. If we did this, we wouldn't even need the `Common.FatalError` module, just using `raise_s` everywhere we need would work.

I believe this also would improve the javascript parser, since currently there internal compiler errors manifest as exceptions rather than populating the `.error` field of the return object

WardBrian:
This closes #1408. There is no real difference in the end result for the binary executable, but this is a huge improvement for stanc3js since internal errors will be cleanly converted to the error string format rather than percolating as javascript exceptions.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

The compiler should display a cleaner error if it encounters an internal error.

## Copyright and Licensing

### DIFF
--- a/src/common/ICE.ml
+++ b/src/common/ICE.ml
@@ -18,7 +18,7 @@ let with_exn_message f =
       else "Backtrace missing." in
     Error
       (Fmt.str
-         "Fatal error:@ @[%a@]@\n\
+         "Internal compiler error:@ @[%a@]@\n\
           %s@\n\
           @\n\
           This should never happen. Please file a bug at \
--- a/src/common/ICE.ml
+++ b/src/common/ICE.ml
@@ -2,13 +2,24 @@
 
 open Core
 
-(** Equivalent to [raise_s] but prepends a stanc specific
-  message asking users to report a bug *)
-let internal_compiler_error message =
-  let augmented =
-    Sexplib0.Sexp.List
-      [ [%message
-          "Fatal error: this should never happen. Please file a bug on \
-           https://github.com/stan-dev/stanc3/issues/new and include the model \
-           that caused this issue."]; message ] in
-  raise_s augmented
+(** An alias of [Core.raise_s]. This used to do more
+    processing, for now it is preserved just as a nicer
+    marker in the code *)
+let internal_compiler_error = raise_s
+
+(** Catch all exceptions at the top-level and convert them into a ['a, string result]
+    where the string contains the exception and a backtrace if present, followed
+    by a link to our bugtracker. *)
+let with_exn_message f =
+  try Ok (f ())
+  with e ->
+    let bt = Printexc.get_backtrace () in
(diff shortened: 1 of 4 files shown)


nhuurre:
Great, this looks good.

My only concern is that there's no way to test it!

How do feel about, for example, adding command line option `--internal-compiler-error-test` (disabled in release builds) that would make statements like
```stan
fatal_error("InternalCompilerError", "parsing");
fatal_error("InternalCompilerError", "typecheck");
fatal_error("InternalCompilerError", "lowering");
```
trigger ICE in various compilation phases?

WardBrian:
I added some code unit tests. End to end tests are tricky because of the idea of the command line option being "(disabled in release builds) " -- OCaml intentionally avoids having easy support for conditional compilation

src/common/ICE.ml
```diff
@@ -2,13 +2,26 @@
...
+    by a link to our bugtracker. *)
+let with_exn_message f =
+  try Ok (f ())
+  with e ->
+    let bt =
+      if Printexc.backtrace_status () then Printexc.get_backtrace ()
+      else "Backtrace missing." in
+    Error
+      (Fmt.str
+         "Fatal error:@ @[%a@]@\n\
```
nhuurre, line 21:
```suggestion
         "Internal compiler error:@ @[%a@]@\n\
```

test/unit/Error_tests.ml
```diff
@@ -0,0 +1,86 @@
...
+                                        ; id_loc= Middle.Location_span.empty }
+                                  ; initial_value=
+                                      Some
+                                        { expr=
+                                            Promotion
+                                              ( { expr= IntNumeral "1"
+                                                ; emeta=
+                                                    { loc=
+                                                        Middle.Location_span
+                                                        .empty } }
```
nhuurre, line 65: This is getting a bit awkward. You could add
```ocaml
let loc = Middle.Location_span.empty in
```
and then make these
```ocaml
emeta={loc}
```

nhuurre (approved):
> conditional compilation

Oh, I didn't even think of that! I would have just made [`let options` in `stanc.ml`](https://github.com/stan-dev/stanc3/blob/282b0946d10b469413a8d477c43fe8d62b1af22c/src/stanc/stanc.ml#L64) something like
```ocaml
if String.is_prefix ~prefix:"%%" version then
  Arg.align (opts @ [("--ice-test", Arg.Set ice_test, "ICE test")])
else
  Arg.align opts
```
Anyway, I think end-to-end tests would be overkill, `let%expect_test` is quite enough.

Issue #1418, opened 2024-03-29
CI: Switch from the ghr script to the github CLI


WardBrian:
We currently use https://github.com/tcnksm/ghr during the 'upload binaries' stage in Jenkins. 

I think it is worth switching to the official GitHub CLI `gh`. The `gh release` subcommand seems to cover all we need: https://cli.github.com/manual/gh_release

PR #1419, opened 2024-04-04
[CI] Switch to gh binary for releases

serban-nicusor-toptal:
#### Submission Checklist

- [ ] Run unit tests
- Documentation

## Release notes

Requested in #1418, this will replace the `ghr` binary with the official `gh` binary from Github.

## Copyright and Licensing

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1024,7 +1024,7 @@ pipeline {
 
                         runShell("""
                             gh release delete ${tagName()} --cleanup-tag -y || true
-                            gh release create ${tagName()} --latest --target master --notes "$(git log --pretty=format:'nightly: %h %s' -n 1)" ./bin/*
+                            gh release create ${tagName()} --latest --target master --notes "\$(git log --pretty=format:'nightly: %h %s' -n 1)" ./bin/*
                         """)
                     }
                 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1023,9 +1023,7 @@ pipeline {
                         unstash 'js-exe'
 
                         runShell("""
-                            wget https://github.com/tcnksm/ghr/releases/download/v0.12.1/ghr_v0.12.1_linux_amd64.tar.gz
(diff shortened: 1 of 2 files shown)


serban-nicusor-toptal:
I've added `gh` in the publish Dockerfile and replaced the `ghr` command, in theory, it should all work smoothly. I think the easiest way to test this is to merge it into master and stay sharp in case anything goes wrong so I can fix it.
If you don't mind Brian to take a look at the flags I added, in case we want to add/remove any. Thanks!

Jenkinsfile
```diff
@@ -1023,9 +1023,7 @@ pipeline {
                         unstash 'js-exe'
 
                         runShell("""
-                            wget https://github.com/tcnksm/ghr/releases/download/v0.12.1/ghr_v0.12.1_linux_amd64.tar.gz
-                            tar -zxvpf ghr_v0.12.1_linux_amd64.tar.gz && rm bin/ghr_v0.12.1_linux_amd64.tar.gz || true
-                            ./ghr_v0.12.1_linux_amd64/ghr -r stanc3 -u stan-dev -recreate ${tagName()} bin/
+                            gh release create ${tagName()} --latest --target master --generate-notes ./bin/*
```
WardBrian, line 1026: The `release create` command fails if the tag already exists, so I think you need 
```
gh release delete ${tagName()} --cleanup-tag -y || true
```
First.

Also, I don't think `--generate-notes` is necessary/generates output we want. Something like 

`--notes "$(git log --pretty=format:'nightly: %h %s' -n 1)"` seems nicer

serban-nicusor-toptal: Ah, I've read this flag:
``` 
--verify-tag 
Abort in case the git tag doesn't already exist in the remote repository
```

I thought it would proceed, misled me a bit.
Agree with the notes.

WardBrian (approved):
Thanks! We can keep an eye on this for a little bit

WardBrian:
Oh, I think you might need to `\$` to escape the git log command

PR #1432, opened 2024-06-21
Allow `#include` statements to function in stanc.js

### LINKED ISSUE #1430: #include support in stancjs
**Is your feature request related to a problem? Please describe.**

stanc.js often runs in environments like v8 where the file system is not available, and so it currently cannot handle #include statements. This limits its ability to be used for tasks like formatting, and compiling requires preprocessing outside of the compiler which leads to bugs and worse error messages. 

**Describe the solution you'd like**

The stanc() function could accept a final argument which is a mapping from file paths to a string of their contents, e.g if i had a stan model foo.stan,
```stan
functions {
#include <helpers/bar.stan>
}
// ...
```

I could invoke `stanc()` like

`stanc("foo", fooCode, arguments, {"helpers/bar.stan": barcode});`

**Describe alternatives you've considered**
In contexts where the file system is actually available, like Node, we could just use it directly. 

**Additional context**

Properly preparing a call to the above would require some work based on the include-paths argument the user supplied. Of course, you could over specify the final argument and include the text of all .stan files in the directory - perhaps a bit slower, but would work. 

@andrjohns @hsbadr - does this sound like something RStan would use if it was supported

WardBrian:
Closes #1430 

This PR can be viewed as several parts:

1. Some boilerplate-y code in stancjs.ml to read a `{ [s: string] : string }` JS object into a ocaml `string String.Map.t`
2. Extracting the logic in the Preprocessor that looks up new included files into a module type that is passed in from the outside
3. Doing the same in the Lexer

An earlier version of this didn't use functors and just had `Preprocessor` store a function pointer that was set by stancjs.ml/stanc.ml, but I had both aesthetic and performance concerns with that approach. These functors are ultimately pretty lightweight

This is currently lightly tested in stanc.js but I intend to add more.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

`stanc.js` can now accept models which contain `#include` statements. A fourth argument is available on the javascript `stanc` function which must be either `undefined` or a object mapping included file names to Stan source code as strings. 

## Copyright and Licensing

### DIFF
--- a/src/frontend/Filesystem_includes.ml
+++ b/src/frontend/Filesystem_includes.ml
@@ -1,0 +1,29 @@
+(** filesystem-based way of opening new lexbufs for #include directive *)
+
+open Core
+
+(** list of allowed directories to look in for the path *)
+let lookup_paths : string list ref = ref []
+
+let find_include fname =
+  let open Lexing in
+  let rec loop paths =
+    match paths with
+    | [] ->
+        let message =
+          let pp_list ppf l =
+            match l with
+            | [] -> Fmt.string ppf "None"
+            | _ -> Fmt.(list ~sep:comma string) ppf l in
+          Fmt.str
+            "Could not find include file '%s' in specified include paths.@\n\
+             @[Current include paths: %a@]" fname pp_list !lookup_paths in
+        raise
+          (Errors.SyntaxError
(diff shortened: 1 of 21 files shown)


WardBrian:
Superseded by #1433

PR #1435, opened 2024-07-01
Implement jacobian+= statement

WardBrian:
Implements https://github.com/stan-dev/design-docs/blob/master/designs/0034-jacobian.md

This required some trickery because `jacobian` is not yet a reserved word, but it isn't too bad IMO. The comments `TODO(2.38)` identify locations where this is being done.

Closes https://github.com/stan-dev/stan/issues/1289
Closes https://github.com/stan-dev/stanc3/issues/979

#### Submission Checklist

- [x] Run unit tests
- Documentation
    
## Release notes

Added the `jacobian +=` statement, an analogue of `target +=` but which can be conditionally disabled by Stan's algorithms when a unadjusted estimate is required.

## Copyright and Licensing


src/frontend/parser.mly
```diff
@@ -215,8 +215,8 @@ identifier:
   | TRUNCATE { build_id "T" $loc}
   | id_and_v = future_keyword
     {
-      let id, v = id_and_v in
-      Input_warnings.future_keyword id.name v $loc;
+      let id, _ = id_and_v in
+      (* Input_warnings.future_keyword id.name v $loc; *)
```
nhuurre, line 219: You removed this to avoid a spurious warning on every use of the word `jacobian`. However,
1. If the commented out code is meant to be restored when `jacobian` becomes a true keyword then I'd like to see a `TODO(2.38)` marker.
2. The only future keyword is `jacobian` so reactivating this warning will not be necessary---it could just be removed now.
3. The future keyword warning would not have this "false positive" if it was on `delc_identifier` rather than `identifier`. And in my opinion that would make more sense anyway. The user does not need the warning on every use of the variable, warning about the declaration is sufficient. (Though note that a for-loop loop variable is another declaration site but does not go through `decl_identifier`)

src/stan_math_backend/Lower_expr.ml
```diff
@@ -102,7 +102,7 @@ let promote_adtype =
 
 let suffix_args = function
   | Fun_kind.FnRng -> ["base_rng__"]
-  | FnTarget -> ["lp__"; "lp_accum__"]
+  | FnTarget | FnJacobian -> ["lp__"; "lp_accum__"]
```
nhuurre, line 105: As far as I know (which isn't much) `lp__` is [only a vestige](https://mc-stan.org/docs/reference-manual/removals.html#lp__-variable) and no current user code can touch it. I don't know if `FnTarget` needs to have a backwards-compatible argument list but `FnJacobian` could be simplified.
```ocaml
 | FnJacobian -> ["lp_accum__"]
```

src/stan_math_backend/Lower_functions.ml
```diff
@@ -409,7 +412,7 @@ let lower_standalone_fun_def namespace_fun
   let internal_fname = namespace_fun ^ "::" ^ fdname in
   let template =
     match fdsuffix with
-    | FnLpdf _ | FnTarget -> [TypeLiteral "false"]
+    | FnLpdf _ | FnTarget | FnJacobian -> [TypeLiteral "false"]
```
nhuurre, line 415: Do stand-alone functions not need to compute the jacobian? It's surprising that `FnTarget` and `FnJacobian` both have `false` here because for `FnTarget` it means **include** normalization constants but for `FnJacobian` it means **exclude** jacobians.

src/stan_math_backend/Lower_expr.ml
WardBrian: I believe it is still used by the constraint code, so I think if we ever expose the built-in constraints through this functionality it will need it.

src/frontend/parser.mly
WardBrian: The future keyword mechanism was (re-)introduced specifically for jacobian in the last release, so I agree it is good to just remove now.

SteveBronder (approved):

PR #1439, opened 2024-07-15
Add support for `@annotations`

WardBrian:
Initial implementation of https://github.com/stan-dev/design-docs/blob/master/designs/0033-declaration-annotations.md:

- Basic parsing of `@annotations` on declaration nodes, propagation of this information throughout compiler.
- Implementation of `@extern` example from the design doc, can be placed on functions to silence the declared-but-undefined typechecking error function-by-function
- Implementation of `@silent`, which allows a parameter which is not output by write_array.  Particularly useful for custom transforms with `jacobian+=` where the "raw" form is not desired. 
  **Note:** This feature does lead to models which prohibit certain features like standalone generated quantities. 

#### Submission Checklist

- [x] Run unit tests
- Documentation
   
## Release notes

`@extern` can now be placed before a function declaration to mark it as defined in an external C++ file
`@silent` can be placed on a variable in one of the output blocks (e.g. `parameters`) to prevent it from appearing in the CSV output.

## Copyright and Licensing


nhuurre:
In the design doc [discussion](https://github.com/stan-dev/design-docs/pull/45), `@silent` was the most controversial annotation; it was agreed that `@gpu` should be the first one.
I think it would be enough for the initial implementation of `@gpu` to check that the mem_pattern is `SoA` after optimization and emit a warning if not. Annotation-aware optimization passes can be added later.

---

I'd prefer formatting the annotation on its own line (like a Python function decorator)
```
@gpu
matrix[N,M] x;
```
This is more or less required for longer parametrized annotations, such as the hypothetical `@extern("file.cpp" :: namespace)`.

---

Checking that the annotation belongs to a list of known annotations may be enough to catch typos, but you should also warn about inapplicable annotations (e.g. `@silent` on a function or non-toplevel variable, `@extern` on a function with a body)

WardBrian:
`@gpu` would make the most sense to be implemented 1) by someone other than myself who understands them more, and 2) as part of something like #1353. In particular, at the moment in the compiler there is a lot of legacy opencl stuff that does _not_ interact with the SoA memory pattern things.

`@silent` was my motivation to getting around to actually implement this, since it seems to be a feature users ask for often ([1](https://discourse.mc-stan.org/t/omit-transformed-parameters-from-output/11758), [2](https://discourse.mc-stan.org/t/monitor-subset-of-parameters-in-cmdstanr/19009), [3](https://discourse.mc-stan.org/t/importing-large-cmdstan-csv-files-to-r/485), [4](https://discourse.mc-stan.org/t/cmdstan-sample-exclude-parameters/23141/3), https://github.com/stan-dev/cmdstan/issues/553), and with the `jacobian +=` changes encouraging more user-defined transforms I imagine it would be even more so. But it does have concerns with other features, so I would be willing to pull it into another PR for separate discussion

Formatting is currently set up with a hv box, so it will fit on one line if it can, but as soon as it can't, every annotation will appear on its own line. We could easily use a vbox instead if that's preferred -- I have a slight preference towards saving lines if they're short annotations

More advanced warnings are definitely something I want to look at before this becomes a non-draft; part of the issue is a lot of annotation logic will inherently live in the lower parts of the compiler, which currently never emit warnings, so wiring it up in a reasonable way requires some thinking

Issue #1295, opened 2023-03-20
[BUG] Compiler optimizations can prevent bounds checks


WardBrian:
### Current Behavior:

Enabling compiler optimizations currently stops emitting a default assignment for a variable if it is immediately assigned to (https://github.com/stan-dev/stanc3/pull/1029). However, this changes the behavior of those very `assign` functions when it comes to bounds checks:

https://github.com/stan-dev/stan/blob/4ec109a21e5bd1cd43eed724ef286aadec3b5e7a/src/stan/model/indexing/access_helpers.hpp#L58

This means the same model:

```stan
parameters {
  vector[3] y;
}

model {
  vector[2] x;
  x = y;
  y ~ std_normal();
  print(x); // prevent dead-code-elimination on x
}
```

Will throw an error at --O0:
```
Unrecoverable error evaluating the log probability at the initial value.
Exception: vector assign rows: assigning variable x (2) and right hand side rows (3) must match in size (in '../../ml/stanc3/bounds.stan', line 12, column 4 to column 10)
Exception: vector assign rows: assigning variable x (2) and right hand side rows (3) must match in size (in '../../ml/stanc3/bounds.stan', line 12, column 4 to column 10)
```
But will sample to completion at `--O1`. 

### Expected Behavior:

Bounds checks are applied uniformly.

WardBrian:
In https://github.com/stan-dev/stan/issues/3271 a user encountered a alternative version of this bug which resulted in an internal error being shown:

```
Chain 1 Rejecting initial value:
Chain 1   Log probability evaluates to log(0), i.e. negative infinity.
Chain 1   Stan can't start sampling from this initial value.
Chain 1 Iteration:   1 / 1000 [  0%]  (Warmup) 
Chain 1 Exception: In serializer: Storage capacity [763] exceeded while writing value of size [16] from position [759]. This is an internal error, if you see it please report it as an issue on the Stan github repository. (in '/var/folders/3d/czqqm9h550g3ktn68hmtkh080000gn/T/RtmpabL6Uf/model-a7f34d9ee239.stan', line 45, column 4 to column 51)
```

SteveBronder:
Talking to @WardBrian we think the best answer here is to have overloads for `assign(...)` that take in the intended sizes as the last arguments.

WardBrian:
I think that will work. This case only occurs when there is a plain variable on the LHS (e.g. no indexing), so it shouldn't need to combinatorially interact with the other overloads which handle that.

PR #1441, opened 2024-07-18
Feature/decl assign params

SteveBronder:
Adds an `assignment` member to the `Decl` record which allows for assignments made inline with the declaration to be a single line. For now this is only being done for parameters. Parameters are also given the auto type so that Eigen matrices and vectors can come in as `Eigen::Map` types. This will stop parameters from being copied when used in later functions.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Make parameters `auto` so that the parameter types for matrices can be Eigen maps

## Copyright and Licensing


SteveBronder:
@WardBrian this seems to be failing because of a failed formatting, but I ran the formatter locally and seems fine? Tried to update all my stuff from opam and the scripts but I'm still getting the same formatting.

src/middle/Stmt.ml
```diff
@@ -25,7 +25,8 @@ module Fixed = struct
           { decl_adtype: UnsizedType.autodifftype
           ; decl_id: string
           ; decl_type: 'a Type.t
-          ; initialize: bool }
+          ; initialize: bool
+          ; assignment: 'a option }
```
nhuurre, line 29: `initialize` is irrelevant if `assignment` is set, right? I'd rather change this to
```suggestion
          ; initialize: 'a decl_init }
```
with
```ocaml
type 'a decl_init = Uninit | Default | Assign of 'a
```
where the first one means `initialize=false`, the second means `initialize=true` and the last one means `assignment=Some`.

nhuurre (commented):
The failure is in `Debug_data_generation_tests.ml`. Some of the randomly generated numbers are off, but only in the least significant bit. Floating-point numbers are notoriously finicky. Could be some difference in rounding mode or whatever?

Running locally, I see the same results as Jenkins. You can just revert `Debug_data_generation_tests.ml`, this PR shouldn't affect that file anyway.

SteveBronder:
> The failure is in Debug_data_generation_tests.ml. Some of the randomly generated numbers are off, but only in the least significant bit. Floating-point numbers are notoriously finicky. Could be some difference in rounding mode or whatever?

Thanks, fixed!

src/middle/Stmt.ml
WardBrian: I agree with the suggestion, but I wonder if it couldn't go further? The optimization detects an assignment in order to set `initialize`, so couldn't it just move the assignment inside the decl ?

SteveBronder: ooh thats a pretty way to do this! I'll update

SteveBronder:
So I absolutely need a thorough set of eyes on the optimizer code. I _think_ I got the lazy code motion correct

src/middle/Stmt.ml
SteveBronder: > I agree with the suggestion, but I wonder if it couldn't go further? The optimization detects an assignment in order to set initialize, so couldn't it just move the assignment inside the decl ?

I agree with this but think having all the decls inline will be enough changes to have a separate PR

src/frontend/Ast_to_Mir.ml
```diff
@@ -458,7 +458,8 @@ let create_decl_with_assign decl_id declc decl_type initial_value transform
             () } in
   let decl =
     Stmt.
-      { Fixed.pattern= Decl {decl_adtype; decl_id; decl_type; initialize= true}
+      { Fixed.pattern=
+          Decl {decl_adtype; decl_id; decl_type; initialize= Default}
```
WardBrian, line 462: I think this should use the logic on the next line which inspects the rhs_assignment to set `Initialize`

src/stan_math_backend/Transform_Mir.ml
```diff
@@ -764,9 +764,27 @@ let add_reads vars mkread stmts =
...
   let add_read_to_decl (Stmt.Fixed.{pattern; _} as stmt) =
     match pattern with
-    | Decl {decl_id; _} when Map.mem var_names decl_id ->
+    | Decl {decl_id; decl_adtype; decl_type; _} when Map.mem var_names decl_id
+      -> (
         let loc, out = Map.find_exn var_names decl_id in
-        stmt :: mkread (Stmt.Helpers.lvariable decl_id, loc, out)
+        let param_reader = mkread (Stmt.Helpers.lvariable decl_id, loc, out) in
+        match param_reader with
+        | [ Stmt.Fixed.
```
WardBrian, line 772: This looks pretty difficult to maintain and should probably be folded into mkread somehow

test/integration/good/code-gen/complex_numbers/cpp.expected
```diff
@@ -240,43 +240,22 @@ class basic_op_param_model final : public model_base_crtp<basic_op_param_model>
     // suppress unused var warning
     (void) function__;
     try {
-      Eigen::Matrix<std::complex<local_scalar_t__>,-1,-1> cmat =
-        Eigen::Matrix<std::complex<local_scalar_t__>,-1,-1>::Constant(N, N,
-          std::complex<local_scalar_t__>(DUMMY_VAR__, DUMMY_VAR__));
-      current_statement__ = 1;
```
WardBrian, line 246: The positioning is going missing here

WardBrian (changes requested):

src/frontend/Ast_to_Mir.ml
SteveBronder: Looking at this code it's not modifying the params. Making changes in places that do no modify params is out of the scope of this PR.

WardBrian: Sounds good -- I think we would really want to fix https://github.com/stan-dev/stanc3/issues/1295 before doing this -- probably by writing a generic-ish `check_size` function in Stan which we can call after the assignment is completed.

test/integration/good/code-gen/complex_numbers/cpp.expected
WardBrian: I found where they had gone

WardBrian (approved):
Ok @SteveBronder, assuming you don't take issue with any of the changes I just pushed this should be good to merge after Jenkins is done

PR #1442, opened 2024-07-26
Add stochastic row and column matrix types

SteveBronder:
#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [ ] If a user-facing facing change was made, the documentation PR is here: Still needs added

## Release notes

Adds row and column stochastic matrix types

```stan
stochastic_row_matrix[10, 10] X;
stochastic_column_matrix[10, 10] X;
```

See the example [here](test/integration/good/code-gen/stochastic_matrices.stan) for more 

## Copyright and Licensing


src/frontend/Ast_to_Mir.ml
```diff
@@ -339,6 +341,16 @@ let rec param_size transform sizedtype =
...
+  | StochasticRow -> (
+      match sizedtype with
+      | SMatrix (mem_pattern, d1, d2) ->
+          SMatrix (mem_pattern, d1, Expr.Helpers.(binop d2 Minus (int 1)))
+      | _ -> sizedtype)
+  | StochasticColumn -> (
+      match sizedtype with
+      | SMatrix (mem_pattern, d1, d2) ->
+          SMatrix (mem_pattern, Expr.Helpers.(binop d1 Minus (int 1)), d2)
+      | _ -> sizedtype)
```
WardBrian, line 353: I think this probably needs a helper like the `shrink_eigen_mat` above so that arrays of matrices work correctly?

src/analysis_and_optimization/Mir_utils.ml
```diff
@@ -61,7 +61,7 @@ let trans_bounds_values (trans : Expr.Typed.t Transformation.t) : bound_values =
        until we define a distribution which accepts a tuple,
        this doesn't matter.
     *)
-   |TupleTransformation _ ->
+   |TupleTransformation _ | StochasticColumn | StochasticRow ->
```
WardBrian, line 64: These can actually be up with `Simplex` since they are bound between 0 and 1, I think?

src/frontend/lexer.mll
```diff
@@ -139,6 +139,8 @@ rule token = parse
                                 Parser.CHOLESKYFACTORCOV }
   | "corr_matrix"             { lexer_logger "corr_matrix" ; Parser.CORRMATRIX }
   | "cov_matrix"              { lexer_logger "cov_matrix" ; Parser.COVMATRIX }
+  | "stochastic_column_matrix"              { lexer_logger "stochastic_column_matrix" ; Parser.STOCHASTICCOLUMNMATRIX }
+  | "stochastic_row_matrix"              { lexer_logger "stochastic_row_matrix" ; Parser.STOCHASTICROWMATRIX }
```
WardBrian, line 143: General question: would it be better to make these e.g. `row_stochastic_matrix`? I'm thinking particularly if we ever support a doubly stochastic type, it would have to have an awkward name if we wanted the word stochastic to go first.

src/middle/Transformation.ml
```diff
@@ -20,10 +20,14 @@ type 'e t =
...
+  | StochasticRow
+  | StochasticColumn
   | TupleTransformation of 'e t list
 [@@deriving sexp, compare, map, hash, fold]
 
 let rec has_check = function
-  | Identity | Offset _ | Multiplier _ | OffsetMultiplier _ -> false
+  | Identity | Offset _ | Multiplier _ | OffsetMultiplier _ | StochasticRow
+   |StochasticColumn ->
+      false
```
WardBrian, line 31: These do have checks though, right? https://github.com/stan-dev/math/pull/3095

src/frontend/parser.messages
WardBrian, line 1: Optional: for the new messages at the bottom, you can collapse any error states with duplicate messages into one section. Easiest way to do this is delete the `##` commpent and put all the `program:` lines together with the error after all of them, then re-run the update_messages script to put back the (now much shorter) comments.

WardBrian (changes requested):
Mostly looks good, a few scattered questions

src/frontend/Ast_to_Mir.ml
WardBrian: Worth adding an array of these to the test, as well!

src/frontend/Ast_to_Mir.ml
```diff
@@ -323,6 +323,18 @@ let rec param_size transform sizedtype =
...
+    | SizedType.SMatrix (mem_pattern, d1, d2) ->
+        SizedType.SMatrix (mem_pattern, f1 d1, f2 d2)
+    | SArray (t, d) -> SizedType.SArray (stoch_size f1 f2 t, d)
+    | SInt | SReal | SComplex | SRowVector _ | SVector _ | STuple _
+     |SComplexRowVector _ | SComplexVector _ | SComplexMatrix _ ->
+        Common.ICE.internal_compiler_error
+          [%message "Expecting SMatrix, got " (st : Expr.Typed.t SizedType.t)]
+  in
+  let min_one d = Expr.Helpers.(binop d Minus (int 1)) in
+  let noop d = d in
```
WardBrian, line 337: [optional] this function has a name -- `Fn.id` -- if you'd prefer to not define it yourself. Seems fine to leave, as this name may be clearer for someone not deeply familiar with the jane street libraries we use

src/frontend/Pretty_printing.ml
```diff
@@ -362,6 +364,8 @@ let rec pp_transformed_type ppf (st, trans) =
     | CholeskyCov -> pf ppf "cholesky_factor_cov%a" cov_sizes_fmt ()
     | Correlation -> pf ppf "corr_matrix%a" cov_sizes_fmt ()
     | Covariance -> pf ppf "cov_matrix%a" cov_sizes_fmt ()
+    | StochasticColumn -> pf ppf "stochastic_column_matrix%a" sizes_fmt ()
+    | StochasticRow -> pf ppf "stochastic_row_matrix%a" sizes_fmt ()
```
WardBrian, line 368: I think these need to be updated -- worth adding these types to a test that gets pretty-printed rather than just code-gen'd

WardBrian (commented):
One testing comment and one optional item but otherwise this looks good to me

src/frontend/lexer.mll
WardBrian: We could also consider `row_stochastic` (drop the `_matrix`) if we wanted a shorter name -- probably worth asking @bob-carpenter

bob-carpenter:
On further reflection, I think the reason we use "simplex" and "corr_matrix" is that's how someone would say them in conversation.  Nobody says "simplex vector" but everyone says "correlation matrix".  We weren't consistent with "ordered", which someone would say as "ordered vector".

WardBrian (approved):

PR #1466, opened 2024-11-13
Fix code generation issue with overloaded distributions in `~` statements

WardBrian:
Initially reported in https://discourse.mc-stan.org/t/use-of-for-overloaded-distribution-statements/37238

I tried a few solutions. Essentially, when lowering from the AST to MIR, a tilde needs to know:
1. Is the distribution user defined
2. Is the distribution a PDF or PMF
 
This information is resolved during typechecking, but currently gets thrown away.

The simplest solution I settled on is 1) adding a separate FnLpmf case to the Fun_kind.suffix type (it always bothered me that FnLpdf lumped these two together, anyway) and then 2) storing this in the Tilde node during typechecking 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes
Fixed an issue where certain overloads of distributions could lead to the `~` statement producing uncompilable C++.

## Copyright and Licensing


test/integration/bad/algebra_solver/stanc.expected
```diff
@@ -335,7 +335,7 @@ where F1 = (vector, vector, array[] real, array[] int) => real
 Available signatures:
 (<F2>, vector) => vector
 where F2 = (vector) => vector
-  The first argument must be a pure function but got a probability density or mass function. These function types are not compatible.
+  The first argument must be a pure function but got a probability density function. These function types are not compatible.
```
nhuurre, line 338: This looks nicer, though.

nhuurre (approved):
> (it always bothered me that FnLpdf lumped these two together, anyway)

They're lumpled together because, as far as the backend is concerned, they behave identically. I'm not sure if having different names in the frontend was a good idea.

Technically, you don't _need_ a new `Fun_kind` variant here either--you could decide which suffix to use based on `UnsizedType.is_discrete_type arg`.

WardBrian:
> I'm not sure if having different names in the frontend was a good idea.

Certainly up for debate. @bob-carpenter agrees, for what it's worth, but it's probably not worth changing now

> Technically, you don't _need_ a new `Fun_kind` variant here either--you could decide which suffix to use based on `UnsizedType.is_discrete_type arg`.

True -- and my first attempt used _only_ that, until I realized that we also need to differentiate between StanLib and UserDefined. If you prefer that solution I can marry the two

PR #1471, opened 2024-11-26
Allow existing uses of _jacobian in function names, with warning 

### LINKED ISSUE #1470: Backwards compatibility: We should allow functions to be named _jacobian that don't use the new feature to be called anywhere for a few versions
We took care to implement `jacobian+=` in a backwards-compatible way, but not `_jacobian` functions, of which there are some in the wild: https://github.com/search?q=language%3Astan+_jacobian%28&type=code

Because these functions don't need the `jacobian +=` feature, ideally they would continue to be treated as `FnPlain`s, with a warning. I'm not sure if this is actually implementable, however, due to the order-independence of the functions block.

First reported by @tillahoffmann https://discourse.mc-stan.org/t/planning-the-2-36-release/36382/15?u=wardbrian

WardBrian:
Closes #1470. This uses the basic rule that if the functions block contains no uses of the new `jacobian +=` statement, it is probably older code, so we should keep allowing it for now. 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fixed a backwards-compatiblity hazard with the new `_jacobian` functions. Existing functions that just happen to have this naming structure will no longer fail to compile, but will issue a warning alerting that this name will change meanings in the future. 

## Copyright and Licensing


nhuurre (approved):
Looks good.

In unrelated news, the following error is not detected until C++ compilation:
```stan
functions {
  // void function
  void foo_jacobian() {
    jacobian += 1;
  }
}
transformed data {
  foo_jacobian();
}
```

WardBrian:
@nhuurre good catch! so did this:

```stan
functions {
    void foo_rng(real x){
        print(normal_rng(0,x));
    }
}

model {
    foo_rng(1.0);
}
```

Fixed both.

nhuurre (approved):
void rngs are useless, but I suppose there's no harm in allowing them.

PR #1475, opened 2024-12-05
 Stancjs: Check input JS types before conversion 

### LINKED ISSUE #1446: Internal compiler error: TypeError: a.charCodeAt is not a function
This happened also a couple of days ago. Now I have the code that caused it:
```r
lambda_true = 3
N = 5
set.seed(111111)
(y = rpois(N, lambda_true))
poisson_dat = list(y = y, N = N)
alpha = beta = 1
```

```stan
data {
	int<lower=0> N;
	array[N] int<lower=0> y;
}

parameters {
	real<lower=0> lambda;
}
model {
	lambda ~ lognormal(0,1);
	y ~ poisson(lambda);
}
```

```r
fit2 <- stan('poisson_lognormal.stan', poisson_dat, iter = 20000, chains = 4)

The error message:
Error in stanc(file = file, model_code = model_code, model_name = model_name,  : 
  0
Internal compiler error:
TypeError: a.charCodeAt is not a function
```
Rstan 2.35.0.9000

This should never happen. Please file a bug at https://github.com/stan-dev/stanc3/issues/new
and include this message and the model that caused this issue.

I removed the package V8 but it didn't change the situation.

WardBrian:
Closes #1446 by improving the error message. Note that the actual bad behavior is in RStan, c.f. https://github.com/stan-dev/rstan/issues/1145

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

stanc.js will now properly complain if it is supplied with bad javascript types for its arguments.

## Copyright and Licensing

### DIFF
--- a/src/stancjs/stancjs.ml
+++ b/src/stancjs/stancjs.ml
@@ -232,11 +232,60 @@ let to_file_map includes =
   | BadJsInput s -> Result.Error s
   | e -> Result.Error (Exn.to_string e)
 
-let stan2cpp_wrapped name code (flags : Js.string_array Js.t Js.opt) includes =
-  let flags =
-    let to_ocaml_str_array a =
-      Js.(str_array a |> to_array |> Array.map ~f:to_string) in
-    Js.Opt.map flags to_ocaml_str_array |> Js.Opt.to_option in
+let get_includes includes =
+  match to_file_map includes with
+  | Result.Ok map -> (map, [])
+  | Result.Error warn ->
+      ( String.Map.empty
+      , [ Fmt.str "Warning: stanc.js failed to parse included file mapping!@ %s"
+            warn ] )
+
+let bad_arg_message ~name ~expected value =
+  Fmt.str
+    "Failed to convert stanc.js argument '%s'!@ It had type '%s' instead of \
+     '%s'."
+    name
+    (Js.typeof value |> Js.to_string)
+    expected
+
+let checked_to_string ~name value =
+  if not (typecheck value "string") then
+    Error (bad_arg_message ~name ~expected:"string" value)
+  else Ok (Js.to_string value)
+
+let checked_to_array ~name value =
+  if not (typecheck value "object" || typecheck value "array") then
+    Error (bad_arg_message ~name ~expected:"array' or 'object" value)
+  else Ok (Js.to_array value)
+
+let process_flags (flags : 'a Js.opt) =
+  let set_backend_args_list flags =
+    (* Ignore the "--o" arg, the stan file and the binary name (bin/stanc). *)
+    let sans_model_and_hpp_paths x =
+      not
+        String.(
+          is_suffix ~suffix:".stan" x
+          && not (is_prefix ~prefix:"filename-in-msg" x)
+          || is_prefix ~prefix:"o=" x) in
+    let stanc_args_to_print =
+      flags |> Array.to_list
+      |> List.filter ~f:sans_model_and_hpp_paths
+      |> List.map ~f:(fun o -> "--" ^ o)
+      |> String.concat ~sep:" " in
+    Lower_program.stanc_args_to_print := stanc_args_to_print in
+  let open Result in
+  let open Result.Let_syntax in
+  let%bind flags =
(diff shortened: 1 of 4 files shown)


src/stancjs/stancjs.ml
```diff
@@ -190,102 +197,132 @@ let wrap_result ?printed_filename ~code ~warnings res =
...
-  | e -> Result.Error (Exn.to_string e)
+          Error (bad_arg_message ~name:"includes" ~expected:"object" includes)
+        else Ok () in
+      let keys = Js.object_keys includes |> Js.to_array |> List.of_array in
+      let value k =
+        let value_js = Js.Unsafe.get includes k in
+        checked_to_string
+          ~name:("includes[\"" ^ Js.to_string k ^ "\"]")
+          value_js in
+      let* values = List.map keys ~f:value |> Result.all in
```
nhuurre, line 249: `Result.all` discards all includes if the object contains any non-string values. This error is not considered fatal so I think you should use `List.partition_result` instead.

src/stancjs/stancjs.ml
```diff
@@ -190,102 +197,132 @@ let wrap_result ?printed_filename ~code ~warnings res =
...
+    match Js.Opt.to_option flags with
+    | None -> Ok None
+    | Some flags ->
+        let* flags_array = checked_to_array ~name:"flags" flags in
+        let* ocaml_flags =
+          Array.mapi flags_array ~f:(fun i v ->
+              checked_to_string ~name:("flags[" ^ string_of_int i ^ "]") v)
+          |> Array.to_list |> Result.all >>| Array.of_list in
+        set_backend_args_list ocaml_flags;
+        Ok (Some ocaml_flags) in
```
nhuurre, line 289:
```suggestion
        let+ ocaml_flags =
          Array.mapi flags_array ~f:(fun i v ->
              checked_to_string ~name:("flags[" ^ string_of_int i ^ "]") v)
          |> Array.to_list |> Result.all >>| Array.of_list in
        set_backend_args_list ocaml_flags;
        (Some ocaml_flags) in
```

src/stancjs/stancjs.ml
```diff
@@ -190,102 +197,132 @@ let wrap_result ?printed_filename ~code ~warnings res =
...
+    | Some flags -> fun flag -> Array.mem ~equal:String.equal flags flag
+    | None -> fun _ -> false in
+  let flag_val =
+    match flags with
+    | None -> fun _ -> None
+    | Some flags ->
+        fun flag ->
+          let prefix = flag ^ "=" in
+          Array.find flags ~f:(String.is_prefix ~prefix)
+          |> Option.bind ~f:(String.chop_prefix ~prefix) in
```
nhuurre, line 301:
```suggestion
          Array.find_map flags ~f:(String.chop_prefix ~prefix) in
```

nhuurre (commented):
> Closes https://github.com/stan-dev/stanc3/issues/1446 by improving the error message.

Will this stanc3 change go out before RStan gets fixed? If not then _technically_ users won't see the new error message and it was the RStan fix that closed issue 1446. If yes, then I think it would be worthwhile to try to fix the issue here as well. If it's just RStan passing array-of-strings instead of a single string, then concatenating the strings should be easy.

WardBrian:
> Will this stanc3 change go out before RStan gets fixed? If not then _technically_ users won't see the new error message and it was the RStan fix that closed issue 1446.

The RStan experimental branch picks up nightly, but this issue is present in the current CRAN release (based on Stan 2.33), so it really has to be fixed by them for most users to benefit, since updating to a more recent version will take another long while I assume. RStan’s string preprocessing of the model code is broken in at least 3 ways right now, and I don’t think the compiler can or should take on the burden of trying to fix it…

Anyway, I think technically finding the cause and opening the RStan issue is the thing that resolves 1446 from the compilers perspective. This is mostly just trying to play a little nicer (and trying to cut down on the noise in the issue list, since it will no longer give a link to open one here)

WardBrian:
I also think we should probably merge this after the release, just because it does touch a lot of the stancjs driver

nhuurre (approved):

Issue #24, opened 2018-12-30
Switch stanc.ml to Core_kernel Command module from Arg module


seantalts:
https://dev.realworldocaml.org/command-line-parsing.html

VMatthijs:
RealWorldOCaml writes

> The Arg module
The Arg module is from the OCaml standard library, which is used by the compiler itself to handle its command-line interface. Command is generally more featureful than Arg (mainly via support for subcommands, the step combinator to transform inputs, and help generation), but there’s absolutely nothing wrong with using Arg either. You can use the Command.Spec.flags_of_args_exn function to convert Arg specifications into ones compatible with Command. This is quite often used to help port older non-Core code into the Core standard library world.

What is the priority of this issue? Should it be listed under the initial release milestone? In what way does the current command line interface not suffice?

seantalts:
Re: why: I think we'll see synergies by switching entirely to Core_kernel and using it whenever we need something it supports. They have the same code and doc standards, interface aesthetics, etc. And I found the Arg thing a bit weird at first - very side-effect-y, for example. I think it will probably work this shouldn't block an initial release, but we can keep this here as a good first issue.

VMatthijs:
OK! Makes sense!

ode33:
Is the idea then to fold `Debugging.ml` into `stanc.ml` with those arguments held inside the definition of `options`? Or would we like `Debugging` to continue to be used as it is now?

seantalts:
We talked a little about this at the meeting - I think we want to move away from the global references used for config and debug flags and what not and instead pass config records to the frontend, middle, and backend. So I think eventually we'd like to get rid of Debugging.ml but we could do that separately from this PR if that's easier.

WardBrian:
This is theoretically possible with #1019  merged, but I'm not sure we can do this still. So far as I can tell, `Core_kernel`'s version of `Command` has no `Command.run` function -- that's only in `Core`, which I believe cannot be cross compiled. This seems really strange to me - `Arg` works fine after all.

PR #1477, opened 2024-12-11
Refactor: Unify majority of stanc.exe and stanc.js code

### LINKED ISSUE #1214: Clean up drivers stanc3, stancjs
These have sort of grown over time to use some odd patterns.

TODO list:

- [ ] `Debugging.ml` is used for printing things like the typed_ast, but not the MIRs or anything else. We should move more of that code in there. We should also generalize `Debugging.ml` to allow its usage in stanc3js so it doesn't always just print
- [ ] There is a lot of nested if-else style logic which shouldn't need to be there if we encapsulate the debug logic into `Debugging.ml`. Because we use the `with_return` pattern in stanc3js (which I don't _love_ but has some benefits in a JS driver), and can exit in stanc, I think the drivers could both just be straight lines of assignments and processing
- [ ] Make sure they both do the same thing - see https://github.com/stan-dev/stanc3/issues/751#issuecomment-1033865785
- [ ] Expose all the same functionality - stanc3js lacks several more of the debug flags

### LINKED ISSUE #1464: Remove `--O` (or alias it to `O1`, not `Oexperimental`)
This name predates us actually recommending anyone use the optimization passes of the compiler, and can continue to cause confusion: https://github.com/stan-dev/cmdstan/issues/1299

WardBrian:
This closes #1214. It introduces a new module, `Driver`, which actually drives code through the compiler's phases. The code for stanc.exe and stanc.js are now just I/O wrappers around this.

**Pros**: Unified codebase, much more difficult to have feature slippages or bugs in one but not the other. 

The code in driver also has a more obvious control flow than either of the existing stanc.ml or stancjs.ml's did, IMO

**Cons:** A bit more abstract code.

In particular, to get the same behavior that we have historically had, all non-C++ outputs are handled via a callback supplied by the caller of the `Driver` code. This callback can exit early, or it can just process the output and then return, allowing the Driver to push forward. In some sense it's a poor man's effect handler, so we could consider using one of those if we updated to OCaml 5+ in the future. 

**Other:**
As part of this, stanc.js got a few of the extra debugging flags it was missing, and this will also close #1464. 

I'm not entirely happy with how the stanc.ml argument parsing looks after this, so I might still look into cleaning that up further, possibly giving #24 another look or looking into `Cmdliner`, which is already in our transitive dependency stack anyway

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

stanc.js and stanc.exe now share significantly more code between them.

## Copyright and Licensing


src/driver/Entry.ml
```diff
@@ -0,0 +1,146 @@
...
+  | Generated of string
+  | Warnings of Warnings.t list
+
+type compilation_result = (string, Errors.t) result
+
+let stan2cpp model_name model (flags : Flags.t) (output : other_output -> unit)
+    : compilation_result =
+  let open Common.Let_syntax.Result in
+  reset_mutable_states model_name flags;
+  if flags.version then output (Info (Fmt.str "%s" version));
```
nhuurre, line 52:
```suggestion
  if flags.version then output (Version (Fmt.str "%s" version));
```

src/driver/Entry.ml
```diff
@@ -0,0 +1,146 @@
...
+    if flags.warn_uninitialized then
+      output (Warnings (Pedantic_analysis.warn_uninitialized mir));
+    if flags.warn_pedantic then
+      output (Warnings (Pedantic_analysis.warn_pedantic mir));
+    let () =
+      match flags.debug_settings.debug_mir with
+      | Off -> ()
+      | Basic ->
+          output (DebugOutput (fmt_sexp [%sexp (mir : Middle.Program.Typed.t)]))
+      | Pretty -> output (DebugOutput (Fmt.str "%a" Program.Typed.pp mir)) in
```
nhuurre, line 93: This pattern occurs three times, first with `mir`, then with `tx_mir`, and then `opt_mir`. Let's make a helper function `debug_output mir`.

src/driver/Flags.ml
```diff
@@ -0,0 +1,98 @@
...
+        (* ------------------------- *)
+        (* extra settings *)
+  ; warn_pedantic: bool
+  ; warn_uninitialized: bool
+  ; filename_in_msg: string option }
+
+and debug_settings =
+  { debug_ast: bool
+  ; debug_typed_ast: bool
+  ; debug_mir: debug_options
```
nhuurre, line 31: I don't like the "`debug_` prefix for every field" convention here. Maybe the prefixes could be more descriptive, like the ones that print intermediate representations could start with `print_`.

src/driver/Flags.ml
```diff
@@ -0,0 +1,98 @@
...
+        || is_prefix ~prefix:"--o=" x) in
+  let stanc_args_to_print =
+    flags |> List.filter ~f:sans_model_and_hpp_paths |> String.concat ~sep:" "
+  in
+  Stan_math_backend.Lower_program.stanc_args_to_print := stanc_args_to_print
+
+let remove_dotstan s =
+  if String.is_suffix ~suffix:".stanfunctions" s then String.drop_suffix s 14
+  else if String.is_suffix ~suffix:".stan" s then String.drop_suffix s 5
+  else s
```
nhuurre, line 98:
```suggestion
  Option.first_some
    (String.chop_suffix ~suffix:".stanfunctions" s)
    (String.chop_suffix ~suffix:".stan" s)
  |> Option.value ~default:s
```

nhuurre (approved):

PR #1478, opened 2024-12-13
Switch stanc.exe to use Cmdliner for cli

WardBrian:
I mentioned this in #1477. The use of imperative style argument parsing in `stanc.exe` led to some weird code patterns and didn't give us type-checking guarantees about each setting actually being available (or, conversely, being used).

This PR switches to using the `cmdliner` library: https://erratique.ch/software/cmdliner. This is by the same author as the `fmt` library we already use, and is used by `opam` itself. I think it might already be installed on most of our switches as a result, but if not it's a very minimal addition. 

#### Submission Checklist

- [x] Run unit tests
- Documentation
    
## Release notes
Improved the command-line argument parser for `stanc`. Existing flags are still supported, but new aliases may be available. See the new and improved `stanc --help` for details.

## Copyright and Licensing


WardBrian:
Okay, I think I finally excised the ghosts in the multiarch jenkins build for this...

WardBrian:
Note for reviewers: The review is probably easiest to do commit-by-commit, with the actual code changes all happening in the first commit. The second is some documentation, which can be reviewed, and the third is the result of a lot of jenkins tweaking to keep the build happy, which can probably be skipped if you're willing to trust the fact that it is building.

src/stanc/stanc.ml
```diff
@@ -381,4 +80,4 @@ let () =
   | Error internal_error ->
       Out_channel.output_string stderr internal_error;
       Out_channel.flush stderr;
-      exit 2
+      exit 125
```
nhuurre, line 83: These changes to exit codes need a bit more explanation.

WardBrian: 125 is what Cmdliner raises for unexpected errors: https://erratique.ch/software/cmdliner/doc/Cmdliner/Cmd/Exit/index.html#val-internal_error

because we’re passing catch:false, we’re not obligated to do exactly the same thing, so I’m happy to do whatever. I don’t expect anyone is relying on the exact exit code of the compiler when it ICEs, and I don’t think we had previously documented it anyway

src/stanc/CLI.ml
```diff
@@ -0,0 +1,455 @@
...
+      ; print_typed_ast
+      ; print_mir
+      ; print_transformed_mir
+      ; print_optimized_mir
+      ; print_mem_patterns
+      ; force_soa
+      ; print_lir
+      ; debug_generate_data
+      ; debug_generate_inits
+      ; debug_data_json= Option.map ~f:In_channel.read_all debug_data_file }
```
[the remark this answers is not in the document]
WardBrian: Fixed this, and a separate issue where it would crash if the file existed, but could not be opened

Jenkinsfile
```diff
@@ -113,7 +115,7 @@ pipeline {
         GIT_AUTHOR_EMAIL = 'mc.stanislaw@gmail.com'
         GIT_COMMITTER_NAME = 'Stan Jenkins'
         GIT_COMMITTER_EMAIL = 'mc.stanislaw@gmail.com'
-        MULTIARCH_DOCKER_TAG = 'multiarch-ocaml-4.14-v2'
+        MULTIARCH_DOCKER_TAG = 'multiarch-ocaml-4.14-v2-and-cmdliner'
```
nhuurre, line 118: I don't know anything about Jenkins, but what's the reason for the tag change?

WardBrian: I needed to re-build the multiarch images to add cmdliner. Because they don't install the developer-only dependencies like `ocamlformat`, I guess they didn't have it already.

This was a headache, and it's definitely non-obvious how the multiarch build works, see #1480

nhuurre (approved):

PR #1488, opened 2025-01-10
Allow model to be supplied on stdin with '-'

WardBrian:
This is a small change that allows some better 'scriptability' of the compiler, e.g.

```
 $ cat test/integration/good/for_early_return.stan | stanc - --auto-format
functions {
  real ends() {
    for (i in 1 : 10) {
      return 1;
    }
  }
}
transformed parameters {
  real a = ends();
}
```

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/845
    
## Release notes

The compiler now accepts `-` as a model argument, in which case it reads from standard input

## Copyright and Licensing

### DIFF
--- a/src/frontend/Errors.ml
+++ b/src/frontend/Errors.ml
@@ -23,20 +23,20 @@ type t =
   | DebugDataError of (Middle.Location_span.t * string)
 
 let get_context ?code Middle.Location.{filename; included_from; _} () =
-  (* If the location is not included from anywhere, it's the orignal code *)
-  let code = if Option.is_none included_from then code else None in
-  (* Otherwise, by the time we are printing an error,
-     all these files are already resolved. *)
-  match !Include_files.include_provider with
-  | FileSystemPaths _ ->
-      (* So we can read directly from the filesystem *)
-      Option.map code ~f:String.split_lines
-      |> Option.value_or_thunk ~default:(fun () ->
-             In_channel.read_lines filename)
-  | InMemory m ->
-      (* Or, we know we can find it in the map *)
-      Option.value_or_thunk code ~default:(fun () -> Map.find_exn m filename)
-      |> String.split_lines
+  (* If the location is not included from anywhere, and we
+     have code provided, use it *)
+  match (included_from, code) with
+  | None, Some code -> String.split_lines code
+  | _ -> (
+      (* Otherwise, by the time we are printing an error,
+         all these files are already resolved. *)
+      match !Include_files.include_provider with
+      | FileSystemPaths _ ->
+          (* So we can read directly from the filesystem *)
+          In_channel.read_lines filename
+      | InMemory m ->
+          (* Or, we know we can find it in the map *)
+          String.split_lines (Map.find_exn m filename))
 
 let pp_context_with_message ?code ppf (msg, loc) =
   Fmt.pf ppf "%a@,%s" (Fmt.option Fmt.string)
@@ -85,5 +85,3 @@ let pp ?printed_filename ?code ppf = function
       else
         let loc = Middle.Location_span.to_string ?printed_filename loc in
         Fmt.pf ppf "@[<v>Error in %s:@ %s@;@]" loc msg
-
-let to_string = Fmt.str "%a" (pp ?printed_filename:None ?code:None)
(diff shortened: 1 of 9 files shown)


src/frontend/Errors.ml
```diff
@@ -22,21 +22,25 @@ type t =
...
   | DebugDataError of (Middle.Location_span.t * string)
 
-let get_context_callback ?code Middle.Location.{filename; included_from; _} () =
-  (* By the time we are printing an error, all these files are already resolved. *)
+let get_context ?code Middle.Location.{filename; included_from; _} () =
+  (* If the location is not included from anywhere, it's the orignal code *)
+  let code = if Option.is_none included_from then code else None in
+  (* Otherwise, by the time we are printing an error,
+     all these files are already resolved. *)
   match !Include_files.include_provider with
```
nhuurre, line 30: Both branches of the `match` map the option, let's just do an early return.
```ocaml
if Option.is_none included_from
  then String.split_lines code
  else match ...
```

WardBrian: `code` is a `string option`, not just a `string`

nhuurre: And if `code` is None this function fails to produce context for stdin source.

Looks like that can only happen if the pretty printer produces incorrect code:
https://github.com/stan-dev/stanc3/blob/528c72c26e3fe466545ba0bf29eac262d7fb69b2/src/frontend/Pretty_print_prog.ml#L56-L58
I don't think this has ever quite worked. (It always prints empty context; `In_channel.read_lines` probably hides a file-not-found error.) That function should use `Errors.pp` instead of `Errors.to_string`.

The only other place with `Errors.pp ~code:None` is
https://github.com/stan-dev/stanc3/blob/528c72c26e3fe466545ba0bf29eac262d7fb69b2/src/stanc/stanc.ml#L78-L80
and that might be better served by exposing a `Errors.pp_debug_data_error`.

WardBrian: We also don't set it in the call for file inputs, a few lines down from there:
https://github.com/stan-dev/stanc3/blob/528c72c26e3fe466545ba0bf29eac262d7fb69b2/src/stanc/stanc.ml#L86

nhuurre: Right, don't know how I missed that.
I don't think I would want to pass the source as `'File|'Code` (I'm not a fan of it in `Parse.parse` either) so this code should stay as is.

I think changing the pretty printer to use `Errors.pp ~code:pretty` is still worth it, even though it is an obscure ICE.

test/integration/cli-args/stanc.t
```diff
@@ -255,3 +255,23 @@ Error when unreadable file is passed
...
+
+Can read from stdin
+  $ echo 'parameters {real y;}' | stanc - --auto-format
+  parameters {
+    real y;
+  }
+  
+
+
+Can filename is set to stdin when reading from stdin
```
nhuurre, line 268:
```suggestion
Filename is set to stdin when reading from stdin
```

src/frontend/Errors.ml
WardBrian: > 'm not a fan of it in Parse.parse either

Do you have something you prefer? (I suppose it would be a minor improvement to make it an actual type rather than a polymorphic variant?)

nhuurre: I do not. If I had, I would have mentioned it when you added it (I reviewed that pull request too and it's only been a month since)

test/unit/Test_utils.ml
```diff
@@ -6,12 +6,14 @@ let untyped_ast_of_string s =
   Fmt.epr "%a" (Fmt.list ~sep:Fmt.nop Warnings.pp) warnings;
   res
 
+let error_to_string = Fmt.str "%a" (Errors.pp ?printed_filename:None ?code:None)
```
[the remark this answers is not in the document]
WardBrian: I removed (I guess just moved) it because of the comment on it that it should basically never be used outside of tests -- so why let it be in scope outside of tests.

PR #1489, opened 2025-01-16
Remove usages of re, clean up libraries fields in dune

WardBrian:
This started with me testing building the compiler under OCaml 5.3.0. After installing the most recent version of our dependencies, the build was still failing, because `re` wasn't installed. It turns out, it's no longer in the Jane Street dependency tree in the next version (0.17.x)

I then realized we were not documenting this dependency anywhere, and it's also completely unnecessary based on its usages. So I pulled it out, and then updated the dune files to not request linking it. It turns out that the `str` library we were listing was also just requesting a link to our own alias of `module Str = Re.Str`, which was... messy, to say the least!

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Removed usages of the `re` library

## Copyright and Licensing

### DIFF
--- a/src/driver/dune
+++ b/src/driver/dune
@@ -1,7 +1,13 @@
 (library
  (name driver)
  (public_name stanc.driver)
- (libraries frontend middle stan_math_backend analysis_and_optimization)
+ (libraries
+  core
+  fmt
+  frontend
+  middle
+  stan_math_backend
+  analysis_and_optimization)
  (instrumentation
   (backend bisect_ppx))
  (inline_tests)
--- a/src/frontend/Typechecker.ml
+++ b/src/frontend/Typechecker.ml
@@ -1081,9 +1081,9 @@ let overlapping_lvalues lvals =
     | LIndexed (l, _), _ -> compare_no_indexing l lv2
     | _, LIndexed (l, _) -> compare_no_indexing lv1 l
     | LVariable id1, LVariable id2 -> String.compare id1.name id2.name
-    | LTupleProjection (lv1, idx1), LTupleProjection (lv2, idx2)
-      when idx1 = idx2 ->
-        compare_no_indexing lv1 lv2
+    | LTupleProjection (lv1, idx1), LTupleProjection (lv2, idx2) ->
+        let idx_comp = Int.compare idx1 idx2 in
+        if idx_comp = 0 then compare_no_indexing lv1 lv2 else idx_comp
     | _, _ ->
         (* remaining cases are not equal, we don't care *)
         Ast.compare_untyped_lval lv1 lv2 in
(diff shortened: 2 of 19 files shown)


src/frontend/Typechecker.ml
```diff
@@ -1081,9 +1081,9 @@ let overlapping_lvalues lvals =
     | LIndexed (l, _), _ -> compare_no_indexing l lv2
     | _, LIndexed (l, _) -> compare_no_indexing lv1 l
     | LVariable id1, LVariable id2 -> String.compare id1.name id2.name
-    | LTupleProjection (lv1, idx1), LTupleProjection (lv2, idx2)
-      when idx1 = idx2 ->
-        compare_no_indexing lv1 lv2
+    | LTupleProjection (lv1, idx1), LTupleProjection (lv2, idx2) ->
+        let idx_comp = Int.compare idx1 idx2 in
+        if idx_comp = 0 then compare_no_indexing lv1 lv2 else idx_comp
```
WardBrian, line 1086: This was a "fun" one I realized after getting a build on 5.3.0, which is as previously implemented the sort here could actually be non-transitive, which led to a newly-optimized version of the `List.find_all_dups` function to miss some duplicates and allow an existing bad test to instead compile

nhuurre: It's reassuring to know that our test coverage is wide enough to catch even such subtle bugs.

test/unit/Error_tests.ml
```diff
@@ -22,7 +22,7 @@ let%expect_test "backtrace indirect test" =
   |> Result.error |> Option.value_exn
   |> fun s ->
     try
-      let _ = Str.search_forward (Str.regexp "^Called from Common") s 0 in
+      let _ = String.substr_index_exn ~pattern:"Called from Common" s ~pos:0 in
       print_endline "Backtrace found in message"
     with _ -> print_endline "FAILED TO FIND BACKTRACE" );
```
nhuurre, line 27: Let's use `if String.is_substring` instead of `try .. _exn .. with`.

src/frontend/Typechecker.ml
WardBrian: That's certainly true -- it was painful to debug, but would have been much worse to not know about it at all

src/driver/dune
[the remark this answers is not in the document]
WardBrian: I wanted the dune files to more accurately serve as documentation for the libraries each folder actually uses -- Driver uses `core` and `fmt`, but didn't list them. Dune by default lets packages re-export, so because frontend and others use core and fmt, driver could see them, even though they weren't listed here.

nhuurre (approved):

PR #1498, opened 2025-03-06
Generate signature map at build time

WardBrian:
This is something that came out of some profiling I did recently. Essentially, for small-medium sized models, the compiler spends most of its time just loading the `Stan_math_signatures` module and the large hashtable it contains of function signatures:

```
Call graph 'stanc test/integration/good/code-gen/8_schools_ncp.stan':
---------------------------------------------------------------------
[   63.78M cycles in     1 calls ]     - 51.02% : load(Stan_math_signatures)
[   27.97M cycles in   135 calls ]     |   - 43.86% : Stan_math_signatures.mk_declarative_sig
[    4.62M cycles in  2119 calls ]     |   |   - 16.52% : Stan_math_signatures.rng_return_type
[    1.63M cycles in  2980 calls ]     |   |   |   - 35.23% : Stan_math_signatures.is_primitive
[    3.38M cycles in   591 calls ]     |   |   - 12.07% : Stan_math_signatures.expand_arg
[    1.42M cycles in  2280 calls ]     |   |   |   - 42.00% : Stan_math_signatures.bare_array_type
[    2.56M cycles in   242 calls ]     |   |   -  9.17% : Stan_math_signatures.all_combinations
[    1.20M cycles in  2041 calls ]     |   |   -  4.30% : Stan_math_signatures.ints_to_real
[   18.66M cycles in    17 calls ]     |   - 29.25% : Stan_math_signatures.add_binary_vec
[   17.41M cycles in    17 calls ]     |   |   - 93.33% : Stan_math_signatures.add_binary_vec_general
[    3.88M cycles in  7480 calls ]     |   |   |   - 22.31% : Stan_math_signatures.bare_array_type
[    2.18M cycles in  3468 calls ]     |   |   |   - 12.54% : Stan_math_signatures.add_unqualified
[    1.71M cycles in  3468 calls ]     |   |   |   -  9.79% : Stan_math_signatures.ints_to_real
[    1.59M cycles in     6 calls ]     |   -  2.50% : Stan_math_signatures.add_binary_vec_int_real
[  516.48K cycles in  1056 calls ]     |   |   - 32.41% : Stan_math_signatures.bare_array_type
[  242.23K cycles in   432 calls ]     |   |   - 15.20% : Stan_math_signatures.add_unqualified
[  990.39K cycles in     3 calls ]     |   -  1.55% : Stan_math_signatures.add_binary_vec_real_real
[  986.71K cycles in     3 calls ]     |   |   - 99.63% : Stan_math_signatures.add_binary_vec_general
[  312.77K cycles in   672 calls ]     |   |   |   - 31.70% : Stan_math_signatures.bare_array_type
[  153.63K cycles in   291 calls ]     |   |   |   - 15.57% : Stan_math_signatures.add_unqualified
[  887.62K cycles in  1232 calls ]     |   -  1.39% : Stan_math_signatures.add_unqualified
[  806.55K cycles in     3 calls ]     |   -  1.26% : Stan_math_signatures.add_binary_vec_real_int
[  251.65K cycles in   528 calls ]     |   |   - 31.20% : Stan_math_signatures.bare_array_type
[  120.48K cycles in   216 calls ]     |   |   - 14.94% : Stan_math_signatures.add_unqualified
[   39.95M cycles in     1 calls ]     - 31.96% : load(Environment)
[   16.87M cycles in 22451 calls ]     |   - 42.22% : Fun_kind.suffix_from_name
[   19.19M cycles in     1 calls ]     - 15.35% : load(stanc)
```

This is particularly annoying because _this hashtable is static data_. After some [discussions on the OCaml forums](https://discuss.ocaml.org/t/generating-a-map-at-compile-time/16217) about options, the following seems like the least intrusive:

- Have a small standalone executable that does the work of computing all the overloads, and generates a small module which contains a compact, binary representation of this
- Have `dune` integrate this into the build seamlessly
- At runtime, just deserialize that representation and proceed as before

This effectively removes 90% of the overhead of "doing nothing" in the compiler, from ~90M cycles above to ~8M cycles:

```
Call graph 'stanc test/integration/good/code-gen/8_schools_ncp.stan':
---------------------------------------------------------------------
[   22.59M cycles in   1 calls ]     - 62.60% : load(stanc)
[    7.55M cycles in   1 calls ]     - 20.91% : load(Generated_signatures)
[    2.37M cycles in   1 calls ]     -  6.57% : load(Lower_expr)
[    1.59M cycles in   1 calls ]     -  4.41% : load(Environment)

```

</p>
</details>

Remarkably, this also doesn't really change the binary size at all, because the generated module is only a few hundred kb as source text.

Besides making the compiler faster for end users, I also observe about a net 20% speed up of `dune runtest`, which is nice as well.

There are some unfortunate parts of doing this:

- Avoiding circular dependencies in `dune` made it necessary to move the `Stan_math_signatures` file out of `Middle` -- though I had wanted to do this anyway.
- To avoid the generator and generate**d** files trying to depend on each other, any common type definitions needed to be moved somewhere. I took this opportunity to give `UnsizedType.ml` a few new helper definitions, which really cleaned things up, but it does mean some code in weird places got touched by this change.
- As a result of the above, the sort order of signatures in the printed output changed. It now sorts by argument type, previously it was sorting by return type. I think the new behavior actually makes more sense. This is the only reason any files in `test/` changed.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Decreased the start-up overhead of the compiler.

## Copyright and Licensing


src/stan_math_signatures/Generate.ml
```diff
@@ -2703,3 +2550,41 @@ let () =
     ~required_fn_rt:UnsizedType.UVector
     ~required_fn_args:[UnsizedType.(AutoDiffable, UVector)]
     ()
+
+(** Print a module definition to [file]
+  that contains the signatures computed above *)
+let generate_module file =
+  let marshal e = Marshal.to_string e [] in
+  (* Core's Hashtbl cannot be safely Marshal'd, so we
+     round trip through an associative list *)
```
nhuurre, line 2559: From the OCaml forum discussion it's unclear if `Stdlib.Hashtbl` can be Marshal'ed. Did that fail? Or is Stdlib undesirable for some other reason? (it doesn't seem to have `Hashtbl.find_multi`...)
What about converting to `Core.String.Map`?

src/stan_math_signatures/Stan_math_signatures.ml
```diff
@@ -0,0 +1,159 @@
...
+  pf ppf "@[<v>%a@]"
+    (list ~sep:cut pp_sigs_for_name)
+    (List.sort ~compare:String.compare (Hashtbl.keys stan_math_signatures))
+
+let pretty_print_all_math_distributions ppf () =
+  let open Fmt in
+  let distributions =
+    String.Map.of_alist_reduce distributions ~f:(fun v1 v2 ->
+        v1 @ v2 |> Set.Poly.of_list |> Set.to_list)
+    |> Map.to_alist in
```
nhuurre, line 143: All this does is dedup `wiener_lpdf`. I think that should be done before marshalling.

nhuurre (commented):
> This is the only reason any files in test/ changed.

That's the only reason `test/integration/{bad,signatures}` changed. Various `test/**/mir.expected` files got additional parens around the newly named `UnsizedType.signature`. 
Also `test/unit/dune` lost a couple of `ppx_deriving`, not sure why they were ever there?

src/stan_math_signatures/Generate.ml
WardBrian: Core's `Map` also has functional values that can't be serialized, unfortunately. 

I've tried both Map and Hashtbl from stdlib, they both perform basically equally to the current code, but require some more ugly contortions to use (because of the absence of `find_multi`, but also the absense of things like `.keys` or `.to_list`...)

The interesting thing this also made me realize is that we almost never do direct lookups into this table. The Typechecker entirely uses `Environment.ml`, which converts it back into an alist first. Especially if we were willing to sort during generation, I suspect we could avoid creating a map/hashtable entirely. But I think that might be better off waiting for another time

WardBrian:
> Also `test/unit/dune` lost a couple of `ppx_deriving`, not sure why they were ever there?

Yes, this was a bit of a drive-by, but every time I learn a bit more about dune I realize we've been doing something unnecessary like over-depending on things we're not using in those modules (like in #1489). 

These were added when I thought doing this would require depending on `ppx_blob` so I was looking at our existing ppx usage. Eventually I realized we didn't need the new dependency if I just had the generated file be a fully-formed `.ml`

src/stan_math_signatures/Generate.ml
WardBrian: I forgot that cons lists don't have constant time access, so you can't just binary search them. Ignore the second half of the above comment

nhuurre (approved):

PR #1507, opened 2025-03-26
Simplify some common patterns in Cpp DSL

WardBrian:
This introduces a couple helpers to make some things we do in the Cpp IR less verbose, similar to the existing helper syntax we've had previously. 

A good example is this snippet from `Lower_program`:

```ocaml
  Expression
    (Assign
       ( Var "vars"
       , Constructor
           (Types.std_vector Double, [Var "num_to_write"; Exprs.quiet_NaN]) ))
```
Becomes
```ocaml
  "vars" := (Types.std_vector Double).:(Var "num_to_write"; Exprs.quiet_NaN)
```

Both of these pieces of OCaml encode the following C++:

```C++
  vars = std::vector<double>(num_to_write,
            std::numeric_limits<double>::quiet_NaN());
```

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Copyright and Licensing

### DIFF
--- a/docs/cpp_ir.mld
+++ b/docs/cpp_ir.mld
@@ -37,7 +37,7 @@ Or, using the DSL constructs, the same expression could be written
 {[
 let open Cpp.DSL in
 let open Types in
-let vector = (row_vector Double).:(Literal "3") in
+let vector = (row_vector Double).:{Literal "3"} in
 let values = [Literal "1"; Var "a"; Literal "3"] in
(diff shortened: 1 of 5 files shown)


WardBrian:
@SteveBronder I agree with your major comments. I've removed the indexing shorthand and swiched to `{}` for construction and I agree it reads better than before.

SteveBronder (approved):

PR #1508, opened 2025-04-01
Stop generating unnecessary omni indexes in C++

WardBrian:
I ran into this while looking at #1295 and it seemed much easier to fix:

If we are using `--O1`, with code like:

```stan
  vector[2] x;
  x[:] = zeros_vector(2);
```
ends up generating an exception complaining that `x` has size `0`. 

This case is correctly (...ish, see #1295!) handled when the code is omits the omni indexing, and when
the _only_ indexing done is omni, the extra checks that get done are completely redundant, so we 
can in fact just omit them entirely. 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fixed a bug with assigning to an entire vector or matrix using `:` with the `--O1` flag.

## Copyright and Licensing


WardBrian:
Once this is merged I plan on adding the test added here to https://github.com/stan-dev/performance-tests-cmdstan/tree/master/optimizer-stress-models -- since this is a runtime error, testing the result of actually running it will be more useful

bob-carpenter:
I couldn't tell from the PR title what was being done here.  I hope the intention is to fix the bug so that the program you list works as intended.  In the end, `x[:]` is equivalent to `x[]` and `x`, so if the proposal's just to avoid generating a needless multi-index, that sounds like a great idea.

I'd look at not just `x[:]`, but also `x` and `x[]`, all of which are synonyms here.

P.S.  I had no idea that there is a `zeros_vector(n)` function!  I see `rep_vector(0, n)` everywhere in my own and others' code.

WardBrian:
> so if the proposal's just to avoid generating a needless multi-index, that sounds like a great idea.

Yes. If the left-hand-side of an assignment is indexed only with what we call "omni" indexes (blank or `:`, they're the same by this part of the compiler), we can leave them out of the generated C++ rather than pass them and end up in a different code path

SteveBronder (approved):

PR #1514, opened 2025-04-30
change to trigger tests

WardBrian:
Investigating a failure I got in #1439


WardBrian:
It was an issue with the `mother` model not putting a density on some of the parameters

Issue #361, opened 2019-10-25
AST refactor & handing user defined vs built-in functions / distributions


enetsee:
I've just started looking at the AST refactor and realized that the current definitions of expression patterns isn't going to play nicely with the `Fixed.Make` functor because we have a type parameter for function kinds:

```ocaml
type ('e, 'f) expression =
  | ...
  | FunApp of 'f * identifier * 'e list
  | CondDistApp of 'f * identifier * 'e list
  | ...
```

My suggestion for this would be:

- have separate union cases for user-defined and built-in functions
- provide a dictionary of built-in function names to the parser so we can syntactically decide if something is a built in function.

This would give us:

```ocaml
type 'e expression = 
   | ...
   | FunApp of identifier * 'e list 
   | UDFunApp of identifier * 'e list
   | ...
```

Since we would be making a change to expressions anyway, this would also be an opportunity be a good opportunity to pull out distributions and make it a separate syntactic object to expressions. Again, we need to provide a dictionary of built-in distributions to the parser. Would have the following for distributions:

```ocaml
type 'e dist = 
  | UDDist of identifier * 'e list 
  | Dist of identifier * 'e list
```

and replace `CondDistApp` with:

```ocaml
type 'e expression = 
  | ... 
  | CondDistApp of 'e * 'e dist
  | ...
```

This design is in-line with SlicStan's syntax and would pave the way for further work on redesigning how we handle groups of functions related to the same distribution and bijectors.

@VMatthijs @seantalts @rybern what are your thoughts? This is blocker for the AST refactor so would be great to agree a way forward.

seantalts:
I think it would be slightly more complicated than specific names - iirc users can override nullary function names, and possibly there are other weird things too... 

I like the separation of the AST as a purely syntactic construct. What do you think about   moving the AST to be more purely syntactic and either sorting this out in AST to Mir or adding another piece of metadata to everything in the decorated AST to hold function kind (and for most expressions it would be empty).

enetsee:
> I think it would be slightly more complicated than specific names - iirc users can override nullary function names, and possibly there are other weird things too...

Ah, I see. Out of interest, do you if that is in the language spec or a bug that people have made use of? I had a quick scan through but couldn't find anything on overloading other than the following:

> Two user-defined functions may not have the same name even if they have different sequences of argument types.

It seems _odd_ that users are allowed to shadow built-in functions but not their own!

[...]

PR #1538, opened 2025-08-12
Cleanup unused code in Common

WardBrian:
This ends up simplifying the MIR by relying less on indirection through functors and module inclusion. The type `Expr.t` is now actually defined in `Expr.ml`, rather than being massaged through a `Fixed` functor.

These functors were originally going to be used more (see #361), but this never happened, so I think it's better to remove the code and use the more direct style. No tests or behavior changes, this is just shuffling code around and removing stuff that was unnecessary afterwards. 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Copyright and Licensing

### DIFF
--- a/src/analysis_and_optimization/Mir_utils.ml
+++ b/src/analysis_and_optimization/Mir_utils.ml
@@ -12,12 +12,6 @@ let rec fold_expr ~take_expr ~(init : 'c) (expr : Expr.Typed.t) : 'c =
 
 let fold_stmts ~take_expr ~take_stmt ~(init : 'c)
     (stmts : Stmt.Located.t List.t) : 'c =
-  (* let rec fold_expr (state : 'c) (expr : Expr.t) =
-   *   Expr.Pattern.fold_left
-   *     ~f:(fun a e -> fold_expr (take_expr a e) e)
-   *     ~init:state
-   *     expr.pattern
-   * in *)
   let rec fold_stmt (state : 'c) (stmt : Stmt.Located.t) =
     Stmt.Pattern.fold
       (fun a e -> fold_expr ~take_expr ~init:(take_expr a e) e)
@@ -29,8 +23,7 @@ let rec num_expr_value (v : Expr.Typed.t) : (float * string) option =
   match v with
   (* internal type promotions should be ignored *)
   | {pattern= Pattern.Promotion (e, _, _); _} -> num_expr_value e
-  | {pattern= Pattern.Lit (Real, str); _} | {pattern= Pattern.Lit (Int, str); _}
-    ->
+  | {pattern= Pattern.Lit ((Real | Int), str); _} ->
       Some (float_of_string str, str)
   | {pattern= Pattern.FunApp (StanLib ("PMinus__", FnPlain, _), [v]); _} -> (
       match num_expr_value v with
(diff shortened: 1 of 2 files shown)


nhuurre:
Most of Dataflow types is gone. Were those types ever used? And what do you think about just moving the remaining ones into Dataflow utils?

> The type `Expr.t` is now actually defined in `Expr.ml`, rather than being massaged through a `Fixed` functor.

Then, in my opinion, it should be named `Expr.t` and not `Expr.Fixed.t` like it currently is. Let's just collapse the intermediate module `Expr.Fixed`. (and `Stmt.Fixed`)

WardBrian:
Actually, Dataflow_types is used in Mir_utils, which is opened by Dataflow_utils, so we can't just combine them without getting a circular dependency.

src/analysis_and_optimization/Mir_utils.ml
```diff
@@ -6,34 +6,33 @@ open Middle.Expr
...
-    ~init expr.pattern
+  Expr.Pattern.fold
+    (fun a e -> fold_expr ~take_expr ~init:(take_expr a e) e)
+    init expr.pattern
 
 let fold_stmts ~take_expr ~take_stmt ~(init : 'c)
     (stmts : Stmt.Located.t List.t) : 'c =
   (* let rec fold_expr (state : 'c) (expr : Expr.t) =
-   *   Expr.Fixed.Pattern.fold_left
+   *   Expr.Pattern.fold_left
```
nhuurre, line 16: `fold_left` doesn't exist anymore, but this code has been commented out for five years, I think it's time to remove it entirely.

src/analysis_and_optimization/Mir_utils.ml
```diff
@@ -6,34 +6,33 @@ open Middle.Expr
...
   List.fold ~f:(fun a s -> fold_stmt (take_stmt a s) s) ~init stmts
 
 let rec num_expr_value (v : Expr.Typed.t) : (float * string) option =
   match v with
   (* internal type promotions should be ignored *)
-  | {pattern= Fixed.Pattern.Promotion (e, _, _); _} -> num_expr_value e
-  | {pattern= Fixed.Pattern.Lit (Real, str); _}
-   |{pattern= Fixed.Pattern.Lit (Int, str); _} ->
+  | {pattern= Pattern.Promotion (e, _, _); _} -> num_expr_value e
+  | {pattern= Pattern.Lit (Real, str); _} | {pattern= Pattern.Lit (Int, str); _}
```
nhuurre, line 32:
```suggestion
  | {pattern= Pattern.Lit (Int | Real, str); _}
```

src/analysis_and_optimization/Mir_utils.ml
```diff
@@ -302,16 +301,16 @@ let rec fn_subst_expr m e =
   match m e with
   | Some e' ->
       (* let print_expr (e:Expr.Typed.t) = *)
-      (* [%sexp (e.pattern : Expr.Typed.t Expr.Fixed.Pattern.t)] |> Sexp.to_string *)
+      (* [%sexp (e.pattern : Expr.Typed.t Expr.Pattern.t)] |> Sexp.to_string *)
       (* in *)
```
nhuurre, line 305: More commented out code.

src/middle/Stmt.mli
```diff
@@ -1,65 +1,77 @@
...
+  -> g:(('a, 'b) t -> ('a, 'b) t)
+  -> ('a, 'b) t
+  -> ('a, 'b) t
+(** [rewrite_bottom_up] specializes [fold] so that the result type
+  ['r1] is equal to the type of the nested fixed-point type
+  i.e. ['r1 = 'a First.t] and the result type ['r2] is equal to the top-level
+  fixed-point type i.e. ['r2 = ('a,'b) t].
+
+  This also means that the function [f] can be written with our nested
+  fixed-point type  ['a First.t] as its argument and [g] can be written with
```
nhuurre, line 54: `First.t`  -> `Expr.t`.

PR #1541, opened 2025-09-04
Allow flags to be passed multiple times 

WardBrian:
This was supported by the old Stdlib.Arg.parse style of CLI, likely by accident. I'm not entirely sure we'd like to duplicate this old behavior, but not doing so could also lead to issues (cf stan-dev/cmdstanr#1111)

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Command line flags like `--use-opencl` can be passed multiple times once more. This is equivalent to passing them exactly once, and is only provided for backwards compatibility with existing code that may not properly prepare the arguments to the compiler.

## Copyright and Licensing

### DIFF
--- a/src/stanc/CLI.ml
+++ b/src/stanc/CLI.ml
@@ -2,9 +2,6 @@ open Core
 open Frontend
 open Cmdliner
 
-let flag_all_to_bool (flags : bool list Term.t) : bool Term.t =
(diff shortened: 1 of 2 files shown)


WardBrian:
Marking as draft because I'm still unsure if this behavior is actually desirable or not

PR #1545, opened 2025-09-09
Statically forbid promotion nodes in the untyped AST

WardBrian:
This adds another type parameter to the ast which can be made empty for the untyped ast, allowing us to replace a couple ICEs in the happy path with refutation cases (`.`) that the compiler proves unreachable. 

This has bugged me since adding the Promotion node a while back.
The main annoyance with doing this is that it would add another (broadly useless) argument to `@@deriving` functions like `map_expression` (and `fold_...`). I get around this by re-defining those helper functions towards the bottom of the file without the arguments we don't use, and cleaning up the usage sites. 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Copyright and Licensing

### DIFF
--- a/src/frontend/Ast.ml
+++ b/src/frontend/Ast.ml
@@ -56,18 +56,18 @@ type ('e, 'f, 'p) expression =
 
 type ('m, 'f, 'p) expr_with =
   {expr: (('m, 'f, 'p) expr_with, 'f, 'p) expression; emeta: 'm}
-[@@deriving sexp, compare, map, hash, fold]
+[@@deriving sexp, compare, hash]
 
 (** Untyped expressions, which have location_spans as meta-data *)
 type located_meta = {loc: (Location_span.t[@sexp.opaque] [@compare.ignore])}
-[@@deriving sexp, compare, map, hash, fold]
+[@@deriving sexp, compare, hash]
 
 (** Uninhabited type. This is supplied as an argument to the Promotion node in
     untyped ASTs so the compiler can deduce it is impossible *)
 type empty = | [@@deriving sexp, compare, hash]
 
 type untyped_expression = (located_meta, unit, empty) expr_with
-[@@deriving sexp, compare, map, hash, fold]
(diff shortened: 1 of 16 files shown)


src/frontend/Ast.ml
```diff
@@ -54,31 +54,35 @@ type ('e, 'f) expression =
...
 (** Untyped expressions, which have location_spans as meta-data *)
 type located_meta = {loc: (Location_span.t[@sexp.opaque] [@compare.ignore])}
-[@@deriving sexp, compare, map, hash, fold]
+[@@deriving sexp, compare, hash]
 
-type untyped_expression = (located_meta, unit) expr_with
-[@@deriving sexp, compare, map, hash, fold]
+(** Uninhabited type. This is supplied as an argument to the Promotion node in
+    untyped ASTs so the compiler can deduce it is impossible *)
+type empty = | [@@deriving sexp, compare, hash]
```
nhuurre, line 67: There's `Core.Nothing.t`, why create a new type?

test/unit/Error_tests.ml
```diff
@@ -27,54 +26,17 @@ let%expect_test "backtrace indirect test" =
...
-                                        ; emeta } } ] }
-                    ; smeta } ]
-              ; xloc }
-      ; comments= [] } in
   Printexc.record_backtrace false;
-  ICE.with_exn_message (fun () -> Typechecker.check_program ast)
+  ICE.with_exn_message (fun () ->
+      Middle.(
+        Expr.Helpers.infer_type_of_indexed UnsizedType.UReal
+          [Index.Single Expr.Helpers.loop_bottom]))
```
nhuurre, line 33: So this test is just looking for some random ICE?

src/frontend/Ast.ml
WardBrian: Ah, I searched around in stdlib and came up empty, neglected to look as closely through Core

PR #1552, opened 2025-09-18
Improve semantic error messages

WardBrian:
This is the first piece of something like #1549, but for the semantic errors this time. For now, just some minor edits to the wording of a few messaging and some bikeshed-y moving of errors between the categories that the Semantic_error module uses internally. 

I also updated the internal definition of `Semantic_error.t` and `Syntax_error.t` to 'lift' the location up. This makes it look a bit more like Warning.t and saves some duplicated type information

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Updated the error messages for type errors to be more consistent with one another. 

## Copyright and Licensing

### DIFF
--- a/src/frontend/Semantic_error.ml
+++ b/src/frontend/Semantic_error.ml
@@ -310,31 +310,20 @@ module TypeError = struct
           "A non-returning function was expected but a non-function value '%s' \
            was supplied."
           fn_name
-    | ReturningFnExpectedUndeclaredIdentFound (fn_name, sug) -> (
-        match sug with
-        | None ->
-            Fmt.pf ppf
-              "A returning function was expected but an undeclared identifier \
-               '%s' was supplied."
-              fn_name
-        | Some s ->
-            Fmt.pf ppf
-              "A returning function was expected but an undeclared identifier \
-               '%s' was supplied.@ A similar known identifier is '%s'"
-              fn_name s)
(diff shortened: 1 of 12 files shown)


src/frontend/Semantic_error.ml
```diff
@@ -597,327 +601,302 @@ module StatementError = struct
...
-  | IdentifierError of Location_span.t * IdentifierError.t
-  | ExpressionError of Location_span.t * ExpressionError.t
-  | StatementError of Location_span.t * StatementError.t
+type err =
+  | TypeError of TypeError.t
+  | IdentifierError of IdentifierError.t
+  | ExpressionError of ExpressionError.t
+  | StatementError of StatementError.t
+
+type t = Location_span.t * err
```
nhuurre, line 626: It's a bit unsightly that `Semantic_error.t` is now an opaque tuple (opaque since users are expected to call our `location` instead `fst` directly) and `pp` only prints half of it.

nhuurre (approved):

src/frontend/Semantic_error.ml
WardBrian: pp only printed half of of the actual content of the type before, it was just one level deeper.

I played around with having pp do the entire job, but it isn't clear where the shared code between Semantic_error and Syntax_error for printing the context would go. It can't stay in Errors, since that would introduce a circular dependency, and it can't be moved entirely into Middle.Location, since it relies on Frontend.Include_files

PR #1556, opened 2025-10-03
Fix/stan3356 tuple overloading codegen

WardBrian:
Closes stan-dev/stan#3356. 

Relies on https://github.com/stan-dev/math/pull/3244
#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Improved code generation of user-defined functions accepting tuples to resolve some C++ compilation errors.

## Copyright and Licensing


WardBrian:
Lessons:
- You cannot use `std::tuple_element<N, T>` in an evaluated context unless you are confident `tuple_size_v<T> < N`, as this triggers static asserts otherwise
- You cannot use `std::tuple_size<T>::value` in an evaluated context unless you are sure `is_tuple_v<T>`, as this leads to errors about incomplete typenames.

With sufficient indirection it is possible to avoid both of these errors by moving them into unevaluated contexts guarded by the proper `requires` templates, but it is not obvious

PR #1513, opened 2025-04-18
Add functors for the embedded Laplace approximation

WardBrian:
This is a draft until https://github.com/stan-dev/math/pull/3097 is updated and merged. It could probably also do with some more code-gen tests. 

Similar to reduce_sum, these functions have too many rules to fit into the general variadic framework, so they have custom typechecking and errors, which make up most of this PR.

Closes #1511 

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: Issue https://github.com/stan-dev/docs/issues/873

## Release notes

Added new functions for the embedded Laplace approximation.

## Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause)

PR #1562, opened 2025-10-16
Tweak the pretty-printing of long function definitions 

WardBrian:
This applies a similar logic to what was added for function _calls_ in #1513. If the function name is long, start the box for it's arguments earlier to allow it to be raggedly formatted below rather than forcing it to indent all the way to the `(`.

This also makes it so each argument gets its own line when there are many arguments -- this helps with readability, IMO, and is common in formatters in other languages

Consider this example from @nsiccha:

```stan
// current formatting
  array[] int vote_countvote_countvote_countvote_count(array[] int rating,
                                                       array[] int item,
                                                       array[] int rater,
                                                       int I, int J) {
    int N = size(rating);
    array[I] int count_by_item = rep_array(1, I); // index 0:5 by 1:6
    for (n in 1 : N) {
      count_by_item[item[n]] += rating[n];
    }
    array[J + 1] int count = rep_array(0, J + 1);
    for (i in 1 : I) {
      count[count_by_item[i]] += 1;
    }
    return count;
  }
```

```stan
// this pr
  array[] int vote_countvote_countvote_countvote_count(
    array[] int rating,
    array[] int item,
    array[] int rater,
    int I,
    int J
  ) {
    int N = size(rating);
    array[I] int count_by_item = rep_array(1, I); // index 0:5 by 1:6
    for (n in 1 : N) {
      count_by_item[item[n]] += rating[n];
    }
    array[J + 1] int count = rep_array(0, J + 1);
    for (i in 1 : I) {
      count[count_by_item[i]] += 1;
    }
    return count;
  }

```

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes
Improved the formatting of long functions (either with long names or many arguments).

## Copyright and Licensing

### DIFF
--- a/src/frontend/Pretty_printing.ml
+++ b/src/frontend/Pretty_printing.ml
@@ -342,7 +342,7 @@ let pp_bracketed_transform ppf = function
     ->
       ()
 
-let rec pp_transformed_type ?end_loc () ppf (st, trans) =
+let rec pp_transformed_type ppf (st, trans) =
   let open Middle in
   let pp_possibly_transformed_type ppf (st, trans) =
     let sizes_fmt =
@@ -387,17 +387,15 @@ let rec pp_transformed_type ?end_loc () ppf (st, trans) =
         (* it isn't perfect with comment placement currently *)
         let transTypes = Middle.Utils.zip_stuple_trans_exn st transforms in
         pf ppf "tuple(@[%a%s@])"
-          (list ~sep:comma (pp_transformed_type ?end_loc ()))
+          (list ~sep:comma pp_transformed_type)
           transTypes
           (if List.length transforms = 1 then "," else "") in
   match st with
   (* array goes before something like cov_matrix *)
   | Middle.SizedType.SArray _ ->
       let ty, ixs = Middle.SizedType.get_array_dims st in
-      let end_loc =
-        match end_loc with
-        | Some x -> x
-        | None -> (List.last_exn ixs).emeta.loc.end_loc in
+      let ({emeta= {loc= {end_loc; _}; _}; _} : untyped_expression) =
+        List.last_exn ixs in
       let ({emeta= {loc= {begin_loc; _}; _}; _} : untyped_expression) =
         List.hd_exn ixs in
       pf ppf "array[@[%a@]]@ %a" pp_list_of_expression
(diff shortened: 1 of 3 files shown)


src/frontend/Pretty_printing.ml
```diff
@@ -472,17 +472,36 @@ and pp_statement ppf ({stmt= s_content; smeta= {loc}} as ss : untyped_statement)
...
-      | b -> pf ppf ") %a" pp_statement b)
-
-and pp_args ppf (at, ut, id) =
-  pf ppf "%a%a %a" pp_autodifftype at pp_unsizedtype ut pp_identifier id
+      let pp_dedent ppf =
+        (* TODO: use pp_print_custom_break for trailing commas? *)
+        if long then Format.pp_print_break ppf 0 (-2) else () in
+      let body = match b with {stmt= Skip; _} -> None | _ -> Some b in
+      pf ppf "%t%a %a%t(%a%a%t)@]%a" pp_open_long_box pp_returntype rt
+        pp_identifier id pp_open_short_box (if' long cut) ()
```
nhuurre: So the box closes at `@]` but it opens inside one of those `%t`. (which one? decided at runtime!) And we have negative indentation? ~~Thanks, I hate it.~~

Can we have a toplevel `if long then pf ppf ... else pf ppf ...` instead of these `Fmt.if'`s? I think that will be a lot easier to read, even if it causes some code duplication. And two nested boxes so we don't need negative indentation.

test/integration/good/overloading/pretty.expected
```diff
@@ -214,13 +214,20 @@ functions {
...
+                    real kappa,
+                    // C_{50}
+                    real gamma,
+                    // recovery rate
+                    real xi,
+                    // bacteria production rate
+                    real delta,
+                    // bacteria removal rate
+                    int unused
+                    // just for testing
```
nhuurre: This is rather unfortunate. Makes it look like the comment applies to next item.

test/integration/good/pretty.expected
```diff
@@ -1365,9 +1365,11 @@ data {
 transformed data {
   int N;
   array[N /*test1*/, //test1.5
-        N] real arr;
+        N] real /*test2*/ arr;
```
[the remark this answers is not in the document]
WardBrian: Well, at least it is an improvement... but you're right, let me see what I can do

test/integration/good/overloading/pretty.expected
WardBrian: I managed to get them back on the same line as the commas (it still doesn't know what the user intended, but at least it preserves previous behavior)

test/integration/good/pretty.expected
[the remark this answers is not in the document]
WardBrian: Got it (and even fixed the other place our tests currently had a `^^^`-marked comment)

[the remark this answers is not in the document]
nhuurre: You tried.
```stan
    array[// 1
    3] // 2
    vector[2 // 3
    ] // 4
    x;
```

src/frontend/Pretty_printing.ml
```diff
@@ -436,8 +438,9 @@ and pp_statement ppf ({stmt= s_content; smeta= {loc}} as ss : untyped_statement)
     =
   match s_content with
   | Assignment {assign_lhs= l; assign_op= assop; assign_rhs= e} ->
-      pf ppf "@[<h>%a %a %a;@]" pp_lvalue l pp_assignmentoperator assop
-        pp_expression e
+      pf ppf "@[<h>%a %a%a %a;@]" pp_lvalue l
+        (pp_comments_spacing get_comments_until_separator)
```
nhuurre: Assignment operators aren't separators so `comment 6` in `tuples./pretty.expected` ends up on the wrong side of it.

PR #1564, opened 2025-10-17
Fix an ICE with include files at the ends of blocks

WardBrian:
Ran into this today, the issue seems to be that the `functions { }` was being omitted entirely. This fix seems simple, and doesn't change any other output, so, maybe it's that easy?

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fixed a crash with nested includes

## Copyright and Licensing


WardBrian:
I’m also suspicious about why the nested include matters. I wouldn’t be surprised if there is a flaw in the preprocessor’s location code that is only exposed when you pop two buffers in a row (though, I tried finding such an issue and could not)

PR #1575, opened 2025-12-10
Try using archival-20260101 repo

WardBrian:
CI runs for https://discuss.ocaml.org/t/opam-repository-archival-next-run-scheduled-2026-01-01/17587


WardBrian:
Looks like we will be fine after requesting the preservation of a couple extra packages.

PR #1578, opened 2026-01-05
Fold @update_messages into build

WardBrian:
Now that there's no additional python requirement, I think it makes sense to just run the `.messages` diff during the build. This won't run often, thanks to build caching, but will prevent someone forgetting when they edit the parser.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Copyright and Licensing

### DIFF
--- a/src/frontend/parser.messages
+++ b/src/frontend/parser.messages
@@ -6,8 +6,9 @@
 # Notably, each example program below is just that -- ONE example of a
 # program ending in that state.
 #
-# When the parser is updated, building the compiler will # automatically update
-# this file, leaving new placeholder messages for new states.
+# When the parser is updated, building the compiler will check this file
+# for completeness and allow you to automatically update it with `dune promote`,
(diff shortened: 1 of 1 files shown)


src/frontend/parser.messages
```diff
@@ -6,8 +6,8 @@
 # Notably, each example program below is just that -- ONE example of a
 # program ending in that state.
 #
-# When the parser is updated, the command `dune build @update_messages` will
-# automatically update this file, leaving new placeholder messages for new states.
+# When the parser is updated, building the compiler will # automatically update
+# this file, leaving new placeholder messages for new states.
```
nhuurre: I tried this and while `dune build` does print the missing messages to console it does not update this file automatically (that requires `dune promote`). Which is good; it would be a bad idea to modify the source when someone might run `dune build` twice; the second time should fail the same way the first did rather than accept the added placeholders as real messages.

Also there's a stray `#`.

PR #1580, opened 2026-01-12
Small syntax error improvements

WardBrian:
Three small things:

1. Make an error message about bad declarations more specific.
2. Add some more example models that have 'improvable' errors (it's difficult to improve with the current menhir tech, but there are other options that could do)
3. Improve the error reporting when the error is at the end of the file. Currently, the `^` points at the line before, which can be confusing.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Improved syntax errors in some cases.

## Copyright and Licensing

### DIFF
--- a/src/frontend/parser.messages
+++ b/src/frontend/parser.messages
@@ -1381,8 +1381,7 @@ program: MODELBLOCK LBRACE REAL IDENTIFIER ASSIGN IDENTIFIER TILDE
 ## In state 126, spurious reduction of production expression -> common_expression
 ##
 
-@{<light_red>Ill-formed declaration.@} Found a declaraction before this point.
-If it was complete, expected a @{<green>";"@} to complete the definition or a @{<green>","@} to start a new declaration of the same type.
+@{<light_red>Ill-formed declaration.@} Expected a @{<green>";"@} to complete the definition or a @{<green>","@} to start a new declaration of the same type.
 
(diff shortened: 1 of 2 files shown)


src/frontend/parser.messages
```diff
@@ -1342,7 +1342,48 @@ program: MODELBLOCK LBRACE MATRIX WHILE
...
+## WARNING: This example involves spurious reductions.
+## This implies that, although the LR(1) items shown above provide an
+## accurate view of the past (what has been recognized so far), they
+## may provide an INCOMPLETE view of the future (what was expected next).
+## In state 111, spurious reduction of production common_expression -> identifier
+## In state 126, spurious reduction of production expression -> common_expression
+##
+
+@{<light_red>Ill-formed declaration.@} Found a declaraction before this point.
+If it was complete, expected a @{<green>";"@} to complete the definition or a @{<green>","@} to start a new declaration of the same type.
```
nhuurre:
```suggestion
@{<light_red>Ill-formed declaration.@} Expected either a @{<green>";"@} to close the definition or a @{<green>","@} to start a new declaration of the same type.
```
"Found a declaration" immediately after "Ill-formed declaration" is redundant and I don't think you need "If it was complete" either.

WardBrian: I was trying to guard against the possibility that the correct solution was to add a `+` or something, but it seems like `real foo = bar baz;` triggers a different error anyhow.

Issue #1482, opened 2024-12-18
Allowing `_lp` in the transformed parameters block and `jacobian` in the model block


spinkney:
The pr #979 added the `jacobian +=` and `_jacobian` function keyword to indicate adding the change of variables correction for a transformed parameter. The `jacobian` functionality can only be used in the transformed parameters block. That same pr also introduced a warning that `_lp` type functions will no longer be allowed in the transformed parameters block and will only be supported in the model block. I think the introduction of the `jacobian` functionality is useful but I believe restricting the `_lp` to only the model block is a step backwards for Stan. 
 
In merging #979, which users will first see in cmdstan 2.36, any user defined function ending with `_lp` that is used in the transformed parameters block will receive a warning that this functionality will not be allowed in future releases. The initial reasoning in the discussion for #979 is that allowing functions to access the `target` was a bug in stanc2 that we continued to support in stanc3. The merged pr is indicating the "fix". However, there are times where it is useful to both add the jacobian correction and to add a log density. I don't believe this is widespread but in matrix parameters I encounter it. For example, below I show the c-vine parameterization of a correlation matrix. I use both the `jacobian` and the `target` keywords here. I'm not sure this will compile because I believe it would need to end in `_jacobian`. I have been using `target` everywhere where `jacobian` now is which does compile in previous versions.   

```stan
  vector lb_ub_lp (vector y, real lb, real ub) {
    target += log(ub - lb) + log_inv_logit(y) + log1m_inv_logit(y);
    return lb + (ub - lb) * inv_logit(y);
  }
  
  real lb_ub_lp (real y, real lb, real ub) {
    target += log(ub - lb) + log_inv_logit(y) + log1m_inv_logit(y);
    return lb + (ub - lb) * inv_logit(y);
  }
  
  matrix corr_constrain_cvine_lp (vector col_one_raw, vector off_raw, real eta) {
    int K = num_elements(col_one_raw) + 1;
    vector[K - 1] z = lb_ub_lp(col_one_raw, -1, 1);
    matrix[K, K] C = identity_matrix(K);
    matrix[K, K] P;
    C[1, 1] = 1;
    C[2:K, 1] = z[1:K - 1];
    C[1, 2:K] = C[2:K, 1]';
    P = C;
    int cnt = 1;
    
    target += beta_lpdf(0.5 * (z + 1) | eta + 0.5 * (K - 1), eta + 0.5 * (K - 1));
[...]

PR #1584, opened 2026-01-20
 Expire deprecation on _jacobian functions

WardBrian:
This expires the _jacobian deprecation and removes a lot of the logic needed for compatibility with old code (which, as an aside, I think was buggy)

Because of https://github.com/stan-dev/stanc3/issues/1482, I'm leaving the _lp deprecation for now. 

#### Submission Checklist

- [x] Run unit tests
- Documentation
    
## Release notes

Starting in this version, functions ending in `_jacobian` will always be treated as needing access to the jacobian adjustment, and thus will only be usable in `transformed parameters` and other `_jacobian` functions.

## Copyright and Licensing

### DIFF
--- a/src/stan_math_backend/Cpp.ml
+++ b/src/stan_math_backend/Cpp.ml
@@ -228,6 +228,7 @@ type stmt =
   | ForEach of (type_ * identifier) * expr * stmt
   | While of expr * stmt
   | IfElse of expr * stmt * stmt option
+  | ConstexprIf of type_ * stmt
   | TryCatch of stmt list * (type_ * identifier) * stmt list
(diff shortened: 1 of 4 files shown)


src/stan_math_backend/Cpp.ml
```diff
@@ -228,6 +228,7 @@ type stmt =
   | ForEach of (type_ * identifier) * expr * stmt
   | While of expr * stmt
   | IfElse of expr * stmt * stmt option
+  | ConstexprIf of type_ * stmt
```
nhuurre: It's `constexpr` not `consttype` so why is the condition field a `type_`?

PR #1587, opened 2026-02-06
Make the syntax error when declaring a function with constrained arguments more specific

WardBrian:
c.f. https://discourse.mc-stan.org/t/passing-an-array-simplex-to-a-function/40900

This may not always be 100% correct advice, but being more specific would require a different error generation strategy like [lrgrep](https://github.com/let-def/lrgrep). It does seem like by far the most likely way

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Improved some parsing errors when a user tries to specify a function argument with a constrained type.

## Copyright and Licensing


WardBrian:
> In my opinion the parser should just accept constrained types in all variable declarations and defer the error to typechecking. That way the error for model { simplex[3] x; can explain that constraints aren't enforced here and suggest vector[3] instead.

I have spent some time trying this, and gave up for the following reasons:

- It's tricky to make sure nothing is sneaking through in the typechecking level
- While it does improve a lot of situations, it unfortunately degrades the errors even further when there is a more basic syntax error, since a lot of different states all end up in the same case. For example, someone mis-spelling `simplax` has to just say "expected a type" rather than specifying if it should be constrained or not, since they're going through the same parser states

> How do you feel about UNREACHABLE token hack?

It's unfortunate that this would have to list every single constrained type token separately to get what we'd want here, (plus arrays, tuples thereof...). Something like LRgrep would let us do this by specifying "If you're in an error and the prefix is reducible to top_level_type".

WardBrian:
> In my opinion the parser should just accept constrained types in all variable declarations and defer the error to typechecking. That way the error for model { simplex[3] x; can explain that constraints aren't enforced here and suggest vector[3] instead.

This is basically what we're doing right now for complex_vector etc. (and it's #614, right?). 
I'd be happy to see it done, but my attempt just now to do so failed because:

- it was surprisingly tricky to make sure nothing 'sneaks through' the typechecker (my first couple attempts rejected existing 'good' tests or accepted existing 'bad' tests). This is ultimately fixable.
- it degrades a lot of _other_ error messages, since the states for 'expected a basic type' and 'expected a constrained type' get merged, so if you see `simplax foo;` the error message can't specify what was expected beyond "a type". I'm not sure this is resolvable. 

> How do you feel about UNREACHABLE token hack?

I'm not inherently opposed to it, but to get really high-quality errors in the situation you're highlighting we would need to make UNREACHABLE states for every possible constrained return type attempt (plus one for sized arrays/tuples of the preceeding...)

This is why I find something like [LRgrep](https://cambium.inria.fr/~fpottier/publis/bour-pottier-lrgrep-2026.pdf) very promising, since we can instead write an 'error rule' that sounds something like "If we error while trying to produce a `function_def` and the prefix of the stack could reduce to `top_var_type`, return some specific message"

All of the `// TODO: would be nice to specialize for this mistake` comments in tests are things I've mentally marked as worth revisiting if/when I get a chance to convert our errors to that style, since it becomes trivial to specialize in those situations

nhuurre (approved):

PR #1603, opened 2026-03-03
Fix pedantic mode false prior counting for distribution arguments

### LINKED ISSUE #746: pedantic mode incorrectly warning about 2 priors?
@rybern Using the latest CmdStanR from github I was trying out pedantic mode (which is really cool!) and I got an unexpected warning:

```r

stan_file <- write_stan_file(
"parameters {
  real x;
  real sigma;
}
model {
  x ~ normal(0, sigma);
  sigma ~ exponential(1);
}"
)
mod <- cmdstan_model(stan_file, stanc_options = list("warn-pedantic"=TRUE))
```
```
Compiling Stan program...
Warning:
  The parameter sigma has 2 priors.
Warning at '/var/folders/h6/14xy_35x4wd2tz542dn0qhtc0000gn/T/RtmpwDnKa2/model-e66b730e6be7.stan', line 6, column 16 to column 21:
  A normal distribution is given parameter sigma as a scale parameter
  (argument 2), but sigma was not constrained to be strictly positive.
Warning at '/var/folders/h6/14xy_35x4wd2tz542dn0qhtc0000gn/T/RtmpwDnKa2/model-e66b730e6be7.stan', line 7, column 2 to column 7:
  Parameter sigma is given a exponential distribution, which has strictly
  positive support, but sigma was not constrained to be strictly positive.
```

The second two warnings make sense but the first one about sigma having 2 priors isn't correct. There's only one prior on sigma. Is this a bug?

ishaan-arora-1:
When a parameter appears as an argument to another parameter's distribution (e.g., sigma in 'x ~ normal(0, sigma)'), pedantic mode was incorrectly counting it as receiving a prior from that factor.

The fix adds an is_factor_subject check to fg_var_priors: for _lpdf/ _lpmf distribution factors, only the first argument (the distribution subject) is considered as receiving a prior. Parameters appearing in other argument positions are distribution arguments, not subjects.

This resolves the false warning 'sigma has 2 priors' when sigma only has one actual prior (sigma ~ exponential(1)) and merely appears as a scale argument in another distribution.

Fixes #746

#### Submission Checklist

- [ ] Run unit tests
- Documentation

## Release notes

## Copyright and Licensing


nhuurre:
Looks ok to me too but unfortunately I'm not familiar with pedantic analysis code either. Maybe reword the warning that says "This means either no prior is provided, or the prior(s) depend on data variables." since this PR is supposed to reduce "data-dependent prior" problems. Or was this PR was only intended to fix the example from #746 which has no data at all? Note that `x ~ normal(0, sigma)` was counted as a prior for `sigma` only if `x` has no other connection to data. Which makes me think of another example where the factor graph heuristic for priors gets it wrong:
```stan
data {
  real a, b;
}
parameters {
  real x, y;
}
model {
  x ~ normal(0,1);
  y ~ normal(x,1); // warning: y has no priors
  a ~ normal(x,1); // unless you delete this line
  b ~ normal(y,1);
}
```

PR #1615, opened 2026-04-15
Fix included positions being wrong for the first token in a new file

### LINKED ISSUE #1614: [BUG] File with includes for entire functions and data blocks fails to auto-format
Hey all, it has been awhile :) 

### Current Behavior:

I have a model where the functions and data blocks are in separate files and I include them in the top-level model. Minimal example is:

```
#include functions.stan
#include data.stan
```

where functions.stan is:
```stan
functions {
  real foo() {
    return 1.0;
  }
}
```
and data.stan is:
```stan
data {
  int N;
}
```

When `--auto-format` is supplied, `stanc` fails with 

```
./bin/stanc --auto-format /home/rstudio/bug_example.stan --include-paths=.
Internal compiler error:
(Invalid_argument "pos + len past end: 0 + 17 > 6")
Raised at Stdlib.invalid_arg in file "stdlib.ml", line 30, characters 20-45
Called from Base__Ordered_collection_common0.check_pos_len_exn in file "src/ordered_collection_common0.ml" (inlined), line 31, characters 7-53
Called from Base__String.sub in file "src/string.ml", line 58, characters 4-84
Called from Middle__Location.pp_context_for.cursor_line in file "src/middle/Location.ml", line 31, characters 6-53
Called from Stdlib__Format.output_acc in file "format.ml", line 1295, characters 4-20
Called from Stdlib__Format.output_acc in file "format.ml", line 1308, characters 32-48
Called from Stdlib__Format.kfprintf.(fun) in file "format.ml", line 1356, characters 16-34
Called from Stdlib__Format.kasprintf.k in file "format.ml", line 1391, characters 4-22
Called from Frontend__Pretty_print_prog.check_correctness in file "src/frontend/Pretty_print_prog.ml", line 58, characters 10-72
Called from Frontend__Pretty_print_prog.pretty_print_program in file "src/frontend/Pretty_print_prog.ml", line 77, characters 2-44
Called from Driver__Entry.stan2cpp in file "src/driver/Entry.ml", line 83, characters 9-383
Called from Dune__exe__Stanc.stanc.(fun) in file "src/stanc/stanc.ml", line 67, characters 4-161
Called from Base__With_return.with_return in file "src/with_return.ml", line 21, characters 12-24
Re-raised at Base__With_return.with_return in file "src/with_return.ml", line 29, characters 12-21
Called from Common__ICE.with_exn_message in file "src/common/ICE.ml", line 13, characters 9-15


This should never happen. Please file a bug at https://github.com/stan-dev/stanc3/issues
and include this message and the model that caused this issue.
```

stanc runs fine without the --auto-format flag. It also runs fine if the top model is

```stan
functions {
  #include functions.stan
}
data {
  #include data.stan
}
```
and the included files have the blocks removed.

### Environment:
I tried it on both a linux machine and macOS, same problem.

WardBrian:
Closes #1614.

After staring at the results of `--debug-parse --debug-lex` on @rok-cesnovar's example, I realized that this was the issue, the very first token was reporting the wrong location.

It turned out that us calling `lexer_lexbuf_to_supplier` was to blame, as this function interacted badly with the fact that our tokenizer can internally change lexbufs. The new `input` in `Parse.ml` is based on it, but accounts for this fact.

It turns out that almost all of our previous location-munging was unnecessary once this small fix was applied. 

This PR does change the location of two existing comments in our test cases. As far as I can tell, one is a previously-correct placement becoming incorrect, and the other is a previously-incorrect placement becoming correct. *shrug*

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fixed a bug in the parser that could lead to crashes when auto-formatting files with many `#includes`.

## Copyright and Licensing


nhuurre:
> one is a previously-correct placement becoming incorrect, and the other is a previously-incorrect placement becoming correct.

Perfectly balanced, as all things should be.

The one becoming is incorrect is
```stan
data {
    #include data.stan
}
```
where `data.stan` is
```stan
real x; // comment
```
Interestingly, a blank line makes the problem go away
```stan
data {
    #include data.stan

}
```

test/integration/cli-args/canonicalize/include/stanc.expected
```diff
@@ -16,6 +16,69 @@ void bar(real y) {
...
 // another line comment b
 [exit 0]
+  $ ../../../../../../install/default/bin/stanc --auto-format --include-paths=. blocks-issue-1614.stan
+Internal compiler error:
+("Pretty-printed program failed to parse"
+  "Syntax error in './data.stan', line 1, column 0, included from\
+ \n'string', line 5, column 2, parsing error:\
+ \n   -------------------------------------------------\
+ \n     1:  data {\
+ \n         ^\
```
nhuurre: I notice that this failing test gives a relatively clean ICE that includes the parsing error but Rok's example double-ICEd while trying to print the error location, from here:

https://github.com/stan-dev/stanc3/blob/d9e4aa6c95f71fa56127e1af76c8261bdefd3207/src/middle/Location.ml#L31

I don't know what causes that difference but just in case this should probably be
```ocaml
      let len = Int.min col_num (String.length highlighted_line) in
      String.sub highlighted_line ~pos:0 ~len
```

WardBrian:
Thanks for the mwe @nhuurre. It is curious, the locations on the parsed file actually all look correct to me, so I think that comment being misplaced is due to the pretty printing code misbehaving, not a parser/lexer issue at this point.

The only difference adding a newline makes is updating the `end_loc` of `prog.data_block.xloc`, which is expected.

nhuurre:
Yes, I assume it's something like, `pp_list_of_statements` forgets to make a trailing `pp_spacing` call when end-of-block location is too close.

WardBrian:
I think it's actually as simple as deleting this other 'historical' workaround:
https://github.com/stan-dev/stanc3/blob/d9e4aa6c95f71fa56127e1af76c8261bdefd3207/src/middle/Location.ml#L69-L75

WardBrian:
> Perfectly balanced, as all things should be.

Bad(?) news, this PR is now strictly out-of-balance and only improves comment positions relative to master

nhuurre (approved):

rok-cesnovar:
Thanks guys! Really appreciate it.

PR #1627, opened 2026-05-27
Improve consistency of assignment rhs and lhs overlap checks

martinmodrak:
Currently, the function `Ast.extract_ids` does not extract variable IDs that appear within an indexing expression. This results in inconsistencies when it is used - I've seen two cases:

The "Assignment of variable to itself." warning is triggered by:
```stan
  array[2] real bar = bar;
```

but not by

```stan
  array[2] int n;
  int k = n[k];
```

And that the "The same variable cannot be both assigned to and read from on the left hand side of an assignment" error is triggered by

```stan
array[2] int x = {1,2};
x[x[1]] = 2;
```

but not by

```stan
array[2] int n = {1, 2};
array[2] int x = {1,2};
x[n[x[1]]] = 2;
```

In this PR I propose to fix this by emitting warning/error also when indexing is used. I think that the warning is clearly justified, though @WardBrian mentioned that emitting the error might not actually be necessary when the expression referencing the lhs has to be evaluated before anything is written to the lhs. 

If that's the case, then it would be consistent to not single out indices, but not emit the error also for references in function calls, distribution calls etc.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] no user-facing changes were made

## Release notes

Improved handling of some edge cases when the same variable is used on both left- and right- hand side of an assignment.

## Copyright and Licensing

### DIFF
--- a/src/frontend/Ast.ml
+++ b/src/frontend/Ast.ml
@@ -331,11 +331,17 @@ let rec expr_of_lvalue {lval; lmeta} =
       | LTupleProjection (l, i) -> TupleProjection (expr_of_lvalue l, i))
   ; emeta= lmeta }
 
+let exprs_in_index = function
+  | All -> []
+  | Single e -> [e]
+  | Upfrom e -> [e]
+  | Downfrom e -> [e]
+  | Between (e1, e2) -> [e1; e2]
+
 let rec extract_ids {expr; _} =
   match expr with
   | Variable id -> [id]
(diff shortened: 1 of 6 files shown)


WardBrian (approved):
Thanks

I think this is probably the safer default, though it does seem like it could actually be useful and well defined to do this in some cases. I guess all this would require is an intermediary variable, so it's not a huge deal

PR #1632, opened 2026-06-30
jenkins: update and migration to new system

dylex:
Migration to new jenkins system and significant cleanup. Mostly tested [here](https://jenkins-new.flatironinstitute.org/job/CCM/job/Stan/job/stanc3/job/jenkins-new/)

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,7 +1,7 @@
 codecov:
   require_ci_to_pass: false
   ci:
-    - jenkins.flatironinstitute.org
+    - jenkins-new.flatironinstitute.org
 
 coverage:
   status:
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,982 +1,467 @@
-@Library('StanUtils')
-import org.stan.Utils
-
-utils = new org.stan.Utils()
-
-def skipExpressionTests = false
-def skipRemainingStages = false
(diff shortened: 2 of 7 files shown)


WardBrian:
I'm guessing this is not in fact a miracle and that these tests are just being skipped or something?
image

dylex:
Yes, when they're run they look like [this](https://jenkins-new.flatironinstitute.org/job/CCM/job/Stan/job/stanc3/job/jenkins-new/40/stages/?selected-node=96). I can add another if to short-circuit that earlier if you want.

By the logic there now, compile only runs if files in `test/integration/good` have changed, and math only if `test/integration/signatures/stan_math_signatures.t` changed. Binaries build if `src` or `Jenkinsfile` changed, and since this PR only changed Jenkinsfile, that's pretty much all that ran.

WardBrian:
@dylex sounds good, I touched a test file in this branch just to observe those pieces of the pipeline.

Assuming that all runs fine, I think that all we'd need to do is update the webhook? Does that need to happen after this is merged into master?

dylex:
The webhook is not critical, can be added whenever, just means builds need to be triggered manually.

FYI we'll point jenkins.flatironinstitute.org to the new system once the old system is retired (and keep "jenkins-new" as well).

PR #1636, opened 2026-07-08
Reword error for using unnormalized distribution in the wrong place

WardBrian:
The existing message doesn't clarify that the normalized version _is_ available, which tripped up @bob-carpenter 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Improved the error message for using an unnormalized function in an invalid context.

## Copyright and Licensing

### DIFF
--- a/src/frontend/Typechecker.ml
+++ b/src/frontend/Typechecker.ml
@@ -268,7 +268,16 @@ let check_postfixop loc op te =
         ~ad_level:te.emeta.ad_level ~type_ ~loc
   | _ -> Semantic_error.illtyped_postfix_op loc op te.emeta.type_ |> error
 
+(** unnormalized _lpdf/_lpmf functions can only be used in _lpdf/_lpmf/_lp udfs
+    or the model block *)
+let verify_unnormalized cf loc id =
+  if
+    Utils.is_unnormalized_distribution id.name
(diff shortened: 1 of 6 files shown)


src/frontend/Typechecker.ml
```diff
@@ -268,7 +268,16 @@ let check_postfixop loc op te =
...
+  if
+    Utils.is_unnormalized_distribution id.name
+    && not (in_udf_distribution cf || cf.current_block = Model)
+  then Semantic_error.invalid_unnormalized_fn loc id.name |> error
+
 let check_id cf loc tenv id =
+  verify_unnormalized cf loc id;
   match Env.find tenv (Utils.stdlib_distribution_name id.name) with
   | [] ->
       Semantic_error.ident_not_in_scope loc id.name
```
nhuurre: Checking unnormalized before in-scope can cause "Did you mean [something that doesn't exist]?" but I suppose that's fine.

nhuurre (approved):

src/frontend/Typechecker.ml
WardBrian: If the goal is to prevent confusion, this seems like a good thing to avoid doing...

nhuurre (approved):

PR #1642, opened 2026-07-10
Replace core dependency with base + stdio

hhugo:
## Summary

Removes the dependency on Jane Street's `core`, replacing it with the much lighter `base` + `stdio` (+ `sexplib0`). This shrinks the dependency footprint and build time, which is particularly relevant for the js_of_ocaml (`stancjs`) build.

- **Dependencies**: `core` → `base`, `stdio`, `sexplib0`. The `ppx_jane` bundle is replaced with only the rewriters actually used: `ppx_base`, `ppx_pipebang`, `ppx_sexp_message`, `ppx_sexp_value`, `ppx_expect`, `ppx_inline_test`.
- **Mechanical changes**: `open Base` instead of `open Core`; I/O goes through `Stdio`; stdlib modules that Base deprecates (`Format`, `Obj`, `Scanf`, `Marshal`, …) are qualified with `Stdlib`.
- **Core-only APIs replaced**:
  - `String.{Map,Set,Table,Hash_set}` → Base's comparator-passing style (`Map.M(String).t`, `Set.of_list (module String)`, `Hashtbl.create (module String)`, …)
  - `Expr.Typed`'s `Map`/`Set` submodules (previously from `Core.Comparable.S`, which Base does not provide) are now small explicit submodules exposing exactly the operations used.
  - `Map.Make_using_comparator`, `Set.of_map_keys`, `fst3` → Base equivalents.
- **New module** `Common.Poly_containers`: Base's `Set.Poly`/`Map.Poly` lack sexp converters, so this provides `sexp_of_t` for both in Core's exact output format — which is why no expect-test output changed.
- The `Generated_signatures` code generator now emits `Base.Hashtbl` instead of `Core.String.Table`.

## Notes for reviewers

- `ppx_jane` silently included **ppx_pipebang** (rewrites `x |> f` to `f x` at parse time), and the codebase depends on it: e.g. `|> UnsizedType.UArray` (constructor after pipe) and several places where record/constructor disambiguation only typechecks with the pipe inlined. It is kept as an explicit dependency.
- Two tiny semantic substitutions: `elt ** 2.` → `elt *. elt` (Base makes `**` integer exponentiation) and `Operator.of_string_opt` builds its `Sexp.Atom` directly instead of going through the deprecated `Sexp.of_string`.

## Testing

- `dune build @all` clean
- `dune runtest` (unit expect tests + integration cram suite): all pass **without a single expected-output change**
- `dune build @runjstest` (stancjs under node): passes
- Smoke-tested the `stanc` binary end-to-end on an integration model

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -15,8 +15,24 @@
  (depends
   (ocaml
    (= 5.5.0))
-  (core
-   (= v0.17.2))
+  (base
+   (= v0.17.3))
+  (stdio
+   (= v0.17.0))
+  (sexplib0
+   (= v0.17.0))
+  (ppx_base
+   (= v0.17.0))
+  (ppx_pipebang
+   (= v0.17.0))
+  (ppx_sexp_message
+   (= v0.17.0))
+  (ppx_sexp_value
+   (= v0.17.0))
+  (ppx_expect
+   (= v0.17.3))
+  (ppx_inline_test
+   (= v0.17.1))
   (menhir
    (= 20260209))
   (ppx_deriving
--- a/src/analysis_and_optimization/Dataflow_types.ml
+++ b/src/analysis_and_optimization/Dataflow_types.ml
@@ -1,4 +1,4 @@
-open Core
+open Base
 
 (***********************************)
 (* Basic datatypes                 *)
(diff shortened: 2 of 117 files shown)


WardBrian:
Thanks @hhugo. I ended up going with a similar solution in #1643, but one that touches fewer files in the process. Helping trim down the list of what was actually needed (especially on the ppx side) was a huge assist

PR #1643, opened 2026-07-10
Remove dependency on Core in favor of small wrapper around Base

### LINKED ISSUE #1640: Ocaml 5.5 upgrade introduced Intl dependency into stanc.js - breaking quickjs
Since the ocaml/js_of_ocaml upgrade, the `stanc.js` output has included a polyfill for the `Temporal` library (L466-L12806 in `stanc-pretty.js`) which relies on the `Intl` lib, which isn't supported by quickjs (so the downstream `rstan` builds are failing).

The polyfill itself is pretty heavy - over 12k LoC (nuts!), is this an option that can be disabled? Given that all seems to work when I shim out the calls or even remove the polyfill entirely, it doesn't seem to be load-bearing as far as I can see

WardBrian:
This should resolve #1640. 

To make the patch as minimal as possible, I added our own `Core.ml` module which shadows all the `open Core`s we have, which just uses base and a few other small changes for compatibility. This means we could rather seamlessly return to using core once 0.18 (which removed the weird Temporal polyfill) is released, if we wished

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

## Copyright and Licensing

### DIFF
--- a/docs/dependencies.mld
+++ b/docs/dependencies.mld
@@ -12,7 +12,7 @@ These are automatically installed through the [scripts/install_ocaml.sh] and
 
 - OCaml 5.5.0
 - Dune 3.5+
-- {{:https://v3.ocaml.org/p/base/latest/doc/index.html}Base} v0.17.0 (a standard library replacement/supplement)
+- {{:https://ocaml.org/p/base/v0.17.3/doc/index.html}Base} v0.17.3 (a standard library replacement)
 - {{:http://gallium.inria.fr/~fpottier/menhir/manual.html}Menhir} 20260209 (a parser generator and parsing library)
 - {{:https://github.com/ocaml-ppx/ppx_deriving}ppx_deriving} 6.1.1 (a tool for generating boilerplate code)
 - {{:https://erratique.ch/software/fmt}fmt} 0.11.0 (a library for pretty-printing of formatted text)
--- a/docs/dependencies.mld
+++ b/docs/dependencies.mld
@@ -12,7 +12,7 @@ These are automatically installed through the [scripts/install_ocaml.sh] and
 
 - OCaml 5.5.0
 - Dune 3.5+
-- {{:https://v3.ocaml.org/p/core/latest/doc/index.html}Core} v0.17.2 (a standard library replacement)
+- {{:https://v3.ocaml.org/p/base/latest/doc/index.html}Base} v0.17.0 (a standard library replacement/supplement)
 - {{:http://gallium.inria.fr/~fpottier/menhir/manual.html}Menhir} 20260209 (a parser generator and parsing library)
 - {{:https://github.com/ocaml-ppx/ppx_deriving}ppx_deriving} 6.1.1 (a tool for generating boilerplate code)
 - {{:https://erratique.ch/software/fmt}fmt} 0.11.0 (a library for pretty-printing of formatted text)
--- a/src/analysis_and_optimization/Monotone_framework.ml
+++ b/src/analysis_and_optimization/Monotone_framework.ml
@@ -24,7 +24,7 @@ let print_mfp to_string (mfp : (int, 'a entry_exit) Map.Poly.t)
     [%sexp (s : Stmt.Located.Non_recursive.t)] |> Sexp.to_string_hum in
   Map.iteri mfp ~f:(fun ~key ~data ->
       print_endline
-        (string_of_int key ^ ":\n "
+        (Int.to_string key ^ ":\n "
         ^ print_stmt (Map.Poly.find_exn flowgraph_to_mir key)
         ^ ":\n " ^ print_set data.entry ^ " \t-> " ^ print_set data.exit))
 
--- a/src/analysis_and_optimization/Monotone_framework.ml
+++ b/src/analysis_and_optimization/Monotone_framework.ml
@@ -6,6 +6,14 @@ open Monotone_framework_sigs
 open Mir_utils
 open Middle
 
+module Expr_set = struct
+  type t = Set.M(Expr.Typed).t
+
+  let empty = Set.empty (module Expr.Typed)
+  let singleton = Set.singleton (module Expr.Typed)
+  let union_list = Set.union_list (module Expr.Typed)
+end
+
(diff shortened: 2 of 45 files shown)


docs/dependencies.mld
```diff
@@ -12,7 +12,7 @@ These are automatically installed through the [scripts/install_ocaml.sh] and
 
 - OCaml 5.5.0
 - Dune 3.5+
-- {{:https://v3.ocaml.org/p/core/latest/doc/index.html}Core} v0.17.2 (a standard library replacement)
+- {{:https://v3.ocaml.org/p/base/latest/doc/index.html}Base} v0.17.0 (a standard library replacement/supplement)
```
nhuurre:
```suggestion
- {{:https://ocaml.org/p/base/v0.17.3/doc/index.html}Base} v0.17.3 (a standard library replacement/supplement)
```
What does "supplement" mean? The doc says
> Unlike other standard library projects, Base is meant to be used as a wholesale replacement of the standard library distributed with the OCaml compiler.

src/core/Core.ml
```diff
@@ -0,0 +1,100 @@
+(** This module shadows [core] from Jane Street with a more lightweight version
```
nhuurre: "shadows" is not the right word.

src/core/Core.ml
```diff
@@ -0,0 +1,100 @@
...
+
+let ( ^^ ) = Stdlib.( ^^ )
+let ( ** ) = Stdlib.( ** )
+let string_of_int = Stdlib.string_of_int
+let int_of_string = Stdlib.int_of_string
+let int_of_string_opt = Stdlib.int_of_string_opt
+let float_of_string = Stdlib.float_of_string
+let float_of_string_opt = Stdlib.float_of_string_opt
+let float_of_int = Stdlib.float_of_int
+let exit = Stdlib.exit
```
nhuurre, line 93: I'd rather not add these.
* we already use `Int.to_string`, `Int.of_string`, and `Float.of_string` in some places. It's more consistent to use those everywhere instead of sometimes calling `Stdlib.string_of_int` etc.
* `(^^)`, `(**)`, `int_of_string_opt`, `float_of_string_opt`, and `exit` are only used once each so defining them here doesn't reduce the diff. (Also if we're using `Float.of_string` then maybe we should use `Float.of_string_opt` instead of `Stdlib.float_of_string_opt`. Is there any difference between those?)
* that leaves only `float_of_int` but it's not that much work to change it to `Float.of_int`, there's only four calls.

src/analysis_and_optimization/Monotone_framework.ml
```diff
@@ -6,6 +6,14 @@ open Monotone_framework_sigs
 open Mir_utils
 open Middle
 
+module Expr_set = struct
+  type t = Set.M(Expr.Typed).t
```
nhuurre, line 10: If the idea is to minimize the diff, shouldn't this module be inside `Expr.Typed`? Is there a reason it has to be defined here?

src/middle/Operator.ml
```diff
@@ -59,7 +59,8 @@ let to_string x = Sexp.to_string (sexp_of_t x) ^ "__"
 
 let of_string_opt x =
   try
-    String.chop_suffix_exn ~suffix:"__" x |> Sexp.of_string |> t_of_sexp |> Some
+    String.chop_suffix_exn ~suffix:"__" x
+    |> Sexp.of_string |> t_of_sexp |> Option.some
   with
```
nhuurre, line 63: Huh? `Some -> Option.some`. Why did this work before and why does it need to change?

nhuurre (commented):
The other PR had several files where the only change is `open Core -> open Base`. If a file doesn't need anything from Core then don't open Core.

src/analysis_and_optimization/Monotone_framework.ml
WardBrian: It doesn't _have_ to be, but this is the only file that actually used it, and writing down the signature in `Expr.mli` was a bit verbose/annoying

src/core/Core.ml
WardBrian: I'm happy to replace the `_to_`/`_of_` with the module versions, but I'd rather leave `^^`/`**`/`exit` -- I think shadowing such basic functionality in the standard library is a dumb move on Base's part, which I suspect at least someone at JS agrees with since they _don't_ do it in Core, despite re-exporting everything else in Base

src/middle/Operator.ml
WardBrian: This one line was the only reason `Middle` depended on `ppx_pipebang`. I'd like to eventually decrease our dependencies on ppxs overall, and this seemed like a small price

WardBrian:
I'd also rather audit whether Core needs to be opened at the top of files in a separate PR, because I suspect the Stdlib improvements in OCaml 5 actually mean a lot of our files don't need Core _or_ Base any longer, but that's a more involved project

hhugo:
How does this PR affects the size of the stanc.js file ?

andrjohns:
> How does this PR affects the size of the stanc.js file ?

A pretty decent amount: ~6mb reduction in unminified/debug bundle, ~700kb reduction in minified/release

nhuurre (approved):

hhugo:
I

> > How does this PR affects the size of the stanc.js file ?
> 
> A pretty decent amount: ~6mb reduction in unminified/debug bundle, ~700kb reduction in minified/release

And in percentage ?

src/core/dune
```diff
@@ -0,0 +1,6 @@
+(library
+ (name core)
```
hhugo, line 2: This name probaby mean it wont be able to link together with core. I dont know if its an issue for you here

WardBrian: This was a feature to me, since it meant I couldn’t accidentally pull it back in somewhere :)

I am intending this to be a temporary solution to unblock our downstream projects before doing a more serious audit of whether we even really need Base. Stdlib has gotten quite a lot better in 5.x as far as I can tell

WardBrian:
> And in percentage ?

~15% reduction

PR #1631, opened 2026-06-24
Update to OCaml 5.5.0

WardBrian:
OCaml 5.5.0 was [recently released](https://ocaml.org/changelog/2026-06-19-ocaml-550). At the same time, the plan is to sunset 4.14 [at the end of the year](https://ocaml.org/changelog/2026-06-15-ocaml-4144), so this seemed like a good opportunity to upgrade.

Even though we're jumping a major version and 5 minor versions, there are actually no breaking changes that affect us directly; the only code changes in this PR are to help with a slowdown related to the Marshal functionality that we use for loading in the compiled list of function signatures. 

There are a few places I think the compiler code could be improved by using OCaml 5's features, but those are all left for future work.

Draft until 
- [x] https://github.com/ocaml-cross/opam-cross-windows/pull/407 is merged
- [x] https://github.com/ocaml/opam-repository/pull/30147 is merged

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [ ] If a user-facing facing change was made, the documentation PR is here: <LINK>
    - [x] OR, no user-facing changes were made

## Release notes

The Stan compiler is now built with OCaml 5.5.0 by default

## Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause)

PR #1644, opened 2026-07-13
Leverage Stdlib additions in OCaml 5

WardBrian:
Follow on to #1631, #1643: The OCaml standard library now includes direct replacements for a few things we wrote ourselves or needed dependencies for:

- `Common.Let_syntax.[Result | Option]`: replaced by `Stdlib.[Result | Option].Syntax`
- `Environment.Distance`: `String` now has a `spellcheck` function
- `Out_channel` and `In_channel` from stdio have sufficient stdlib replacements

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Copyright and Licensing


src/frontend/Environment.ml
```diff
@@ -56,81 +56,19 @@ let find env key = Map.find_multi env key
...
-               ~f:(fun n -> "_" ^ n)
-               cumulative_distribution_suffices_w_rng)
-        |> List.map ~f:(fun suffix -> name ^ suffix)
-        |> List.filter ~f:(Map.mem env)
-        |> List.hd)
-    with _ -> None in
+  let open Stdlib.Option.Syntax in
+  let max_dist s =
+    let length = String.length s in
+    if length < 2 then 0 else if length < 10 then 2 else 4 in
```
nhuurre: This is correct translation of the previous code but edge cases don't really make much sense. Two-character identifier shouldn't allow two edits, and the transition from 2 to 4 feels a bit sudden. Might be better to just calculate the maximum distance as length divided by two, rounded down, but no larger than 4.

src/stanc/stanc.ml
```diff
@@ -17,7 +17,8 @@ let dump_math_dists () =
 
 let write filename data =
   try
-    Out_channel.write_all filename ~data;
+    ( Out_channel.with_open_text filename @@ fun oc ->
```
nhuurre: So this adds OS-specific newline handling? Do we really want that?

WardBrian: I'm not sure it mattered, but since core was using binary mode I'll change these

src/frontend/Environment.ml
WardBrian: I like the simpler criteria. I added it, with the extra caveat that I decided to base it on the length minus the length of any known suffix like `_lpdf`, since `foo_lupdf` getting 4 characters of leeway led to some weirder test results

test/integration/bad/removed_features/stanc.expected
```diff
@@ -157,6 +157,7 @@ Semantic error in 'if_else.stan', line 9, column 26 to column 54:
    -------------------------------------------------
 
 A returning function was expected but an undeclared identifier "if_else" was supplied.
+Did you mean "inverse"?
```
nhuurre, line 160: Ha ha, close enough I guess.

nhuurre (approved):

PR #1645, opened 2026-07-13
Update ICE handling with new Format functions

WardBrian:
One of the more interesting additions in 5.5 is the `l*printf` family of functions (https://github.com/ocaml/ocaml/pull/13372). These make it much easier to create functions that "act like" pretty-printers that would have previously required continuations or been impossible to provide a type for. 

This PR replaces `Common.ICE.internal_compiler_error` with two functions, one that just accepts a string, and the other that uses `Format.lasprintf` to allow additional formatting. This is more flexible than the existing sexp-based function, and doesn't require the use of any PPX to invoke

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Copyright and Licensing

### DIFF
--- a/src/analysis_and_optimization/Monotone_framework.ml
+++ b/src/analysis_and_optimization/Monotone_framework.ml
@@ -14,20 +14,6 @@ module Expr_set = struct
   let union_list = Set.union_list (module Expr.Typed)
 end
 
-(** Debugging tool to print out MFP sets **)
-let print_mfp to_string (mfp : (int, 'a entry_exit) Map.Poly.t)
-    (flowgraph_to_mir : (int, Stmt.Located.Non_recursive.t) Map.Poly.t) : unit =
-  let print_set s =
-    [%sexp (Set.Poly.map ~f:to_string s : string Set.Poly.t)]
-    |> Sexp.to_string_hum in
-  let print_stmt s =
-    [%sexp (s : Stmt.Located.Non_recursive.t)] |> Sexp.to_string_hum in
-  Map.iteri mfp ~f:(fun ~key ~data ->
-      print_endline
-        (Int.to_string key ^ ":\n "
-        ^ print_stmt (Map.Poly.find_exn flowgraph_to_mir key)
-        ^ ":\n " ^ print_set data.entry ^ " \t-> " ^ print_set data.exit))
-
 (** Calculate the free (non-bound) variables in an expression *)
 let rec free_vars_expr (e : Expr.Typed.t) =
   match e.pattern with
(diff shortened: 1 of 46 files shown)


src/frontend/Pretty_printing.ml
```diff
@@ -82,11 +82,11 @@ let skip_comments loc =
...
              noisy rather than ever producing anything invalid for these. *)
-          Common.ICE.internal_compiler_error
-            [%message
-              "Unable to format #include in this position!"
-                (l : string list)
-                (loc : Middle.Location_span.t)] [@coverage off]
+          Common.ICE.(
+            internal_errorf
+              "Unable to format #include in this position! Lines: %t Loc: %t"
+              [Fmt.list Fmt.string $ l; Middle.Location_span.pp $ loc])
```
nhuurre: This is the one "internal" error we know users can see so I think it's worth phrasing more thoughtfully. Location should be first, like with all other errors, and "lines" is always just one item, the filename.

src/common/Nonempty_list.mli
```diff
@@ -1,11 +1,9 @@
 (** A non-empty list type. This can be constructed with the usual list syntax
     when the compiler is inferring the type, e.g.
 
-    [let x : _ Nonempty_list.t = [1;2;3]]
+    [let x : _ Nonempty_list.t = [1;2;3]] *)
 
-    In Core v0.18, we can replace this with [Core.Nonempty_list] *)
```
nhuurre, line 6: Update to Core v0.18 cancelled, I guess.

src/analysis_and_optimization/Monotone_framework.ml
```diff
@@ -14,20 +14,6 @@ module Expr_set = struct
   let union_list = Set.union_list (module Expr.Typed)
 end
 
-(** Debugging tool to print out MFP sets **)
-let print_mfp to_string (mfp : (int, 'a entry_exit) Map.Poly.t)
-    (flowgraph_to_mir : (int, Stmt.Located.Non_recursive.t) Map.Poly.t) : unit =
-  let print_set s =
```
nhuurre, line 20: Seeing this kind of changes reminds me that unfortunately MFP (and our optimization suite) is effectively unmaintained.

nhuurre (approved):

src/analysis_and_optimization/Monotone_framework.ml
WardBrian: Yes, unfortunately. I own a copy of [the book](https://books.google.com/books/about/Principles_of_Program_Analysis.html?id=RLjt0xSj8DcC) it was supposedly implemented out of, but I haven't so much as read the intro yet.

I've long been tempted to try to understand enough of it to consider replacing it with [`Fix`](https://ocaml.org/p/fix/latest/doc/index.html), a library by the author of menhir that (I think) is solving the same problem. But this hasn't happened. 

If I do make a pass at actually removing `base` as a dependency it will probably be the hardest part, since it is absolutely chock full of `Map.Poly` and `Set.Poly`

PR #1649, opened 2026-07-18
Compiler support for `laplace_latent_solve(_tol)`

florence-bockting:
#### Submission Checklist

This PR refers to math [PR#337](https://github.com/stan-dev/math/pull/3337) in which we implemented `laplace_latent_solve` and `laplace_latent_solve_tol`. The new functions are currently only supported in the generated quantities block, therefore we update also the Typechecker in this PR.

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/968

## Release notes

+ Adds Stan compiler support for Math’s `laplace_latent_solve(_tol)`, which return `tuple(vector, matrix)` (posterior mean and Cholesky factor).

## Copyright and Licensing

This PR used AI assistance (Cursor).


WardBrian:
Thanks @florence-bockting -- question about 6e2c28bb0f0babc8c2e92729d8153ea9f0d89ef4 -- is it actually a requirement that these functions only be called in generated quantities? If so we will want to enforce that in the typechecker

florence-bockting:
> Thanks @florence-bockting -- question about [6e2c28b](https://github.com/stan-dev/stanc3/commit/6e2c28bb0f0babc8c2e92729d8153ea9f0d89ef4) -- is it actually a requirement that these functions only be called in generated quantities? If so we will want to enforce that in the typechecker

Thank you for pointing me towards the typechecker! 
Short about the background: As `laplace_latent_solve` currently only goes through `laplace_base_rng` it inherits the constraint on CovarArgs and LLArgs:
```
template <bool ReturnMeanAndCovCholesky = false, typename LLFunc,
          typename LLArgs, typename CovarFun, typename CovarArgs,
          bool InitTheta, typename RNG,
          require_t<is_all_arithmetic_scalar<CovarArgs, LLArgs>>* = nullptr>
```

For the `model` block this fails as some of the arguments are variables. Here an example snippet of a corresponding model and generated quantities block:

```
model {
  rho ~ inv_gamma(rho_location_prior, rho_scale_prior);
  alpha ~ inv_gamma(alpha_location_prior, alpha_scale_prior);
  eta ~ normal(0, 1);

  target += laplace_marginal(ll_function, (eta, log_ye, y), hessian_block_size,
                             K_function, (x, n_obs, alpha, rho));

  target += laplace_marginal_tol(ll_function, (eta, log_ye, y), hessian_block_size,
                                 K_function, (x, n_obs, alpha, rho),
                                (theta_0, tolerance, max_num_steps,
                                solver, max_steps_line_search, allow_fallthrough));
}
generated quantities {
  tuple(vector[n_obs], matrix[n_obs, n_obs]) mean_chol
      = laplace_latent_solve(ll_function, (eta, log_ye, y), hessian_block_size,
                             K_function, (x, n_obs, alpha, rho));
}
```

I am not sure whether there are use cases where a user wants to call `laplace_latent_solve` in the model block (@avehtari )? In this case, we would have to change the definition in math. Otherwise I would add the corresponding constraint in the `Typechecker.ml`.

SteveBronder:
So since we never actually use the rng for `laplace_latent_solve` idt there is any reason we could not have it in the model block. So language wise I don't see an issue arising because of a user calling this in the model block.

florence-bockting:
> So since we never actually use the rng for `laplace_latent_solve` idt there is any reason we could not have it in the model block. So language wise I don't see an issue arising because of a user calling this in the model block.

The definition of `laplace_latent_solve` is as follows:

```
template <typename LLFunc, typename LLArgs, typename CovarFun,
          typename CovarArgs>
inline auto laplace_latent_solve(LLFunc&& ll_fun, LLArgs&& ll_args,
                                 int hessian_block_size,
                                 CovarFun&& covariance_function,
                                 CovarArgs&& covar_args, std::ostream* msgs) {
  auto options = laplace_options_default{hessian_block_size};
  internal::laplace_unused_rng unused_rng;
  return laplace_base_rng<true>(std::forward<LLFunc>(ll_fun),
                                std::forward<LLArgs>(ll_args),
                                std::forward<CovarFun>(covariance_function),
                                std::forward<CovarArgs>(covar_args),
                                std::move(options), unused_rng, msgs);
}
```
So, we actually call `laplace_base_rng` and therefore inherit the constraints on the arguments.

An alternative, to circumvent calling `laplace_base_rng`, could be to factor out the code for computing the mean and variance in an extra function and then to call this function in `laplace_base_rng` and in `laplace_latent_solve`.

On the other hand, I talked yesterday with Aki and he said, he could only come up with theoretical examples for calling `laplace_latent_solve` in the model block and thus would be fine to restrict it to the generated quantities block.

I don't know whether you prefer to keep it simple and leave it as is (with the missing Typechecker) or redefining `laplace_latent_solve` such that it generalizes to more situations.

test/integration/bad/embedded_laplace/bad_solve_return.stan
```diff
@@ -0,0 +1,39 @@
...
+}
+parameters {
+  real<lower=0> alpha;
+  real<lower=0> rho;
+  real<lower=0> eta;
+}
+
+generated quantities {
+  // laplace_latent_solve returns tuple(vector, matrix), not vector
+  vector[n_obs] theta = laplace_latent_solve(ll_function, (eta, log_ye, y), 1,
```
nhuurre: What's the point of this test? The error comes from typechecking the assignment, not `laplace_latent_solve`.
And if the return type was wrong, good code-gen tests would fail to compile.

nhuurre (commented):
I think that `is_all_arithmetic_scalar<..>` restriction in C++ corresponds to `DataOnly` in OCaml. You'll need to add
```ocaml
if UnsizedType.is_autodifftype cov_tupl.emeta.ad_level then error (...)
```
where the error says something like "the argument tuple for covariance function must be data-only."
And same for `lik_tupl`.


> idt there is any reason we could not have it in the model block.

> An alternative, to circumvent calling `laplace_base_rng`, could be to factor out the code for computing the mean and variance in an extra function and then to call this function in `laplace_base_rng` and in `laplace_latent_solve`.

Relaxing the data-only restriction requires a C++ implementation that can compute derivatives. I don't know how much extra work that is but if there's no known use case it's probably not worth it.

SteveBronder:
Oh sorry I missed the requires that stops autodiff types on `laplace_base_rng`

```
require_t<is_all_arithmetic_scalar<CovarArgs, LLArgs>>* = nullptr
```

In that case lets just keep it to generated quantities for now. Later we can always change the impl so that `laplace_latent_solve` can be in the model block.

test/integration/bad/embedded_laplace/bad_solve_return.stan
florence-bockting: Thank you for pointing this out. I agree that this is a weak test and deleted it.

src/frontend/Typechecker.ml
```diff
@@ -919,6 +929,8 @@ and check_laplace_fn ~is_cond_dist loc cf tenv id tes =
         lik_args @ (hbs_arg :: cov_fun_type :: cov_tupl :: control_args) in
       let return_type =
         if String.is_suffix id.name ~suffix:"_rng" then UnsizedType.UVector
+        else if String.is_substring id.name ~substring:"_solve" then
+          UnsizedType.UTuple [UVector; UMatrix]
         else UnsizedType.UReal in
```
WardBrian, line 947: It's a bit unusual, but especially since we think we may one day lift the restriction, I'd prefer to have the cf.current_block check occur here, where we already know we are looking at one of these functions

nhuurre:
I don't see a reason to disallow the function in `transformed data`. Sure, it might be useless there but the relevant issue is autodiff, and I think the typechecker should just check autodiffability directly instead of inferring it from the current block.

And actually, I think it should be allowed in `functions` block, but only in functions that don't use autodiff, so checking the current block is not good enough.

WardBrian:
It might be simpler to update the C++ than teach the compiler about a new exception, especially if it is not a real technical limitation

nhuurre:
Here you go: https://github.com/nhuurre/stanc3/commit/63c5213f187ee0920dd9492cd2794ed7ecde97f5

WardBrian:
@florence-bockting -- I think if you applied @nhuurre's patch in the comment above this one this would be ready to merge

WardBrian (approved):
Thank you!

PR #1650, opened 2026-07-27
Use a Dynarray instead of a hashtable for the preprocessor locations

WardBrian:
This hashtable only ever had keys that were numeric and growing, which means it was essentially just a growable array in a language without good support for those. Now that the standard library has a `Dynarray` module, we can use that and save on some storage

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Copyright and Licensing

### DIFF
--- a/src/frontend/Preprocessor.ml
+++ b/src/frontend/Preprocessor.ml
@@ -11,26 +11,26 @@ let include_stack = Stack.create ()
 let included_files : string list ref = ref []
 let size () = Stack.length include_stack
 
-let locations_map : (string * Middle.Location.t option) String.Table.t =
-  String.Table.create ()
+let locations_map : (string * Middle.Location.t option) Dynarray.t =
+  Dynarray.create ()
 
 let new_file_start_position buf filename included_from =
   (* Lexing.position does not have a field to store `included_from` so we store
-     it in a global hashmap instead and put the hashmap key in `pos_fname`. The
-     keys are arbitrary unique strings. (Filenames are not good keys because the
-     same file could be included multiple times.)
+     it in a global array and put the key in `pos_fname`. The keys simply count
+     up. (Filenames are not good keys because the same file could be included
+     multiple times.)
 
      Prefixing the key with NUL allows us to do a little optimization: when
-     `included_from` is None we don't need to access the hashmap and can store
-     the filename directly in `pos_fname`. Filenames never begin with NUL so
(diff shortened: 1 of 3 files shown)


src/frontend/Preprocessor.ml
```diff
@@ -11,34 +11,34 @@ let include_stack = Stack.create ()
...
+  Dynarray.create ()
 
 let new_file_start_position buf filename included_from =
   (* Lexing.position does not have a field to store `included_from` so we store
-     it in a global hashmap instead and put the hashmap key in `pos_fname`. The
-     keys are arbitrary unique strings. (Filenames are not good keys because the
-     same file could be included multiple times.)
+     it in a global array and put the key in `pos_fname`. The keys simply count
+     up. (Filenames are not good keys because the same file could be included
+     multiple times.)
```
nhuurre:
```suggestion
     it in a global array and put its index in `pos_fname`.
```

src/frontend/Preprocessor.ml
```diff
@@ -11,34 +11,34 @@ let include_stack = Stack.create ()
...
      `location_of_position` only needs to check the first character to know
-     whether to do a hashmap lookup. *)
+     whether to do a lookup. *)
   let pos =
     if Option.is_none included_from then
       {Lexing.pos_fname= filename; pos_lnum= 1; pos_bol= 0; pos_cnum= 0}
     else
-      let key = "\u{0}" ^ Int.to_string (Hashtbl.length locations_map) in
-      Hashtbl.add_exn locations_map ~key ~data:(filename, included_from);
+      let key = "\u{0}" ^ Int.to_string (Dynarray.length locations_map) in
```
nhuurre: `Int.(to_string,of_string)` converts to decimal but the representation here doesn't really matter and hexadecimal might be more efficient?
You could even store the bytes directly:
```ocaml
let decode s =
  Stdlib.Bytes.get_uint16_ne (Stdlib.Bytes.of_string s) 1

let encode i =
  let s = Stdlib.Bytes.create 3 in
  Stdlib.Bytes.set s 0 Stdlib.Char.Ascii.min;
  Stdlib.Bytes.set_uint16_ne s 1 i;
  Stdlib.Bytes.to_string s
```

WardBrian: Good suggestion. I used the int32 versions just to be extra safe...

nhuurre (approved):

PR #1658, opened 2026-08-10
Stanc.Std: Move drivers over

WardBrian:
#### Submission Checklist

- [ ] Run unit tests
- Documentation

## Release notes

## Copyright and Licensing

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


src/std/Return.ml
```diff
@@ -0,0 +1,14 @@
+let with_return : 'a 'b. (('a -> 'b) -> 'a) -> 'a =
+ fun (type a) f ->
```
nhuurre: Oof, this is some advanced type syntax I've never seen before. I guess `'a 'b.` is a universal quantifier and, whaa, `fun (type a)`???
...
Okay, after reading about _locally abstract types_, I think this is equivalent to
```suggestion
let with_return (type a) (f : (a -> 'b) -> a) : a =
```

WardBrian: You're correct that it is equivalent, but I didn't mean for it to be. The idea is that the `return` callback should have a universally quantified return type such that it can be used anywhere, similar to `failwith`, which was previously only supported in [record types](https://github.com/janestreet/base/blob/master/src/with_return.ml#L5). 

It appears the only usages we have of this functionality happen to be happy deducing the same type in every invocation, but this is not the most general/useful version. I'll try to see if I can get it actually working, or I'll use the simpler version

nhuurre (approved):

PR #1661, opened 2026-08-10
Stanc.Std: Move over Analysis module

WardBrian:
The most notable change here is the addition of our own vendored polymorphic set. This is based on OCaml's set implementation, but with polymorphic comparison swapped in. Remarkably, I've found that it is dramatically faster than Base.Set.Poly in our uses, presumably because `Base.Set.Poly` is storing the comparison function in the runtime representation still, even though it never changes

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Copyright and Licensing

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/test/integration/good/compiler-optimizations/cpp.expected
+++ b/test/integration/good/compiler-optimizations/cpp.expected
@@ -34886,7 +34886,6 @@ class off_dce_model final : public model_base_crtp<off_dce_model> {
             "assigning variable lcm_sym9__");
           double qT = std::numeric_limits<double>::quiet_NaN();
           lcm_sym16__ = stan::math::prod(lcm_sym9__);
-          lcm_sym19__ = static_cast<double>(1);
           stan::model::assign(psi_con, ((lcm_sym14__ * lcm_sym16__) /
             stan::math::fma(lcm_sym14__, lcm_sym16__, (1 - lcm_sym14__))),
             "assigning variable psi_con", stan::model::index_uni(1));
@@ -34897,7 +34896,6 @@ class off_dce_model final : public model_base_crtp<off_dce_model> {
                 stan::model::index_uni(1)), base_rng__),
             "assigning variable z", stan::model::index_uni(1));
         } else {
-          lcm_sym19__ = static_cast<double>(1);
           stan::model::assign(psi_con, 1, "assigning variable psi_con",
             stan::model::index_uni(1));
           current_statement__ = 10;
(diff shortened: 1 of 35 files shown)


test/integration/good/compiler-optimizations/cpp.expected
```diff
@@ -34940,8 +34937,8 @@ class off_dce_model final : public model_base_crtp<off_dce_model> {
               "assigning variable z", stan::model::index_uni(i));
           } else {
             current_statement__ = 19;
-            stan::model::assign(psi_con, lcm_sym19__,
-              "assigning variable psi_con", stan::model::index_uni(i));
+            stan::model::assign(psi_con, 1, "assigning variable psi_con",
+              stan::model::index_uni(i));
```
nhuurre, line 34941: Looks like constant propagation got a bit smarter, but why?

WardBrian: I spent a few hours on trying to debug this and I couldn’t pinpoint the change. 

I also threw whatever free credits of Claude github gives me, and it spun in circles but couldn’t identify anything either after spending the entire month’s allocation. 

In the end, I think the fact that this wasn’t being propagated in master was a bug, and the relevant optimizations are only turned on for Oexperimental, so I gave up looking for the cause… if you can find anything without wasting too much time, I would still be curious!

nhuurre: Yes, this was probably a bug in master and not worth chasing down. However, note that the anomaly happens in constant propagation, which is enabled for O1, even though it didn't trigger without some experimental passes first. If you're curious and want to waste even more time I found a "minimal" example:
```ocaml
let%expect_test "constant propagation" =
  let mir =
    Test_utils.mir_of_string {|
data {
  vector[2] X;
  int sum_y;
  real alpha_occ;
  vector[2] logit_p;
}
generated quantities {
  vector[2] logit_psi = alpha_occ + alpha_occ * X;
  array[2] real psi_con;
  if (1) {
    real psi = inv_logit(logit_psi[1]);
    real qT = prod(logit_p);
    psi_con[1] = inv(psi * qT + 1);
  } else {
    psi_con[1] = 1;
  }
  real psi = inv_logit(logit_psi[2]);
  real qT = prod(logit_p);
  psi_con[2] = inv(psi * qT + 1);
}     |} in
  let open Analysis_and_optimization.Optimize in
  let mir = expression_propagation mir in
  let mir = lazy_code_motion mir in
  let mir = constant_propagation mir in
  Fmt.str "@[<v>%a@]" Middle.Program.Typed.pp mir |> print_endline;
  [%expect {| ... |}]
```

WardBrian: The `if (1)` seems to be the load-bearing part. If you change it to something semantically equivalent, like `if (2)`, then the constant does indeed get moved down

WardBrian: My theories:

- Some part of the optimization pipeline is iteration-order-dependent by accident, and this branch gets a favorable order
- There is something in `master` that is accidentally sensitive to the location metadata, and isn't here (though I've tried to find the likely places this would be, and came up empty)


Assuming we're both happy not knowing, what do you think of the rest of the changes?

nhuurre: Accidental order-dependence sounds plausible. The rest of the changes look good.

nhuurre (approved):

Issue #702, opened 2020-09-23
Analysis and vectorization of for loops 


adamhaber:
Following the discussion on #594, I’m trying to detect for loops which can be automatically vectorised. The original motivation was to find cases for which we can generate TFP code that can use vmap instead of Python for loops, but perhaps this might be interesting in a broader context.

I've started with the two simplest cases I could think of: A for loop with a single statement, that's either an assignment (in which `lhs` is not in `rhs`) or a target lp increment. The initial attempt is here:

https://github.com/adamhaber/stanc3/blob/b785b7b79869f93798b699a5b82ee1fc03b43f49/src/tfp_backend/Code_gen.ml#L88-L120

Couple of things:

1. I'm sure there's a better way to do the extraction of `rhs_vars` - couldn't find reasonable documentation for Set.Poly.t.
2. The `tf__.range(%a-1, %a)` is a hacky workaround - we probably need a more robust solution to the fact that Stan is 1-based and Python is 0-based.
3. I've tested this on some simple programs and I _think_ the generated TFP code is doing what it should be doing, but I haven't thoroughly tested it yet. I'm also getting an `OP_REQUIRES failed at functional_ops.cc:374 : Internal: No function library` warning from TFP, I'll try to contact the TFP team about it.

Would be happy to hear your thoughts on this.

adamhaber:
@rok-cesnovar any thoughts on this, in general / tfp-backend contexts?

rok-cesnovar:
Will take a look this week, detecting simple for loops would definitely be useful in the stan-to-c++ world as well.

seantalts:
@adamhaber I would say that if at all possible we want to be doing these transformations on the MIR itself in one of the optimization passes rather than as the code is printed in any/all of the backends. This is because the pretty printers are already too complicated but also because this work is likely backend agnostic and multiple backends can benefit. If there's some TFP specific stuff, that's fine to put in the TFP backend but I wouldn't throw it inline with the rest of the code gen; I'd propose creating a separate phase as part of lowering.

> The tf__.range(%a-1, %a) is a hacky workaround - we probably need a more robust solution to the fact that Stan is 1-based and Python is 0-based.

This is the same hack we use in Stan_math_code_gen, for what it's worth :P It's a little bit tricky but definitely open to suggestions.

seantalts:
[...]

Issue #356, opened 2019-10-23
automated vectorization of simple for loops


seantalts:
I want to start turning loops like this:
```stan
  for (i in 1:I)
    y[i] ~ bernoulli_logit(a[i] * (theta - b[i]));
```
into 
```stan
y ~ bernoulli_logit(a * (theta - b));
```

@rybern or @VMatthijs where should this code live? any tips for me as I embark on my first journey into the analysis and optimization package?

WardBrian:
I think this is superseded by #702 which has some more details and discussion

PR #1666, opened 2026-08-15
Add vectorize_loops optimization: scalar density loops become vectorized densities

seantalts:
Closes the ancient #356; related to #702.

This adds a `vectorize_loops` optimization pass that rewrites loops whose body is a single scalar density statement into the vectorized density call Stan Math already provides:

```stan
for (n in 1:N) target += normal_lpdf(y[n] | mu[n], sigma);
// becomes
target += normal_lpdf(y | mu, sigma);
```

Both forms sum the same terms; the vectorized call shares subcomputations across elements (log(sigma) once instead of N times) and allocates O(1) autodiff nodes instead of O(N). Tilde statements are already this shape in the MIR, so `y[n] ~ normal(mu[n], sigma)` is covered by the same pattern and keeps its proportionality flag.

The pass is enabled only at `--Oexperimental` for now. Follow-up PRs extend it to indirectly indexed arguments (`a[county[n]]`) and elementwise assignment loops, and would make the case for `--O1` with benchmarks; keeping this one small should make the soundness argument easy to review.

**How it stays sound.** Every argument must be a scalar loop invariant or exactly `x[n]`. An `x[n]` argument becomes the slice `x[lower:upper]`, which needs no size proof, or bare `x` when the range provably spans the declaration (data and parameters only, since nothing else is immutable). At least one argument must vary with the loop, or a loop of N identical terms would collapse to one. Invariant arguments must be scalars: an invariant container is summed over anew by every iteration but would be zipped elementwise by the vectorized call, and both forms typecheck. Finally the rewritten call must re-typecheck against the Stan Math signatures or the loop is left alone, so the signature table decides which densities vectorize and user-defined densities never do. Truncated tildes lower to multi-statement bodies and are excluded by the single-statement match.

**Measurements** (Apple M-series, per-gradient via `log_prob_propto_jacobian` + `grad`, interleaved A/B against the same build with the pass disabled, medians of 3): across the 120 posteriordb models this pass changes generated C++ for 5 (radon_pooled, normal_mixture, soil_incubation, covid19imperial_v2/v3); the other 115 are byte-identical. radon_pooled's gradient goes 327us to 92us (**3.54x**). The other four claim only prior loops or are ODE-dominated and measure 0.99x to 1.00x. Log probability and gradients agree with the pass off to all printed digits in every case. Compile time is unchanged: clang differences are within noise and stanc itself gets slightly faster on claimed models since later passes see fewer statements.

I used AI but I have read the code and vouch for it.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] OR, no user-facing changes were made (only `--Oexperimental` behavior changes; happy to add user-facing docs when this graduates to `--O1`)

## Release notes

New optimization `vectorize_loops` (enabled at `--Oexperimental`): loops whose body is a single scalar density statement are rewritten to the vectorized density call, e.g. `for (n in 1:N) y[n] ~ normal(mu[n], sigma);` becomes `y ~ normal(mu, sigma);`.

## Copyright and Licensing


src/analysis_and_optimization/Optimize.ml
```diff
@@ -642,6 +641,135 @@ let unroll_loop_one_step_statement _ =
...
+     |SMatrix (_, d, _)
+     |SArray (_, d) ->
+        Some d
+    | SInt | SReal | SComplex | SComplexVector _ | SComplexRowVector _
+     |SComplexMatrix _ | STuple _ ->
+        None in
+  (* Outer sizes that stay true wherever the variable appears: data and
+     parameters cannot be assigned to or shadowed, and parameters are
+     re-materialized at exactly their declared size. Local and transformed-data
+     containers can be overwritten whole, possibly at another size, so they are
```
nhuurre: Surely assignments are always checked for correct size, even when no other constraints are enforced? And this optimization triggers on `target +=` so it can only happen in the `model` block (or `transformed parameters` if `_lp` functions have been inlined) which is after transformed data has been frozen and constraint-checked.

src/analysis_and_optimization/Optimize.ml
```diff
@@ -642,6 +641,135 @@ let unroll_loop_one_step_statement _ =
...
+    List.filter_map mir.input_vars ~f:(fun (name, _, st) ->
+        Option.map (outer_size st) ~f:(fun d -> (name, d)))
+    @ List.filter_map mir.output_vars ~f:(fun (name, _, ov) ->
+        match ov.out_block with
+        | Parameters ->
+            Option.map (outer_size ov.out_constrained_st) ~f:(fun d ->
+                (name, d))
+        | TransformedParameters | GeneratedQuantities -> None)
+    |> String.Map.of_alist_exn in
+  let slice sizes ~lower ~upper (base : Expr.Typed.t) =
```
nhuurre: IMO it would be cleaner to fuse this `slice` function into `classify_arg` (which I guess should then be renamed `vector_arg` or something...)
Keep the full_range check as a separate helper function, though.

src/analysis_and_optimization/Optimize.ml
```diff
@@ -642,6 +641,135 @@ let unroll_loop_one_step_statement _ =
...
+      Expr.Typed.equal lower Expr.Helpers.loop_bottom
+      &&
+      match base.pattern with
+      | Var v -> Option.exists (Map.find sizes v) ~f:(Expr.Typed.equal upper)
+      | _ -> false in
+    if full_range then base
+    else
+      let type_ =
+        Expr.Helpers.infer_type_of_indexed (Expr.Typed.type_of base)
+          [Between (lower, upper)] in
```
nhuurre: This is always `Expr.Typed.type_of base`.

src/analysis_and_optimization/Optimize.ml
```diff
@@ -642,6 +641,135 @@ let unroll_loop_one_step_statement _ =
...
+        { pattern= Indexed (base, [Index.Between (lower, upper)])
+        ; meta= {base.meta with type_} } in
+  let classify_arg ~loopvar (arg : Expr.Typed.t) =
+    if not (Set.mem (expr_var_names_set arg) loopvar) then
+      (* Only scalar invariants broadcast. An invariant container argument is
+         summed over anew by every iteration of the scalar loop, but the
+         vectorized call would zip it against the sliced arguments elementwise
+         and both forms typecheck, so this must be ruled out here. *)
+      match Expr.Typed.type_of arg with
+      | UInt | UReal | UComplex -> Some (`Invariant arg)
```
nhuurre: I would write this logic as
```ocaml
match arg.pattern with
 | Indexed (({pattern= Var _; _} as base), [Single {pattern= Var v; _}])
    when v = loopvar ->
     Some (`Sliced base)
 | _ when UnsizedType.is_scalar_type (Expr.Typed.type_of arg)
       && not (Set.mem (expr_var_names_set arg) loopvar) ->
    Some (`Invariant arg)
 | _ -> None
```
Though apparently `is_scalar_type` thinks `UComplex` is not a scalar? Well, then you could also write
```ocaml
match arg with
 | {pattern=Indexed (...);_} -> ...
 | {meta={type_= UInt | UReal | UComplex; _}; _}
    when not (Set.mem (expr_var_names_set arg) loopvar) ->
      Some (`Invariant arg)
 | _ -> None
```
And technically you should also check `cannot_remove_expr arg` because the "invariant" expression might have side effects.

src/analysis_and_optimization/Optimize.ml
```diff
@@ -642,6 +641,135 @@ let unroll_loop_one_step_statement _ =
...
+           ; _ } as e) ->
+          Some (e, name, suffix, mem, args)
+      | _ -> None in
+    let* classified = List.map args ~f:(classify_arg ~loopvar) |> Option.all in
+    let* () =
+      Option.some_if
+        (List.exists classified ~f:(function
+          | `Sliced _ -> true
+          | `Invariant _ -> false))
+        () in
```
nhuurre, line 767: If all arguments are invariant, the loop can still be "vectorized." For example,
```
for (i in lower:upper)
  target += dist_lpdf(args);
```
turns into
```stan
if (upper >= lower)
  target += (upper - lower + 1) * dist_lpdf(args);
```

nhuurre (commented):
Please see [Stan AI Contribution policy](https://github.com/stan-dev/stan/wiki/AI-Contribution-Policy). It's not mentioned in README.md or CONTRIBUTING.md (we don't even _have_ a CONTRIBUTING.md) but there is a link in the documentation.

More vectorization-related discussion: https://github.com/stan-dev/stanc3/discussions/877

This would be more useful if it could vectorize multiple statements in a single for loop, e.g.
```
for (i in 1:N) {
  x[i] ~ std_normal();
  y[i] ~ normal(x[i], 1);
}
```

And what about vectorizing truncated distributions?
```
for (i in 1:N) {
  x[i] ~ std_normal() T[y,];
}
```
is shorthand for
```
for (i in 1:N) {
  target += std_normal_lupdf(x[i]);
  if (x[i] > y)
    target += -std_normal_lccdf(y);
  else
    target += negative_infinity();
}
```
Even if you can't vectorise the LCDF part it would be good to pull the LPDF part out of the loop.

RELEASE-NOTES.txt
WardBrian, line 1: We update the changelog during releases, not individual PRs, to avoid too many conflicts on this file

seantalts:
This is still a draft so review is premature. But thank you for the comments! This is the first in a series of PRs that have been broken up for easier review. The full capabilities of loop re-rolling / "vectorization" are not included in the first PR.

src/analysis_and_optimization/Optimize.ml
seantalts: You're right. I tested it and whole-variable assignments are size-checked at runtime, for vectors and arrays both. So declared sizes can't change. The latest push trusts transformed data and all output variables and fixes the comment.

src/analysis_and_optimization/Optimize.ml
seantalts: Done, following your sketch.

src/analysis_and_optimization/Optimize.ml
seantalts: True. Switched to add_int_index. Turns out that helper types vector[MultiIndex] as real, fix in #1667.

src/analysis_and_optimization/Optimize.ml
seantalts: Done. Cleaner, and apparently a bug fix :) an invariant _lp argument would have bumped target once instead of N times. Added a regression test.

src/analysis_and_optimization/Optimize.ml
seantalts: Yeah, that works. Leaving it out to keep this PR small, but it's on the list.

RELEASE-NOTES.txt
seantalts: Removed.

seantalts:
Multi-statement bodies are coming later in the series. The next PR does indirect indexing like a[county[n]] and assignment loops. Pulling the lupdf term out of truncation expansions is a nice idea, added it to the plan for the multi-statement PR.

src/analysis_and_optimization/Optimize.ml
```diff
@@ -642,6 +641,156 @@ let unroll_loop_one_step_statement _ =
...
+let cannot_duplicate_expr ?(preserve_stability = false) (e : Expr.Typed.t) =
+  let pred e =
+    can_side_effect_top_expr e
+    || (match e.pattern with
+      | FunApp ((UserDefined (_, FnRng) | StanLib (_, FnRng, _)), _) -> true
+      | _ -> false)
+    || (preserve_stability && UnsizedType.is_autodiffable e.meta.type_) in
+  expr_any pred e
+
+let cannot_remove_expr (e : Expr.Typed.t) = expr_any can_side_effect_top_expr e
```
seantalts, line 673: moved from below

seantalts:
ok ready!

src/analysis_and_optimization/Optimize.ml
```diff
@@ -642,6 +641,154 @@ let unroll_loop_one_step_statement _ =
...
+let vectorize_loops (mir : Program.Typed.t) =
+  let outer_size st = List.hd (SizedType.get_dims st) in
+  (* Declared outer sizes never change at runtime. Whole-variable assignments
+     are size-checked by stan::model::assign, and nothing else can resize. So a
+     declared size holds wherever the name appears. *)
+  let trusted_sizes =
+    List.filter_map mir.input_vars ~f:(fun (name, _, st) ->
+        Option.map (outer_size st) ~f:(fun d -> (name, d)))
+    @ List.filter_map mir.output_vars ~f:(fun (name, _, ov) ->
+        Option.map (outer_size ov.out_constrained_st) ~f:(fun d -> (name, d)))
```
nhuurre: You need to filter out generated quantities.
```stan
model {
  vector[3] x = rep_vector(1.0, 3);
  for (i in 1:2)
    x[i] ~ std_normal();
}
generated quantities {
  vector[2] x;
}
```

seantalts: Good catch, fixed. Generated quantity names are no longer trusted, only data, parameters, and transformed parameters, whose names can't be reused. Your example now vectorizes to the slice form, x[1:2] ~ std_normal(). Added it as a regression test.

nhuurre (approved):

PR #1667, opened 2026-08-16
Fix infer_type_of_indexed handling of MultiIndex

seantalts:
`Expr.Helpers.infer_type_of_indexed` mis-types multi-indexing. `vector[MultiIndex]` returns `real` (it falls through to the scalar arm), `matrix[MultiIndex]` returns `vector` (it is grouped with the `Single` arms), and `matrix[MultiIndex, Single]` matches no arm and raises an internal error. The rule is that `Single` reduces a dimension while `All`, `Upfrom`, `Between`, and `MultiIndex` preserve it.

The bug is latent today because codegen ignores the type metadata. It matters for anything that trusts it: `Memory_patterns.matrix_set` stops recursing at non-Eigen types, so an expression wrongly typed `real` can leave a variable out of the SoA demotion set. I hit this while working on #1666, where a compiler-built gather typed `real` produced C++ that assigned an SoA value into an AoS slice and did not compile. `Ast_to_Mir.copy_indices` also feeds arbitrary index lists into this function and could hit the internal error.

The fix moves `MultiIndex` into the dimension-preserving arm and adds `[MultiIndex, Single]` to the matrix arms next to `[All, Single]`. New cases in the inline expect test cover vector, row vector, matrix, matrix-then-column, and arrays. No other test output changes.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

Fixed the inferred type of multi-indexed expressions in the compiler's middle intermediate representation.

## Copyright and Licensing

I used AI but understand the code and think it makes sense.

### DIFF
--- a/src/middle/Expr.ml
+++ b/src/middle/Expr.ml
@@ -207,26 +207,34 @@ module Helpers = struct
     match (ut, indices) with
     | _, [] -> ut
     | _, [Index.All] | _, [Upfrom _] | _, [Between _] | _, [MultiIndex _] -> ut
-    | UnsizedType.UMatrix, [All; Single _]
-     |UMatrix, [Upfrom _; Single _]
-     |UMatrix, [Between _; Single _]
-     |UMatrix, [MultiIndex _; Single _]
-     |UMatrix, [Single _] ->
+    | ( (UnsizedType.UMatrix | UComplexMatrix)
+      , [ (All | Upfrom _ | Between _ | MultiIndex _)
+        ; (All | Upfrom _ | Between _ | MultiIndex _) ] ) ->
+        ut
+    | UMatrix, [Single _; (All | Upfrom _ | Between _ | MultiIndex _)] ->
+        URowVector
+    | UComplexMatrix, [Single _; (All | Upfrom _ | Between _ | MultiIndex _)] ->
+        UComplexRowVector
+    | ( UMatrix
+      , ([(All | Upfrom _ | Between _ | MultiIndex _); Single _] | [Single _]) )
+      ->
         UVector
-    | UComplexMatrix, [All; Single _]
-     |UComplexMatrix, [Upfrom _; Single _]
-     |UComplexMatrix, [Between _; Single _]
-     |UComplexMatrix, [MultiIndex _; Single _]
-     |UComplexMatrix, [Single _] ->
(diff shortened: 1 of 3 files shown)


nhuurre:
Thanks. It seems lines 224, 226, and 227 implicitly rely on the wildcard index matching only `Single _` because all others were already handled on line 209. (which was not the case for `MultiIndex _` as you discovered)

I'd be more confident that this fixes all the corner cases if the final internal error arm was an exhaustive match instead of wildcard. I believe it should be
```ocaml
    | ( (UInt | UReal | UComplex | UTuple _ | UFun _ | UMathLibraryFunction)
      , _ :: _ )
     |(UVector | URowVector | UComplexVector | UComplexRowVector), _ :: _ :: _
     |(UMatrix | UComplexMatrix), _ :: _ :: _ :: _ ->
        ICE.(internal_errorf "Can't index %t" [UnsizedType.pp $ ut])
        [@coverage off]
```

seantalts:
Done. Making it exhaustive also forced two arms the wildcard used to swallow: matrix with a Single row and a range column is a row_vector, and matrix with two range indices is a matrix. Both used to hit the internal error. Added test cases for those.

src/middle/Expr.ml
```diff
@@ -207,26 +207,34 @@ module Helpers = struct
...
+    | ( (UnsizedType.UMatrix | UComplexMatrix)
+      , [ (All | Upfrom _ | Between _ | MultiIndex _)
+        ; (All | Upfrom _ | Between _ | MultiIndex _) ] ) ->
+        ut
+    | UMatrix, [Single _; (All | Upfrom _ | Between _ | MultiIndex _)] ->
+        URowVector
+    | UComplexMatrix, [Single _; (All | Upfrom _ | Between _ | MultiIndex _)] ->
+        UComplexRowVector
+    | ( UMatrix
+      , ([(All | Upfrom _ | Between _ | MultiIndex _); Single _] | [Single _]) )
```
nhuurre: This looks like it covers both `x[ : ,1]` and `x[1]`. But `x[1]` is the same as `x[1, : ]`. I believe this is a pre-existing bug. `Matrix[Single]` should be a row vector.

nhuurre (approved):

PR #1668, opened 2026-08-16
vectorize_loops: gathers, assignment loops, and widened arguments

seantalts:
Rebased onto master after #1666 merged. Single commit.

Widening is now recursive. `x[n]` becomes the slice `x[lower:upper]`, or `x` when the range provably spans the declaration. `x[idx[n]]` becomes the multi-index gather `x[idx[lower:upper]]`. A StanLib call widens its arguments and is re-typechecked. It has to come back as a container: the loop was using the scalar signature, so the container overload is the elementwise map, and functions like dot_product that typecheck at a scalar are rejected. Operators fall back to their elementwise variant, so `x[n] * w[n]` becomes `x .* w`. Suffixed calls are excluded since a vectorized `_lpdf` value is the summed lp.

This covers density arguments with arithmetic in them:

```stan
for (n in 1:N) y[n] ~ normal(a[county[n]] + mu[n] / 2, sigma);
```

and assignment loops:

```stan
for (i in 1:N) y_hat[i] = a[county[i]];
// becomes
y_hat[1:N] = a[county];
```

For assignments the written variable can't appear in the right-hand side or the loop bounds, and the widened right-hand side has to match the target's declared type. Gathers are typed as their base container directly instead of through `infer_type_of_indexed`, see #1667.

Measurements, same protocol as #1666: 11 of 120 posteriordb models change and the other 109 are byte-identical. election88_full goes 716us to 453us per gradient (1.58x). surgical_model, losscurve_sislob, and pilots are 1.05x to 1.08x. log10earn_height is 1.01x. radon_county is 0.92x, the sliced assignment costs one extra pass over the data compared to its scalar loop, and the fix for that belongs in codegen. Log probability and gradients agree with the pass off in every model. stanc runs a little faster on the changed models and clang compile times don't move.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] OR, no user-facing changes were made (only `--Oexperimental` behavior changes)

## Release notes

The `vectorize_loops` optimization (enabled at `--Oexperimental`) now handles indirectly indexed arguments such as `a[county[n]]`, density arguments built from elementwise arithmetic, and elementwise assignment loops.

## Copyright and Licensing

I used AI.


seantalts:
I'm just going to close this, the earlier PR was reverted.

nhuurre:
It was not reverted, it just required a small fix: #1672

WardBrian:
Apologies for the confusion @seantalts -- it wasn't reverted (and if it had been, it was only going to be for the weekend until I could fix the build issue). This just needs a rebase on master I believe

seantalts:
Ah I see, thanks for fixing! I'm not very familiar with github anymore and used the new automerge button. I find I kinda don't care enough to make the comments pithy and poignant, but I will instruct Fable in that way and see how it goes :P

seantalts:
Continued in #1678 after a rebase onto master. GitHub can't reopen a PR whose branch was force-pushed after closing.

Issue #177, opened 2019-06-14
Tool to generate model diagram


seantalts:
It would be nice to be able to generate a plate or graph diagram of a Stan model, assuming the model specified is equivalent to a DAG. 

Some discussion here: https://discourse.mc-stan.org/t/tool-to-auto-generate-model-diagrams/7361/

VMatthijs:
And in case, we aren't a generative model, we can always visualize a factor graph. This shouldn't be super hard to do a conservative approximation of. However, as it would all depend on a dependence analysis, things would get trickier when it comes to conditional independence of different array elements (as that requires an array element level dependence analysis, which can be done decently in special cases, but not in general). Note that it is fine, however, to produce a graphical model which is overly conservative in the sense of capturing some but not all independence relationships. This is precisely what happens with dependence analyses as well: they are conservative. We are guaranteed to find a superset of all the actual dependencies in the program.

PR #1672, opened 2026-08-21
Fix Optimize.ml to not depend on Base

nhuurre:
#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [ ] If a user-facing facing change was made, the documentation PR is here: <LINK>
    - [x] OR, no user-facing changes were made

#1666 branched off before #1663 so it was still depending on Base, and consequently broke builds (#1670)

## Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause)

PR #1673, opened 2026-08-21
Add flag to print compiler's over-approximation of factor graph

collin-cademartori:
#### Submission Checklist

- [ ] Run unit tests
- Documentation

## Summary

Adds a debug flag that causes the compiler to compute and print it's factor graph approximation in DOT format.

The compiler already has machinery for computing a conservative over-approximation of the factor graph for the model(s) that a Stan program implements in `Factor_graph.ml`. I believe this was implemented in response to [#177](https://github.com/stan-dev/stanc3/issues/177), and while the module contains a function to print this graph to DOT format, this function is never called anywhere in the compiler or exposed to users. While the approximation is coarse, the information can still be useful.

Tests: The test `integration/cli-args/stanc.t` currently fails. I was unsure whether I should edit that test file manually. I think I recall @WardBrian saying something about a tool for automatically updating expected test output, but I couldn't figure out how to run that. I'm happy to update the test in whatever way is standard practice.

Docs: While the change is user-facing, I noticed that not all debug flags are currently documented in the [stanc CLI doc](https://mc-stan.org/docs/stan-users-guide/using-stanc.html), so I defaulted to not adding documentation so as to not clutter the doc. Also happy to add that if desired.

## Release notes

Add debug flag `--debug-print-factor-graph` to print the compiler's conservative over-approximation of the factor graph for the model(s) implemented in the Stan program.

## Copyright and Licensing

### DIFF
--- a/src/driver/Entry.ml
+++ b/src/driver/Entry.ml
@@ -91,8 +91,8 @@ let stan2cpp model_name model (flags : Flags.t) (output : other_output -> unit)
   if flags.warn_pedantic then
     output (Warnings (Pedantic_analysis.warn_pedantic mir));
   if flags.debug_settings.debug_print_factor_graph then
-    print_endline (Factor_graph.factor_graph_to_dot 
-      (Factor_graph.prog_factor_graph mir));
+    print_endline
+      (Factor_graph.factor_graph_to_dot (Factor_graph.prog_factor_graph mir));
   debug_output_mir output mir flags.debug_settings.print_mir;
   let* generation_context =
     match flags.debug_settings.debug_data_json with
--- a/src/driver/Entry.ml
+++ b/src/driver/Entry.ml
@@ -90,6 +90,9 @@ let stan2cpp model_name model (flags : Flags.t) (output : other_output -> unit)
     output (Warnings (Pedantic_analysis.warn_uninitialized mir));
   if flags.warn_pedantic then
     output (Warnings (Pedantic_analysis.warn_pedantic mir));
+  if flags.debug_settings.debug_print_factor_graph then
+    print_endline (Factor_graph.factor_graph_to_dot 
+      (Factor_graph.prog_factor_graph mir));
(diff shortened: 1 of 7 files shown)


WardBrian:
Thanks @collin-cademartori!


The relevant test command is “dune runtest”, which will show a diff. If the diff looks good, “dune promote” will update your local files. 

The current build failure in Jenkins will be fixed by #1672, so don’t worry about that specific error too much

nhuurre:
Why is this considered a "debug" flag?

WardBrian:
I think it falls into the same category as our existing flags for inspecting the MIR and optimizer, but I could be convinced it deserves a more “permanent” feeling name

WardBrian:
@collin-cademartori since you've been looking at this output already, you might be in a good position to weigh in on if https://github.com/stan-dev/stanc3/pull/1603 improves the approximation or not. If you do have thoughts we'd appreciate it!

collin-cademartori:
Happy to look at it in the next few days! And thanks for pointing out `dune promote`, I've updated the test file.

As for whether this goes behind a debug flag, a couple thoughts:

- The factor graph plays a role in the analysis routines, but also carries some information about the statistical model, so I think it straddles internal/user-facing (though less purely internal than the other debug flags).
- If the factor graph output weren't behind a debug flag, I imagine it would be added to the CLI docs, and it would be good to have a written explanation somewhere of exactly what over-approximation to the model factor graph this is computing (e.g. no visibility into non-scalar quantities, how control flow-dependent structure is handled, etc). I think this can be done, but I'd need to carefully read and test it to be confident about it.

I think the main reason to keep it behind a debug flag is that the implementation seems scoped to what the compiler internals need. Because of the limitations of the approximation, it doesn't consistently produce the same statistical information across Stan model implementations. 

That said, I'm happy to reconsider the scope if there's disagreement about exposing the factor graph computation in its current incarnation.

WardBrian (approved):
Thank you for your contribution @collin-cademartori!

PR #1346, opened 2023-08-16
Add tuple-returning special functions

WardBrian:
Counterpart to https://github.com/stan-dev/math/pull/2931

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/664
    
## Release notes

Exposed new functions `qr_thin`,  `eigendecompose_sym`, `eigendecompose`, `complex_schur_decompose`, `svd`, and `csr_extract`.

## Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause)

PR #1674, opened 2026-08-23
Fuse adjacent eigenvectors_sym/eigenvalues_sym pairs into one eigendecompose_sym call (--O1+), add pedantic warning

sims1253:
The natural two-call idiom (and essentially all pre-2.34 GP / latent-factor code):

```stan
matrix[N, N] Q = eigenvectors_sym(A);
vector[N] w = eigenvalues_sym(A);
```

compiles to two independent calls, each constructing and running a full `SelfAdjointEigenSolver` in `ComputeEigenvectors` mode — `eigenvalues_sym` too, because its reverse-mode adjoint `V diag(ḡ_w) Vᵀ` needs the vectors. Per gradient evaluation, for every symmetric matrix treated this way: 2 full decompositions where 1 suffices (each discarding half the result) and 2 reverse-pass callbacks instead of 1. On a Kronecker-GP-class model (two 30×30 symmetric eigendecompositions and one 2×2 per gradient), eigh is 39.3% of total program instructions and the model runs 4 decompositions per gradient where 2 would suffice.

The combined primitive already exists — `eigendecompose_sym`, one solver run, one callback, both adjoints (math PR #2931 / stanc3 PR #1346, language support since CmdStan 2.34) — and the hand rewrite is bit-identical at the model level (the two-callback adjoints and the combined callback accumulate into the same zero-initialized operand adjoint; values come from the same Eigen solver either way — structural, not luck). But the compiler never generates it from the two-call idiom, so the existing corpus keeps paying double. This PR teaches stanc3 the rewrite.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] If a user-facing facing change was made, the documentation PR is here: https://github.com/stan-dev/docs/pull/972

## Release notes

Adds a peephole optimization (enabled at `--O1` and above) that fuses an adjacent `eigenvectors_sym(A)` + `eigenvalues_sym(A)` pair into a single `eigendecompose_sym(A)` call, halving the number of full eigendecompositions in models that use both on the same matrix (e.g. GP kernels). Measured −15.6% time per gradient on a kronecker-GP model, with bit-identical output. Also adds a pedantic-mode warning for the same pattern (fires when the pair is not fused, e.g. at `--O0` or with non-adjacent/impure arguments), recommending `eigendecompose_sym`; documented in the Stan Users Guide.

## Copyright and Licensing


WardBrian:
This is a good idea, but note:

- Stan does not allow fully agentic PRs https://github.com/stan-dev/stan/wiki/AI-Contribution-Policy
- there are a small handful of other functions with exactly the same structure, not least of all the non _sym versions but also functions like qr decomposition

PR #1678, opened 2026-08-24
vectorize_loops: gathers, assignment loops, and widened arguments

seantalts:
Rebased onto master after #1679 merged.

This extends `vectorize_loops` in three related ways:

- `x[idx[n]]` becomes the multi-index gather `x[idx[lower:upper]]`.
- Scalar Stan Math expressions are lifted recursively when their container arguments re-typecheck. Operators retry their elementwise form, so `x[n] * w[n]` becomes `x .* w`. Calls that still return a scalar, such as `dot_product`, are rejected.
- Elementwise assignment loops become whole-container or sliced assignments. For example:

```stan
for (i in 1:N) y_hat[i] = a[county[i]];
// becomes
y_hat[ : ] = a[county];
```

The assignment target uses the whole-container index only when the range is proven to cover its current declared outer dimension. A partial range uses a slice such as `y_hat[2:N]`.

The loop body may contain multiple assignments or density increments. They move out of the loop only when every statement vectorizes, assignment targets are unique, and the statements' read/write sets do not interfere. Otherwise the original loop remains.

Correctness guards added during review:

- The rewritten assignment uses the unindexed target type already carried by MIR `Assignment`.
- Outer-dimension facts follow MIR scope: `SList` threads them forward, while `Block` and `Profile` contain declarations. Assignments, including assignments nested under control flow, invalidate dimension facts that read the written variable.
- A reassigned size therefore produces the required slice, such as `z[:] = y[1:M]`, rather than an incorrect whole-container read.
- The written variable cannot occur in the right-hand side or loop bounds.
- Calls with effects, including RNG calls, cannot have their evaluation count changed.
- The lifted right-hand side must match the target's declared container type.

Current-head validation:

- The focused Optimize unit partition and the full stanc3 integration suite pass locally.
- A current-master versus current-branch `--Oexperimental` census compiled all 120 posteriordb models; seven models had different optimized MIR and 113 were byte-identical.
- An end-to-end pass-off/pass-on run used the exact current PR head and the stanli runtime with full-span assignment support. It covered 131 models and 393 semantic points: 12 models had different portable MIR, with zero semantic failures and zero measurement-infrastructure failures.
- The pass changed 24 finite values across five models through arithmetic order. Every result remained within its CmdStan reference gate, all compared shapes and nonfinite categories agreed, and write-array values were identical. The largest bounded gradient on/off ratio was 1.0080 and the smallest was 0.9966; timing ratios are descriptive.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] OR, no user-facing changes were made (only `--Oexperimental` behavior changes)

## Release notes

The `vectorize_loops` optimization (enabled at `--Oexperimental`) now handles indirectly indexed arguments such as `a[county[n]]`, density arguments built from elementwise arithmetic, independent statements within one loop, and elementwise assignment loops.

## Copyright and Licensing

I used AI.


seantalts:
I think there are some places we can re-use some of the analysis here and maybe cache it between passes. For a later PR.

test/unit/Optimize.ml
```diff
@@ -3833,8 +3833,304 @@ let%expect_test "vectorize bail: indirect indexing (vectorized in a later PR)" =
...
+  [%expect
+    {|
+    real alpha;
+    real beta;
+    {
+      FnValidateSize__("mu", "N", N);
+      vector[N] mu;
+      FnValidateSize__("v", "N", N);
+      vector[N] v;
+      mu[1:N] = (alpha + (beta * x));
```
WardBrian: I'm a bit surprised these are not just `mu = ...`. Is that just one step too far in the analysis? I don't think the result ultimately matters all that much, but curious

test/unit/Optimize.ml
```diff
@@ -3833,8 +3833,304 @@ let%expect_test "vectorize bail: indirect indexing (vectorized in a later PR)" =
...
+      FnValidateSize__("v", "N", N);
+      vector[N] v;
+      for(n in 1:N) {
+        v[n] = (x[n] * u[n]);
+      }
+      target += normal_lpdf(v, promote(0, real, data), promote(1, real, data));
+    }
+    |}]
+
+let%expect_test "vectorize bail: doubly indirect indexing" =
```
WardBrian, line 4107: I guess I would have expected this case to work when you say "Widening is now recursive.", but I think it's probably a rare case anyway?

WardBrian:
> Patch coverage is `93.20388%` with `7 lines` in your changes missing coverage. Please review.

These seven lines all look like interesting cases worth testing (besides maybe the eltpow one)

test/unit/Optimize.ml
SteveBronder: If mu is size N I thought we had another pass that collapsed these types of indices?

seantalts: some codegen overloads of assign(x, y, name) don't exist for all the uses we have; some models stopped compiling. Maybe a bug to fix there in another PR. I can try `mu[:] = ...` instead.

test/unit/Optimize.ml
seantalts: No signature of * or .* takes array, vector. Could do a to_vector version of this in the future?

test/unit/Optimize.ml
seantalts: mu[:] works and is way faster lol

WardBrian:
Is "widened" an existing term of art or is this one of those cases of the llm inventing jargon

seantalts:
I haven't worked with this exact optimization but we often say widening when extending a type like i32 -> i64 or using more vectorization lanes, 4xf32 -> 8xf32. It makes sense to me but if you would prefer other language we can change it.

test/unit/Optimize.ml
SteveBronder: If we supported array, vector .* I wonder if we should expose it to users or just leave it as a thing for the compiler?

SteveBronder:
I thought it was referred to as loop folding?

WardBrian:
Huh, TIL. I'd always just heard it called (auto-) vectorization 

> These instructions and registers are usually called "wide" or "vector" instructions / registers (which is why you may see vectorization being referred as widening)

https://sbaziotis.com/performance/a-beginners-guide-to-vectorization-by-hand-part-1.html

seantalts:
imo everyone tries to invent slightly different new names for everything, especially in compilers, and I always personally prefer describing what is happening in ~5 words instead of using a given name

src/analysis_and_optimization/Optimize.ml
```diff
@@ -671,64 +671,122 @@ let cannot_duplicate_expr ?(preserve_stability = false) (e : Expr.Typed.t) =
...
+   becomes [x[lower:upper]], or [x] alone when the range provably spans the
+   declaration. Loops that do not match are left unchanged. *)
 let vectorize_loops (mir : Program.Typed.t) =
   let outer_size st = List.hd (SizedType.get_dims st) in
-  let trusted_sizes =
-    List.filter_map mir.input_vars ~f:(fun (name, _, st) ->
-        Option.map (outer_size st) ~f:(fun d -> (name, d)))
+  let global_env =
+    List.map mir.input_vars ~f:(fun (name, _, st) ->
+        (name, (SizedType.to_unsized st, outer_size st)))
```
nhuurre: So the code later tracks local variables sizes and `global_env` is only needed to track the sizes from the other blocks, right? Not entirely sure how these optimization passes work but I _think_ `global_env` might not be necessary if you use `transform_program` instead of `transform_program_blockwise`?

Regardless, I actually think that using known sizes to simplify indexing is quite separate from loop vectorization and could instead be part of partial evaluation.

src/analysis_and_optimization/Optimize.ml
```diff
@@ -738,43 +796,87 @@ let vectorize_loops (mir : Program.Typed.t) =
...
+      else Index.Between (lower, upper) in
+    Some
+      (Stmt.Pattern.Assignment
+         ((LVariable v, [lhs_index]), Expr.Typed.type_of rhs', rhs')) in
+  let vectorize_for env ~loopvar ~lower ~upper body =
+    let open Stdlib.Option.Syntax in
+    let* stmt = singleton body in
+    match vectorize_density env ~loopvar ~lower ~upper stmt with
+    | Some p -> Some p
+    | None -> vectorize_assign env ~loopvar ~lower ~upper stmt in
```
nhuurre: Kind of klunky to chain them like this. At least match on `stmt.pattern` before delegating to `vectorize_density`/`vectorize_assign`!

Also not a fan of `singleton`. The ultimate goal is to collect all non-interfering statements from the loop, but you're leaving that for a later PR to make review easier, right? I'd rather see how you tackle that as early as possible because I think it will inform the high-level structuring of this code.

src/analysis_and_optimization/Optimize.ml
```diff
@@ -738,43 +796,87 @@ let vectorize_loops (mir : Program.Typed.t) =
...
+  let rec go env (stmt : Stmt.Located.t) =
+    match stmt.pattern with
+    | Block stmts -> {stmt with pattern= Block (go_list env stmts)}
+    | SList stmts -> {stmt with pattern= SList (go_list env stmts)}
```
nhuurre: This obviously can't be right -- there must be _some_ difference between `SList` and `Block`!

The difference is that `SList` should not discard the `env`.

src/analysis_and_optimization/Optimize.ml
```diff
@@ -738,43 +796,87 @@ let vectorize_loops (mir : Program.Typed.t) =
...
+        | Some pattern -> {stmt with pattern}
+        | None ->
+            {stmt with pattern= For {loopvar; lower; upper; body= go env body}})
+    | pattern -> {stmt with pattern= Stmt.Pattern.map Fun.id (go env) pattern}
+  and go_list env stmts =
+    List.rev
+      (snd
+         (List.fold_left stmts ~init:(env, []) ~f:(fun (env, acc) s ->
+              let env = track_decl env s.Stmt.pattern in
+              (env, go env s :: acc)))) in
```
nhuurre: `go` and `go_list` are rather nondescript names, and I find this code a bit difficult to follow. `track_decl` is a better name but the definition is quite far from here and the function doesn't even do that much.

Assignments can invalidate known sizes. e.g.
```stan
data {
  int N;
  vector[N] x;
} transformed data {
  int M = N;
  vector[M] y = x;
  M = N - 1;
  vector[M] z; // not the same size as y
  for (i in 1:M) z[i] = y[i]; // not z[:] = y;
  print(z);
}
```

src/analysis_and_optimization/Optimize.ml
```diff
@@ -738,43 +796,87 @@ let vectorize_loops (mir : Program.Typed.t) =
...
+    let reads_v e = Set.Poly.mem v (expr_var_names_set e) in
+    let* () =
+      Option.some_if
+        ((not (reads_v rhs)) && (not (reads_v lower)) && not (reads_v upper))
+        () in
+    let* () = Option.some_if (varies_with loopvar rhs) () in
+    let* rhs' = widen env ~loopvar ~lower ~upper rhs in
+    let* v_type, _ = String.Map.find_opt v env in
+    let* () =
+      Option.some_if (UnsizedType.equal v_type (Expr.Typed.type_of rhs')) ()
```
nhuurre: You extract `v_type` from the `env` map but looks like the Assignment statement also contains the unindexed type of the target variable.
```ocaml
pattern= Assignment (lvar, v_type, rhs)
```

src/analysis_and_optimization/Optimize.ml
seantalts: Addressed in cf29b9af: rewrite_sequence threads dimension facts through SList. Block and Profile scope their inner declarations, while assignments inside them conservatively invalidate affected outer facts.

src/analysis_and_optimization/Optimize.ml
seantalts: Addressed in cf29b9af: dispatch now starts by matching stmt.pattern. loop_body_statements handles a list, and multiple statements vectorize only when every statement widens and their read/write sets prove that they do not interfere. Repeated targets and cross-statement dependencies retain the loop; unit tests cover both cases.

src/analysis_and_optimization/Optimize.ml
seantalts: Addressed in cf29b9af: vectorize_assign captures the unindexed target type directly from Assignment and uses it for both the compatibility check and the rewritten assignment.

src/analysis_and_optimization/Optimize.ml
seantalts: Addressed in cf29b9af: the pass now uses transform_program, removes global_env, and records outer dimensions only from declarations encountered in scope. Assignments invalidate any dimension fact that reads the assigned variable. The remaining dimension lookup is limited to deciding whether an exact 1:upper range can use Index.All; broader index simplification stays out of this PR.

src/analysis_and_optimization/Optimize.ml
seantalts: Addressed in cf29b9af: the traversal is now rewrite_statement/rewrite_sequence, and assignments nested in control flow or lexical blocks invalidate dimension facts used later. The counterexample now rewrites safely to z[:] = y[1:M], never z[:] = y.

src/analysis_and_optimization/Optimize.ml
```diff
@@ -671,64 +671,106 @@ let cannot_duplicate_expr ?(preserve_stability = false) (e : Expr.Typed.t) =
...
-  let vectorize_for sizes ~loopvar ~lower ~upper (body : Stmt.Located.t) =
+  let widen sizes ~loopvar ~lower ~upper rhs =
+    let open Stdlib.Option.Syntax in
+    let rec widen_expr (e : Expr.Typed.t) =
+      if not (varies_with loopvar e) then
+        if cannot_duplicate_expr e then None else Some e
+      else
+        let scalar_lane =
+          match Expr.Typed.type_of e with UInt | UReal -> true | _ -> false
+        in
```
WardBrian: Since all the branches of the subsequent match guard on this, might be clearer to just nest a conditional

WardBrian: (Also, this is `UnsizedType.is_scalar_type`)

src/analysis_and_optimization/Optimize.ml
```diff
@@ -671,64 +671,106 @@ let cannot_duplicate_expr ?(preserve_stability = false) (e : Expr.Typed.t) =
...
+    else DataOnly in
+  let varies_with loopvar e = Set.Poly.mem loopvar (expr_var_names_set e) in
+  let elt_variant name =
+    match Operator.of_string_opt name with
+    | Some Times -> Some (Operator.to_string EltTimes)
+    | Some Divide -> Some (Operator.to_string EltDivide)
     | _ -> None in
-  let vectorize_for sizes ~loopvar ~lower ~upper (body : Stmt.Located.t) =
+  let widen sizes ~loopvar ~lower ~upper rhs =
+    let open Stdlib.Option.Syntax in
```
WardBrian:
```suggestion
    let open Option.Syntax in
```

src/analysis_and_optimization/Optimize.ml
```diff
@@ -671,64 +671,106 @@ let cannot_duplicate_expr ?(preserve_stability = false) (e : Expr.Typed.t) =
...
+    | None -> Frontend.Typechecker.stan_math_return_type name arg_types in
+  let adlevel_over args =
+    if UnsizedType.any_autodiff (List.map ~f:Expr.Typed.adlevel_of args) then
+      UnsizedType.AutoDiffable
+    else DataOnly in
+  let varies_with loopvar e = Set.Poly.mem loopvar (expr_var_names_set e) in
+  let elt_variant name =
+    match Operator.of_string_opt name with
+    | Some Times -> Some (Operator.to_string EltTimes)
+    | Some Divide -> Some (Operator.to_string EltDivide)
```
WardBrian, line 720: Even though it is equivalent currently, I'd prefer to leave the EltPow case in this match

src/analysis_and_optimization/Optimize.ml
```diff
@@ -671,64 +671,106 @@ let cannot_duplicate_expr ?(preserve_stability = false) (e : Expr.Typed.t) =
...
+                  | None -> None) in
+            Some
+              Expr.
+                { pattern= FunApp (StanLib (name', FnPlain, mem), args')
+                ; meta= {e.meta with type_; adlevel= adlevel_over args'} }
+        | _ -> None in
+    widen_expr rhs in
+  let loop_body_statements (body : Stmt.Located.t) =
+    match body.Stmt.pattern with
+    | Block stmts | SList stmts -> Some stmts
```
WardBrian: Might as well include `Profile` blocks, no?

src/analysis_and_optimization/Optimize.ml
```diff
@@ -738,43 +780,140 @@ let vectorize_loops (mir : Program.Typed.t) =
            ; _ } as e) ->
           Some (e, name, suffix, mem, args)
       | _ -> None in
-    let* classified =
-      List.map args ~f:(vectorize_arg sizes ~loopvar ~lower ~upper)
+    let* () = Option.some_if (List.exists args ~f:(varies_with loopvar)) () in
```
WardBrian: This is an odd way to write an if statement. I guess it avoids a layer of nesting?

src/analysis_and_optimization/Optimize.ml
```diff
@@ -738,43 +780,140 @@ let vectorize_loops (mir : Program.Typed.t) =
...
+    (* The generated for statement re-evaluates its bounds every iteration. *)
+    let reads_v e = Set.Poly.mem v (expr_var_names_set e) in
+    let* () =
+      Option.some_if
+        ((not (reads_v rhs)) && (not (reads_v lower)) && not (reads_v upper))
+        () in
+    let* () = Option.some_if (varies_with loopvar rhs) () in
+    let* rhs' = widen sizes ~loopvar ~lower ~upper rhs in
+    let* () =
+      Option.some_if (UnsizedType.equal v_type (Expr.Typed.type_of rhs')) ()
```
WardBrian: Huh, the llm must really like this code pattern!

I don't actively hate it, but I'd be curious what it would look like if not written in this style

src/analysis_and_optimization/Optimize.ml
```diff
@@ -738,43 +780,140 @@ let vectorize_loops (mir : Program.Typed.t) =
...
+    let* () = Option.some_if (statements_do_not_interfere stmts) () in
+    let* patterns =
+      List.map stmts ~f:(vectorize_statement sizes ~loopvar ~lower ~upper)
+      |> Option.all in
+    match
+      List.map2 stmts patterns ~f:(fun (stmt : Stmt.Located.t) pattern ->
+          {stmt with pattern})
+    with
+    | [stmt] -> Some stmt.Stmt.pattern
+    | stmts -> Some (Stmt.Pattern.SList stmts) in
```
WardBrian: I believe we have a separate pass later that removes singleton SLists, so no need to do it here?

test/integration/good/compiler-optimizations/vectorize-loops.stan
WardBrian, line 1: Can you include more of the edge cases from the unit tests in this file? Things like reassigning a side variable etc as additional code-gen

src/analysis_and_optimization/Optimize.mli
```diff
@@ -15,9 +15,9 @@ val one_step_loop_unrolling : Program.Typed.t -> Program.Typed.t
     continue statements in their body at the top level *)
 
 val vectorize_loops : Program.Typed.t -> Program.Typed.t
-(** Rewrite a loop whose body is a single scalar density statement into the
-    vectorized density call, when the Stan Math signatures have one. Loops that
-    do not match are left unchanged *)
+(** Rewrite a loop whose body is a single scalar density statement or
```
WardBrian: I don't think it has to be a single statement any longer?

WardBrian (commented):
Overall behavior looks pretty good to me, most of these comments are stylistic

src/analysis_and_optimization/Optimize.ml
seantalts: Yeah, it's basically just a non-nesting early return in the option monad, like the following in c
```
if (not cond) return;
```

In general I like as many non-nesting early returns as possible. Specifically here I had the LLM rewrite all of these to merge into pattern match `when` clauses though, which is less code and nicer to read probably.

WardBrian: I’m on board with the monadic let operators generally, but I feel like using them and .some_if at the same time just ends up being very indirect. 


I wouldn’t be opposed to introducing a local let operator over bools directly that did the same thing, e.g. `let*? () = cond in`?

I’ll have to look at what you’ve done most recently to see, but I’m guessing I’ll prefer the matches

src/analysis_and_optimization/Optimize.ml
seantalts: Added it back. Heads up that coverage will flag it since it's unreachable atm.

test/integration/good/compiler-optimizations/vectorize-loops.stan
```diff
@@ -21,4 +42,27 @@ model {
   for (n in 3 : N) {
     y[n] ~ normal(alpha, sigma);
   }
+  for (n in 1 : N) {
+    y[n] ~ normal(a[county[n]], sigma);
+  }
+  for (n in 1 : N) {
+    scratch[n] = alpha + eta[n];
```
seantalts, line 49: Is this what you meant by a side variable?

test/unit/Optimize.ml
seantalts: Maybe @bob-carpenter has an opinion. Arrays and vectors are different types to help users avoid certain classes of errors, I think. Making it easy to pollinate them might weaken that. We don't need to expose it to users to make this work in the compiler.

src/analysis_and_optimization/Optimize.ml
seantalts: Oh definitely, working it into existing guards is way cleaner imo, I feel like you'll agree.

src/analysis_and_optimization/Optimize.ml
```diff
@@ -671,110 +671,245 @@ let cannot_duplicate_expr ?(preserve_stability = false) (e : Expr.Typed.t) =
...
-      top_down_map_rec_stmt_loc (vectorize_statement sizes))
+      (List.map2 stmts patterns ~f:(fun (stmt : Stmt.Located.t) pattern ->
+           {stmt with pattern})) in
+  let written_names stmt =
+    fold_stmts
+      ~take_expr:(fun names _ -> names)
+      ~take_stmt:(fun names (stmt : Stmt.Located.t) ->
+        match stmt.pattern with
+        | Assignment (lhs, _, _) -> Stmt.Helpers.lhs_variable lhs :: names
+        | TargetPE _ -> "target" :: names
```
nhuurre: What???
... This makes no sense to me.


In accordance with our AI Contribution policy, you said that you have reviewed and understood the code. Please explain the reason for the inclusion of `"target"` here. What is the expected behaviour?

seantalts: I'm kinda confused about how confused you are. If we see a TargetPE statement then we are clearly writing to target, no?

nhuurre: This code tracks writes that prevent vectorization. Writes to `target` do not conflict with vectorization. The only effect adding it here is that for loops with multiple `target+=` cannot vectorize.

src/analysis_and_optimization/Optimize.ml
nhuurre: You did not really "address" the second paragraph which suggests to [rip it all out and put it elsewhere](https://github.com/stan-dev/stanc3/commit/fa2c8967d905ca657767004eb27c52de7b7d7aea). I'd prefer keeping _all_ index simplification out of this PR.

src/analysis_and_optimization/Optimize.ml
seantalts: Ah, I think I understand what you might be getting at - read references to target are actually FnApp(_, FnTarget) and handled in `can_side_effect_top_expr` / `cannot_duplicate_expr` so we don't need this line. I'll remove it. Regarding the AI accusations: I'm home sick with covid rn so that's why I have time to work on this at all. If you would prefer I not send PRs upstream, please let me know - I understand that reviewing AI-generated code is harder than human-written code. Or if Brian still wants these he can take over reviewing. I'm only submitting them because he asked me to - Stanli doesn't need this to be upstreamed.

nhuurre:
Honestly, it feels easier to [just write the whole thing myself](https://github.com/stan-dev/stanc3/commit/21fef0b8103a2c2d1a135e4d4ed1623c8f63a476) than review this.

src/analysis_and_optimization/Optimize.ml
seantalts: Oh your comment just loaded in as soon as I hit "send." This line doesn't need to be here but I think you are also misunderstanding the code here. 

This line doesn't block vectorization of two target+= statements; there is a test for that in the next PR.

`written` exists only to be checked against reads - we don't want to read a var that was written earlier in the same loop. 
Reads from target are excluded because we are (probably - perhaps this could be tightened up?) modifying target throughout the loop and vectorizing that would not give the same values of `target` as it would have each iteration.

nhuurre: You mean there is a [test on you next PR](https://github.com/seantalts/stanc3/blob/b054434624e5f13b9f5b924220aaa2bddb82b3fe/test/unit/Optimize.ml#L4667-L4691). Try copying that test here and see what happens.

WardBrian:
I am still in favor of these PRs. That being said, I’m also happy if we end up deciding it’s easier to treat the vibe coded bits as initial drafts/ “4 minute miles” and then someone does their own implementation that they feel like they can take ownership of and help maintain going forward

src/analysis_and_optimization/Optimize.ml
seantalts: You're right. Sorry, yeah I was getting confused with the next PR, which gets rid of this. I had the AI backport the test and get rid of this restriction here in this PR and go ahead and get rid of the "two writes" check & convert to set immediately.

seantalts:
> Honestly, it feels easier to [just write the whole thing myself](https://github.com/stan-dev/stanc3/commit/21fef0b8103a2c2d1a135e4d4ed1623c8f63a476) than review this.

That's almost always true and the perennial battle in open source :) AI makes it much worse. 

Feel free to close these PRs.

seantalts:
Discussed w Brian offline - will close this and let you go ahead with your versions. I'm glad there's someone out there with the time to do the hand-written one and that the compiler will finally get this feature after like 6 years :)

PR #1679, opened 2026-08-26
Expose typed MIR from Driver.Entry

seantalts:
## Summary

Expose `Driver.Entry.stan2mir`, which runs the existing frontend, Stan Math backend transformation, and configured MIR optimization suite, then returns the typed `Middle.Program.Typed.t` before C++ lowering.

This gives alternate in-process consumers a typed boundary without duplicating the driver phase sequence or parsing debug MIR output. `stan2cpp` now composes over `stan2mir`; CLI and stanc.js behavior remain unchanged.

The returned value is deliberately the exact backend-transformed and optimized MIR currently consumed by `Lower_program.lower_program`, rather than raw frontend MIR.

## Testing

- focused `Entry_tests` partition (optimized MIR and frontend error propagation)
- `dune build @fmt @check src/stanc/stanc.exe`
- `dune runtest test/integration/cli-args/debug-flags.t test/integration/good/code-gen`
- `dune runtest test/stancjs`

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [x] No user-facing changes were made

## Release notes

Expose a typed, optimized MIR entry point in `Driver.Entry` for alternate compiler consumers.

## Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause).

I used AI to prepare this PR but I vouch for the code.

### DIFF
--- a/src/driver/Entry.mli
+++ b/src/driver/Entry.mli
@@ -1,11 +1,15 @@
-(** The main entrypoint for Stan -> C++ compilation *)
+(** Stan compiler entrypoints *)
 
 open Frontend
 
 (** Either the C++ a model compiled to, or an error *)
 type compilation_result = (string, Errors.t) result
 
-(** The type of all non-C++-code outputs from the compiler *)
+(** Either a model compiled to Stan Math backend-transformed and optimized MIR,
+    or an error *)
+type mir_compilation_result = (Middle.Program.Typed.t, Errors.t) result
+
+(** The type of all auxiliary outputs from the compiler *)
 type other_output =
   | Formatted of string
   | DebugOutput of string
@@ -14,6 +18,16 @@ type other_output =
   | Version of string
   | Generated of string
   | Warnings of Warnings.t list
+
+val stan2mir :
+     string
+  -> [`Code of string | `File of string]
+  -> Flags.t
+  -> (other_output -> unit)
+  -> mir_compilation_result
+(** Compile a model through Stan Math backend transformation and MIR
+    optimization, without lowering it to C++. Takes the model's name, model
+    code, compiler settings, and a callback for auxiliary output. *)
 
 val stan2cpp :
      string
--- a/test/unit/Entry_tests.ml
+++ b/test/unit/Entry_tests.ml
@@ -1,0 +1,39 @@
+open Middle
+
+let compile_mir code flags =
+  match
+    Driver.Entry.stan2mir "entry_test_model" (`Code code) flags (fun _ -> ())
+  with
+  | Ok mir -> mir
+  | Error error -> failwith (Test_utils.error_to_string ~code error)
+
+let%expect_test "stan2mir returns transformed and optimized MIR" =
+  let flags =
+    { Driver.Flags.default with
+      optimization_level= Analysis_and_optimization.Optimize.O1 } in
+  let mir = compile_mir "model { target += 1 + 2; }" flags in
+  Fmt.pr "%a@." Fmt.(list ~sep:cut Stmt.Located.pp) mir.log_prob;
+  [%expect {|
+    { target += 3;
+    }
+    |}]
+
+let%expect_test "stan2mir reports frontend errors" =
+  let code = "model { target += missing; }" in
+  match
+    Driver.Entry.stan2mir "entry_test_model" (`Code code) Driver.Flags.default
+      (fun _ -> ())
+  with
+  | Ok _ -> print_endline "unexpected success"
+  | Error error ->
+      print_endline (Test_utils.error_to_string ~code error);
+      [%expect
+        {|
+        Semantic error in 'string', line 1, column 18 to column 25:
+           -------------------------------------------------
+             1:  model { target += missing; }
+                                   ^
+           -------------------------------------------------
+
+        Identifier "missing" not in scope. Did you mean "is_inf"?
+        |}]
--- a/test/unit/Entry_tests.ml
+++ b/test/unit/Entry_tests.ml
@@ -20,13 +20,10 @@ let%expect_test "stan2mir returns transformed and optimized MIR" =
 
 let%expect_test "stan2mir reports frontend errors" =
   let code = "model { target += missing; }" in
-  match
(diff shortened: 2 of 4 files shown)


WardBrian:
I’m not inherently opposed to this, but note that we will not really be guaranteeing any stability of either the MIR type or the Errors.t representation, so anything linking to this outside of OCaml is going to need to be extremely careful 


(So in particular, I wouldn't call it a "stable typed boundary")

seantalts:
Well, probably more stable than the debug output generated during cpp codegen? Haha

seantalts:
But yes, I don't have any expectations around versioning or stability for this API. I think this api endpoint makes sense for downstream consumers in general, but I foresee always using AI to do updates when the types change, new signatures are added, etc.

Removed the word stable from the PR desc.

test/unit/Entry_tests.ml
```diff
@@ -0,0 +1,39 @@
+open Middle
+
+let compile_mir code flags =
```
WardBrian, line 3: This helper is used in one of the two tests below but could be used in both, or just removed

(This entire file is redundant with our other tests, I believe, but I don't mind it)

src/driver/Entry.mli
```diff
@@ -15,6 +19,16 @@ type other_output =
   | Generated of string
   | Warnings of Warnings.t list
 
+val stan2mir :
+     string
+  -> [`Code of string | `File of string]
+  -> Flags.t
+  -> (other_output -> unit)
+  -> mir_compilation_result
+(** Compile a model through Stan Math backend transformation and MIR
```
WardBrian, line 28: It is a bit unfortunate that our pipeline does optimizations after the Stan-Math specific lowerings. This is what something like stanli wants, obviously, but it makes this kind of an odd API endpoint (What if someone wants just the normal MIR, for example? That is the more "portable" thing after all)

seantalts:
I should add this is all part of the plan you proposed to move the relevant parts of stanli upstream.

src/driver/Entry.mli
seantalts: Yeah. I think both endpoints would be fine. Why not be as useful as you can be to other packages in the OSS ecosystem? Sort of a modified Unix philosophy in my head :) 

I think it's inherent that we run some opts after deciding Stan math is the backend fwiw. Not sure where the exact line lies today but it makes sense to have backend specific opts for sure.

test/unit/Entry_tests.ml
seantalts: I think this file is testing the new endpoint, which isn't tested in older tests, right?

WardBrian: Since the “old” endpoint defers to this one for the first part of the pipeline, all those tests also cover the new function

(I guess this isn’t very “black box” of me to reason about — it’s certainly fine to have a few redundant tests here or there)

src/driver/Entry.mli
WardBrian: Agreed, and in practice it’s extremely difficult to actually do anything that isn’t the stan math backend just because the standard library is quite large and you’ll need to support a lot of weird functions out of the box 


It occurs to me that, with this endpoint, we’re halfway to something approaching a plugin structure. If we also provided an easy way to hand the MIR back to stanc, we’d be able to implement arbitrary optimizations as plugins (I think LLVM has something like this)

We would absolutely want to make some sort of API guarantee about that, so it’s probably not a good idea at the moment, but interesting to think about

WardBrian:
> I should add this is all part of the plan you proposed to move the relevant parts of stanli upstream.

Yes, and I still think that’s a good idea! I just wonder if the actual OCaml MIR type is the best layer to provide interop, or if we’d want to define some new thing that we actually could keep stable(ish)

In some sense I think this might be less stable than the s-expressions, because we intentionally remove some information like locations from those. 

I am planning a PR after this next release which changes the internal representation of the locations, which wouldn’t affect the sexps at all but will be incompatible for this function, for example

test/unit/Entry_tests.ml
seantalts: I think it's a good idea to add tests for the new api functions if you're ok with that?

src/driver/Entry.mli
seantalts: Yeah, LLVM provides a bunch of pass classes you can instantiate in various ways and you build your own pipeline with whatever passes you want. Of course you can use the default pipeline as part of that too. It's a cool idea to expose something like that in the future. I'm already planning to use this like that, I'll get the typed MIR and then run the upstream vectorization pass on it :)

seantalts:
What do you mean by "incompatible?" If you're just changing how they're represented that seems totally fine, I'll adapt. I'm planning to add my own OCaml stable API layer downstream. I think that's WAY less maintenance burden on upstream. Idk how much experience you've had with versioned APIs but versioned ASTs are way more of a pita lol. That said if upstream wants to design one and implement it stanli could evaluate switching to that down the line. It's a fairly big project to just design the stable API imo.

test/unit/Entry_tests.ml
WardBrian: Yep, the only actionable piece here is that I think you should either use this helper in both tests or delete it :)

WardBrian:
> What do you mean by "incompatible?"

It's most likely going to get a new field or have one of the existing fields change meaning to better track actual byte locations in files, which is not a backwards-compatible change for any code touching the current type. Nothing major, but just the kind of thing that is a hazard here

test/unit/Entry_tests.ml
seantalts: Oops, sorry, will do.

seantalts: Updated: both tests now call `compile_mir`; the frontend-error test catches the helper error and preserves the existing diagnostic assertion.

WardBrian (approved):

seantalts:
Thank you!

PR #1683, opened 2026-08-30
vectorize_loops refactor

### LINKED ISSUE #287: Implement optimization pass to automatically vectorize functions
For example, to automate things like this:
```
for (n in 1:240)
  y[n] ~ normal(mu, sigma);
```
becomes 
```
y[1:240] ~ normal(mu, sigma);
```

### LINKED ISSUE #702: Analysis and vectorization of for loops

nhuurre:
This is a continuation of / alternatve to #1678 and keeps all tests from there but rewrites `vectorize_loops` logic in a way I think is more maintainable.
Although loop vectorization is `--Oexperimental` optimization, note that this PR also changes `--O1` by adding index simplifications there.

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

The vectorize_loops optimization (enabled at --Oexperimental) now handles indirectly indexed arguments such as a[county[n]], density arguments built from elementwise arithmetic, independent statements within one loop, and elementwise assignment loops.

## Copyright and Licensing


WardBrian:
I didn't exactly follow all the review discussion in #1678, but did #1681 get folded in at some point? Just curious where this sits in terms of the overall plan


On a related note, I don't think I clocked that this was only at `--Oexperimental`. Since that setting also has some optimizations that we more or less know are broken, I'm hesitant to recommend users actually use it for testing... should we make an `O2`, or consider updating Oexperimental to basically be a copy of O1 with this new pass?

I guess I don't understand the risk behind putting it in O1 directly, either. We could do that immediately following the next release (2wks) if we wanted more time for developer testing

nhuurre:
It's `--Oexperimental` because that's where Sean put it in his first PR. I believe this may be robust enough for `--O1` but haven't tested anything. I agree `--Oexperimental` is too unstable to recommend. I guess we could put it behind `--O1` for the release candidate and see if early tester report any problems.

As for #1681 no, it is not included here, but offers three changes:
1. "temporaries": inlining scalar temporaries is nothing more than `expression_propagation` we already have as an experimental optimization pass. It just needs to be applied before loop vectorization. (but of course "experimental" means it may be completely broken...)
2. "dependent statements": allows loop to vectorize even when it has multiple writes to the same container and reads from written-to containers. This is a subtle dataflow analysis and I'm not planning to implement anything like it. The current PR only vectorizes order-independent statements.
3. "truncation": when a loop contains both vectorizable and non-vectorizable statements, don't just bail out entirely but move the vectorizable ones out. Implemented on this PR.

src/analysis_and_optimization/Optimize.ml
```diff
@@ -642,139 +642,240 @@ let unroll_loop_one_step_statement _ =
...
+        `Invariant e
+    | UInt | UReal -> widen_inner e
+    | _ -> `None
+  and widen_inner e =
+    match e.pattern with
+    | Indexed (base, [Single idx])
+      when not (Set.Poly.mem loopvar (expr_var_names_set base)) -> (
+        match (idx.pattern, widen idx) with
+        | Var idx, _ when String.equal idx loopvar ->
+            `Widened (Expr.Helpers.add_int_index base (Between (lower, upper)))
```
nhuurre: There's a bug here. The code is trying to be clever by not requiring that `base` is a variable and instead allows any loop-invariant expression, since that should be sufficient for the correctness of this widening.
But `Expr.Helpers.add_int_index` only accepts `Var`, `Indexed`, or `TupleProjection`.

src/analysis_and_optimization/Optimize.ml
```diff
@@ -642,139 +642,240 @@ let unroll_loop_one_step_statement _ =
 let one_step_loop_unrolling mir =
   transform_program_blockwise mir unroll_loop_one_step_statement
 
-let rec expr_any pred (e : Expr.Typed.t) =
-  match e.pattern with
-  | Indexed (e, is) -> expr_any pred e || List.exists ~f:(idx_any pred) is
-  | _ -> pred e || Expr.Pattern.fold (accum_any pred) false e.pattern
+let idx_any = Partial_evaluator.idx_any
+let cannot_duplicate_expr = Partial_evaluator.cannot_duplicate_expr
+let cannot_remove_expr = Partial_evaluator.cannot_remove_expr
```
WardBrian: Could these be in MIR_utils instead?

SteveBronder:
I think once all of it is in we move it to `--O1`. Else I think we could just have some of the optimizations we think always break turned off even at `--Oexperimental`

src/analysis_and_optimization/Optimize.ml
```diff
@@ -642,139 +642,263 @@ let unroll_loop_one_step_statement _ =
 let one_step_loop_unrolling mir =
   transform_program_blockwise mir unroll_loop_one_step_statement
 
-let rec expr_any pred (e : Expr.Typed.t) =
-  match e.pattern with
-  | Indexed (e, is) -> expr_any pred e || List.exists ~f:(idx_any pred) is
-  | _ -> pred e || Expr.Pattern.fold (accum_any pred) false e.pattern
+let can_duplicate_expr e = not (Mir_utils.cannot_duplicate_expr e)
+let cannot_remove_expr = Mir_utils.cannot_remove_expr
```
WardBrian: This file has `open Mir_utils`

src/analysis_and_optimization/Optimize.ml
```diff
@@ -821,7 +945,7 @@ let constant_propagation ?(preserve_stability = false) =
 let expression_propagation ?(preserve_stability = false) mir =
   propagation
     (Monotone_framework.expression_propagation_transfer ~preserve_stability
-       (cannot_duplicate_expr ~preserve_stability))
+       (Mir_utils.cannot_duplicate_expr ~preserve_stability))
```
WardBrian: This is similarly unnecessary

src/analysis_and_optimization/Optimize.ml
```diff
@@ -642,139 +642,263 @@ let unroll_loop_one_step_statement _ =
...
-   shape, and the proportionality flag is part of the function suffix, so they
-   are covered too.
+let stan_math_return_type name args =
+  let arg_types = List.map ~f:Expr.Typed.fun_arg args in
+  match Operator.of_string_opt name with
+  | Some op ->
+      Option.map
+        (Frontend.Typechecker.operator_stan_math_return_type op arg_types)
+        ~f:fst
+  | None -> Frontend.Typechecker.stan_math_return_type name arg_types
```
WardBrian: An extremely similar function occurs starting at line 101 of Partial_evaluator, might make sense to combine?

src/analysis_and_optimization/Optimize.ml
```diff
@@ -642,139 +642,263 @@ let unroll_loop_one_step_statement _ =
...
+        | `Scalar ls, Single e when is_loop_invariant_expr e ->
+            `Scalar (Index.Single e :: ls)
+        | `Scalar ls, Single {Expr.pattern= Var i; _}
+          when String.equal i loopvar ->
+            `Wide (Between (lower, upper) :: ls)
+        | `Scalar ls, Single e -> (
+            match widen e with
+            | `Widened e -> `Wide (MultiIndex e :: ls)
+            | _ -> `Error)
+        | (`Scalar _ | `Wide _ | `Error), _ -> `Error)
```
WardBrian: I'd find this a lot easier to read with a named type over these polymorphic variants. I can't even tell from looking at it if `Widened` and `Wide` are two different values in the same union or if there is a mapping between types happening here

test/unit/Optimize.ml
```diff
@@ -3686,7 +3686,7 @@ let%expect_test "vectorize: lpmf over an int outcome array" =
     }
     |}]
 
-let%expect_test "vectorize bail: no argument varies with the loop variable" =
+let%expect_test "vectorize: no argument varies with the loop variable" =
   (* Collapsing this loop would add one lp term to the target instead of N. *)
```
WardBrian: comment out of date

test/unit/Optimize.ml
```diff
@@ -3870,6 +4162,68 @@ let%expect_test "vectorize bail: invariant container argument" =
...
+      array[real, N] v;
+      v[:] = exp(x);
+      target += normal_lpdf(v, promote(0, real, data), sigma);
+    }
+    |}]
+
+let%expect_test "vectorize bail: array target with a vector right-hand side" =
+  (* The loop assigned real into real, but the widened rhs is a vector and the
+     target an array, so the sliced assignment would not typecheck.
+     GLM_Binomial_model's generated quantities have this shape. *)
```
WardBrian, line 4196: future: this special case might be common enough to justify inserting a `to_array` call?

test/unit/Optimize.ml
```diff
@@ -4041,3 +4395,552 @@ let%expect_test "vectorize: generated quantity names are not trusted sizes" =
...
+    }
+    |}]
+
+let%expect_test "vectorize bail: rng bounds" =
+  print_vectorized
+    {|
+      functions {
+        void loops_rng(int N, vector a) {
+          vector[N] x;
+          vector[poisson_rng(10)] w;
```
WardBrian, line 4852: this typechecks? that's fascinating

test/unit/Optimize.ml
```diff
@@ -4041,3 +4395,552 @@ let%expect_test "vectorize: generated quantity names are not trusted sizes" =
...
+          for (j in 1 : M) {
+            g[j] = x[j];
+          }
+          for (k in 1:M) {
+            g[k] ~ std_normal();
+            if (1) {
+              break;
+            } else {
+              continue;
+            }
```
WardBrian: This very complicated test being the only one that shows break/continue is a bit of a shame. Could we add a much simpler `bail` test just showing something like 

```
for (i in 1:N) {
 if (x[i] < 0) continue;
 x[i] ~ std_normal();
}
```
doesn't get vectorized erroneously?

test/unit/Optimize.ml
```diff
@@ -4041,3 +4395,552 @@ let%expect_test "vectorize: generated quantity names are not trusted sizes" =
...
+      matrix[N, N] w3;
+      FnValidateSize__("w4", "N", N);
+      FnValidateSize__("w4", "N", N);
+      matrix[N, N] w4;
+      a[:, 1, 1] = x;
+      for(i in 1:N) {
+        target += std_normal_lupdf(y[i, x[i]]);
+        w2[:, i] = y[i];
+        w3[i, :] = z[i];
+        target += ((N - 0) * std_normal_lupdf(y[i, i]));
```
WardBrian, line 4964: As an aside, I'm surprised that `x - 0` isn't partial evaluated to `x`

WardBrian (commented):
Mostly minor comments, the one about the polymorphic variants is the only one preventing me from saying I fully understand this code. The behavior looks good based on the tests however

src/analysis_and_optimization/Optimize.ml
nhuurre: There's a mapping because `Widened` contains an expression but `Wide` a list of indexes. You wouldn't want to mix those.

But yes, polymorphic variants are terrible. I was confused when these lines
https://github.com/stan-dev/stanc3/blob/21fef0b8103a2c2d1a135e4d4ed1623c8f63a476/src/analysis_and_optimization/Optimize.ml#L777-L779
were originally
```ocaml
| `Invariant e -> (invariant, Some e)
| `Widen e -> (false, Some e)
| _ -> (false, None))
```
and it compiled fine but nothing would vectorize. Ocaml didn't even complain that the second match case was unused, like it usually does!

test/unit/Optimize.ml
nhuurre: It really shouldn't--the generated code is wrong. But at least it's not optimization that breaks it.

WardBrian: Huh, seems like an easy fix though. Opened it as https://github.com/stan-dev/stanc3/pull/1691

WardBrian (approved):
Looks great, thanks!

src/analysis_and_optimization/Optimize.ml
```diff
@@ -642,139 +642,251 @@ let unroll_loop_one_step_statement _ =
...
+(* Rewrites e.g. [for (n in 1:N) target += normal_lpdf(y[n] | mu[n], sigma)] to
+   [target += normal_lpdf(y | mu, sigma)], and [for (n in 1:N) mu[n] = alpha +
+   beta * x[n]] to [mu[:] = alpha + beta * x]. Tilde statements have the same
+   MIR shape as target increments, so they are covered too.
+
+   An expression widens when every loop-varying node is [x[n]], [x[idx[n]]], or
+   a StanLib application of widened children that re-typechecks at a container
+   return type, with operators retrying as their elementwise variant. [x[n]]
+   becomes [x[lower:upper]], or [x] alone when the range provably spans the
+   declaration. Loops that do not match are left unchanged. *)
```
WardBrian:
```suggestion
*)
```

This comment still feels kind of slop-y, and I think the 'provably spans' part isn't true anymore anyway

PR #1711, opened 2026-09-13
Remove fmt.tty dependency with Stdlib

andrjohns:
#### Submission Checklist

- [x] Run unit tests
- Documentation

`stanc` currently uses the `fmt.tty` library to setup terminal outputs and formatting but this pulls in `base-unix` as a runtime dependency. This PR removes the `fmt.tty` dependency in favour of a small helper providing the same formatting and setup

## Release notes

Removed `fmt.tty` dependency

## Copyright and Licensing


WardBrian:
The name base-unix is a bit misleading because it is really part of the standard library, it’s just conditionally linked. Is there a reason you’d like to avoid this?

andrjohns:
Nothing too essential or critical, just exploring the C bytecode bundling again (for `stanli`). `Unix` is the only extra library required to be linked in addition to the core runtime (thanks to all of the other dependency removals!), dropping it simplifies the build a little - but only as a nice-to-have, so I'm not too fussed if it's not worth the manual implementation

WardBrian:
I think I'd rather not do this here, but I did flag it upstream https://github.com/dbuenzli/fmt/issues/70

PR #1611, opened 2026-04-06
Track additional locations for errors

WardBrian:
This updates the typechecker to store secondary locations for several types of errors. For example, if an overload is ambiguous, besides the primary location (the call site), it also adds information to the error type about where the misbehaving signatures were defined.

This is not currently utilized in the output, but I hope to soon integrate [grace](https://github.com/johnyob/grace), the primary advantage of which being that it does nicely support multiple locations in one error

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [ ] If a user-facing facing change was made, the documentation PR is here: <LINK>
    - [x] OR, no user-facing changes were made

## Release notes

Added additional location tracking in type errors.

## Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause)

PR #1414, opened 2024-03-25
use tabs to position error message cursor when appropriate

nhuurre:
On error, stanc prints the problematic source line and a caret that points to the horizontal location of the error but if the source line contains tabs instead of spaces the caret was at the wrong location.
A simple fix is to duplicate the tabs from the source to the line that contains the caret.

#### Submission Checklist

- [x] Run unit tests
- Documentation
    - [ ] If a user-facing facing change was made, the documentation PR is here: <LINK>
    - [x] OR, no user-facing changes were made

## Copyright and Licensing
Copyright holder: Niko Huurre

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause)

PR #1724, opened 2026-09-18
Overhaul compiler diagnostics

WardBrian:
This PR adds a dependency on the [`grace`](https://github.com/johnyob/grace) diagnostic library. Grace in essence implements a much more robust form of our `pp_context_for` function, with support for extra locations and overall nicer formatting, with a lot of flexibility to customize the errors for specific situations. 

This work began with #1611, but needed some [upstream improvements](https://github.com/johnyob/grace/pulls?q=is%3Apr+state%3Aclosed+author%3AWardBrian) before we could really utilize it. 

`grace` has no dependencies that we do not already directly depend on ourselves. 

Note that the actual code changes are only +800 -500, the rest of the diff is all of the tests in `test/integration/bad` having their expected outputs changed. 

Example:

### Before
image

### After

image

Note that tools such as the stan-language-server will have a difficult time adjusting to parsing the new error format; I have a follow-on PR that exposes the errors in a JSON format which will alleviate this. 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Release notes

## Copyright and Licensing


src/frontend/Diagnostic.ml
```diff
@@ -0,0 +1,98 @@
...
+let get_source ?printed_filename ?code loc =
+  let code = get_context code loc in
+  let name =
+    if Option.is_none loc.included_from then
+      Option.first_some printed_filename (Some loc.filename)
+    else Some loc.filename in
+  let content =
+    (* Tab alignment hack, c.f. https://github.com/johnyob/grace/issues/83 *)
+    String.replace_all ~sub:"\t" ~by:" " code in
+  let source : Source.t = `String {name; content} in
```
nhuurre: The reason to always use `'String` here instead of `'File` is (1) tab alignment and (2) filename override. I'm not sure I care about tabs so it would be nicer if `get_context` returned `Source.t`. It's pretty easy to convert to `'String` here if the filename needs changing.

src/frontend/Diagnostic.ml
```diff
@@ -0,0 +1,98 @@
...
+
+(** Generate secondary diagnostics identifying where a file was included *)
+let rec included_diagnostic ?printed_filename ?code
+    Middle.Location.{included_from; filename; _} : Diagnostic.Label.t list =
+  let range_of_loc loc =
+    let source = get_source ?printed_filename ?code loc in
+    let start = loc.byte_num in
+    let end_ =
+      start + 8
+      (* '#include' *) in
```
nhuurre:
```suggestion
      start + String.length "#include" in
```

src/frontend/Diagnostic.ml
```diff
@@ -0,0 +1,98 @@
...
+  Config.
+    { default with
+      styles=
+        { Style_sheet.default with
+          header_warning= [`Bold; `Fg `Magenta]
+        ; primary_label_warning= [`Fg `Magenta]
+        ; source_border= [`None; `Faint]
+        ; line_number= [`Fg `Yellow] }
+    ; chars= {Chars.unicode with source_border_left_break= "⋯"}
+    ; num_contextual_lines= 1
```
nhuurre, line 86: Previously we had two lines of context, I think. One is probably enough, but was there any particular reason for the change?

src/frontend/Diagnostic.ml
```diff
@@ -0,0 +1,98 @@
...
+    let end_ = clamp_int ~min:start ~max:(Source.length source) end_ in
+    Range.create ~source (Byte_index.of_int start) (Byte_index.of_int end_)
+  in
+  match included_from with
+  | None -> []
+  | Some loc ->
+      let label =
+        Diagnostic.Label.secondaryf ~range:(range_of_loc loc)
+          "file '%s' included here" filename in
+      label :: included_diagnostic ?printed_filename ?code loc
```
nhuurre: I'm not sure adding these as secondary labels works as intended. For example, try this model behind multiple levels of `#include`
```stan
model {
    int x = 1.5;
}
```
Included files are not listed in the correct order and the "file included here" messages are duplicated.

src/frontend/Preprocessor.ml
```diff
@@ -113,16 +113,15 @@ let find_include_fs lookup_paths fname =
...
           let keys = String.Map.to_list l |> List.map ~f:fst in
           if List.is_empty keys then Fmt.string ppf "None"
           else Fmt.(list ~sep:comma string) ppf keys in
-        Fmt.str
-          "Could not find include file '%s'.@ stanc was given information \
-           about the following files:@ %a"
-          fname pp_list map in
-      include_error message
+        Fmt.str "@[stanc was given information about the following files: %a@]"
+          pp_list map in
```
nhuurre: This really should say something more informative than `None` if the `map` is empty, like, "stanc was not given information about any files."

src/frontend/Semantic_error.ml
```diff
@@ -1,10 +1,61 @@
...
 
+(* location helpers, set by to_grace to avoid needing to pass them to every
+   single internal function *)
+let printed_filename_ref : string option ref = ref None
+let code_ref : string option ref = ref None
+let loc_ref : Location_span.t ref = ref Location_span.empty
+
+(* grace helpers *)
+
+(** This is the real workhouse function of this module. It is in charge of
```
nhuurre: I'm pretty sure the idiom is "workho**r**se"

src/frontend/Semantic_error.ml
```diff
@@ -515,112 +661,162 @@ module ExpressionError = struct
...
-        Fmt.pf ppf
-          "Ill-typed arguments supplied to prefix operator %a. Available \
-           signatures: %s@[<h>Instead supplied argument of incompatible type: \
-           %a.@]"
-          Operator.pp op
-          (Stan_math_signatures.pretty_print_math_lib_operator_sigs op
-          |> String.concat ~sep:"\n")
-          UnsizedType.pp ut
+        let sigs =
+          Stan_math_signatures.make_assignmentoperator_stan_math_signatures op
```
nhuurre: `assignmentoperator`? But these are for a prefix operator! I guess it works for all operators and the function name is just some legacy thing...

src/frontend/Semantic_error.ml
```diff
@@ -672,31 +879,34 @@ module StatementError = struct
           | LVariable id -> string ppf id.name
           | LIndexed (l, _) -> pf ppf "%a[%t]" pp_lvalue l ellipsis
           | LTupleProjection (l, ix) -> pf ppf "%a.%n" pp_lvalue l ix in
-        Fmt.pf ppf
+        make_error
+          ~summary:(Message.create "Ill-typed assignment statement.")
           "@[<v2>The same value cannot be assigned to multiple times in one \
            assignment:@ @[%a@]@]"
           Fmt.(list ~sep:comma pp_lvalue)
           lvs
```
nhuurre: Let's update this; the multiple conflicting lvalues should display their locations.

src/frontend/Syntax_error.ml
```diff
@@ -94,29 +90,56 @@ let pp_styled_text : styled_text Fmt.t =
...
-  | Parsing message -> pp_styled_text ppf message
-  | Lexing -> Fmt.pf ppf "Invalid character found.@."
-  | UnexpectedEOF -> Fmt.pf ppf "Unexpected end of input.@."
-  | Include message -> Fmt.pf ppf "%s@." message
+let to_grace ?printed_filename ?code ((loc, err) : t) =
+  let open Grace.Diagnostic in
+  let range, included =
+    Diagnostic.range_of_loc_span ?printed_filename ?code loc in
+  let primary = Label.primaryf ~range "here." in
+  let summary, notes =
```
nhuurre, line 98: I don't like "here". Could it just be the `summary` instead?

src/frontend/Typechecker.ml
```diff
@@ -529,29 +549,38 @@ let check_normal_fn ~is_cond_dist loc tenv id es =
...
+                  Env.find tenv (prefix ^ "_lpmf") |> List.map ~f:Env.location
+                in
+                if List.is_empty sugs then default_error ()
                 else
-                  Semantic_error.returning_fn_expected_undeclaredident_found loc
-                    id.name
-                    (Env.nearest_ident tenv id.name))
+                  Semantic_error.returning_fn_expected_wrong_dist_suffix_found
+                    loc (prefix, suffix, sugs)
+            | _ -> default_error ())
```
nhuurre:
```suggestion
            let suggestions =
              match suffix with
              | "lpmf" | "lupmf" ->
                  Env.find tenv (prefix ^ "_lpdf") |> List.map ~f:Env.location
              | "lpdf" | "lupdf" ->
                  Env.find tenv (prefix ^ "_lpmf") |> List.map ~f:Env.location
              | _ -> [] in
            if not (List.is_empty suggestions) then
              Semantic_error.returning_fn_expected_wrong_dist_suffix_found loc
                (prefix, suffix, suggestions)
            else default_error ()
```

src/frontend/Typechecker.ml
```diff
@@ -2066,10 +2095,16 @@ and verify_pmf_fundef_first_arg_ty loc id arg_tys =
...
   match List.find_a_dup ~cmp:Ast.compare_identifier arg_names with
   | None -> ()
-  | Some dup -> Semantic_error.duplicate_arg_names loc dup |> error
+  | Some dup ->
+      let dups =
+        List.filter ~f:(fun {name; _} -> String.equal dup.name name) arg_names
+      in
+      let prev = List.hd_exn dups in
+      let dup = List.nth dups 1 in
+      Semantic_error.duplicate_arg_names dup.id_loc prev |> error
```
nhuurre:
```suggestion
and verify_fundef_distinct_arg_ids arg_names =
  let _ =
    List.fold_left arg_names ~init:String.Map.empty ~f:(fun seen id ->
        match String.Map.find_opt id.name seen with
        | None -> String.Map.add seen ~key:id.name ~data:id
        | Some prev ->
            Semantic_error.duplicate_arg_names id.id_loc prev |> error) in
  ()
```

nhuurre (commented):
This looks fantastic but `grace` seems quite lacking in documentation. I pretty much just have to guess what these functions do.

---

By the way, Location comparator can be simplified
https://github.com/stan-dev/stanc3/blob/42d3cdfe90f14b9facda060f590e7a6902a1f04f/src/middle/Location.ml#L77-L81
```ocaml
        let x = Int.compare hd1.byte_num hd2.byte_num in
        if x <> 0 then x else go (tl1, tl2) in
```

src/frontend/Diagnostic.ml
WardBrian: `grace` has a better definition of context than we did -- it will print the entire region spanned by the error, plus this many lines on each side. Our previous code would just print two lines on either side of the _start_ of the error span, which was sometimes too much context and sometimes too little

The exact value I picked here is mostly based on liking how the output looked. Especially when longer error spans or secondary labels get involved, 2 lines of context on each side ends up being a lot of output

src/frontend/Diagnostic.ml
WardBrian: Can you elaborate? It is indeed true that `grace` doesn't let you decide on an ordering when multiple files are involved, which isn't too bothersome IMO, but duplicated messages sound like a problem

WardBrian:
> This looks fantastic but grace seems quite lacking in documentation

The `Grace` module [has ok docstrings](https://ocaml.org/p/grace/latest/doc/grace/Grace/index.html), but the Grace_ansi_renderer is indeed undocumented at the top level. I'm sure they'd accept more doc upstream, so if there are any functions that really stand out as needing it in this diff let me know and I can submit a patch

src/frontend/Diagnostic.ml
nhuurre: I mean thing like this:
https://github.com/stan-dev/stanc3/blob/5f4a3adee34280b089039f5caa08771f6c7e96af/test/stancjs/stancjs.expected#L285-L286


I think printing include stack makes sense only if the files are listed in the order of inclusion.

src/frontend/Diagnostic.ml
WardBrian: > I'm not sure I care about tabs

[Oh how the times change](https://github.com/stan-dev/stanc3/pull/1414)!

I do think it's fine if we just punt this to upstream (https://github.com/johnyob/grace/issues/83). I think 'spaces' pretty definitively won the war anyway.

nhuurre: [Times did not change](https://github.com/stan-dev/stanc3/pull/1413#discussion_r1537954885) -- I only did that PR because (1) it was very easy and (2) you thought it would be quite difficult.

WardBrian: It seems my memory is good, but not perfect...

src/frontend/Diagnostic.ml
WardBrian: Huh, I think this is actually a bug in the preprocessor code which detects loops. It only detects the loop the _next_ time it finds an include

nhuurre: No, that's just the only existing test that triggers the issue. The duplication happens for any included file that generates multiple labels.
Write this in `example.stan`
```stan
model {
    int x = 1.5;
}
```
and this in `example2.stan`
```stan
#include example.stan
```
Compiling `example2.stan` prints:
```
error: Ill-typed assignment statement.
    ┌─ ./example.stan:2:5
  1 │  model {
  2 │      int x = 1.5;
    │      ^^^^^^^^^^^^
    │      │
    │      Expected the right hand side to have a type matching the destination.
    │      Instead found type real.
    │      Left hand side has type int.
  3 │  }
    ┌─ example2.stan:1:1
  1 │  #include example.stan
    │  --------
    │  │
    │  file './example.stan' included here
    │  file './example.stan' included here
```
The last line is doubled.

nhuurre: By the way, that example also shows the location metadata is wrong for ill-typed assignments. It should be
```
  2 │      int x = 1.5;
    │          -   ^^^ Expected the right hand side to have a type matching the destination.
    │          │       Instead found type real.
    │          │
    │          Left hand side has type int.
```

WardBrian: Ah, the same include is being reported for both labels. In general, these actually could be in different files (less so for this example, but for something like the callback Secondary it definitely could). But we can rather simply dedupe them

src/frontend/Semantic_error.ml
WardBrian: Actually this was a copy/paste mistake that was generating bad messages, thanks for catching!

src/frontend/Semantic_error.ml
WardBrian: My latest push attempts to do so for this and the message immediately following which is similar. I think the errors are better, but the code is a bit messy and I'd appreciate your feedback

src/frontend/Syntax_error.ml
WardBrian: I personally think that looks visually too noisy for a lot of messages.

Compare:
image


image

I do think we could in general do better, but I think that would really want more custom error content on the parser errors side.

src/frontend/Semantic_error.ml
nhuurre: It's not _too_ messy but if you want to clean it up IMO you don't need to report anything other than the first error (cf. `DuplicateArgNames`).

src/frontend/Semantic_error.ml
```diff
@@ -113,7 +195,18 @@ module TypeError = struct
...
+  let rec expected_types : UnsizedType.t Nonempty_list.t Fmt.t =
+    let ust = expected_style UnsizedType.pp in
+    fun ppf l ->
+      match l with
+      | [t] -> ust ppf t
+      | [t1; t2] -> Fmt.pf ppf "%a or %a" ust t1 ust t2
+      | [t1; t2; t3] -> Fmt.pf ppf "%a,@ %a,@ or %a" ust t1 ust t2 ust t3
+      | t :: ts ->
+          Fmt.pf ppf "%a,@ %a" ust t expected_types
+            (ts |> Nonempty_list.of_list_exn)
```
nhuurre, line 207:
```suggestion
      | t1 :: t2 :: ts -> Fmt.pf ppf "%a,@ %a" ust t1 expected_types (t2 :: ts)
```

src/frontend/Semantic_error.ml
WardBrian: I'm happier with it now than a few commits ago. This is a situation where I think reporting multiple errors is good (there's no chance the additional error is spurious given the first), and it's a great demo of how improved the reports can be now:

Screenshot from 2026-09-22 15-55-40
Screenshot from 2026-09-22 15-56-02

nhuurre (approved):

PR #1725, opened 2026-09-21
Allow Dependency Analysis to Track Index Access

SteveBronder:
## Summary

This is the first PR in a breakdown of PRs for #1718. The overall goal is to have a dependency analysis framework that will allow us to do all of the optimizations listed [here](https://gist.github.com/SteveBronder/60649d86527e6f43f0c1b2216755632f).

The goal of this PR is to get everything in the dependency analysis setup so that we can track index access of objects. That will allow us to use the dependency analysis for the loop vectorizer and fix some false positives in the pedantic and factor graph analysis.

Currently `Dependence_analysis.ml` is used for the factor graph and pedantic analysis. My goal with this PR is to make it possible to use the same dependency analysis for the for-loop vectorization that comes next. To do that the analysis needs to understand which elements of a container each statement reads and writes, and in which iteration of the enclosing loop those elements are read or written to. In `main` the `VVar` type was a placeholder. By comments in the code it seems like there was a plan to flesh `VVar` out so that it would also track index dependencies. That never happened, so the analysis tracks variables by name only. 

Currently in the dependency analysis, a write to any element of a variable counts as a definition reaching a read of any other element of the variable. And a write at the bottom of a loop body counts as reaching a read above it in the same iteration. Both produce edges that cannot exist. And they surface in both consumers of the dependency analysis. Pedantic mode reports control flow that does not depend on a parameter, and the factor graph counts `target` terms as priors on a parameter they never touch.

For example, in the code below pedantic mode will give several false positives for `a`.

```stan
parameters {
  real a;
}
model {
  vector[2] theta;
  theta[1] = a;
  theta[2] = 1;
  if (theta[2] > 0) {   // main warns here: theta "depends on a"
    target += 1;
  }
  if (theta[1] > 0) {   // the warning that should fire
    target += 1;
  }
  a ~ normal(0, 1);
}
```

Running `stanc --warn-pedantic` on this model with a build of `main` gives three warnings: `a` has 3 priors, and both `if` statements depend on `a`. The third "prior" is the `target += 1` inside the first `if`, which the factor graph connects to `a` through the same false edge. With this PR the output `a` has 2 priors and only the second `if` depends on `a`. 

Adding index tracking to the dependency analysis will also fix cases like the below that gives false positives for `a` again.

```stan
for (n in 1:(N - 1)) {
  c[n] = b[n + 1];              // b[n] below is written in a later iteration
  d[n] = b[n];                  // b[n] below is written later in this iteration
  b[n] = a;
  if (c[n] > 0) target += 1;    // never depends on a
  if (d[n] > 0) target += 1;    // never depends on a
  if (b[n] > 0) target += 1;    // depends on a
}
```

On `main` all three `if` statements are reported and `a` has 3 priors. With this PR only `if (b[n] > 0)` is reported and there is no priors warning, because `a` now has exactly the one `target` term that reads an element written from it.

The same name-only tracking is what stops a vectorizer from using the depenency analysis. 
For an optimization to decide if a loop's statements can be vectorized it needs to be able to track reads and writes of indices across pairs of statements. The dependency analysis now has extra members in its nodes such that tools built on top of it can check that they are in a loop and query information related to index access inside of the loop. 

I will open another PR on top of this one tmrw that shows how we can rewrite the current loop vectorization such that we can optimize the current code + deeper dependencies. 

#### Submission Checklist

- [x] Run unit tests
- Documentation

## Copyright and Licensing


src/analysis_and_optimization/Dependence_analysis.ml
```diff
@@ -10,60 +10,292 @@ open Monotone_framework
...
+(* ---- Element test: can two accesses name the same element? ---- *)
+
+(** The loops enclosing both accesses, outermost first; [None] is a [While]. *)
+type frame = string option list
+
+let free = {directions= Set.Poly.of_list [Lt; Eq; Gt]; distance= None}
+
+(** Unconstrained at every level: any two iterations may name one element. *)
+let confused (frame : frame) : dependence =
+  Dependent (List.map frame ~f:(fun _ -> free))
```
WardBrian: This kind of thing is completely inscrutable to me. `frame`? "Unconstrained at any level"?

src/analysis_and_optimization/Dependence_analysis.ml
```diff
@@ -10,60 +10,292 @@ open Monotone_framework
...
+    factor graph asks which data and parameters feed each [target] term.
+
+    Four layers, each a map keyed by [label], an [int] naming one MIR statement:
+    + {b Statement map} ([Dataflow_utils.build_statement_map]): every statement
+      in pre-order, children replaced by the children's labels;
+    + {b Control flow} ([Dataflow_utils.build_cf_graphs]): predecessors, and the
+      {e control parents} ([if], [while], [for]) deciding whether a node runs;
+    + {b Reaching definitions} ([Monotone_framework.reaching_definitions_mfp]):
+      pairs [(variable, label')], "[label'] may be the last to have assigned
+      [variable]", keyed by name only;
```
WardBrian: Seems odd to document 3 other module's responsibilities here

src/analysis_and_optimization/Dependence_analysis.ml
```diff
@@ -10,60 +10,292 @@ open Monotone_framework
...
+      for a read of [theta[2]].
+
+    Running example, labels in the margin:
+    {[
+      theta[1] = a;          // 5
+      theta[2] = 1;          // 6
+      if (theta[2] > 0) ...  // 7
+    ]}
+    The definitions of [theta] reaching node 7 are the declaration and nodes 5
+    and 6; the element test drops node 5, so the [if] does not depend on [a]. *)
```
WardBrian: In other words, the definitions reaching node 7 are only the declaraction and 6? This is written very oddly; "the element test" is presumably the new functionality, but we should just document what it does now, not the path to get there historically

src/analysis_and_optimization/Dependence_analysis.ml
```diff
@@ -193,31 +419,216 @@ let mir_uninitialized_variables (mir : Program.Typed.t) :
                    (Set.Poly.union arg_vars globals)
                    fdbody))) ]
 
+(***********************************)
+(* Building the map                *)
+(***********************************)
+
+(** [left ± right]; [None] when [linear] cannot hold the result: two symbols,
```
WardBrian: These unicode symbols in comments are unnecessary

src/analysis_and_optimization/Dependence_analysis.ml
```diff
@@ -193,31 +419,216 @@ let mir_uninitialized_variables (mir : Program.Typed.t) :
                    (Set.Poly.union arg_vars globals)
                    fdbody))) ]
 
+(***********************************)
+(* Building the map                *)
+(***********************************)
```
WardBrian:
```suggestion
```

src/analysis_and_optimization/Dataflow_types.ml
```diff
@@ -7,15 +8,70 @@ open Std.Sexp_conv
...
 (** A 'reaching definition' (or reaching_defn or RD) statement (v, l) says that
-    the variable v could have been affected at the label l. *)
-type reaching_defn = vexpr * label [@@deriving sexp_of]
+    the variable named v could have been affected at the label l. *)
+type reaching_defn = string * label [@@deriving sexp_of]
+
+(***********************************)
+(* Access model                    *)
+(***********************************)
+
```
WardBrian: It looks like all the new types below this point are only used in the dependence analysis files? In which case, I think we should define them there. This Dataflow_types file is brought into scope several places and I think having a type like `level` floating around would be confusing/distracting

src/analysis_and_optimization/Dataflow_types.ml
```diff
@@ -7,15 +8,70 @@ open Std.Sexp_conv
...
+[@@deriving sexp_of, compare]
+
+(** One integer index expression as a function of the enclosing loop variables
+    (Goff, Kennedy and Tseng 1991, stride 1). *)
+type point =
+  | Affine of linear
+      (** invariant when [loopvar] is [None] (ZIV), else [loopvar + offset] for
+          one enclosing loop (SIV); neither mentions the written set *)
+  | Varying of varying_kind
+      (** not comparable by the element test; any pair is [confused] *)
```
WardBrian: Can you do a human pass over all of these comments on the types? I think understanding what each of these types is really meant to do would make reviewing a lot easier, but I don't think the comments here help me do that at all

src/analysis_and_optimization/Dependence_analysis.mli
```diff
@@ -13,76 +13,96 @@ open Dataflow_types
...
       indicates global dependence.
-    - Indexed variables are currently handled as monoliths
     - No probabilistic dependency, I'll do that elsewhere **)
 
-(** Sufficient information about each node to build the dependency graph.
-
-    Label dependence doesn't need the exit RD set, but variable dependence does.
-*)
+(** What the analysis knows about one statement; the overview and a worked
+    example are at the top of the implementation. *)
```
WardBrian: It is nice if that information was here (or both places if you feel strongly), because this file is what gets added to [the docs](https://mc-stan.org/stanc3/stanc/Analysis_and_optimization/Dependence_analysis/index.html#type-node_dep_info)

src/analysis_and_optimization/Dependence_analysis.ml
```diff
@@ -193,31 +419,216 @@ let mir_uninitialized_variables (mir : Program.Typed.t) :
...
+let node_accesses ~(loopvars : string Set.Poly.t) ~written_vars
+    (stmt : (Expr.Typed.t, 'substatement) Stmt.Pattern.t) : access list =
+  (* the loop variables are induction variables here, not written symbols *)
+  let written_vars = Set.Poly.diff written_vars loopvars in
+  let reads_in = reads_in_expr ~loopvars ~written_vars in
+  let reads_in_all exprs = List.concat_map exprs ~f:reads_in in
+  let increment_target = {var= "target"; subs= []; kind= Increment} in
+  match stmt with
+  | Assignment (((lbase, indices) as lhs), _, rhs) ->
+      (* a tuple projection is a write of the whole variable *)
```
WardBrian: This is incorrect in general, right? It's a fine simplifying assumption to make, but we should say "is **treated** as a write" if so

src/analysis_and_optimization/Dependence_analysis.ml
```diff
@@ -193,31 +419,216 @@ let mir_uninitialized_variables (mir : Program.Typed.t) :
...
+  |> Option.bind ~f:(fun parent ->
+      match pattern_of parent with
+      | Stmt.Pattern.For _ | While _ -> Some parent
+      | _ -> enclosing_loop statement_map parents parent)
+
+(** The accesses of one statement alone, reads before the write; an [if] or a
+    loop contributes only the condition or bounds; [_lp] calls increment. *)
+let node_accesses ~(loopvars : string Set.Poly.t) ~written_vars
+    (stmt : (Expr.Typed.t, 'substatement) Stmt.Pattern.t) : access list =
+  (* the loop variables are induction variables here, not written symbols *)
```
WardBrian: Comments like this need to be clarified/cleaned up to be less slop-sounding

WardBrian (changes requested):
This is indeed more manageable. I got pretty tripped up getting started on the review by some of the comments, so most of my review is asking for clarity there before I try to understand the rest of the code changes.

I reviewed everything outside of `Dataflow_types` and `Dependence_analysis` and have no complaints about those changes, so it's just those files left.

src/analysis_and_optimization/Dependence_analysis.ml
SteveBronder: Yes I ripped this whole thing out. Now we just have an `Unknown` type in `dependence` that gets rid of this

src/analysis_and_optimization/Dataflow_types.ml
SteveBronder: moved over

src/analysis_and_optimization/Dataflow_types.ml
SteveBronder: I went through each comment in this file to make sure it was less slop wording

src/analysis_and_optimization/Dependence_analysis.mli
SteveBronder: I'll move all of the docs over to the mli file

src/analysis_and_optimization/Dependence_analysis.ml
SteveBronder: I fixed this so parts of tuples are handled correctly

PR #1726, opened 2026-09-22
Evaluate reused scalar arguments once when inlining

seantalts:
Bind repeated, nontrivial `real` arguments to locals when inlining. For example, `f(exp(theta))` with `f(x) = x * x` evaluates `exp(theta)` once. Argument evaluation order is preserved.

The three regression tests, 100 optimization unit tests, and O0/O1/experimental C++ checks pass. The C++ expectation changes affect one fixture, mostly temporary-name renumbering. One unrelated generated-data test still has the same floating-point mismatch seen on unmodified master on this Mac.


WardBrian:
This seems like a reasonable change. Is there a reason to only do it for real types?

SteveBronder:
@seantalts would you mind writing a simpler stan model to show the effect in the C++? I tried reading the `--Oexperimental` expected C++ code and I had a hard time following.

I think it would be nice if we had something like `-finlining` etc for each optimization so it was easier to have tests that showed each optimization happening

seantalts:
@WardBrian updated to do more scalars; containers need more care.
@SteveBronder did you see the .ml expect tests already?

test/unit/Inline_arguments.ml
```diff
@@ -0,0 +1,103 @@
...
+  | _ -> events
+
+let events_stmt events (s : Stmt.Located.t) =
+  match s.pattern with
+  | NRFunApp (CompilerInternal FnPrint, _) -> events @ ["print"]
+  | _ -> events
+
+let print_events statements =
+  fold_stmts ~take_expr:events_expr ~take_stmt:events_stmt ~init:[] statements
+  |> String.concat ~sep:", " |> print_endline
```
WardBrian: I find this a pretty opaque way of testing compared to printing the MIR and visually inspecting.

WardBrian:
> containers need more care

Could you elaborate on this?

test/integration/good/compiler-optimizations/inline-arguments.t/run.t
WardBrian, line 1: I think this fits better as a test in test/unit/Optimize.ml, we haven't really used cram tests for this kind of thing before

test/unit/Inline_arguments.ml
WardBrian, line 1: Similarly, these could all be moved Optimize.ml to be with the others like them

src/analysis_and_optimization/Optimize.ml
```diff
@@ -254,6 +254,42 @@ let compute_suffix_and_name propto suffix fname =
...
+
+    These are syntactic reads, including reads in mutually exclusive branches.
+    For example, [f(exp(theta))] with [f(x) = x * x] becomes
+    [tmp = exp(theta); tmp * tmp], avoiding duplicate computation, autodiff
+    nodes, and effects. [f(theta)], [f(2.0)], other argument types, and
+    parameters used at most once retain direct substitution. Keep the binding
+    with the argument's own statements to preserve the existing evaluation
+    order. Container bindings would also need to preserve the original call's
+    reference handling. *)
+let bind_repeated_scalar_actual fname name body
```
WardBrian: I don't think we need to introduce "actual" as a new piece of terminology.

src/analysis_and_optimization/Optimize.ml
```diff
@@ -254,6 +254,42 @@ let compute_suffix_and_name propto suffix fname =
...
+    - its parameter is read at least twice in the already-inlined function body.
+
+    These are syntactic reads, including reads in mutually exclusive branches.
+    For example, [f(exp(theta))] with [f(x) = x * x] becomes
+    [tmp = exp(theta); tmp * tmp], avoiding duplicate computation, autodiff
+    nodes, and effects. [f(theta)], [f(2.0)], other argument types, and
+    parameters used at most once retain direct substitution. Keep the binding
+    with the argument's own statements to preserve the existing evaluation
+    order. Container bindings would also need to preserve the original call's
+    reference handling. *)
```
WardBrian: This is still not entirely clear to me, since all the arguments are const in the function. Is the concern about eigen expression templates that could potentially be unevaluated?

WardBrian (changes requested):

test/integration/good/compiler-optimizations/inline-arguments.t/run.t
seantalts: Moved this Stan example into `test/unit/Optimize.ml` as a MIR expect test and removed the Cram fixture.

test/unit/Inline_arguments.ml
seantalts: Moved all four tests into `test/unit/Optimize.ml`, using its existing reset helper and the standard MIR printer. All 105 tests in that file pass.

src/analysis_and_optimization/Optimize.ml
seantalts: Yes—the concern is evaluation and copying cost. Eigen arguments use `const auto& x = stan::math::to_ref(x_arg__)`, which evaluates expressions as needed and can preserve references to storage or views. An owning local could add a copy. The `const` binding is fine; I meant to leave this performance question for a separate change. I haven't measured a container regression, and I've clarified the comment.

WardBrian: In principle there’s nothing preventing us from generating the same sort of thing for the inlined variables, but we’d need a little more expressibility in the MIR I think. Worth keeping for later

src/analysis_and_optimization/Optimize.ml
```diff
@@ -357,6 +392,24 @@ let rec inline_function_expression propto adt fim (Expr.{pattern; _} as e) =
...
+and inline_function_args propto adt fim kind es =
+  let inline = inline_function_expression propto adt fim in
+  let definition =
+    match kind with
+    | UserDefined (fname, suffix) ->
+        let _, name = compute_suffix_and_name propto suffix fname in
+        String.Map.find_opt name fim
+    | _ -> None in
+  match (kind, definition) with
+  | UserDefined (fname, _), Some (_, args, body) ->
```
WardBrian: I think this would be a lot cleaner and avoid the double matching if it just happened inside the `| UserDefined ` branch on line 313. The existing calculation of `es` can just be inlined into the other branches

